PR #690, opened 2015-01-10
Move Bounds with other public PDAL files.

connormanning:


connormanning:
No longer valid - resolved by 1c016a75b65844204d71af7cf701a6a0e11753df.

PR #712, opened 2015-01-20
initial commit of PDALConfig.cmake

chambbj:
Those wishing to link against PDAL will not need a FindPDAL.cmake file,
they will only need to set PDAL_DIR to point to PDALConfig.cmake, which
is installed by PDAL (typically in /usr/local/lib/pdal/cmake).

This is a minimal approach for now, only addressing a very basic PDAL
build (i.e., no plugins). It also does not propagate flags, definitions,
etc. as it probably should in the end.

A version of the writing tutorial from the docs is provided as an
example.


chambbj:
forgot this needed a little more work on linux

PR #714, opened 2015-01-20
Pdal config

chambbj:
Initial commit was a little messed up on travis - pdalcpp and pdal_util needed to be exported together. Should be good now.


hobu:
This reminds me we should remove the PDAL package-config stuff or people will think it works and is supported :)

PR #719, opened 2015-01-21
Put version information in 'pdal info' output

hobu:


chambbj:
@hobu Are you ready to merge this?

:+1: to the version fix in doc building

hobu:
Yeah, I was going to squash it because it is messy. Doing so now.

PR #733, opened 2015-01-30
Remove CMAKE_BUILD_WITH_INSTALL_RPATH=TRUE for OSX

gadomski:
This option was breaking the out-of-the-box build on OSX, specifically by breaking each test so it couldn't find libpdalcpp.dylib.

As far as I understand rpath (which is not very far), CMAKE_BUILD_WITH_INSTALL_RPATH meant that the libraries and executables in our build tree (un-installed) would have their rpaths set to CMAKE_INSTALL_PREFIX directories. On a system w/o an installed PDAL, this meant that each test executable had no libpdalcpp.dylib to dynamically link against.

Removing this line falls back to the setting from above, where CMAKE_BUILD_WITH_INSTALL_RPATH was set to FALSE (IMO the correct setting).

@hobu I know that you were mucking with rpath stuff, so I don't know if you had a reason for setting CMAKE_BUILD_WITH_INSTALL_RPATH as such. FYI this problem cropped up in da40bce8fba074170703abe3c85155c2ed8ca802 when we stopped setting INSTALL_NAME_DIR for OSX systems in our macros.


hobu:
I'm supportive of this, but @chambbj has some unique requirements with building pdal as a PDAL.app/ to make OSX happy when doing GUI/VTK stuff. 

@chambbj do you want to merge this for now, and then #ifdef a .app solution afterward?

abellgithub:
I had the same experience.  Seems important to allow it to build/run correctly out of the box.  I ended up setting LD_LIBRARY_PATH.

chambbj:
I think we can go ahead and merge. If it breaks something for the app bundles, we'll address that separately.

PR #735, opened 2015-01-30
Add kernel plugin for Coherent Point Drift

gadomski:
This plugin depends on the C++ Coherent Point Drift library: https://github.com/gadomski/cpd/. It adds a kernel function to run nonrigid CPD on two source files, and write out the results to an output file. CPD uses Gaussian Mixture Models via Expectation-Maximization to nonrigidly transform one set of points to align to another set of points. Original work: https://sites.google.com/site/myronenko/research/cpd

The result file contains the registered points. If the output file type supports it, it also contains the point displacements in the {X|Y|Z}Velocity dimensions.

The plugin also has some extra logic to support (single-threaded) chipped CPD. Basically, we use PDAL's chipper to break the source dataset into chips, run each chip through CPD, then mush them all back together.

This commit only includes nonrigid-lowrank registration, for now. CPD itself can do rigid and affine registrations as well, but I've only needed nonrigid-lowrank so far.

I have not included CI integration for these changes because CPD itself has a medium-weight dependency chain of its own. It wouldn't be bad to script up the dependency install, but I wanted to potentially start a discussion about plugin CI testing (since, as I browse through the source tree, more than a couple plugins have tests). Do we want to move towards building and testing plugins on Travis, and accept the added complexity and build/test time? Or keep punting and accept that plugin tests can regress without Travis finding out? Or (this may be overkill) find some way to test PDAL plugins in a separate Travis repo (e.g. by hosting the CPD plugin external to the main source tree in PDAL-cpd or something like that)?


hobu:
:+1:

PR #738, opened 2015-01-30
Disable pgpointcloud plugin by default

gadomski:
I think users on a blank or close-to-blank system should see as few cmake errors as possible — since this was on by default, a fresh build on a mostly fresh system would throw an error.


strk:
ideally, the default should be based on the availability of dependencies on the system, building it if possible, and not if not possible

gadomski:
Something like: flip on the plugin if cmake's `find_package(PostgreSQL)` succeeds, but keep it off if it fails?

strk:
Something like that, but I don't know what dependencies the pgpointcloud driver has exactly.

gadomski:
I've updated the pull request to quietly search for PostgreSQL, and use POSTGRESQL_FOUND to initialize the option.

strk:
works for me (builds it, as desired)

PR #743, opened 2015-02-02
Add passthrough filter

chambbj:
This could obviously be more complicated, but as it is, I think it's a nice
alternative to the Python filters. The documentation, sample pipeline, and unit
tests should adequately explain the current feature set (basically
multi-dimensional, min/max/exact value criteria for passing points through to
the output buffer).


chambbj:
Just to clarify, this is not intended to be a full replacement for Python, just a small subset for simple filtering.

hobu:
Could this be called `filters.sieve`?

chambbj:
I have no strong feeling on the name. Passthrough is just the analogue in PCL.

hobu:
ah, that's fine to keep it aligned then. I think our challenge is that `filters` itself is probably wrong, since they don't really/always "filter".

chambbj:
I still have it in the back of my head to possible split the filters from the non-filters. I struggle a little with what to call the non-filters. I believe @abellgithub processed "operations" or something similar.

gadomski:
My two cents: `sieve` does make more sense than `passthrough` for me — I had to go look at the source to figure out what `passthrough` did, but I think I could have intuited `sieve`.

chambbj:
Discussed offline -- this basically just filters a range, and we don't have plans to make it any more complicated than that, so call it a range filter.

chambbj:
I'd missed a file rename on this. Otherwise, assuming it passes Travis, it's all ready to go.

PR #752, opened 2015-02-06
Initial GeoWave driver for PDAL

jwomeara:
This is the initial stab at adding GeoWave as a back end for PDAL.  This code uses Jace to run GeoWave through the JVM.  

By default, data is stored in Accumulo with multiple points per entry, conceptually similar to the chipping filter which is already in use.  Optionally, you can configure the driver to store a single point per Accumulo entry.  

There is still some documentation work which needs to be done to describe how to generate the C++ proxy files which are needed by Jace to interact with GeoWave.  That mainly involves pulling GeoWave from Github, pointing it to your PDAL source directory and running Maven with a specific profile.  In the future, we will look into providing a pre-generated version of these files for each release.


chambbj:
@jwomeara Glad to see you've been making progress with this! Looks like your driver is going to have enough dependencies that are unique to it that it should probably be treated as a "plugin" rather than a native driver. I'd suggest taking a look at the plugins directory for examples. Not too much will change.

Changes that come to mind include:

1) you'll create a CMake option called `BUILD_PLUGIN_GEOWAVE` that defaults to false
2) you can move all of your find_package calls to the new plugin CMakeLists so that people don't have to toggle the dependencies off and on individually, i.e., if you want to build the plugin you probably need to find them all, if not, you probably don't need any of them
3) you will no longer have to hardcode geowave driver details into `include/pdal/Drivers.hpp` and `src/StageFactory.cpp`
4) your reader/writer .cpp files will require a new macro `CREATE_READER_PLUGIN` (and likewise for the reader)

There may be more, but that's probably the bulk of it. I'd be happy to walk you through it if you run into issues.

jwomeara:
Hey Brad, 

I reconfigured GeoWave to run as a plugin.  I tested everything on my end, and it seems to be working.  I definitely need to write some documentation on setting everything up correctly.  I'll try to get that uploaded soon.  Thanks for your help.

jwomeara:
I added documentation which describes the GeoWave plugin.  I also created some documentation on the GeoWave io page [here](https://ngageoint.github.io/geowave/docs/pdal.html) to describe how to setup and configure the GeoWave plugin.  I think that's it...

jwomeara:
FYI, I created a GeoWave ticket to remind me to update your Vagrant config to include GeoWave.  Thanks!

PR #753, opened 2015-02-07
Some performance optimizations for naive readers.

connormanning:
- Add no-arg `append()` to push back an empty point onto the PointBuffer to be filled in later.  Returns index of the added point.
- Store the PointContext's pointSize when dimensions are added so the client doesn't have to store this separately to gain the speed benefits over looping through the dimensions.
- Add `set/getPoint` to the RawPtBuf (and allow access to it via a pointBuffer), which lets a client bulk-set/get an entire point without looping through the dimensions.  Only applicable to PointBuffers with identical PointContexts.

The latter two modifications sped up my test application considerably (about 40% time saved).

### DIFF
--- a/include/pdal/PointBuffer.hpp
+++ b/include/pdal/PointBuffer.hpp
@@ -107,6 +107,12 @@ public:
         m_size += buf.size();
         clearTemps();
     }
+    point_count_t append()
+    {
+        const PointId rawId(m_context.rawPtBuf()->addPoint());
+        m_index.push_back(rawId);
+        return m_size++;
+    }
 
     /// Return a new point buffer with the same point context as this
     /// point buffer.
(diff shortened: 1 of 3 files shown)


hobu:
Ready to merge?

abellgithub:
Didn't remember it was sitting out there.  Let me look at something first.

On Wed, Feb 11, 2015 at 4:43 PM, Howard Butler <notifications@github.com>
wrote:

> Ready to merge?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/PDAL/PDAL/pull/753#issuecomment-73971039>.
>



-- 
Andrew Bell
andrew.bell.ia@gmail.com

abellgithub:
@connormanning Where is the code that uses this?

PR #756, opened 2015-02-10
requiere boost but not geotiff

vmora:
Compiles fine without geotif include in test


hobu:
I'm merging @99f5f88162a9e423844004ddfd242351f6ca6b96 instead of this one to tease out the geotiff include issue separately from the Boost REQUIRED one.

hobu:
See #758 for that item.

PR #775, opened 2015-02-26
Add yaml pipeline parser

gadomski:
Add a yaml pipeline reader, so users can do the following:

```sh
$ cat > pipeline.yml
readers.las:
    filename: infile.las
filters.decimation:
    step: 2
writers.las:
    filename: outfile.las
$ pdal pipeline pipeline.yml
```

This is an itch I've wanted to scratch for a while, since a vast majority of my pipelines (and, I expect, other folks' pipelines) have no branches and no nested options. This yaml pipeline reader could be extended to support nested options as well if there were interest, but I didn't want to sink a lot of work into it up front.

Yaml pipeline support is implemented as an default-off `PipelineParserYaml` — the always-on `PipelineParserXml` is the way we've always done things. If you enable the `WITH_YAML_PIPELINES` option, yaml pipeline support is built using the vendored `yaml-cpp` library. I made the yaml pipeline parser default-off because I haven't tested it at all on Windoze — it wouldn't be hard to flip it to default-on or smart-on.

`yaml-cpp` depends on a few boost headers, so if we nuke boost we'll have to consider that. The xml pipeline parser depends on boost as well, so I figured it wouldn't hurt to add a bit more boost to a spot that already has a ton of it.

I haven't documented yaml pipeline support because I wanted to get feedback first. If it looks like something we do want to integrate I can write up some information on its usage.

Comments, suggestions, and complaints welcome.


abellgithub:
YAML?  XML, JSON and now YAML?  That's my only complaint.  Make it JSON and I'd think it's great.  I've done YMAL before.  It's fine.  I have nothing against it.  It's just proliferation of MLs.

gadomski:
I hear you, and I thought the same thing at first. I ended up doing a yaml b/c I wanted to implement something that didn't have closed braces/tags, which makes the file itself much easier to create on the fly. This yaml doesn't have to creep beyond the PipelineParser stage, either — I think yaml is less useful as an output format. It's probably my favorite config file format.

I think its worth discussing, but in my opinion the ease-of-use-to-the-end-user win of yaml is worth the added complexity.

abellgithub:
Would you be open to some other methodology to do simple pipelines?  I've been thinking about it, but I'm not sure that I've reached a conclusion yet.

gadomski:
Yup, I'm not married to this way. It was just The Thing That I Thought Of. Very open to other alternatives as well, willing to shelve this for a bit if you need time to put something together.

gadomski:
Well, turns out `yaml-cpp` is an unreliable dependency anyways — I'm getting inconsistent behavior from simple usages between OSX and Ubuntu. I'm going to close the PR for the time being, but count the exercise as useful if it started thought on ways to simplify pipelines :beer:.

PR #776, opened 2015-02-27
Geowave vagrant

jwomeara:


jwomeara:
I verified that this works on an AWS Ubuntu Trusty64 node, but not officially through Vagrant as that's something I can only really try at home.  I'll give Vagrant a go sometime this weekend, but I'm pretty confident that it will work the same.  

I updated the install scripts to first clone pdal, then build GeoWave-Jace bindings against pdal, and then install pdal.  The install script also sets up geowave service which kicks off a 2 node MiniAccumulo cluster with GeoWave installed.  The default connection parameters are:

<Option name="zookeeperUrl">localhost:2181</Option>
<Option name="instanceName">geowave</Option>
<Option name="username">root</Option>
<Option name="password">password</Option>

Let me know if you have any questions.

jwomeara:
I almost forgot, I updated the GeoWave plugins to check to see if a JVM is running before trying to spawn a new JVM.  I ran into an issue with this when running the command - pdal info input.xml - with the GeoWave Reader.  

I noticed that GeoWaveReader's initialize method was being called twice before any data was read.  In InfoKernel.cpp the dump method calls PipelineManager's prepare method and then calls PipelineManager's execute method.  Looking at the execute method of PipelineManager, it also calls it's prepare method.  I think that's how initialize was being called twice for my reader.  

Anyway, this is a JNI specific thing.  You can only spawn one JVM per application, and once you destroy that, you can't get another (JNI won't let you).  Either way, that's a situation that we should be able to handle, so I added the quick check to ensure that the JVM isn't running before we try to initialize a new one.

abellgithub:
There is a bug in the info kernel where prepare gets called too many times.  It's on my list to fix.

PR #782, opened 2015-03-04
Add reader for Optech (csd) files

gadomski:
Optech provides ALS discrete-return data in Corrected Sensor Data files, extension .csd. These files provide scan angle and range information along with IMU data, boresight calibration values, and other sundry data bits. The Optech reader does the georeferencing necessary to convert the GNSS+IMU+SOCS data into latitudes, longitudes, and heights.

The Optech documentation states that the latitudes and longitudes are provided in the wgs84 reference frame, so the spatial reference system of the Optech reader is fixed at EPSG:4326.

The Optech reader is provided as an optional plugin. No additional dependencies or external libraries are required.


gadomski:
I'm going to keep this open, but per discussion with @hobu I'm going to:

- make this a native driver rather than a plugin
- pluck out and genericize the georeferencing stuff to libpdal or libpdal_utils

gadomski:
@abellgithub I've updated per your suggestions. I'll let the tests run while I cook up some docs for the Optech reader – we shouldn't merge until those docs are done.

gadomski:
Okay, docs are done, this thing is ok to be merged. Don't know if it should go now or if there's about to be a lot of **master** movement that I should wait for?

PR #784, opened 2015-03-04
Add RialtoWriter

chambbj:
RialtoWriter is a tiled format that is currently under development. See
http://lists.osgeo.org/pipermail/pointdown/2015-February/000001.html for more
information.

In addition to the RialtoWriter, we have added several functions to FileUtils
to facilitate handling of directories (existance, recursive removal, and
creation).

We have also fixed a bug in TranslateKernel, in which the output_srs option was
not being passed to the reprojection filter.


hobu:
Rebase this to catch up to the plugin refactor and then merge at your discretion.

Issue #783, opened 2015-03-04
Purge `Utils::read_n`


gadomski:
It is superseded by IStream and Extractor, and should be purged to simplify our API.

PR #791, opened 2015-03-07
Purge read_n and read_field from Utils.hpp

gadomski:
V.a.v. #783, `Utils::read_n` and `Utils::read_field` have been superseded by the capabilities of `IStream`, `Extractor`, and friends. This pull request removes the superseded methods.

There are a series of commits in this pull request, all related to `read_n` and `read_field` cleanup.

- Add a new method to `IStream` that allows seeking by position (allows seek to end)
- Added new classes `ISwitchableStream` and `SwitchableExtractor` to support the wonky qfit format. Qfit files sometimes are big endian and sometimes are little endian, and you can't tell until you read the file itself. The `*Switchable*` flavors of `IStream` and `Extractor` provide `switchToBigEndian()` and `switchToLittleEndian()` methods, which switches the read mode of the `IStream` or `Extractor`. I did not use virtual methods because I wanted to avoid polluting the interface of the existing `IStream` and `Extractor` classes.
- Refactored the qfit reader to use `*Switchable*` `IStream` and `Extractor`
- Add tests for the `TerrasolidReader` (it was untested)
- Fixed two bugs in the `TerrasolidReader`; one was a critical bug which trapped the reader in an infinite loop when reading the test file, one was a bug with the way the TerrasolidReader handles return numbers
- Refactor the TerrasolidReader to use `IStream` and `Extractor`
- Remove the offending methods, `read_n` and `read_field`

### DIFF
--- a/include/pdal/util/Extractor.hpp
+++ b/include/pdal/util/Extractor.hpp
@@ -240,4 +240,158 @@ public:
     }
 };
 
+
+/// Stream wrapper for input of binary data that converts from little-endian
+/// to host ordering.
+class PDAL_DLL SwitchableExtractor : public Extractor
+{
+public:
+    static const bool DefaultIsLittleEndian = true;
+
+    SwitchableExtractor(const char* buf, std::size_t size)
+        : Extractor(buf, size)
+        , m_isLittleEndian(DefaultIsLittleEndian)
+    {
+    }
+
+    SwitchableExtractor(const char* buf, std::size_t size, bool isLittleEndian)
+        : Extractor(buf, size)
+        , m_isLittleEndian(isLittleEndian)
+    {
+    }
+
+    bool isLittleEndian() const { return m_isLittleEndian; }
+    void switchToLittleEndian() { m_isLittleEndian = true; }
+    void switchToBigEndian() { m_isLittleEndian = false; }
+
+    using Extractor::get;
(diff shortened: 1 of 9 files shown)


gadomski:
I've updated this pull request for the new plugin regime. Since there are non-trivial changes to native readers, could someone take a quick look and sign off, just to sanity check before merge? Thanks.

include/pdal/util/Extractor.hpp
```diff
@@ -240,4 +240,158 @@ class PDAL_DLL LeExtractor : public Extractor
     }
 };
 
+
+/// Stream wrapper for input of binary data that converts from little-endian
+/// to host ordering.
+class PDAL_DLL SwitchableExtractor : public Extractor
```
abellgithub: The reason not to make this switchable is performance.  If there are separate classes, you eliminate all the if (little_endian) stuff.  It's in code like this where people seem to care about such things.

gadomski: Understood. Qfit was already doing some goofy switching logic in its reader, which I tried to encapsulate with this `SwitchableExtractor`. I don't *think* this should introduce any performance hits that weren't already present in the qfit reader (though of course I could be wrong there).

abellgithub:
I don't notice anything interesting.  If the tests pass, seems fine to me.

PR #803, opened 2015-03-10
move references to HIGHEST and LOWEST from .hpp to .cpp to make msvc hap...

chambbj:
...py


chambbj:
Hold off on merging this one. `BOX3D::grow()` will need to move as well.

PR #804, opened 2015-03-10
catch up PCL kernels to new plugin changes #774

hobu:


plugins/pcl/kernel/GroundKernel.cpp
```diff
@@ -136,7 +117,19 @@ int GroundKernel::execute()
     readerOptions.add<bool>("debug", isDebug());
     readerOptions.add<uint32_t>("verbose", getVerboseLevel());
```
chambbj: Are the previous two lines necessary, given the new block below?

plugins/pcl/kernel/GroundKernel.cpp
```diff
@@ -136,7 +117,19 @@ int GroundKernel::execute()
...
 
-    std::unique_ptr<Stage> readerStage = makeReader(readerOptions);
+    if (isDebug())
```
chambbj: This block could probably be dropped into `setCommonOptions`, which we use for the writer, but not the reader (don't remember why). It's a little crude, but cuts out some redundant code.

PR #815, opened 2015-03-13
Bash completion

strk:
Provide a bash completion script, to allow TAB-based commandline completion for the "pdal" command.


strk:
NOTE: I hadn't touched cmake scripts to attempt at automatically installing the completion. Would be nice to have that too.

strk:
Allows for: ```pdal i<TAB> myfile<TAB>```

strk:
Now that I build with plugins I noticed that obtaining the list of commands is annoyingly slow, even if a single plugin exists. You notice because plugins are loaded on hitting <TAB> ...

PR #837, opened 2015-03-18
Properly quote PG identifiers

strk:
Includes test.
Allows naming tables after filenames with no "pre-washing".

### DIFF
--- a/plugins/pgpointcloud/io/PgCommon.hpp
+++ b/plugins/pgpointcloud/io/PgCommon.hpp
@@ -41,6 +41,7 @@
 #include <pdal/pdal_error.hpp>
 #include <pdal/Options.hpp>
 #include <pdal/Compression.hpp>
+#include <pdal/Utils.hpp>
 
 namespace pdal
 {
@@ -144,6 +145,13 @@ inline PGresult* pg_query_result(PGconn* session, std::string const& sql)
     return result;
 }
 
+inline std::string pg_quote_identifier(std::string const& name) {
+  return std::string("\"") + Utils::replaceAll(name, "\"", "\"\"") + "\"";
+}
+
+inline std::string pg_quote_literal(std::string const& lit) {
+  return std::string("'") + Utils::replaceAll(lit, "'", "'") + "'";
+}
 
 
 } // pdal
--- a/plugins/pgpointcloud/io/PgReader.cpp
+++ b/plugins/pgpointcloud/io/PgReader.cpp
@@ -108,11 +108,11 @@ point_count_t PgReader::getNumPoints() const
         return m_cached_point_count;
 
     std::ostringstream oss;
-    oss << "SELECT Sum(PC_NumPoints(" << m_column_name << ")) AS numpoints, ";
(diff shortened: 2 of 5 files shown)


abellgithub:
There is already Utils::replaceAll().  Also, should probably just run stuff once and use that, rather than embed the function call in every usage.

strk:
I just tried re-implementing the quoters with replaceAll but it enters
an infinite loop.

```
inline std::string pg_quote_identifier(std::string const& name) {               
  //return pg_quote(name, '"');                                                 
  return Utils::replaceAll(name, "\"", "\"\"");                                 
}    
```

Running the quoting once, upfront, would require storing 2 different
versions of each identifier (schema, table, column) as they are both
used as literals (for pointcloud_columns query) and identifiers.

strk:
does travis not build or test pgpointcloud ?

strk:
I've a fix for Utils::replaceAll, whcih should also make it generally faster. Will commit shortly.

strk:
There, now replaceAll is used (once fixed). I hadn't found an existing test for replaceAll and didn't add one.

abellgithub:
Sorry, I already committed something with regard to replaceAll.  Can you explain why you'd get quotes in the filenames?  Seems an odd case.

strk:
Quotes in the table name were put intentionally, to test the SQL quoting.
Once we quote, it's good to do it correctly.

The issue with replaceAll was that it did enter an infinite loop when
the string to be replaced is a substring of the replacement one.
Was that case fixed by your change ?

--strk;

strk:
I've seen your change. It does fix the loop (and shows mine was bogus). Thanks!
I'll rebase this PR against master and force-push.

strk:
Rebased. It's 3 commits. First one adds the quoting methods and makes use of it.
Second and third ensure all configuration options are always used (not directly related to quoting but new code does proper quoting so needs the quoting methods)

abellgithub:
I still don't understand what we're doing.  I get that if you have an embedded single-quote in SQL, it needs to be escaped by using two single-quotes.  But I don't understand why we'd need to escape the double quotes in the context in which we're using them.  They're embedded in the final string that gets passed to the postgres function, right?  These aren't literals that we're embedding.

We need a good test case that shows failure, so that we can make sure that any change we're making actually fixes the problem in question.

strk:
My filename is:

 ``20h64_1x1_raw_004_2010_3003-tcxyzissssa``

Unquoted identifiers cannot contain dashes.

``
strk=# create table a-b (a int);
ERROR:  syntax error at or near "-"
LINE 1: create table a-b (a int);
                      ^
``

There are other characters that you cannot put in unquoted identifiers.
Quoting identifiers allows you to use any character.

``
strk=# create table "a-b" (a int);
CREATE TABLE
``

Double-quotes inside quoted identifiers need be escaped by
using two double-quotes.

strk:
BTW, the test is included. Try running the updated test (pgpointcloudtest.cpp)
against master and you'll get the failure.

Beware of #840, which may actually make testing easier ...

PR #841, opened 2015-03-18
Drop partial support for "capacity" option in pgpointcloud.

strk:
See https://github.com/pgpointcloud/pointcloud/issues/48


abellgithub:
Could you please update doc/stages/writers.pgpointcloud.rst as part of this change?

strk:
pushed update of the .rst file

PR #842, opened 2015-03-18
Windows clean-up (mostly warnings)

chambbj:
Compiler is VC 11.0. Compiler warning is referenced in parens.

Stop getting PluginManager instance when it is not needed. Remove initialized, but unreferenced variables (C4189).

Use `EXPECT_DOUBLE_EQ` vice `EXPECT_FLOAT_EQ` where appropriate. Rearrange comparisons such that the expected value comes first (not strictly necessary, just a convention). Compute some expected values rather than typing (and truncating) them.

Change `point_count_`t and `PointId` to `uint32_t`. Change datatypes throughout the codebase to try to adhere to this convention, using `PointId` for point indices and `point_count_t` for point counts.

Standardize handling of `-Wfloat-equal` on various compilers, generally when sucking in GDAL/OGR headers.

Stop casting values to bool. Compare `non_bool_value != 0` instead (C4800).

Safer tolower/toupper implementation (so sayeth stackoverflow). See http://stackoverflow.com/questions/19876746/stdtolower-and-visual-studio-2013.

Omit termination test from for loop (rather than using `true`) to avoid warning about constant conditional expression (C4127).

Globally suppress warning about nameless union/struct (C4201).

Globally suppress warning about exception specifications. These are used extensively in Nitro (C4290).


filters/colorization/ColorizationFilter.hpp
```diff
@@ -39,12 +39,23 @@
 
 #include <boost/array.hpp>
```
abellgithub: Clang handles GCC diagnostics just fine.  I would think they'd be silently ignored by MSVC.  But like Howard said, I thought we were pitching this.

include/pdal/Metadata.hpp
```diff
@@ -560,7 +560,7 @@ class PDAL_DLL MetadataNode
     }
 
     bool hasChildren() const
-        { return m_impl->m_subnodes.size(); }
+        { return (m_impl->m_subnodes.size()) != 0; }
```
abellgithub: Gah.  I hate this so much.  Why must we state the obvious?  Can we suppress whatever warning this is?

filters/colorization/ColorizationFilter.hpp
chambbj: If we're ditching it, that's fine. Was just shooting for a common handling.

include/pdal/Metadata.hpp
chambbj: We can. Didn't know how overboard we wanted to go with disabling warnings.

include/pdal/Utils.hpp
```diff
@@ -149,17 +149,17 @@ namespace Utils
 
     inline std::string tolower(const std::string& s)
```
abellgithub: Curious what the problem was the the old implementation.  Seems like it's doing exactly the same thing.  And what's with the 'X' bit?

test/unit/Support.cpp
```diff
@@ -224,7 +219,7 @@ uint32_t Support::diff_files(std::istream& str1, std::istream& str2,
     uint32_t numdiffs = 0;
     char p, q;
 
-    for (uint32_t i = 0; true; ++i)
+    for (uint32_t i = 0; ; ++i)
```
abellgithub: Not that I care, but why, oh why would anyone issue a warning for this?

include/pdal/Utils.hpp
chambbj: I'll have to pull it back out to reproduce the actual warning/error that led me to look at this. The stackoverflow link in the commit message has some commentary on this approach, having to do in part with signed vs. unsigned char's. The 'X' bit is just me initializing the output string to be the proper length. Maybe I could have resized instead. It must be the right size before using transform though, or bad things happen.

include/pdal/Metadata.hpp
abellgithub: I guess I'm old.  The casting of ints to bools seems superfluous.  If you don't know that this is part of the language, then it seems like you don't know the language.  I think someone just got assigned some task like "Make the compiler emit lots of warnings."

I'm done now :)

test/unit/Support.cpp
chambbj: This is certainly one of the smaller ones. We could gloss over it, suppress it, whatever. The overall intent of the PR is to get the Windows build output to be significantly quieter so that we can see the actual errors through the slew of warnings.

include/pdal/Metadata.hpp
chambbj: No, that's fine. It's not always ints to bools FWIW.

include/pdal/Utils.hpp
chambbj: The original way led to...

`warning C4244: 'argument' : conversion from 'int' to 'char', possible loss of data`

repeated literally hundreds of times. We could ignore if that's preferred.

chambbj:
Maybe the answer is to back off the project wide warning level? We've been at level 4 for as long as I can remember. Backing off to level 3, we'll be left to address/suppress C4244 (convert int to char), C4290 (exception specification), C4800 (casting to bool).

hobu:
The only strong opinion I have is we want to see relevant warnings on the windows builds so we can react to them now that we have automated windows builds. Backing up to /W3 is fine with me.

include/pdal/Utils.hpp
abellgithub: I don't care that much.  I think a cast of the int that was resulting from (std::toupper) would have been fine, too.  Mostly was curious what was wrong.

chambbj:
@hobu @abellgithub There is a little less upheaval in this updated version. Global `/W3` instead of `/W4`. Added `/wd4800` to ignore the bool stuff.

abellgithub:
Very nice.

chambbj:
Rebased with master.

PR #844, opened 2015-03-24
Request ctest to load plugins from the build dir

### LINKED ISSUE #840: tests use system-installed plugins
Rather than using the built plugins, testcases use the installed ones.
I'm pretty sure about this at least for the "pgpointcloudtest" case, not sure about the others.

I could tell as, after fixing a bug, "make test" failed, "sudo make install; make test" succeeded.

strk:
Fix #840


gadomski:
:+1: from me. I like the per-target property configuration.

Issue #839, opened 2015-03-18
Have travis run pgpointcloud tests


strk:
It seems to me that Travis is not running pgpointcloud tests.
I know because it gave a green on #837 while still known to be broken.

strk:
To confirm pgpointcloud is not tested: https://travis-ci.org/PDAL/PDAL/jobs/54878637#L1323-L1333

strk:
On my local build, where PGPOINTCLOUD support is built, the "PostgreSQL" package is listed under "REQUIRED" packages that have been found. On travis, it is still listed as "found", but under the "OPTIONAL" packages. What's the deal ?

gadomski:
I believe it is because we actually search for PostgreSQL in [two](https://github.com/PDAL/PDAL/blob/master/CMakeLists.txt#L139) [spots](https://github.com/PDAL/PDAL/blob/master/cmake/postgres.cmake#L5). The first is to auto-enable (or disable) PostgreSQL support, the second is to ensure that PostgreSQL is available if the user has asked for it.

Since the first `find_package` is non-required, PostgreSQL will show up as optional if the plugin [is not enabled](https://github.com/PDAL/PDAL/blob/master/scripts/ci/script.sh#L36).

strk:
I actually found the exact spot where PGPOINTCLOUD is _explicitly_ disabled:
https://github.com/PDAL/PDAL/blob/master/scripts/ci/script.sh#L36

strk:
For the record, I'm using PR #846 for testing enabling pgpointcloud tests  in travis

strk:
@gadomski indeed enabling PGPOINTCLOUD makes it listed as REQUIRED: https://travis-ci.org/PDAL/PDAL/jobs/55809540#L1897

It also shows the additional message I was missing before:
https://travis-ci.org/PDAL/PDAL/jobs/55809540#L1865

PR #846, opened 2015-03-25
Try installing postgresql server headers

strk:
Empirical attempt at having travis test pgpointcloud (#839) -- not yet ready for merge

### DIFF
--- a/scripts/ci/before_install.sh
+++ b/scripts/ci/before_install.sh
@@ -66,6 +66,15 @@ then
     cd libgeotiff-1.4.0
     ./configure --prefix=/usr && make && sudo make install
     cd $TRAVIS_BUILD_DIR
+
+    # install pgpointcloud from sources
+    sudo apt-get install postgresql-server-dev-9.1
+    wget https://github.com/pgpointcloud/pointcloud/archive/master.tar.gz
+    tar -xzf master.tar.gz
+    cd pointcloud-master
+    ./autogen.sh && ./configure
+    make && sudo make install
+    cd $TRAVIS_BUILD_DIR
 fi
 
 gcc --version
--- a/scripts/ci/script.sh
+++ b/scripts/ci/script.sh
@@ -33,7 +33,7 @@ cmake \
     -DBUILD_PLUGIN_OCI=OFF \
     -DBUILD_PLUGIN_P2G=OFF \
     -DBUILD_PLUGIN_PCL=OFF \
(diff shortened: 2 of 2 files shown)


strk:
Ready for merge as soon as Travis goes green (should be green already)

hobu:
hmm, it failed. Did it run out of cpu time?

gadomski:
/me restarts the build

We've been over this ground before, but it seems like we're again bumping up against the "how much effort to we go through to CI the plugins?" This patch doesn't seem too bad (just `apt-get`ting the postegres server and installing pgpointcloud from source) but there's a couple of other plugins that might be a similar level of complexity that we *aren't* testing (e.g. sqlite, p2g, hexer...).

I wonder how much pain would we inflict upon ourselves by making a `PDAL/PDAL-plugins` repository, where we could stash all plugins that we *don't* CI (e.g. rivlib, mrsid...). Just musing...the seemingly arbitrary way we test some plugins but not others makes my anal-rentitive side twitch a bit :-P...

strk:
It now passed. Can't tell about a separate repo.

chambbj:
@gadomski This probably deserves it's own issue/thread, but I tend to agree with testing more plugins if we can. I fear that separating them might lead to neglect. I know that some of the closed plugins are already outdated. This probably settles down some after a formal release, and is certainly mitigated if we set up a corresponding CI configuration for the separate plugins repo.

If we keep the plugins in PDAL/PDAL, I think we may be able to save some Travis clock cycles by doing the following:

1) Pretty sure we no longer need to build cmake and ninja from source. I'm pretty sure this explains the delta between gcc and clang build times.
2) Though we lose some granularity, we could combine the optional components ON/OFF tests in the build script for each compiler. This would really just save on machine setup. I'm guessing we could also make sure that if optional components OFF fails, we do not proceed to optional components ON.

Are these steps alone enough for us to turn on more dependencies? Maybe. PCL without a current distribution will probably always time out. Even with binaries, it likely won't be built the way we need it.

strk:
Could it make sense to also build with optimizations turned off ? (-O0)

gadomski:
Moved plugin discussion to #851. This PR could be merged, or could be delayed until we come to some decision on #851. I have no preference either way.

chambbj:
Assuming we changed nothing with the plugin testing strategy for now, it looks like there is very little overhead in turning this one on.

PR #863, opened 2015-04-02
add ability for pdal::Kernel to add hidden switches, use to implement --pointcloudschema output for pdal info and pdal pipeline

hobu:


sambaldwin05:
Need to have `--pointcloudschema` available in `pdal info` as well.

hobu:
It's there in `pdal info`, but STDOUT doesn't work as an output type.

PR #874, opened 2015-04-10
Updated vagrant provisioning to use common GeoWave jace bindings

jwomeara:


jwomeara:
There is now a common set of Jace C++ headers & library for GeoWave.  We thought this would be better than having to first clone PDAL, then build GeoWave, then build PDAL.  This will segue into rpm & deb packaging at some point in the not too distant future.  

I also updated the GeoWave plugin to handle errors when trying to connect to Accumulo.  Also, we ran some benchmarks and the FeatureDataAdapter is faster than the FeatureCollectionDataAdapter, so we changed the default for the reader & writer.

hobu:
I will just blindly merge this, as I don't have a Vagrant instance alive right now to test. Please report back with a new PR if there's still issues.

jwomeara:
Will do!  I'll test a fresh Vagrant install once more tonight.  Thanks!

PR #886, opened 2015-04-25
Load only one plugin when guessing plugin names

gadomski:
In `PluginManager::guessLoadByPath`, the driverName was not being entirely used when loading plugins. Only the plugin type (filter, reader, writer) was used. This patch checks the actual name of the driver (e.g. hexbin) against the filename of the library, and only loads the plugin if the name matches.

Since a plugin only needs to be loaded once, the return value of the `loadByPath` call is returned from `guessLoadByPath`. If no plugins are found, the `guessLoadByPath` now returns -1.

cc @chambbj

### DIFF
--- a/src/PluginManager.cpp
+++ b/src/PluginManager.cpp
@@ -246,6 +246,12 @@ int32_t PluginManager::guessLoadByPath(const std::string& driverName)
             if (ext != dynamicLibraryExtension)
                 continue;
 
+            std::string stem = full_path.stem().string();
+            std::string::size_type pos = stem.find_last_of('_');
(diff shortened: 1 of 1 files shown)


chambbj:
Good catch. I obviously overlooked this. @connormanning has some edits in a branch that I'm reviewing as well. It may be worthwhile to sync the two efforts prior to merging.

gadomski:
Agreed. Probably wouldn't be hard for @connormanning to just pull in this commit to his work. Connor if you do that, just close this PR? Thanks.

hobu:
Is this PR dead @connormanning?

connormanning:
The fix here has been merged into #887, which made some other changes in this area.

PR #893, opened 2015-05-04
Reading "off end of file" when las header point count is too big

gadomski:
If a las header says that the file has more points than it really does, the reader will zero-fill the `PointTable` up to the header-specified number of points. This changed in c63f198db2f08caa2a47b34f4c7409a248e22a12 — previously, we stopped reading at the end of file. Since we moved to chunked reading of the lasfile, we're essentially reading off the end of the file into our chunk buffer.

I'm opening this pull request with a test that illustrates the issue; we happened to have a file among our test data that advertised 1065 points in its header but only has 1064 points in the file itself (`las/1.2-with-color-clipped.las`). The "offending" read is [here](https://github.com/PDAL/PDAL/blob/master/io/las/LasReader.cpp#L622). The summary of the problem is presented below, line numbers from `io/las/LasReader.cpp`.

```cpp
count = std::min(count, getNumPoints() - m_index);  // L556, getNumPoints() spits out the header's point count
size_t bufsize = std::min<size_t>((point_count_t)1000000,  count * pointByteCount);  // L587, L588
std::vector<char> buf(bufsize);  // L589
point_count_t blockpoints = buf.size() / ptLen;  // L619
m_istream->read(buf.data(), blockpoints * ptLen);  // L622, possible read off end of m_istream
```

I haven't put in a fix because I'm not sure how we want to handle this scenario.

cc @abellgithub b/c git-blame

### DIFF
--- a/test/unit/io/las/LasReaderTest.cpp
+++ b/test/unit/io/las/LasReaderTest.cpp
@@ -392,3 +392,21 @@ TEST(LasReaderTest, callback)
     EXPECT_EQ(count, (point_count_t)1065);
 }
 
+
+// The header of 1.2-with-color-clipped says that it has 1065 points,
+// but it really only has 1064.
+TEST(LasReaderTest, LasHeaderIncorrentPointcount)
+{
+    PointTable table;
+
+    Options readOps;
+    readOps.add("filename", Support::datapath("las/1.2-with-color-clipped.las"));
+    LasReader reader;
+    reader.setOptions(readOps);
+
+    reader.prepare(table);
+    PointViewSet viewSet = reader.execute(table);
+    PointViewPtr view = *viewSet.begin();
+
+    EXPECT_EQ(1064, view->size());
+}


hobu:
@gadomski That file is very intentionally in the suite to test a somewhat common issue of having a misrepresented point count.

gadomski:
Well in that case it worked! I was using that file just by chance to test the ply reader->writer round trip and found out that our las reader was adding in that zero point for that file.

One crappy solution would be to check the stream for a bad state after the batch read. If the stream is bad, go back and read the points one by one into the buffer till you get to the end. Better ideas?

abellgithub:
Well, we can certainly get the file size and use that.  Part of the point of the batched read was to avoid checking stream state.  I'll think about it.

hobu:
How bad is it to check after each batch? And if you're then eof() or bad(), compare where you actually are with where you should be?

abellgithub:
Probably not bad, but is it any better than using the file size?  If you're OK with not reading every possible point if there's a mismatch, then just stopping if we can an eof() seems fine, but to have to add logic to go back look for the end sounds ugly.

hobu:
The end isn't the end in LAS 1.4 with eVLRs and such, so this scenario is probably only for 1.3 files and below because the file size comparison is no good there.

Old liblas used to compute an expected point count and complain if it looked like there wasn't going to be enough bytes. Annoying.

It is a scenario that should make the log() at some WARNING level.

Maybe just stopping on eof() is enough...

hobu:
#902 merges this test with an attempt at accounting for it by checking eof(). Further effort on this is now there.

PR #894, opened 2015-05-04
Add a ply writer

gadomski:
Note that the `ready->write->done` sequence is a little corrupted due to the fact that I can't change the header's point count once I start writing values. Instead in `write` I save the points to an internal `PointView` and then actually write them in `done`. Awkward, but doesn't seem like this is the only time this will come up (there might very well be other writers that have trouble going back and changing the point count in the header once points have started being written).

Will let this pass tests then will merge unless objections are raised.


abellgithub:
Perhaps we should provide a way to get the active views from the table, or something like that.

Issue #897, opened 2015-05-07
Intended semantics of BOX3D?


mpgerlek:
```
    BOX3D boxA(0.0, 0.0, 100.0, 100.0);  // a "2D" box
    BOX3D boxB(50.0, 50.0, 3.1, 51.0, 51.0, 3.14);  // a "3D" box, wholly inside boxA
    
    bool aContainsB = boxA.contains(boxB); // returns false??
    bool bContainsA = boxB.contains(boxA); // returns false
    bool aContainsA = boxA.contains(boxA); // returns true
    bool bContainsB = boxB.contains(boxB); // returns true
```

Is this intended behavior?

I expected that `aContainsB` would have been true, but I believe that since a "2D" Box3D has is explicitly constructed to have an invalid Z, the result is actually false.

hobu:
Yes, this is probably busted. I think I've only ever done 2D contains calls in the context of database drivers, so this isn't well tested code.

abellgithub:
I'd vote this as user error.  I just don't think it's defined sufficiently (semantically speaking) to yield proper values.

Perhaps we should disable the 2D ctor?

mpgerlek:
It's arguable either way.

Do you consider a 2D box to be a point or an infinite column?

.mpg

On May 7, 2015, at 4:11 PM, Howard Butler <notifications@github.com<mailto:notifications@github.com>> wrote:


Yes, this is probably busted. I think I've only ever done 2D contains calls in the context of database drivers, so this isn't well tested code.

abellgithub:
Since it's arguable either way, it seems wrong.  Can we just make a BOX2D type or something?

mpgerlek:
'Point2d' would be better?

There's probably a library we can use for this, rather than rolling our own..?

Either way, such a change is liable to have to touch a lot code.

For now I'll just add the test case and put a bit of documentation in there.

.mpg

On May 7, 2015, at 4:14 PM, Andrew Bell <notifications@github.com<mailto:notifications@github.com>> wrote:


Since it's arguable either way, it seems wrong. Can we just make a BOX2D type or something?

abellgithub:
Well, I don't care about the name.

As far as rolling our own, the code is trivial (it's essentially already there, too).

PR #900, opened 2015-05-08
added note and test case for #897

mpgerlek:
Also took the liberty of moving the 2D ctor from cpp file to the hpp
file, to be consistent with the other ctors.

### DIFF
--- a/include/pdal/util/Bounds.hpp
+++ b/include/pdal/util/Bounds.hpp
@@ -79,7 +79,13 @@ public:
         minz(box.minz), maxz(box.maxz)
     {}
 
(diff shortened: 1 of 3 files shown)


chambbj:
@mpgerlek I believe your Appveyor failure has to do with the 2D ctor move. It needs to be in the cpp file where `HIGHEST` and `LOWEST` are defined.

mpgerlek:
This should not be merged due to MSVC fail:

https://github.com/PDAL/PDAL/commit/9b4f3629a1543b9fafd789de60831c7dbe2241a0

mpg will submit new PR

PR #902, opened 2015-05-08
Read mislabeled LAS files where the header says there are more

hobu:
points than the file actually has

### DIFF
--- a/io/las/LasReader.cpp
+++ b/io/las/LasReader.cpp
@@ -111,7 +111,7 @@ void LasReader::processOptions(const Options& options)
 static PluginInfo const s_info = PluginInfo(
     "readers.las",
     "ASPRS LAS 1.0 - 1.4 read support. LASzip support is also \n" \
-        "enabled through this driver if LASzip was found diring \n" \
+        "enabled through this driver if LASzip was found during \n" \
         "compilation.",
     "http://pdal.io/stages/readers.las.html" );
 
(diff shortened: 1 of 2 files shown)


gadomski:
:+1:

io/las/LasReader.cpp
```diff
@@ -619,7 +619,17 @@ point_count_t LasReader::readFileBlock(std::vector<char>& buf,
     point_count_t blockpoints = buf.size() / ptLen;
 
     blockpoints = std::min(maxpoints, blockpoints);
+    if (m_istream->eof())
+        throw pdal::invalid_stream("stream is done");
```
abellgithub: Not sure I like throwing an exception here.  Also, I've been trying to get rid of all the subclassed exceptions that we're not using anyway (of course I haven't TOLD anyone that I've been doing this.)

hobu: I threw that because the thing calling it was explicitly catching it and dropping it on the floor right away. 

This seemed the least disruptive, but if you have other ideas, please push them to this branch.

abellgithub: OK.  Looks fine.

hobu: >  Also, I've been trying to get rid of all the subclassed exceptions that we're not using anyway (of course I haven't TOLD anyone that I've been doing this.)

Admittedly not much thought has been put in that other than to subclass std::runtime_error, which is what most end up having to catch (or the top-level PDAL-specific one, pdal::error).

PR #906, opened 2015-05-12
Add a Gridder filter and a Grid kernel

oscarmartinezrubi:
I have added a filter (and an app/kernel) similar to the Splitter but it grids the data according to a grid fully defined in the options (extent and number of cells)
I have also added a test class which works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,6 +102,7 @@ test/data/filters/range_z_classification.xml
 test/data/filters/reproject.xml
 test/data/filters/sort.xml
 test/data/filters/splitter.xml
+test/data/filters/gridder.xml
 test/data/filters/stats.xml
 test/data/hole/crop.xml
 test/data/icebridge/pipeline.xml
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -25,6 +25,7 @@ include_directories(
     ${PROJECT_SOURCE_DIR}/kernels/random
     ${PROJECT_SOURCE_DIR}/kernels/sort
     ${PROJECT_SOURCE_DIR}/kernels/translate
+    ${PROJECT_SOURCE_DIR}/kernels/grid
 )
 
 #------------------------------------------------------------------------------
--- a/doc/apps.rst
+++ b/doc/apps.rst
@@ -21,6 +21,7 @@ application currently contains six commands:
 * :ref:`translate <translate_command>`
 * :ref:`view <view_command>`
 * :ref:`split <split_command>`
+* :ref:`grid <grid_command>`
 
 Applications are run by invoking the *pdal* application along with the
 command name:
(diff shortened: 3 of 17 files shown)


abellgithub:
I don't understand this.  I'm not sure what this is accomplishing that wouldn't be achieved by setting an origin for the current splitter filter.  Can you explain?

oscarmartinezrubi:
I need it because if you run it for multiple files (so different PDAL runs) with splitter I get an irregular grid but I would like a regular grid as end result .

In the splitter, where each point goes depends on:

```
int xpos = (inView->getFieldAs<double>(Dimension::Id::X, idx) - xOrigin) / m_length;
int ypos = (inView->getFieldAs<double>(Dimension::Id::Y, idx) - yOrigin) / m_length;
```
This means that the edges of the final parts/cells are not defined. So, if you have two files and run the splitter in both of them the produced cells form an irregular grid.
Also the m_length is more or less capacity (if density is similar) but there is no way to control the number of generated cells.

What I want is a regular grid, that the edges of each cell are defined (then not all of them have same number of points), so when running it for multiple files I can combine the generated files and I will have a nicely looking grid, i.e.:

```
int xpos = (inView->getFieldAs<double>(Dimension::Id::X, idx) - min_x) * num_x / (max_x - min_x);
int ypos = (inView->getFieldAs<double>(Dimension::Id::Y, idx) - min_y) * num_y / (max_y - min_y);
```

The output file names are a bit annoying and I still need to so some trick out of PDAL to know exactly where each file generated by PDAL should go in my regular grid (to change that in PDAL was a bit more messy so I did not do it)

abellgithub:
1) If you want a min/max, use the existing range filter.
2) If you have can specify an origin in the splitter (I have the code done, but a test fails and I had to work on other things), all the PointViews will create cells with the same origin, offset by length in either direction.

I'm not sure what you mean about controlling the number of generated cells.  The only way to control the number of cells is by dropping points or changing the length, unless you're OK with non-regular cells, which it seems you aren't.

oscarmartinezrubi:
I want to have a grid but I want to know/control the spatial extent of each cell and with the function used in splitter I can not do that.

abellgithub:
I just pushed the origin support to the splitter filter.  As far as I can tell, this will provide the same capability as the gridder.  The difference is only in the WAY that the grid is specified and a limitation that the cells are always square.  In the gridder, the user specifies a min max and cell count.  In the splitter, the user specifies an origin and cell length (X and Y are the same length).  To get the same effect, one simply needs to do the division ((max - min) / count -> length) and pass min as origin.  Does this make sense.  I still need to integrate the origin support into the split kernel.

Am I missing something?

oscarmartinezrubi:
I think it makes sense. I saw that you changed the length to be a double which was the thing that was making me doubt...and in that case I think I can do the same as with gridder.

I would like to test with your suggestion and see if I find any reason to keep the gridder. Let you know.

thanks

hobu:
Is this PR no longer valid?

oscarmartinezrubi:
I think it is not. However the split kernel is not updated yet so I have
not yet tested that it does the same as the grid kernel. I can update the
split kernel if you guys want

Is this PR no longer valid?

—
Reply to this email directly or view it on GitHub
<https://github.com/PDAL/PDAL/pull/906#issuecomment-103226486>.

hobu:
Overtaken by #959

PR #911, opened 2015-05-12
misc SQLite refactoring and cleanups

mpgerlek:
- regularized the log() function
- added getVersion functions
- lifted a couple spatialite operations up into the header
- fixed a bug when column type is null (hit by spatialite version call)
- passing “1” to the spatialite init function now, which yields a big
speedup (http://hub.qgis.org/issues/8340)


mpgerlek:
Have fixed @abellgithub's issues.

PR #913, opened 2015-05-13
Add support for owning stages in the StageFactory.

abellgithub:


hobu:
What prompts this addition? Clarification of ownership, obviously, but what situations merit not just always letting the StageFactory own them?

mloskot:
The stages are created by the stage-defined named constructors (`create` functions),
then they are deallocated by the default deleter of `unique_ptr`, instead of corresponding named destructors (`destroy` functions). (Unless, I'm missing something, the `destroy` functions are hardly used.)
I see potential problems:
* allocator/deallocator mismatch - a plugin developer has freedom to allocate stage using whatever allocator he desires in the `create`, no?
* allocation/deallocation across shared module boundaries - nasty issue on Windows, in particular.

abellgithub:
Some plugin stuff maybe needs a second look.  Perhaps we should discuss later.  I prefer simple and I'm not sure the create(), destroy() functions are necessary at all.  They came with the plugin infrastructure that was pulled from elsewhere.

mloskot:
@abellgithub The functions may still be needed, especially for shared plugins, to allow smooth allocations across module boundaries.

mloskot:
For the archive: we've just discussed this issue with @abellgithub and @hobu at the Paris sprint and agreed on a few possible simplifications:

* `create`/`destroy` functions may go and the `PluginManager::l_createObject` may simply `new` to create a stage object.
* `StageFactory::createStage` boolean flag may also go, then the `StageFactory` always owns stages, or, if necessary (use cases to be confirmed) the flag semantic will flip and `StageFactory::destroyStage` function is needed to ensure `delete` used corresponds exactly to the allocator used by `new` in the `createStage`.

Issue #915, opened 2015-05-14
Options::remove() missing


mpgerlek:
Options::remove(string) is declared but doesn't appear to be defined anywhere.

PR #919, opened 2015-05-15
added missing Options::remove() for #915

mpgerlek:


abellgithub:
Do we need this?  Can't one just fetch the existing option by reference and reset the value?

mpgerlek:
Is there a fetch-by-ref? I had looked, but did not see one --but then I saw that there was a remove() already declared in there so I decided to happily just use that.

mpgerlek:
(If there is a fetch-by-ref, /me will change this PR to delete the remove() declaration.)

hobu:
Is this one dead?

mpgerlek:
No, it's still a thing.

.mpg

On Jun 7, 2015, at 11:28 AM, Howard Butler <notifications@github.com<mailto:notifications@github.com>> wrote:


Is this one dead?

abellgithub:
I thought it was unnecessary given the usage.  I don't think anyone calls it.

abellgithub:
Handled elsewhere.

PR #928, opened 2015-05-26
Sqlite error handling

### LINKED ISSUE #927: Improve SQLiteCommon error handling
The sqlite wrapper class should handle all possible return codes from the underlying sqlite object, and it should handle them in the same way.

There is a missing finalize() call at the end of insert(), which causes random bits of havoc when mixing insert() and query() operations, since m_statement is shared between them. We should explicitly null out m_statement, just to make durned sure it's okay to reuse.

mpgerlek:
fixes #927

- handle error returns consistently
- always call finalize()
- fix test case to flush files

### DIFF
--- a/plugins/sqlite/io/SQLiteCommon.hpp
+++ b/plugins/sqlite/io/SQLiteCommon.hpp
@@ -47,31 +47,6 @@
 namespace pdal
 {
 
-
-    class sqlite_driver_error : public pdal_error
-    {
-    public:
-        sqlite_driver_error(std::string const& msg)
-            : pdal_error(msg)
-        {}
-    };
-
-    class connection_failed : public sqlite_driver_error
-    {
-    public:
-        connection_failed(std::string const& msg)
-            : sqlite_driver_error(msg)
-        {}
-    };
-
-    class buffer_too_small : public sqlite_driver_error
-    {
-    public:
-        buffer_too_small(std::string const& msg)
-            : sqlite_driver_error(msg)
-        {}
-    };
-
 class column
 {
 public:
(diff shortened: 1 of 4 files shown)


abellgithub:
A bunch of the error messages don't really say what's wrong, just where the error occurred.  OK for a programmer, but less OK for others.  Perhaps something in English that might make sense along with something to help when creating a ticket?

mpgerlek:
I didn't intend to change any of the error messages -- which ones are you talking about?

abellgithub:
Maybe they got moved and I didn't realize.  I just saw stuff like: "sqlite3_step" and "sqlite3_finalize".  Not sure the best thing to display, though.

mpgerlek:
Okay, yes, I added those to indicate which functions were causing the problem - admittedly a programmer-biased view, but more than we had before.

If you can point me to an example of the preferred style for error messages, I'd be happy to upgrade these ones.

abellgithub:
Well, you'd probably know better.  Maybe something like "Error inserting point tile into SQLite3 database (sqlite3_step: <error_code>)"  I'm just making stuff up.  If there are any errors to anticipate along with solutions, that's always nice, but I don't know if you've got that in this case.

mpgerlek:
@abellgithub See if you like that better... If we want to take this further, we should first define a PDAL-wide style guide for error handling, and then I'll conform to it.

(side note: I can't say things like "error inserting point" because this SQLite class is not specific to any kinds of data, e.g. points.)

abellgithub:
I'm not sure I understand.  What else might we be storing in the DB beside points?

mpgerlek:
Whatever you want to store... The SQLiteCommon class is a generic wrapper around sqlite3 database operations. It doesn't know anything about any kind of data or schemata.

The SQLiteReader/Writer classes use the SQLiteCommon class to work with databases containing points using a schema system Howard defined. My Rialto classes, developed later, use the SQLiteCommon class to do a sqlite database with a different schema.

PR #948, opened 2015-06-25
Hexer on travis

gadomski:
I've updated the PPA build of hexer to use gcc 4.8, so hopefully tests pass ok. Some things of note:

- The PPA package is called `hexboundary` instead of `hexer` because `hexer` was already taken
- One commit removes a bit of testing from the HexbinFilterTest, since apparently hexer is nondeterministic with regards to the ordering of its output, leading to test failures for functionally identical outputs. The Right Way to do this would be to compare the actual output geometries, but, well, I didn't do that.


abellgithub:
On Thu, Jun 25, 2015 at 4:48 PM, Pete Gadomski <notifications@github.com>
wrote:

> I've updated the PPA build of hexer to use gcc 4.8, so hopefully tests
> pass ok. Some things of note:
>
>    - The PPA package is called hexboundary instead of hexer because hexer
>    was already taken
>    - One commit removes a bit of testing from the HexbinFilterTest, since
>    apparently hexer is nondeterministic with regards to the ordering of its
>    output, leading to test failures for functionally identical outputs. The
>    Right Way to do this would be to compare the actual output geometries, but,
>    well, I didn't do that.
>
> I'm aware of the test issue.  It would be fixed if it was easy, but I
didn't want to incur overhead to make the tests pass.  Other stuff would be
more work and nobody was complaining :)

-- 
Andrew Bell
andrew.bell.ia@gmail.com

gadomski:
Yeah, my instinct was to file it under "something for someone to clean up someday when they have oodles of time."

PR #959, opened 2015-07-17
Add origin selection in split kernel (it was already done for the filter but not for the kernel)

oscarmartinezrubi:
Now the split kernel/app can be used with selection of X and Y origin. 

So, there are 3 modes to run the split kernel now as shown in the 3 examples below where the extent of the initial file is minX=85000 minY=446250 maxX=86000 maxY=447500:

## With origin selection and length which now has to be a double
``
pdal split -i /pak2/usrdata/ahn2/benchmarks/000020m/ahn_bench000020.las -o test_pdal1.laz --origin_x 84000 --origin_y 445000 --length 375.0
``
pdal1

## Without origin selection
``
pdal split -i /pak2/usrdata/ahn2/benchmarks/000020m/ahn_bench000020.las -o test_pdal3.laz --length 375.0
``
pdal2

## With capacity (in this case the Chipper filter is used)
pdal split -i /pak2/usrdata/ahn2/benchmarks/000020m/ahn_bench000020.las -o test_pdal3.laz --capacity 1000000 

pdal3

### DIFF
--- a/filters/splitter/SplitterFilter.cpp
+++ b/filters/splitter/SplitterFilter.cpp
@@ -65,7 +65,7 @@ void SplitterFilter::processOptions(const Options& options)
 Options SplitterFilter::getDefaultOptions()
 {
     Options options;
-    Option length("length", 1000, "Splitter length");
+    Option length("length", 1000.0, "Splitter length");
     options.add(length);
 
     return options;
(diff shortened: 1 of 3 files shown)


abellgithub:
Looks good!  Thank you.

oscarmartinezrubi:
No problem!
Now I can definitively drop the Gridder filter ;)

PR #965, opened 2015-08-11
A bunch of junk

hobu:
My self loathing requires that I at least dump it into a branch and make a PR instead
of just slamming it into master


filters/crop/CropFilter.cpp
```diff
@@ -113,10 +115,19 @@ void CropFilter::processOptions(const Options& options)
...
 
+            if (!m_assignedSRS.empty())
+            {
+                // FIXME: This should use the PointTableRef's SpatialReference
+                pdal::gdal::SpatialRef in_ref(m_assignedSRS);
+                pdal::gdal::SpatialRef out_ref(getSpatialReference().getWKT());
+                pdal::gdal::Geometry geom(poly, in_ref);
+                geom.transform(out_ref);
+                poly = geom.wkt();
+            }
```
abellgithub: I would think we should also transform a bounds box.  Also, we should probably put this bit somewhere else, as I would think it's used in other places, too.

plugins/python/plang/Environment.cpp
```diff
@@ -66,14 +66,37 @@ EnvironmentPtr Environment::get()
...
+void wake_up_numpy()
+#endif
+{
+    import_array();
+#if (PY_VERSION_HEX >= 0x03000000)
+    return 0;
+#else
+#endif
+}
+
```
abellgithub: I think I cleaned all of the dependency stuff up so that you don't need to have this function.

plugins/python/plang/Environment.cpp
```diff
@@ -66,14 +66,37 @@ EnvironmentPtr Environment::get()
...
     char *path = Py_GetPath();
-    import_array();
+#endif
+
+    // import_array got macro'ized and now returns
+    // a value in Python 3+. We can't return stuff in
+    // constructors and we want to import the redirector
+    // anyway
+    wake_up_numpy();
     PyImport_ImportModule("redirector");
```
abellgithub: Don't think this is necessary.

src/Kernel.cpp
```diff
@@ -460,26 +460,26 @@ void Kernel::setCommonOptions(Options &options)
...
             {
-                options.add<double >("offset_x", offsets[0]);
-                options.add<double >("offset_y", offsets[1]);
-                options.add<double >("offset_z", offsets[2]);
+                options.add<std::string>("offset_x", offsets[0]);
+                options.add<std::string>("offset_y", offsets[1]);
+                options.add<std::string>("offset_z", offsets[2]);
             }
         }
     }
```
abellgithub: I don't understand this change.  Was there a problem?  We should parts to double just fine, right?

plugins/python/plang/Environment.cpp
hobu: I'm not sure what you mean? We still need to have import_array happen somewhere in the plugin, right? The problem is Python 3's import_numpy is a macro that just returns crap, and our clang settings error out in that situation. I put it in its own function in case we ever need to do something else.

src/Kernel.cpp
hobu: If the options are string, you can pass in `auto` via command line.

plugins/python/plang/Environment.cpp
abellgithub: Well, if nothing else we should move it back to Environment, because it's really unclear why it's here.  Should be able just to do: (void)import_array(), I would think and lose all the ifdef stuff.

src/Kernel.cpp
abellgithub: Yes.  Forgot that.

plugins/python/plang/Environment.cpp
hobu: import_array has a different signature for Python 3

hobu:
this stuff has all been merged.

PR #974, opened 2015-08-17
Add DerivativeWriter

chambbj:
Adds a derivative writer to compute slope, among other things.

This PR also vendors the header-only Eigen library. This shouldn't be necessary, but I was having trouble getting Travis to detect Eigen otherwise. We can remove it in the future if need be.


hobu:
:+1: to merge whenever you are ready.

chambbj:
Okay, I'll merge. We may enhance it a bit later per our slack discussion, but this will be a good starting point.

Issue #962, opened 2015-08-03
Add omni kernel (was "Add colorization kernel")


hobu:
It would be handy to be able to specify

`pdal color --input myfile.las --raster myfile.tif --output output.las`

to drive the `filters.colorization` filter.

abellgithub:
Seems like we could just do something like:

`pdal filter -filter=filters.colorization --filters.colorization.raster myfile.tif ...`

Maybe this would suffice for other existing kernels?  Would have to think on it.

hobu:
I like this. A kernel that dynamically defines one reader, one filter, and one writer (plus all options being able to be created via command line) would be quite powerful.

chambbj:
#975 created in support of this, but I haven't tried colorization specifically just yet...maybe tomorrow morning.

abellgithub:
@chambbj Can we close this?

chambbj:
@abellgithub Yes, closing.

PR #975, opened 2015-08-17
Create Omni kernel

chambbj:
Just specify `--input` and `--output` for the input/output file names, and `--filter` for the filter type you'd like. Substitution can be used to modify filter-specific parameters.


abellgithub:
Isn't this really the same as the translate kernel without the special options that the translate kernel has?

chambbj:
Similar I guess. Though I don't remember translate being able to
dynamically specify the filter as was suggested in #962.

abellgithub:
On Mon, Aug 17, 2015 at 2:52 PM, chambbj <notifications@github.com> wrote:

> Similar I guess. Though I don't remember translate being able to
> dynamically specify the filter as was suggested in #962.
>

No.  Just wondering if it should.  We could perhaps get rid of the
specialized stuff.

-- 
Andrew Bell
andrew.bell.ia@gmail.com

chambbj:
Closing for now. Expect to reopen in the future.

PR #980, opened 2015-08-25
Update translate kernel to accept any supported filter or chain of filters

chambbj:
This replaces #975

### DIFF
--- a/doc/apps.rst
+++ b/doc/apps.rst
@@ -8,21 +8,22 @@ PDAL Applications
 :Contact: howard at hobu.co
 :Date: 6/1/2015
 
-PDAL contains consists of a single application, called *pdal*. The `pdal`
+PDAL contains consists of a single application, called ``pdal``. The ```pdal``
 application currently has the following subcommands:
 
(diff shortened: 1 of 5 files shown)


msmitherdc:
@chambbj should this syntax work?

```
pdal translate -i mypc.laz -o mydem.tif --writer=writers.p2g  --writers.p2g.grid_dist_y=0.5 --writers.p2g.grid_dist_x=0.5 --writers.p2g.output_format=tif --writers.p2g.output_type=idw --filter=filters.range --filters.range.dimension=classification --filters.range.dimension.equals=2
```

chambbj:
@msmitherdc The nested options stuff does not work currently. I would consider this a separate issue though, as the translate kernel doesn't do the recursive parsing of the stage options itself, and you shouldn't expect this to work even in master. It needs to be fixed in the [Kernel class](https://github.com/PDAL/PDAL/blob/b5e548087dc081c705ce0394a06d28cca08cb09f/include/pdal/Kernel.hpp#L129). I don't know enough about how options and nested options work at the moment to put together a quick fix for this.

Issue #754, opened 2015-02-08
PointStream complement to PointBuffer


hobu:
It is a frequent use case that folks use PDAL just for reading data. They might use one or two filters inline with operations, but that's it.

The current model requires reading the entire set of data into memory due to the fact that some filters like `filters.chipper` require the entire set to be able to begin computing. Our entire model currently satisfies this somewhat edge scenario. 

What if our API could default to supporting streaming operations when the entire list of intermediate filters (and maybe writers too) advertise that they can support it? Any filters would advertise their 
"streamability", and if any filter in a pipeline was not streamable, the stream `Stage::execute()` method would throw an exception.

```
Stage* reader = OpenReader(filename);
Stage filter = MakeReprojectionFilter; 
reprojection->setOptions(opts); 
reprojection->setInput(reader);
PointContextRef ctx;
filter->prepare(ctx);

PointStream strm(ctx);

// Throws if any stages cannot be read as a stream
strm = filter->execute(strm);

```

gadomski:
:+1:

Just to make sure I'm understanding, "streamable" stages, e.g. filters, would have a method something like this?

```cpp
virtual void filter(PointStream& ptstream);
```

And the `PointStream` would support some an iterator interface that would let you do this?

```cpp
for (auto it = ptstream.begin(); it != ptstream.end(); ++it)
{
   *it; // should be a point reference or something?
}
```

mpgerlek:
+1

This is similar to what we had back in the day, when we proceeded the data in continuously-emitted chunks/strips/blocks. Makes PDAL palatable for doing stuff like format-to-format translations (e.g. LAS to LAZ) or for reprojections (LAS/WGS84-to-LAS/somethingorother).

-mpg

On Feb 8, 2015, at 11:52 AM, Howard Butler <notifications@github.com<mailto:notifications@github.com>> wrote:


It is a frequent use case that folks use PDAL just for reading data. They might use one or two filters inline with operations, but that's it.

The current model requires reading the entire set of data into memory due to the fact that some filters like filters.chipper require the entire set to be able to begin computing. Our entire model currently satisfies this somewhat edge scenario.

What if our API could default to supporting streaming operations when the entire list of intermediate filters (and maybe writers too) advertise that they can support it? Any filters would advertise their
"streamability", and if any filter in a pipeline was not streamable, the stream Stage::execute() method would throw an exception.

Stage* reader = OpenReader(filename);
Stage filter = MakeReprojectionFilter;
[...]

PR #982, opened 2015-09-03
Sample chunked reader-only interface.

connormanning:
Here's a sample of what I think might be nice for API users - a simple chunked reading API.  See `reader.hpp` for the only public API change proposed.

It allows reading data from a file with a simple API - all you need is a `pdal::Reader` and an `onData` function to act on a `pdal::PointView` to consume your points.  Behind the scenes, the memory for the points will be reclaimed/reused after your data function completes.  No overloading of PDAL types necessary.

This operates in a reader-only fashion, but it's easy to snap together the rest of a full pipeline.  For example, one of my use-cases creates a Reprojection filter and applies it to the chunks that come back.

I've provided a sample usage program [here](https://gist.github.com/connormanning/3c1205b7602ae6250ba0) that shows the current point-at-a-time callback API as well as the chunked version I'm proposing.  I've seen between 9 and 11% speedups using the chunked version between this and various other testing.

Pros:
- Don't need to maintain all the points from the file in RAM.
- Fast.  Only one allocation, so less overhead there - also probably better caching since the same buffer is reused.
- No need to override any PDAL classes to get basic chunked functionality that I think many API users would want.
- Point-at-a-time semantics are available through this API as a special case (if `chunkBytes = 1` or is less than `PointLayout::pointSize``), without holding onto the points that have already been read.

Cons:
- No control over the PointTable, since it uses a private/internal type to allow for memory reuse and chunking stuff.
- Probably others.

Thoughts?


abellgithub:
What does putting this code in the Reader do for you?  Couldn't it just sit outside?  Isn't it just an example of a user implementation using a custom PointTable and callback?  I probably didn't read carefully enough.

Not sure that I agree that many users would want anything chunked.  What we've seen is that they want to read and process a single point at a time, which means even less copying that what's going on here.

connormanning:
I would be fine with making the derived PointTable public, but I figured it was mostly an implementation detail so I privatized it.  For the chunking not being wanted, there are multiple asks for it in #754.  Even without chunking, since this method supports point-at-a-time, you still get the performance benefits.  Allocation overhead is gone, and caching is improved since it's always working with the same few bytes of memory.

There's also the current thread on the mailing list looking for ways to better manage memory.  That thread talks about LAS files of 10+ billion points - that's 40 GB of allocation *just* for the PointView's `m_index` overhead without even considering the point data (hence the private addition of `PointView::clear`).  Add in the actual point data and you're looking at ~400 GB of live memory.  That seems a little unreasonable if you're only looking at tens of bytes at a time.

As for it being a method on a Reader rather than requiring users to implement it separately, I think that it provides a meaningful and simple API for this common use case.  No need to construct a custom PointTable, mess about with options, set your callback, call prepare/execute with that same PointTable, make sure your CB has access to your CustomPointTable (which you now have to care about in addition to the PointView that you *actually* care about since you need to clear the table), make sure you call your `CustomPointTable::clear` method in the read CB (which still leaves dangling unusable indices in the PointView you don't have access to clear), wonder if you just hosed your PointView by clearing the PointTable that's backing it, and then convince yourself it's probably fine since your test app doesn't segfault...

Or skip that and get faster speeds with negligible memory usage in a few lines:

```
void onInit(PointTable& table) { }
void onData(PointView& view) {
    for (uint32_t i(0); i < view.size(); ++i) doStuffWithPoints();
}
LasReader r;
r.read("file.las", onInit, onData);
```

abellgithub:
Maybe I'm not getting it.  Other than the clear() function in PointView, is there any reason that all of this code can't be stuck in it's own source file?  That's my suggestion.  I understand the potential benefit to some users.

connormanning:
Moved `VectorPointTable` could to PointTable.hpp (see commit).  I had just added it to the anonymous namespace to keep it out of any public header.

Any recommendations on where the `Reader::read` functions would live if not on the Reader?  I do like them there because it seems like a natural method for a Reader to have, if the alternative is a free function that accepts a Reader as an argument, for example.

Could maybe rename them to `Reader::stream` to be more explicit and differentiate from the standard `read`.

abellgithub:
That's not what I meant.  Is there any reason that any of this need be in the PDAL library as such (with the exception of PointView::clear())?  The whole point of the callback/PointTable derivation is so that this kind of thing can be built without messing with the library.

Seems like if we had a "contrib" area or something like that, it would go there perfectly, as a package -- one file.  We could make a bunch of things like this.  You could make an explicit one that just does a single point and processes in the PointTable with no callback at all, for example.  That's what makes the most sense to me.

hobu:
> Is there any reason that any of this need be in the PDAL library as such

Arm's reach convenience. While it absolutely increases our API surface, without stuff like this, external applications looking to leverage PDAL have to build it all themselves. It is very much a PDAL user desire to have portions of the API oriented toward "give me the data from a reader and then leave me alone", and we need to have good support for that.

The danger with stuffing it in a "contrib" bucket is that it signals to potential users that it isn't first class to the library. Is what it uses always going to be supported? Is it always installed? Can I depend upon it working? Making it part of the proper library answers those questions and signals to users what the state of the approach is going to be going forward.

abellgithub:
>>Is there any reason that any of this need be in the PDAL library as such

>Arm's reach convenience. While it absolutely increases our API surface, without stuff like this,
external applications looking to leverage PDAL have to build it all themselves. It is very much a PDAL
user desire to have portions of the API oriented toward "give me the data from a reader and then leave me alone", and we need to have good support for that.

Well, I disagree with the first bit here.  I can see no reason to enlarge the API.  It's simply unnecessary.  I'm not asking anyone to build this themselves.  It's nice for us to provide it as a convenience.

>The danger with stuffing it in a "contrib" bucket is that it signals to potential users that it isn't first class to the library. Is what it uses always going to be supported? Is it always installed? Can I depend upon it working? Making it part of the proper library answers those questions and signals to users what the state of the approach is going to be going forward.

OK, but that's what it is.  If you don't like the name "contrib", well, we can stick it in a place with some name that makes it clear that it's endorsed and part of PDAL.  Stick it in some library that people can link in.  I thought the point was to encourage contribution.  If contribution means enlarging the API, then we're going to end up with a mess.  If there was a technical reason to enlarge the API, then fine, but there isn't.

Also, if you're going to use this, you have to know what you're doing.  It doesn't follow all the rules, and you can't use it with all the stages as we've described in the documentation.  For instance, you can't just attach this to the splitter and write tiles.  In that respect, it isn't "first class."  That doesn't mean it's not useful for people.

connormanning:
Even if we had a contrib area (or any other name), I wouldn't think this would belong there - personally I think it belongs in the core API.  "Give me `n` points at a time (where n may be 1), and **don't** use all my memory please" is such a fundamental task that I think it should deserves native API support, and I don't think this argument applies to other things that might belong in a 'contrib' area.

You mention it not being 'first class' functionality because you can't pipe the results of a streamed interface to any arbitrary PDAL component, since some special ones need all the points at once.  However I don't think that prohibits it from being first class *reader* functionality - which I think it is, and is why I stuck it on the reader class.

hobu:
> "Give me n points at a time (where n may be 1), and don't use all my memory please" is such a fundamental task that I think it should deserves native API support

:+1:

abellgithub:
> "Give me n points at a time (where n may be 1), and don't use all my memory please" is such a fundamental task that I think it should deserves native API support"

I understand this.  Not sure this is the way, though it may be.  If this is an important task, then perhaps we should think about how to best handle it.  There's a bunch of stuff that we do now that perhaps we don't need to do at all if what we are wanting to implement is a stand-alone, fast, file-reading interface.  You could eliminate the PointView altogether, the data type conversion, dimension handling on registration, etc.  You could simply have the user provide a place for data to be stuck.  There's also no need, in general, to copy the data twice (a set and a get).  Also many people may not even want to load data other than say X, Y and Z.  I think Chris Foster just wanted to load the values as floats into video memory or something like that.  Do we even want to necessarily normalize data (apply scale/offset to X,Y,Z, convert scan angle values, etc.)?

abellgithub:
We could also break out a lot of the datatype conversion code to make it easier for people to build with that if the wanted.  It doesn't have to sit in the classes in most cases.

connormanning:
Streaming functionality was added in c6bf9c269a13ab33fa7da998049bad618de30409.

PR #988, opened 2015-09-10
Be sure to set the compression flag when inferring LAZ

chambbj:


chambbj:
Seeing as how this didn't show up on Travis, and I'm only on Windows at the moment, someone may want to verify that it indeed fixes things.

PR #991, opened 2015-09-11
Fix memleak in KernelSupport::makePipeline() in case of exceptions

rouault:
For examples fixes
valgrind --leak-check=full ./bin/pdal  info --all ../test/data/nitf/autzen-utm10.ntf
and
valgrind --leak-check=full ./bin/pdal  info --all cmake_install.cmake


abellgithub:
I'm going to close this and replace the raw pointer if that's OK.

rouault:
> I'm going to close this and replace the raw pointer if that's OK.

I trust your judgement. I'm just discovering the code base

PR #992, opened 2015-09-11
Fix failure on PgpointcloudWriterTest

rouault:
If Postgres is available at runtime but not properly configured for database
creation, or pointcloud extension not available, the tests fail.
Here we sheat by skipping the tests in case of those errors.
It would be good if the infrastructure had a cleaner way of skipping tests
in case a requirement isn't met at runtime.


abellgithub:
Agree that this is not the most beautiful way to handle such things.  Perhaps someone else will chime in and have a better way to manage this?

Also, could you please try to follow the conventions in place...functions start with lowercase, if() formatting and so on.  There may be something written up on the pdal.io site, but I think it's very out-of-date.  Best to just follow along from context.

hobu:
Clean up any styling issues after merge into PDAL.

PR #993, opened 2015-09-11
Tutorial mentions rst and xml output

rouault:
It seems there's only JSon right now.


hobu:
Yes, only JSON for now. It is simpler to only support one output encoding, and JSON seems to be the one everyone wants first.

PR #996, opened 2015-09-17
readers.gdal stage for reading pixel data as point clouds

hobu:


filters/colorization/ColorizationFilter.cpp
```diff
@@ -183,72 +161,25 @@ void ColorizationFilter::ready(PointTableRef table)
...
+        int i(0);
         double x = view.getFieldAs<double>(Dimension::Id::X, idx);
         double y = view.getFieldAs<double>(Dimension::Id::Y, idx);
 
-        if (!getPixelAndLinePosition(x, y, m_inverse_transform, pixel,
-                line, m_ds))
-            continue;
+        bool bRead = m_raster->read(x, y, data);
+
+        if (!bRead) continue;
```
abellgithub: I think we're generally avoiding one-line ifs and I think we've also pretty much gotten away from the bBoolean bit.  Is there some good reason to do it?

include/pdal/GDALUtils.hpp
```diff
@@ -78,13 +80,19 @@ class SpatialRef
         { return m_ref.get() != NULL; }
     OGRSpatialReferenceH get() const
         { return m_ref.get(); }
-    bool empty() const
+    std::string wkt() const
     {
         char *pszWKT = NULL;
         OSRExportToWkt(m_ref.get(), &pszWKT);
         bool valid = (bool)*pszWKT;
```
abellgithub: Doesn't look like this is used anymore.

include/pdal/GDALUtils.hpp
```diff
@@ -78,13 +80,19 @@ class SpatialRef
...
+        std::string output(pszWKT);
         CPLFree(pszWKT);
-        return (!valid);
+        return output;
+    }
+
+    bool empty() const
+    {
+        return (wkt().size() != 0);
```
abellgithub: There's wkt().empty().

include/pdal/GDALUtils.hpp
```diff
@@ -166,6 +174,69 @@ class PDAL_DLL ErrorHandler
...
+{
+
+public:
+    Raster(const std::string& filename);
+    ~Raster();
+    bool open();
+    void close();
+
+    bool read(double x, double y, std::vector<double>& data);
+    inline std::vector<pdal::Dimension::Type::Enum> getPDALDimensionTypes() const
```
abellgithub: Don't need "inline"

io/gdal/GDALReader.cpp
```diff
@@ -0,0 +1,216 @@
...
+    if (!m_raster)
+        m_raster = std::unique_ptr<gdal::Raster>(new gdal::Raster(m_filename));
+
+    m_raster->open();
+    setSpatialReference(m_raster->getSpatialRef());
+    m_count = m_raster->m_raster_x_size * m_raster->m_raster_y_size;
+
+}
+
+void GDALReader::processOptions(const Options& options)
```
abellgithub: This can go away.

io/gdal/GDALReader.cpp
```diff
@@ -0,0 +1,216 @@
...
+            m_raster->pixelToCoord(row, col, coords);
+            view->setField(Dimension::Id::X, nextId, coords[0]);
+            view->setField(Dimension::Id::Y, nextId, coords[1]);
+            nextId++;
+        }
+    }
+
+    std::vector<uint8_t> band;
+    std::vector<pdal::Dimension::Type::Enum> band_types = m_raster->getPDALDimensionTypes();
+    for (int b = 0; b < m_raster->m_band_count; ++b)
```
abellgithub: Might be cleaner to index from 1.

io/gdal/GDALReader.cpp
```diff
@@ -0,0 +1,216 @@
...
+        std::stringstream oss;
+        oss << "band-" << b+1;
+        log()->get(LogLevel::Info) << "Read band '" << oss.str() <<"'" << std::endl;
+        pdal::Dimension::Id::Enum d = view->layout()->findProprietaryDim(oss.str());
+        size_t dimSize = pdal::Dimension::size(band_types[b]); // count from 0
+
+        uint8_t* p = band.data();
+        for (point_count_t i = 0; i < count; ++i)
+        {
+            double test =  convert<uint8_t>(p);
```
abellgithub: Used?

io/gdal/GDALReader.cpp
```diff
@@ -0,0 +1,216 @@
...
+                band_types[b] == pdal::Dimension::Type::Double )
+            {
+                if (dimSize == 4)
+                    view->setField(d, i, convert<float>(p));
+                else
+                    view->setField(d, i, convert<double>(p));
+            }
+            else
+            {
+                if (dimSize == 1)
```
abellgithub: If the data is properly aligned, you can just cast the pointer...
*(uint8_t *) p
*(uint16_t *) p
*(uint32_t *) p

PDAL will do the conversion to double for you if necessary.

io/gdal/GDALReader.cpp
```diff
@@ -0,0 +1,216 @@
...
+            }
+            p = p + dimSize;
+        }
+    }
+
+    return view->size();
+}
+
+
+void GDALReader::done(PointTableRef table)
```
abellgithub: This can go away.

src/GDALUtils.cpp
```diff
@@ -100,6 +100,282 @@ ErrorHandler::~ErrorHandler()
...
+            {
+                std::ostringstream oss;
+                oss << "unable to read block for ("<<iXBlock <<","<< iYBlock <<")";
+                throw pdal::pdal_error(oss.str());
+            }
+
+        }
+
+    }
+    return true;
```
abellgithub: Does this ever return anything but true?

src/GDALUtils.cpp
```diff
@@ -100,6 +100,282 @@ ErrorHandler::~ErrorHandler()
...
+        {
+            std::ostringstream oss;
+            oss << "Unable to get band " << i+1 <<
+                " from data source!";
+            throw pdal_error(oss.str());
+        }
+
+        GDALDataType t = GDALGetRasterDataType(band);
+        int x(0), y(0);
+        GDALGetBlockSize(band, &x, &y);
```
abellgithub: Is this being used?

src/GDALUtils.cpp
```diff
@@ -100,6 +100,282 @@ ErrorHandler::~ErrorHandler()
...
+{
+
+    if (!m_ds)
+        throw pdal::pdal_error("Unable to read() because raster data source is not open");
+
+    int32_t pixel(0);
+    int32_t line(0);
+    data.resize(m_band_count);
+
+    std::array<double, 2> pix = { {0.0, 0.0} };
```
abellgithub: Is this initialization OK on Windows?

io/gdal/GDALReader.cpp
hobu: What specifically can go away?

hobu: oh, processOptions. I will be adding more GDAL-specific options in the near future.

io/gdal/GDALReader.cpp
hobu: GDAL indexes bands from 1 and everything else from 0. It seemed safer to use 0 and let GDAL tell me there's no "band-0" instead of doing alist[-1] and blowing up.

io/gdal/GDALReader.cpp
hobu: lint.

src/GDALUtils.cpp
hobu: no, but the other read() method does, and I made them both return a bool for consistency.

src/GDALUtils.cpp
hobu: The `x` and `y` are used two lines after fetching it with this GDALGetBlockSize call...

src/GDALUtils.cpp
hobu: Has seemed to work in the past when this code was in `filters.colorization`

io/gdal/GDALReader.cpp
hobu: I wasn't sure what GDAL is going to fill in. I presume it will always be aligned though...

src/GDALUtils.cpp
chambbj: I thought it would barf on this too. Looking again, I think you can do array initialization lists in VC11, but not vector initialization lists.

src/GDALUtils.cpp
abellgithub: Confused on this one.  What does this do?  It looks like you're fetching x and y and tossing.  All that follows in the PR is stuff about the type.

hobu: I'm fetching the x and y sizes for the block sizes for the band. Each band might have a different block size configuration (though most don't). You want to read on block boundaries in GDAL for optimal access performance.

abellgithub: We must be looking at different code.  x and y are never used in what I'm looking at.  Sorry.

Issue #1009, opened 2015-10-20
Add `cull_unprojectable_points` option to `filters.reprojection`


hobu:
There are cases where a point's coordinates will not reproject into the target coordinate system, but most of the rest of the points in the file might happily do so. Adding this option to `filters.reprojection` means putting a try...catch around the reprojection code, and that's slow, so we only want to do it if the user asked for it.

abellgithub:
@hobu: Can this be closed?

PR #1011, opened 2015-10-21
add capability to cull bad reprojection #1009

hobu:


abellgithub:
You could avoid this by moving up the exception.  You could just return an error from transform() and then only throw the exception if you're not culling.  There's also a leftover debug thing.

PR #1020, opened 2015-10-29
ReprojectionFilter updates

connormanning:
- Reintroduce `ReprojectionFilter::filter`, which is used by the FilterWrapper to modify a PointView in-place
- Remove per-point try-catch
- Minor performance tweaks like allowing `transform` to be inlined


hobu:
Merging this for now, and I'll fix the exception thing in another PR.

PR #1021, opened 2015-10-29
Reprojection re-culling

connormanning:
I made lots of tweaks in #1020 based on a misunderstanding.  This reverts those, but keeps the addition of `filter` from that PR.  Behavior should be the same as before, except with the `m_cullBadPoints` logic negated.

### DIFF
--- a/filters/reprojection/ReprojectionFilter.cpp
+++ b/filters/reprojection/ReprojectionFilter.cpp
@@ -187,7 +187,7 @@ bool ReprojectionFilter::transform(double& x, double& y, double& z)
     catch (pdal::pdal_error& e)
     {
         if (m_cullBadPoints) return false;
-        else throw e.what();
+        else throw e;
     }
 }
 
--- a/filters/reprojection/ReprojectionFilter.cpp
+++ b/filters/reprojection/ReprojectionFilter.cpp
@@ -164,7 +164,7 @@ void ReprojectionFilter::ready(PointTableRef table)
 }
 
 
-bool ReprojectionFilter::transform(double& x, double& y, double& z, bool bThrowOnFailure)
+bool ReprojectionFilter::transform(double& x, double& y, double& z)
 {
     try
     {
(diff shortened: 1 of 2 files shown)


abellgithub:
I'm really confused by all this.  Why do we implement both run() and filter()?  That seems wrong.  Who's calling filter()?

connormanning:
The [FilterWrapper](https://github.com/PDAL/PDAL/blob/master/include/pdal/StageWrapper.hpp#L39).

abellgithub:
Yeah, but that's just for test.

connormanning:
I've been depending on that for my reader-only-with-callback functionality.  That's why it was initially added - because I complained about `filter` being private, which I needed at the time (maybe by now things have changed such that I'd be able to use `run` instead - I don't know).  Then @hobu added `run`, and removed `filter`, not realizing it was being used.  So I added it back.

It's also slightly more efficient than `run` is currently, since it doesn't have per-point try-catch overhead.

abellgithub:
OK.  Well, we need a comment, because it's confusing.  If this is special for Greyhound, it should say so.  As far as I can tell, it is.  I still think the *right* thing to do is to turn off the exception in OCTTransform() and just snag the return code, and then just throw if we're not culling.  But maybe that's more than we want to mess with right now.

connormanning:
I think "special for Greyhound" is overly narrow.  It's "special" for anyone that is using the provided callback functionality to implement a stream-ish API, which may just be my stuff for now - but it's necessary for using the callback stuff with any power behind it.  And we know more people than just me want to use streaming stuff from previous tickets (maybe this all becomes moot once the full streaming API is in place). 

Turning off the exception in OCTTransform would be fine with me, but that also sounds a little messy since we'd need to install a special handler in the Trampoline stuff, I think.

I would be fine with either a comment mentioning its necessity for the streaming API, and/or doing custom error handler stuff so that call won't throw.

abellgithub:
What's misleading is that it's providing true filtering semantics through `run()` by culling points AND the non-filtering semantics through `filter()`, which ignores the cull option.  This is kind of weird.  I don't see a good reason to implement `filter()` at all.  It's less powerful.  You can still call `run()` from a wrapper.  If this is a performance thing, then we should address that separately.

Will new streaming stuff make this a non-issue?  We're going to have to deal with point-at-a-time semantics anyway.

If we're really concerned about exception overhead (which is really quite small), we should just not create exceptions at all for GDAL/OGR errors and handle the return codes.  This is potentially a pain, but if we put it all in one place, it shouldn't be too bad.

hobu:
Let's throw the culling stuff out of here altogether. In retrospect, I think it is misplaced. We could develop a `filters.valid` that could actually check for stuff like "does this reproject into the correct coordinate space?", and deal with the intricacies of that issue there.

abellgithub:
We could just subclass...

PR #1023, opened 2015-10-29
Test changes for SRS setting in prepare().

abellgithub:
OK, this passes tests, but I don't like it because I'm not sure that SRS is special.  Is there some reason we shouldn't promise, say, that all the header data is loaded when prepare() is complete?  We could do this too, but I'd have to change some more stuff to allow access to the table's private metadata.

We should really decide and write down what we're promising at this point since it seems semi-important.

Thoughts?

### DIFF
--- a/io/las/LasReader.cpp
+++ b/io/las/LasReader.cpp
@@ -128,7 +128,7 @@ QuickInfo LasReader::inspect()
 }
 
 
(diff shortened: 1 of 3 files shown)


hobu:
:+1: to merge this instead of #1022

PR #1002, opened 2015-09-24
provide a Height Above Ground processing example

hobu:

PR #1030, opened 2015-11-11
Add PCL-based height (above ground) filter

chambbj:
Add a height filter in-part as a tutorial, but also for users who may want to do this kind of thing without a pre-existing raster as in #1002.


plugins/pcl/filters/HeightFilter.cpp
```diff
@@ -0,0 +1,182 @@
...
+}
+
+void HeightFilter::addDimensions(PointLayoutPtr layout)
+{
+    m_heightDim = layout->registerOrAssignDim("Height", Dimension::Type::Double);
+}
+
+void HeightFilter::filter(PointView& view)
+{
+    if (!view.hasDim(Dimension::Id::Classification))
```
abellgithub: This check is better in prepared(), as it will get called before any execution starts and points are read.

plugins/pcl/filters/HeightFilter.cpp
```diff
@@ -0,0 +1,182 @@
...
+    PluginInfo("filters.height", "Height Filter", "");
+
+CREATE_SHARED_PLUGIN(1, 0, HeightFilter, Filter, s_info)
+
+std::string HeightFilter::getName() const
+{
+    return s_info.name;
+}
+
+Options HeightFilter::getDefaultOptions()
```
abellgithub: This is unnecessary -- unless you're going to fill it in later.

chambbj:
Thanks for comments @abellgithub. Changes made.

doc/tutorial/calculating-normalized-heights.rst
```diff
@@ -0,0 +1,188 @@
...
+    ground->indices.reserve(view.size());
+
+    std::vector<PointId> nonground;
+    nonground.reserve(view.size());
+
+    for (PointId id = 0; id < view.size(); ++id)
+    {
+        double c = view.getFieldAs<double>(Dimension::Id::Classification, id);
+
+        if (c == 2)
```
hobu: need a little comment that 2 == ground (in LAS)

chambbj: @hobu Added a `.. note::` in the text above with a link to the spec.

PR #1032, opened 2015-11-12
BPF should use auto offsets by default for XYZ dimensions

chambbj:
@abellgithub Not sure if this is how you would have handled it, but it seems to work as expected.


abellgithub:
Looks good.  I just had one thought.  The new BPF spec supports other projections.  Do we care?

chambbj:
I didn't realize that it did. Still doesn't seem like a bad idea to use the offsets, but I couldn't speak to any implications. Do we know of anybody using it with other projections? Or was it thrown in just because?

abellgithub:
Oh, I doubt anyone's using it at all.  It's just in the new spec.  Was just throwing it out there.

chambbj:
@hobu @abellgithub Any objections to me going ahead and merging this PR?

PR #1036, opened 2015-11-13
Updating reader tutorial.

scott-lewis-nsidc:
Modifying reader tutorial to discuss the steps needed for a reader,
including a sample input file and pipeline file to use.

Uses a fake dummy format for the file just to exercise the steps needed to have the writer work.


hobu:
Keep on rockin' our tutorials! It is much appreciated.

PR #1039, opened 2015-11-16
Check point view size in `getFieldInternal`

gadomski:
`PointView::getFieldInternal` uses `deque::operator[]` to map index values to raw point ids. This means that a user can silently access beyond the end of of the `m_index` deque with an ill-advised `getFieldAs` call, leading to either garbage results or a segfault. I recokon we should sanitize this situation, since `PointView::getFieldAs` is part of our oft-used public API.

This patch proposes the simplest possible solution, a switch to `deque::at` inside `getFieldInternal`. It also adds a test case.

NB I haven't done an audit to see if there is other unsafe m_index access in the codebase.

### DIFF
--- a/include/pdal/PointView.hpp
+++ b/include/pdal/PointView.hpp
@@ -381,6 +381,7 @@ template <class T>
 inline T PointView::getFieldAs(Dimension::Id::Enum dim,
     PointId pointIndex) const
 {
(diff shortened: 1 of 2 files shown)


abellgithub:
Thanks!

Issue #1038, opened 2015-11-16
`pdal info --summary` broken for RxpReader


gadomski:
NB this is a possible duplicate of #1012, @abellgithub feel free to close this if it is.

`pdal info` works just fine:

``` bash
$ pdal info test/data/0916_2014_girdwood35.rxp | head
{
  "filename": "test/data/0916_2014_girdwood35.rxp",
  "pdal_version": "1.0.1 (git-version: 1394a1)",
  "stats":
  {
    "statistic":
    [
      {
        "average": 0.407355579,
        "count": 257576,
```

But `pdal info --summary` is bunk:

``` bash
$ pdal info --summary test/data/0916_2014_girdwood35.rxp
{
  "filename": "test/data/0916_2014_girdwood35.rxp",
  "pdal_version": "1.0.1 (git-version: 1394a1)",
  "summary":
  {
    "bounds":
    {
      "X":
      {
        "max": -1.797693135e+308,
        "min": 1.797693135e+308
      },
      "Y":
      {
        "max": -1.797693135e+308,
        "min": 1.797693135e+308
      },
      "Z":
      {
        "max": -1.797693135e+308,
        "min": 1.797693135e+308
      }
    },
    "dimensions": "",
    "num_points": 0,
    "spatial_reference": ""
  }
}
```

I'm guessing this has to do with the fact that the `RxpReader` doesn't know anything about its points until it reads them, since rxp is a streaming format with no concept of a "header" or anything. If that is the case, it's an open question what the appropriate response of `pdal info --summary` should be — either bomb, saying summary doesn't work with this reader, or read all the points then summarize (maybe with a warning message about what it's doing)?

(if we used issue priority levels, this one would be tagged "low priority")

abellgithub:
You have to implement inspect() to get output.  `pdal info --summary` should certainly tell you if it can't do a summary, or it should do things the hard way and report that.  The way it is now is bad.

gadomski:
Is `inspect` expected to fill all member variables of `QuickInfo`? I.e. if
I can't fill m_pointCount without reading the whole file, what should I do?
Read the whole file and fill it in? Leave `inspect` unimplemented?

On Mon, Nov 16, 2015 at 11:25 AM Andrew Bell <notifications@github.com>
wrote:

> You have to implement inspect() to get output. pdal info --summary should
> certainly tell you if it can't do a summary, or it should do things the
> hard way and report that. The way it is now is bad.
>
> —
[...]

PR #1043, opened 2015-11-17
Issues/1012 misleading summary

gadomski:
If a reader does not implement `inspect`, `pdal info --summary` produces misleading results. This patch fixes the problem by forcing readers to opt-in to summary information by explicitly implementing `inspect`. If a reader does not implement `inspect`, the default implementation returns `boost::none`, which is then upgraded to a `pdal_error` by `Stage::preview`.

This is not a perfect solution to the problem — it may be overly strict. But it is an attempt to clarify the policy for `Stage::inspect`, which [was ambiguous](https://github.com/PDAL/PDAL/issues/1038#issuecomment-157238262).

The second commit removes the default constructor and the `m_valid` flag from `QuickInfo`, since they are no longer needed (I think).

@abellgithub this is just one idea I had, feel free to modify/reject as needed.

### DIFF
--- a/include/pdal/QuickInfo.hpp
+++ b/include/pdal/QuickInfo.hpp
@@ -49,13 +49,6 @@ public:
     SpatialReference m_srs;
     point_count_t m_pointCount;
     std::vector<std::string> m_dimNames;
-    bool m_valid;
-
-    QuickInfo() : m_pointCount(0), m_valid(false)
-        {}
-
-    bool valid() const
-        { return m_valid; }
 };
 
 } // namespace pdal
--- a/include/pdal/Stage.hpp
+++ b/include/pdal/Stage.hpp
@@ -75,7 +75,13 @@ public:
     {
         l_processOptions(m_options);
         processOptions(m_options);
-        return inspect();
+        auto quickInfo = inspect();
+        if (!quickInfo) {
+            std::stringstream ss;
+            ss << "Cannot preview stage '" << getName() << "' because it does not implement `inspect()`";
+            throw pdal_error(ss.str());
+        }
+        return *quickInfo;
     }
     void prepare(PointTableRef table);
     PointViewSet execute(PointTableRef table);
@@ -155,8 +161,8 @@ private:
     virtual void writerProcessOptions(const Options& /*options*/)
         {}
     void l_initialize(PointTableRef table);
-    virtual QuickInfo inspect()
-        { return QuickInfo(); }
+    virtual std::unique_ptr<QuickInfo> inspect()
+        { return nullptr; }
     virtual void initialize(PointTableRef /*table*/)
         { initialize(); }
     virtual void initialize()
--- a/io/bpf/BpfReader.cpp
+++ b/io/bpf/BpfReader.cpp
@@ -63,32 +63,31 @@ void BpfReader::processOptions(const Options&)
 }
 
 
-QuickInfo BpfReader::inspect()
+std::unique_ptr<QuickInfo> BpfReader::inspect()
 {
-    QuickInfo qi;
+    std::unique_ptr<QuickInfo> qi(new QuickInfo());
 
     initialize();
-    qi.m_valid = true;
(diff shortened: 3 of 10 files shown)


connormanning:
Personally I'd rather see a `unique_ptr` here than introducing boost stuff here, assuming that the problem you're solving is never having a dereferenceable `QuickInfo` object whose members are not valid.

E.g.
`virtual std::unique_ptr<QuickInfo> inspect() { return std::unique_ptr<QuickInfo>(); }`

And:
```
auto quick(r.inspect());
if (quick) { /* Use quick->m_srs, etc. */ }
```

gadomski:
The problem I'm trying to solve is the idea that a `Stage` may or may not provide a meaningful `inspect` implementation, which right now is a binary case — either a stage can `inspect`, or it can't. I agree that introducing moar boost is a negative, but I felt as though `optional` exactly captured what I was trying to say, so I used it.

To me, it feels like a `unique_ptr` is just trying to be an `optional` with some extra confusion (is `QuickInfo` big, and therefore we don't want to copy it?), but in the end it serves the same purpose. I'd be ok with `unique_ptr`, I guess I just am mostly wishing we had a standard optional to work with for exactly this scenario. In the end, whatever ya'll think is best.

abellgithub:
I guess I'd be for putting more into QuickInfo to let people know what you've got.  Maybe that would include error messages or indicators about things.  Don't like boost::optional at all, but we can achieve the same thing in some other way.  Some formats don't have SRS, for example.  It would be nice to say that, rather than just having an empty string for the SRS.  But perhaps that belongs in the SRS rather than in the QuickInfo.  Mostly thinking out loud here...

connormanning:
How about stack-allocating QuickInfo like before, but with `unique_ptr` members?  Then a QuickInfo instance is always valid, but a Reader can construct only the fields it knows how to inspect, leaving the other fields as invalid.

E.g.
```
QuickInfo MyCoolReader::inspect()
{
    QuickInfo q;
    q.m_bounds.reset(new BOX3D(readHeaderBounds()));
    return q;
}
```

abellgithub:
That’s just the same as boost::optional, really, without boost::optional.  Doesn’t give you any extra information other than availability.  Is that enough?

gadomski:
@connormanning that's a way to implement a finer-grained QuickInfo. If that's what we want, ok. However, that's going to be more impactful on existing code, as each existing implementation of `inspect` will need to be updated with the new data member assignment ritual.

IMO it seems simpler to make each stage choose "either I implement `inspect`, or I don't". For finer-grained control, most data members of QuickInfo object already support the idea of being "unset", e.g. BOX3D can just be numeric limits, and the srs can be empty. So to me putting them all behind optional/unique_ptr seems a bit redundant?

hobu:
@abellgithub please merge or adjust this as necessary.

abellgithub:
Where are we with this?  Is there some reason why setting/checking the valid flag was insufficient?

We're working to remove boost, so I think boost::optional is a no-go.

Thoughts?

gadomski:
> Is there some reason why setting/checking the valid flag was insufficient?

The valid flag wasn't being checked in `Stage::preview`, which (in my opinion) demonstrates why a `unique_ptr` approach is more robust than a flag approach. It's too easy to forget to check the flag.

> We're working to remove boost, so I think boost::optional is a no-go.

I agree. I've updated this PR to use `std::unique_ptr`.

@connormanning and I were going back and forth on how fine-grained the "optional" parts of the `QuickInfo` should be. I don't have much of an opinion, so this PR keeps things as low-impact as possible by not changing the character of the members of QuickInfo. If we wanted to make each QuickInfo member a (nullable) pointer as well to capture their "optional-ness", that'd also be an option.

hobu:
@abellgithub are we keeping this?

abellgithub:
Up to you.  I don't care.

gadomski:
Closing as interest-lost.

PR #1047, opened 2015-11-18
Update tutorial writing a writer

scott-lewis-nsidc:
The previous tutorial looked like just a copy of the Kernel tutorial.  I updated it so that it was a true Writer tutorial, albeit for a pretty simple and arbitrary output text format.

### DIFF
--- a/doc/tutorial/writing-reader.rst
+++ b/doc/tutorial/writing-reader.rst
@@ -208,5 +208,5 @@ In the directory with those two files, run
 called ``output.txt``, which will have the same data as in the input file,
 except in a CSV style format, and with the Z values scaled by .001.
 
-.. _`pipeline-myreader.xml`: https://github.com/PDAL/PDAL/examples/writing-reader/pipeline-myreader.xml?raw=true
-.. _`test-reader-input.txt`:https://github.com/PDAL/PDAL/examples/writing-reader/test-reader-input.txt?raw=true
+.. _`pipeline-myreader.xml`: https://github.com/PDAL/PDAL/blob/master/examples/writing-reader/pipeline-myreader.xml?raw=true
+.. _`test-reader-input.txt`: https://github.com/PDAL/PDAL/blob/master/examples/writing-reader/test-reader-input.txt?raw=true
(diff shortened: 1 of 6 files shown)


abellgithub:
Hi Scott,

I just had a chance to look at this and I thought I'd offer an alternative.  I'm not saying it's better or worse, but it seems more natural to me.  Of course, I'm the guy who wrote most of the code, so I'm naturally biased :)  I haven't taken the time to compile or test, but I thought I'd pass along.

Thanks again!

```
MyReader : public pdal::Reader
{
public:
    std::string getName() const
        { return "MyReader"; }

    virtual void addDimensions(PointLayoutPtr layout)
    {
        layout->registerDim(Dimension::Id::X);
        layout->registerDim(Dimension::Id::Y);
        layout->registerDim(Dimension::Id::Z);
    }

    virtual point_count_t read(PointViewPtr view, point_count_t count)
    {
        for (PointId i = 0; i < 1000 && i < count; ++i)
        {
            view->setField(Dimension::Id::X, i, -93.0 + i*0.001)
            view->setField(Dimension::Id::Y, i, 42.0 + i*0.001);
            view->setField(Dimension::Id::Z, i, 106.0 + i);
        }
    }
}


int main(int argc, char* argv[])
{
    using namespace pdal;

    Options options;
    options.add("filename", "myfile.las");

    PointTable table;

    MyReader reader;

    StageFactory factory;
    Stage *writer = factory.createStage("writers.las");

    writer->setInput(reader);
    writer->setOptions(options);
    writer->prepare(table);
    writer->execute(table);
}
```

scott-lewis-nsidc:
Thanks for the feedback!  I did have a question, though - the code you describe here looks like it would be an alternative for the "Writing" tutorial, basically simplifying it a little bit, as opposed to the tutorial this pull request fixed (the "Writing a writer" tutorial).  But perhaps I'm missing something?

In any case, the Writing tutorial could certainly be modified to this; in that tutorial, I only did a few minor tweaks because the code didn't work (I think the core PDAL code had changed underneath it, and some objects/references got changed).

Anyway, I hope that makes sense.  I admit, I'm still learning the ins-and-outs of the codebase myself so there are likely many things I'm not quite using properly.

PR #1050, opened 2015-11-21
Rename Stage::getDefaultOptions to Stage::getAnsweredOptions

hobu:
The previous method name is still available as a public method, but
it is aliased to getAnsweredOptions. We are going to do this for a
while to attempt not to break plugins for a version or two.

### DIFF
--- a/apps/pdal.cpp
+++ b/apps/pdal.cpp
@@ -139,7 +139,7 @@ void outputOptions(std::string const& n)
     std::cout << n << std::endl;
     std::cout << headline << std::endl;
 
-    std::vector<Option> options = s->getDefaultOptions().getOptions();
+    std::vector<Option> options = s->getAnsweredOptions().getOptions();
     if (options.empty())
     {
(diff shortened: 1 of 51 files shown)


hobu:
@abellgithub and @chambbj I'm interested in your thoughts on this particular name and whether or not you have a better one.

chambbj:
For me, I'm not sure this rename is very helpful. I think new users are probably tripped up by the fact that calling `getOptions` on a stage wouldn't just give you this answer. Maybe `getStageOptions` is better?

Beyond the naming issue, it would also be nice to address the fact that `getDefaultOptions` is not always implemented and we can end up with inadvertently hidden arguments, as well as the redundant and occasionally inconsistent definition of "default" values within `options.add` in `getDefaultOptions` and `options.getValueOrDefault` within `processOptions`.

hobu:
ok, I'll close this for when we have a better idea of what we should do.

PR #1052, opened 2015-11-22
Implement readers.tindex

hobu:


include/pdal/SpatialReference.hpp
```diff
@@ -75,10 +75,14 @@ class PDAL_DLL SpatialReference
     bool operator<(const SpatialReference& other) const
         { return m_wkt < other.m_wkt; }
 
-    // Returns true iff the object doesn't contain a valid srs.
+    // Returns true if the object doesn't contain a valid srs.
```
abellgithub: The original is math-speak, not a typo, I don't think :)

io/las/LasReader.cpp
```diff
@@ -67,6 +67,7 @@ class invalid_stream : public pdal_error
 
 void LasReader::processOptions(const Options& options)
 {
+    m_skipWktVLR = options.getValueOrDefault<bool>("skip_vlr_wkt", false);
```
abellgithub: Don't think this is necessary, is it?  The code I'm going to commit for the other fix will take care of this.  Is this just a temp thing?

abellgithub: Do tests pass with this?  Because mine fail, which is why the fix hasn't been merged.

hobu: I think this is sloppy merging.

PR #1058, opened 2015-12-02
Refactor Travis configuration to use Docker to provide the base set of dependencies.

hobu:
* Travis only builds clang/ninja now.
* Travis builds ALL and NONE configuration
* Dependencies are fetched from https://hub.docker.com/r/pdal/dependencies/
* Dependencies Dockerfile is at scripts/docker/dependencies


hobu:
squash merged in @2b9cf8a

PR #1062, opened 2015-12-07
Create transform for ReprojectionFilter as soon as necessary inputs are available.

connormanning:
After `processOptions`, if an input SRS was explicitly provided, we can create the transform here instead of waiting until `run`, which the `FilterWrapper` does not call.  If this is the case, this PR creates the transform during `initialize` so the filter wrapper may be used to manage a reprojection filter.

### DIFF
--- a/filters/reprojection/ReprojectionFilter.cpp
+++ b/filters/reprojection/ReprojectionFilter.cpp
@@ -126,6 +126,9 @@ void ReprojectionFilter::initialize()
         throw pdal_error(oss.str());
     }
 
+    // If we have our input and output spatial references, create the transform
+    // now.  Otherwise, if this filter is used via the FilterWrapper, it will
+    // not be initialized properly since run() won't get called.
     if (!m_inferInputSRS)
     {
         createTransform(0);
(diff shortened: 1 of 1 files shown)


filters/reprojection/ReprojectionFilter.cpp
```diff
@@ -132,7 +134,13 @@ void ReprojectionFilter::createTransform(PointView *view)
 {
     if (m_inferInputSRS)
     {
+        if (!view)
```
abellgithub: Don't think this is necessary.  It's a code bug, not a run-time bug.

filters/reprojection/ReprojectionFilter.cpp
```diff
@@ -125,6 +125,8 @@ void ReprojectionFilter::initialize()
             "for the 'out_srs' option.";
         throw pdal_error(oss.str());
     }
+
+    if (!m_inferInputSRS) createTransform(nullptr);
```
abellgithub: No one-line if/then's please and `nullptr` shouldn't be used as it isn't supported on some platforms.

abellgithub:
Do we need the FilterWrapper anymore?  If so, why?

connormanning:
Streaming stuff is not on master and the branch doesn't yet support a predefined Layout.  The FilterWrapper is necessary to perform filtering using the callback mechanism.

abellgithub:
Always in a hurry...  Put a comment in to say why it's there.

PR #1068, opened 2015-12-10
Ilvis2 reader update

scott-lewis-nsidc:
Modified the ILVIS reader to allow either HIGH or LOW value to be used, or ALL, which will generate two points if the elevations of HIGH and LOW differ.  CENTROID will not be used at this time, per discussion with the LVIS science team.

Also modified the longitude conversion.  Values 0-180 should remain "as is"; values start from 0 and go east, so past 180 you have wrapped around the dateline; to get the correct longitude, you thus subtract 360 from values that have "wrapped" to get a negative value, representing negative degrees east.

Modified the ILVIS Reader Test to accommodate these changes, as well as to add a few more test cases.

NOTE: The reader will also be implementing a method to read a metadata XML file.  This will be done via a separate pull request.


io/ilvis2/Ilvis2Reader.cpp
```diff
@@ -50,14 +50,16 @@ std::string Ilvis2Reader::getName() const { return s_info.name; }
 Options Ilvis2Reader::getDefaultOptions()
 {
     Options options;
-    Option mapping("mapping", "LOW");
-    options.add(mapping);
+    options.add("mapping", "ALL", "The point to extract from the shot: LOW, HIGH, or ALL.  ALL creates 1 or 2 points per shot, depending on if LOW and HIGH are the same.");
```
abellgithub: I *think* we're generally advertising fixed option values as lowercase, but it doesn't much matter.  Look around and see what you find.  I'm almost always forcing case when I check to avoid failures based only on case.

Also, in this case, probably best to use an enum internally, rather than a string.

io/ilvis2/Ilvis2Reader.cpp
```diff
@@ -123,11 +125,83 @@ T convert(const StringList& s, const std::string& name, size_t fieldno)
...
+    return longitude;
+  }
+}
+
+
+void Ilvis2Reader::readPoint(PointViewPtr view, PointId nextId, StringList s, std::string pointMap)
+{
+  PointLayoutPtr layout = view->layout();
+
+  std::string name("LVIS_LFID");
```
abellgithub: I know that you're just moving this, but there's no reason I can think of not to make these real dimensions (add to Dimension.hpp).  It gets away from all the string stuff, dealing with the layout and the dimension finding.  It'll let you just say `setField(Dimension::Id::LongitudeCentroid,...)

io/ilvis2/Ilvis2Reader.cpp
scott-lewis-nsidc: I can make that change.  I missed that LASWriter/LASReader was using a toupper call on the fixed param it uses (compression), so I'll implement that.

I will also modify the code to internally use an ENUM, I agree it's cleaner (and slightly faster internally than using string comparisons on every iteration of the point view).

io/ilvis2/Ilvis2Reader.cpp
scott-lewis-nsidc: I could do that, although that would mean adding 6 new dimensions (lat/lon for high/low/centroid), and they do seem kind of LVIS-specific.  I may leave it for now, unless there are really strong feelings for making them general dimensions.

abellgithub: Up to you, but I'll just say that I was going to make this change anyway and probably will at some point if you don't.  Just because they're in the list doesn't mean they're general.  It means that we know about them from the specification -- they aren't variable strings that we don't know until runtime.  I allowed for 65,000+ known dimensions.  We have a long way to go :)

scott-lewis-nsidc: Ok, fair point :)  I'll add them then.  Would you recommend, then, that I add all of these (the lat/lon/elev values for centroid/low/high, as well as the LVIS_LFID and SHOTNUMBER)?  Or just the lat/lon/elev values?

abellgithub: I'd add them all.

scott-lewis-nsidc:
I've implemented the changes as suggested in the previous discussions.

abellgithub:
Thanks for this!

PR #1075, opened 2015-12-28
Support GDAL 2.0+, improve GDAL configuration detection

hobu:


cmake/modules/FindGDAL.cmake
```diff
@@ -89,6 +89,7 @@ set (GDAL_VERSION_COUNT 3)
 
 SET(GDAL_NAMES gdal)
 
+MESSAGE(STATUS "I'm here!!!!!!!!!!!!!!!")
```
abellgithub: Ummmm.

cmake/modules/FindGDAL.cmake
```diff
@@ -202,20 +208,22 @@ ELSEIF(UNIX)
             set (GDAL_VERSION_COMPATIBLE true)
         endif()
 
+        MESSAGE(STATUS "Found version: ${GDAL_VERSION_STRING}")
```
abellgithub: Should this be left?

PR #1079, opened 2015-12-29
Lvis metadata file reading

scott-lewis-nsidc:
Added a file to process the XML metadata file the accompanies the main data file.  Including this file is optional but using it will import a slew of metadata related to the file, including a tighter convex hull boundary and a list of platforms/instruments on the flight.


scott-lewis-nsidc:
After submitting this, I realized I had meant to go and add some comments to my code.  I tried to make it somewhat self-explanatory, but I think some extra comments may do well.  I'll submit those comments shortly.

scott-lewis-nsidc:
It looks like I need to figure out how to get it to find libxml/parser.h.  I'll work on that tomorrow.

scott-lewis-nsidc:
For what it's worth, I realize there is a little redundancy on the checks for element names; it checks BEFORE it enters the parsing method for an element, as well as at the beginning of the parsing element method itself.

I've been toying with that in my mind; I may remove the one at the top of the parsing element method, going off the assumption of "it wouldn't be here if it wasn't the right element."

scott-lewis-nsidc:
I decided (for now) to leave the redundancy in there just as an extra safeguard that if the code gets changed down the road that somehow a wrong function call doesn't end up causing issues; it checks it right off the bat.  However, I can still see an argument for removing one of the redundant calls.

hobu:
Squash merged into master as c8e1eb8d8539d20a6d7080449f56825b5110c6e5

PR #1091, opened 2016-01-25
Make GEOS error and debug handling work like GDAL

hobu:


src/CMakeLists.txt
```diff
@@ -200,7 +201,7 @@ endif()
 # Targets installation
 
 #
-# On Linux, we install a linker script as libpdalcpp.so.  That file
+# On Lixxxxxxxxxxxxxxxxxxnux, we install a linker script as libpdalcpp.so.  That file
```
abellgithub: On purpose?

src/util/FileUtils.cpp
```diff
@@ -41,8 +41,7 @@
 
 #include <pdal/util/FileUtils.hpp>
 #include <pdal/util/Utils.hpp>
-
-using namespace std;
+#include <pdal/pdal_types.hpp>
```
abellgithub: Not that it makes  any difference, but I don't think "using namespace std" hurts anything here and just means you have std:: all over the place.

plugins/hexbin/filters/HexBin.cpp
```diff
@@ -35,7 +35,7 @@
 #include "HexBin.hpp"
 
 #include <hexer/HexIter.hpp>
-#include <pdal/Geometry.hpp>
+#include <pdal/GEOSUtils.hpp>
```
abellgithub: Not sure why this is how it is.  My thought behind Geometry was to hide the implementation, which this undoes.  Not sure if there's some advantage to exposing this.

PR #1093, opened 2016-01-29
Prevent segfaults in outlier filters

chambbj:
If the radius or statistical outlier filters would remove all points, log it, and return the input cloud. This is especially problematic if `filters.radiusoutlier.radius` is chosen too small, and the minimum number of neighbors cannot be satisfied.


abellgithub:
Seems good.

Issue #1098, opened 2016-02-05
problems with dropping tables with pgpointcloud driver


smathermather:
Using a table name like 54001640PAN_heightasz_0-1.5 loads fine, but does not reload fine, resulting in an error like:

```LINE 1: DROP TABLE IF EXISTS 54001640PAN_heightasz_0-1.5```
I think this is an issue with escaping the name appropriately. I get a similar error in psql if I fail to quote my table name:
```
gisuser=# drop table 54001640PAN_heightasz_0-1.5;
ERROR:  syntax error at or near "54001640"
LINE 1: drop table 54001640PAN_heightasz_0-1.5
                   ^
gisuser=# drop table "54001640PAN_heightasz_0-1.5";
DROP TABLE
```

hobu:
Done in #1098

smathermather:
Awesome. Thanks!

PR #1099, opened 2016-02-05
Quote delete statement for pgpointcloud #1098

hobu:


abellgithub:
Perhaps we can add a test that uses a crazy table name for this?  Could be exactly the same otherwise, I'm thinking.

PR #1103, opened 2016-02-08
Output more convenient metadata for bbox and SRS

hobu:
Output bbox information in `filter.stats`:
```
    "bbox":
    {
      "EPSG:4326":
      {
        "bbox":
        {
          "maxx": -123.0625001,
          "maxy": 44.0624972,
          "maxz": 178.73,
          "minx": -123.0749695,
          "miny": 44.0500086,
          "minz": 123.93
        }
      },
      "native":
      {
        "bbox":
        {
          "maxx": -123.0625001,
          "maxy": 44.0624972,
          "maxz": 178.73,
          "minx": -123.0749695,
          "miny": 44.0500086,
          "minz": 123.93
        }
      }
    },
```

### DIFF
--- a/filters/stats/StatsFilter.cpp
+++ b/filters/stats/StatsFilter.cpp
@@ -38,10 +38,8 @@
 
 #include <pdal/pdal_export.hpp>
 #include <pdal/Options.hpp>
-/**
+#include <pdal/Polygon.hpp>
 #include <pdal/PDALUtils.hpp>
-#include <pdal/util/Utils.hpp>
-**/
 
 namespace pdal
 {
@@ -107,7 +105,7 @@ void StatsFilter::filter(PointView& view)
 
 void StatsFilter::done(PointTableRef table)
 {
-    extractMetadata();
+    extractMetadata(table);
 }
 
 
@@ -168,9 +166,9 @@ void StatsFilter::prepared(PointTableRef table)
         m_stats.insert(std::make_pair(layout->findDim(dv.first),
             Summary(dv.first, dv.second)));
 }
-    
 
-void StatsFilter::extractMetadata()
+
+void StatsFilter::extractMetadata(PointTableRef table)
 {
     uint32_t position(0);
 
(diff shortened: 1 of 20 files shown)


filters/stats/StatsFilter.cpp
```diff
@@ -182,6 +180,39 @@ void StatsFilter::extractMetadata()
         t.add("position", position++);
         s.extractMetadata(t);
     }
+
+    // If we have X, Y, & Z dims, output bboxes
+    auto xs = m_stats.find(Dimension::Id::X);
```
abellgithub: Do we always want to require the Z dimension?

include/pdal/util/Bounds.hpp
```diff
@@ -189,6 +202,10 @@ class PDAL_DLL BOX3D : private BOX2D
     BOX3D()
        { clear(); }
 
+    BOX3D(const BOX2D& box)
```
abellgithub: We got rid of this on purpose.  It might be OK with "explicit".  Why is it here?

hobu: I added it so I don't have to make two pdal::Polygon constructors.

filters/stats/StatsFilter.cpp
hobu: I'm 50/50 on requiring Z's. My first pass didn't require it, and my second pass did. Interested in others' thoughts on the topic...

include/pdal/util/Bounds.hpp
abellgithub: I'd recommend making two constructors.  If you need to do some common stuff, just call a function.   You could also make the ctor explicit.  You just don't want that conversion to happen in most cases.

hobu: I do, but I'm weird I guess. I rather don't like that there's BOX2D and BOX3D at all to be honest. What we have now is more explicit, to be sure.

PR #1123, opened 2016-02-16
GEOS is no longer optional

mloskot:
geos.cmake specifies REQUIRED property

It is confirmed by CMake error while configuring VS2015 32-bit build, if GEOS is not found:

```
-- The following REQUIRED packages have not been found:

 * GEOS (required version >= 3.3)
   Provides general purpose geometry support

CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
GEOS_INCLUDE_DIR
...
GEOS_LIBRARY
    linked by target "pdalcpp" in directory D:/dev/PDAL/src
```


hobu:
GEOS and GDAL are indeed no longer optional.

PR #1125, opened 2016-02-17
Remove variable-length array use from ILVIS2 reader

mloskot:
Fixes compilation using Visual C++ (VS2015 Update 1) which has no support for C99 VLAs.

-----
Note, it does not correct potential memory leak in `Ilvis2MetadataReader::parseGPolygon`:
```
    if (numBoundaries > 1)
    {
        fullPoly = GEOSGeom_createCollection(
                GEOS_MULTIPOLYGON, poly.data(), numBoundaries);
    }
    else
    {
        fullPoly = poly[0];
    }
```

While `GEOSGeom_createCollection` takes ownership of the geometries passed in `poly` array, then in the `else` case, it is unclear who deallocates the single polygon.

### DIFF
--- a/io/ilvis2/Ilvis2MetadataReader.cpp
+++ b/io/ilvis2/Ilvis2MetadataReader.cpp
@@ -328,7 +328,7 @@ void Ilvis2MetadataReader::parseGPolygon(xmlNodePtr node, MetadataNode * m)
 
     // The number of boundaries is essentially the number of sub-polygons
     int numBoundaries = countChildElements(node, "Boundary");
-    GEOSGeom poly[numBoundaries];
+    std::vector<GEOSGeom> poly(numBoundaries); // size shall never change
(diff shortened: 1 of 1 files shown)


abellgithub:
Totally unclear to me who's responsible for ownership in general with GEOS.  In both cases it seems like fullPoly is the result of some GEOSGeom_create...() or equivalent, so I'm not sure the two cases you cite are different.

mloskot:
@abellgithub Ownership in GEOS is a challenge indeed and I confused myself here. 

AFAIS, it is `fullPoly` which seems to be never deallocated. AFAIK, correct sequence likes like this:
```
GEOSGeom fullPoly;
fullPoly = GEOSGeom_create...
...
GEOSGeom_destroy(fullPoly);
```
whereas the `destroy` step seem to be missing in the ilvis2 reader.

PR #1129, opened 2016-02-17
Fix UB caused by incrementing pass-the-end iterator

mloskot:
Iterator returned from `std::find_if` may point pass-the-end, but incrementing pass-the-end iterator is an undefined behaviour.

The bug was revealed by `stringsplit` case of the `OptionsTest`, where input string has separator(s) at as well as `UtilsTest`, where final `std::find_if` returns end iterator which must not be incremented.

### DIFF
--- a/include/pdal/util/Utils.hpp
+++ b/include/pdal/util/Utils.hpp
@@ -229,14 +229,17 @@ namespace Utils
         if (s.empty())
             return result;
 
-        auto it = s.begin();
-        decltype(it) endIt;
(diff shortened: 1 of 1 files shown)


abellgithub:
I guess I don't care, but I what harm can come from incrementing the iterator past the end as the result is unused in this case?  Make more more complicated code.

mloskot:
UB means anything can happen.

The code is effectively one extra test but it reads more verbose since it avoids multiple function calls to `end()`.

PR #1130, opened 2016-02-17
Fix out of range access bug in trimTrailing

mloskot:
If input string is empty, `size() - 1` overflows causing character access beyond valid range.

### DIFF
--- a/src/util/Utils.cpp
+++ b/src/util/Utils.cpp
@@ -149,6 +149,9 @@ void Utils::trimLeading(std::string& s)
 
 void Utils::trimTrailing(std::string& s)
 {
+    if (s.empty())
+        return;
+
     size_t pos = s.size() - 1;
     while (isspace(s[pos]))
     {


connormanning:
Something like: 

```
while (!s.empty() && std::isspace(s.back())
{
    s.pop_back();
}
```

seems much clearer to me than the current function.

abellgithub:
That seems fine, too.  Was old code that came from elsewhere before you could "pop_back()" on strings.  Current *may* be more efficient, but I doubt it matters in any of our cases.

mloskot:
@connormanning I prefer clang/LLVM coding style where early exits are recommended to *simplify code*. Also, `!s.empty()` test in your code is confusing as it is irrelevant to actual iteration.
I guess, it is irrelevant to discuss coding style here. I don't care what it is changed too.

connormanning:
To me, the current implementation separates "is the initial string empty?" and "is the modified string empty?" into two cases for no reason, which I don't think simplifies the code - since we always just want to check the current value regardless of whether it's the first iteration or not.  Consolidating into one case makes sense to me, in which case the `empty` check is required because `back` is UB if `empty` is true.

+1 on the current implementation as far as I'm concerned, I just thought that all the branched cases made it more confusing to follow.

mloskot:
@connormanning Yes, you are right. As I re-stated my explanation, I stick to minimal changes and not mixing fixes with improvements.

Issue #333, opened 2014-05-01
C++11 and Windows


chambbj:
We've been making more use of C++11 features (e.g., range-based for loops) lately. We offer no workarounds for users on older versions of Visual Studio however. Are we sure we want to force the new standard and shut out these individuals?

PR #1133, opened 2016-02-17
Fix too small sprintf buffer in Uuid::unparse

mloskot:
sprintf always appends terminating null character, so the character buffer used with sprintf needs to accommodate it.

### DIFF
--- a/include/pdal/util/Uuid.hpp
+++ b/include/pdal/util/Uuid.hpp
@@ -163,9 +163,10 @@ public:
 
     std::string unparse() const
     {
-        std::vector<char> buf(36);
+        std::vector<char> buf(36 + 1);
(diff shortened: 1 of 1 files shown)


connormanning:
Can you elaborate on the added TODO - PDAL does require C++11, so is the issue a lack of MSVC compliance for a specific C++11 feature?

mloskot:
Despite looking at #333, it is unclear to me what is minimal supported version of Visual C++, so I can't assume `_MSC_VER >= 1900` is true and `std::snprintf` is always available.
I also don't like the idea of polluting the code with typical `#define snprintf _snprintf` workarounds.

So, I've decided to leave the comment to the core developers (also to address possible questions "why not improving it further with snprintf" :)).

abellgithub:
Code copied from libuuid.  Sorry.

mloskot:
@connormanning I'm doing too many things at ones, so apology for yet another single comment written in multiple iterations :)

PR #1137, opened 2016-02-18
Reimplement getFilename using filesystem::path

### LINKED ISSUE #1131: FileUtilsTest assumes Unix paths failing on Windows
Regardless of OS, the FileUtilsTest always sticks to Unix paths (or normalised paths) with `/` separator, while `FileUtils::getFilename` (and `stem` too) behaviour is OS-specific and always uses `\` on Windows.

This mismatch makes the test fail.

I guess, the tests should use OS-specific test values, right?

mloskot:
Previously, on Windows, `getFilename` always looked for preferred separator only while it should look for both, preferred (backslash) and portable (slash) which are equally supported by Windows. PDAL in general seems to prefer portable slash separator anyway, so this change only aligns the impl. with the tests, docs, etc.

Fixes #1131.

### DIFF
--- a/src/util/FileUtils.cpp
+++ b/src/util/FileUtils.cpp
@@ -270,16 +270,7 @@ string toAbsolutePath(const string& filename, const string base)
 
 string getFilename(const string& path)
 {
-#ifdef _WIN32
-    char pathsep = '\\';
-#else
-    char pathsep = '/';
-#endif
-
-    string::size_type pos = path.find_last_of(pathsep);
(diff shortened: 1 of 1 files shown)


abellgithub:
Would it work just to change to:

```
#ifdef _WIN32
    std::string pathsep("\\/");		
 #else		
   std::string pathsep("/");
#endif	
```
?

mloskot:
New boost-free proposal submitted in #1139

Thanks @abellgithub

PR #1139, opened 2016-02-18
Make getFilename aware of portable path separator on Windows.

### LINKED ISSUE #1131: FileUtilsTest assumes Unix paths failing on Windows

mloskot:
Previously, on Windows, getFilename always looked for preferred
separator only while it should look for both, preferred (backslash)
and portable (slash) which are equally supported by Windows.
PDAL in general seems to prefer portable slash separator anyway,
so this change only aligns the impl. with the tests, docs, etc.

Add more related test cases to FileUtilsTest.

Fixes #1131.

### DIFF
--- a/src/util/FileUtils.cpp
+++ b/src/util/FileUtils.cpp
@@ -271,7 +271,7 @@ string toAbsolutePath(const string& filename, const string base)
 string getFilename(const string& path)
 {
 #ifdef _WIN32
-    char pathsep = '\\';
+    std::string pathsep("\\/");
 #else
     char pathsep = '/';
 #endif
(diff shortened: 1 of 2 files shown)


abellgithub:
Looks fine.

_Wonders why Travis doesn't seem to be running on this._

mloskot:
Hmm, no idea. It's not even listed in https://travis-ci.org/PDAL/PDAL/pull_requests

mloskot:
I added the little `#else` case for Unix as attempt to nudge the Travis and seems it's worked :)

PR #1140, opened 2016-02-18
Compare bundle value up to the length of expected value.

mloskot:
On Windows, for reasons unknown to me, outbuf receives some extra bytes at the end of buffer (ASCII: 10 and 204).

-----
For example, the line `outbuf = Utils::base64_decode(n.find(findbundle1).value());` fills the buffer with the garbage (?) at the end:

```
outbuf.size()
16
outbuf.data(),[outbuf.size()]
0x0000023f4cc2e070 "This is a test\nÌ"
    [0]: 84 'T'
    [1]: 104 'h'
    [2]: 105 'i'
    [3]: 115 's'
    [4]: 32 ' '
    [5]: 105 'i'
    [6]: 115 's'
    [7]: 32 ' '
    [8]: 97 'a'
    [9]: 32 ' '
    [10]: 116 't'
    [11]: 101 'e'
    [12]: 115 's'
    [13]: 116 't'
    [14]: 10 '\n'
    [15]: 204 'Ì'
```

I have no knowledge about BPF, so perhaps this patch actually hides a bug, so I'd appreciate review.

### DIFF
--- a/test/unit/io/bpf/BPFTest.cpp
+++ b/test/unit/io/bpf/BPFTest.cpp
@@ -433,13 +433,13 @@ TEST(BPFTest, bundled)
     auto findbundle1 = [](const MetadataNode& m)
         { return m.name() == "bundle1"; };
     outbuf = Utils::base64_decode(n.find(findbundle1).value());
-    EXPECT_EQ(memcmp(outbuf.data(), "This is a test",
-        outbuf.size() - 1), 0);
+    std::string expect = "This is a test";
+    EXPECT_EQ(memcmp(outbuf.data(), expect.data(), expect.size()), 0);
     auto findbundle2 = [](const MetadataNode& m)
         { return m.name() == "bundle2"; };
     outbuf = Utils::base64_decode(n.find(findbundle2).value());
-    EXPECT_EQ(memcmp(outbuf.data(), "This is another test",
-        outbuf.size() - 1), 0);
+    expect = "This is another test";
+    EXPECT_EQ(memcmp(outbuf.data(), expect.data(), expect.size()), 0);
 }
 
 TEST(BPFTest, inspect)


mloskot:
@abellgithub :+1:

abellgithub:
I pushed a change to reading of files to be bundled that might address the issue, but I don't currently have a test environment.  If you can test without this fix and let me know, I'd appreciate it.

mloskot:
@abellgithub The `outbuf` content has changed now, but it still does not pass the test:

```
outbuf.size()
16
outbuf.data(),[outbuf.size()]
0x00a46d38 "This is a test\r\n"
    [0]: 84 'T'
    [1]: 104 'h'
    [2]: 105 'i'
    [3]: 115 's'
    [4]: 32 ' '
    [5]: 105 'i'
    [6]: 115 's'
    [7]: 32 ' '
    [8]: 97 'a'
    [9]: 32 ' '
    [10]: 116 't'
    [11]: 101 'e'
    [12]: 115 's'
    [13]: 116 't'
    [14]: 13 '\r'
    [15]: 10 '\n'
```

abellgithub:
Maybe I don't understand.  The file that we were testing contains 15 bytes on my system (just a \n terminator).  When you check it out on yours with GIT, do you get 16 bytes?  Does the \n change to a \r\n?  If so, then we either need to change the attribute on the file to binary in GITland or we should use your fix which just ignores the terminators.  Thoughts?

mloskot:
> When you check it out on yours with GIT, do you get 16 bytes?

Yes

> Does the \n change to a \r\n?

Yes!

```
$ file test/data/bpf/bundle1
test/data/bpf/bundle1: ASCII text, with CRLF line terminators
```

It must be due to the default Git for Windows setting:
```
$ git config --get core.autocrlf
true
```

I wonder, shouldn't all the test data be treated as binary? Perhaps an entry like this one in `.gitattributes` would do the trick:
```
/test/data/* -text -diff
```

Obviously, converting `test/data/bpf/bundle1` and `test/data/bpf/bundle2` to LF makes the BPF test pass, obsoleting this PR.

PR #1141, opened 2016-02-19
Allow derivative writer to produce output for multiple primitive types in one command

abellgithub:
@chambbj: Can you look this over please?


chambbj:
:+1:

PR #1158, opened 2016-02-25
Expect one of multiple possible PROJ.4 outputs

mloskot:
PROJ.4 output may change between version and checking all possibilities is easier than determining which one corresponds with particular PROj.4 version.

Clean up float/double mismatch warning.


abellgithub:
waiting on travis fix

PR #1161, opened 2016-02-25
Make StageFactory a sole owner of any created stages.

mloskot:
Remove `Kernel::ownStage` and `m_stages`, cache factory as stage owner.
Remove uses of smart pointers capturing `StageFactory::createStage` result.
Make `StageFactory` the only place where `std::xxx_ptr<Stage>` is allowed.

This introduces a very simple contract:
* always create stage with `StageFactory::createStage`
* interact with stages as an observer
* never become an owner of any stages
* if you need to extend lifetime stages, just extend lifetime of `StageFactory`.

Since there is no point in controlling lifetime of individual stages, especially if used with pipeline, this change should simplify PDAL interface.

### DIFF
--- a/apps/pdal.cpp
+++ b/apps/pdal.cpp
@@ -132,7 +132,7 @@ void outputOptions(std::string const& n)
     // Force plugin loading.
     StageFactory f(false);
 
-    std::unique_ptr<Stage> s(f.createStage(n));
+    Stage* s = f.createStage(n);
     if (!s)
     {
         std::cerr << "Unable to create stage " << n << "\n";
--- a/examples/writing/tutorial.cpp
+++ b/examples/writing/tutorial.cpp
@@ -54,7 +54,7 @@ int main(int argc, char* argv[])
 
     // Set second argument to 'true' to let factory take ownership of
     // stage and facilitate clean up.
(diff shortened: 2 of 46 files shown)


mloskot:
Perhaps someone could have a look at these two direct constructions of stages:
```
plugins\pcl\kernel\PCLKernel.cpp(103):    std::shared_ptr<Stage> pclStage(new PCLBlock());
plugins\pcl\kernel\SmoothKernel.cpp(100):    std::shared_ptr<Stage> smoothStage(new PCLBlock());
```
I guess, they are sound and no need to go via the `StageFactory`, but I'd like to double check.

Second, `shared_ptr` could be replaced there with `unique_ptr`.

chambbj:
@mloskot They should be fine via `StageFactory`.

mloskot:
@chambbj Thanks. I've update the PR.

mloskot:
The PR should be sound now.
The AppVeyor is half-green with one job succeeded
https://ci.appveyor.com/project/hobu/pdal/build/1.0.1337/job/uq63aay3ai0e1v3s
while the other one terminating com,pilation for mysterious reason
https://ci.appveyor.com/project/hobu/pdal/build/1.0.1337/job/exwe9ll85ds4qgo9

PR #1164, opened 2016-02-26
Prefer std::experimental::filesystem if available.

mloskot:
Add `FileUtils::replaceExtension`, thin wrapper for `filesystem::replace_extension`.
Remove remaining uses of `pdalboost::filesystem` and make `FileUtils.cpp` the sole place where `xxx::filesystem` is used.

### DIFF
--- a/include/pdal/util/FileUtils.hpp
+++ b/include/pdal/util/FileUtils.hpp
@@ -106,12 +106,15 @@ namespace FileUtils
     // toAbsolutePath(base)
     PDAL_DLL std::string toAbsolutePath(const std::string& filename,
         const std::string base);
-    
+
     PDAL_DLL void fileTimes(const std::string& filename, struct tm *createTime,
         struct tm *modTime);
 
     /// Return the extension of the filename, including the separator (.).
     PDAL_DLL std::string extension(const std::string& path);
+
+    /// Replaces the extension of the filename with replacement or removes it if the default value is used.
+    PDAL_DLL std::string replaceExtension(std::string const& path, std::string const& replacement = {});
 
     /// Return the filename stripped of the extension.  . and .. are returned
     /// unchanged.
--- a/io/derivative/DerivativeWriter.cpp
+++ b/io/derivative/DerivativeWriter.cpp
@@ -44,8 +44,6 @@
 #include <iostream>
 #include <limits>
 
-#include <boost/filesystem.hpp>
-
 #include "gdal_priv.h" // For File I/O
 #include "gdal_version.h" // For version info
 #include "ogr_spatialref.h"  //For Geographic Information/Transformations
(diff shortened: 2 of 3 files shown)


abellgithub:
What's the reason for this?  Can't get rid of boost in PCL anyway?  Any good reason to hassle with experimental if we can't depend on it existing?

mloskot:
I thought it could be a step towards cutting down on the use of `pdalboost`, so at some point it may be eliminated.
AFAIK, PCL requires external Boost, that's why we install Boost on Travis CI and AppVeyor, am I correct?

VS2015, GCC 5.3.0, clang 3.7.1 all those offer the `filesystem` feature:

```
// g++ test -std=c++11 -lstdc++fs
// clang++ test -std=c++11 -lstdc++fs
#include <experimental/filesystem>
namespace fs = std::experimental::filesystem;
int main() { fs::exists("/usr"); }
```

I mean, it's a loose proposal and I'm not strongly attached to it.
If accepted though, `#if (defined(_MSC_VER) && _MSC_VER >= 1900) || defined(__cpp_lib_experimental_filesystem)` will need to be improved, since the last symbol has been removed from the FS TS document.

hobu:
> Can't get rid of boost in PCL anyway? 

PCL is a common, but not hard, requirement of PDAL. Anything that gets Boost out of the core of PDAL is good as far as I'm concerned.

> Any good reason to hassle with experimental if we can't depend on it existing?

I don't think we should do `experimental` stuff yet.  Is this generally available at c++14?

abellgithub:
On Mon, Feb 29, 2016 at 9:09 AM, Howard Butler <notifications@github.com>
wrote:

> Can't get rid of boost in PCL anyway?
>
> PCL is a common, but not hard, requirement of PDAL. Anything that gets
> Boost out of the core of PDAL is good as far as I'm concerned.
>
To me, this is trading one dependency for two and it makes two set of
things to deal with -- std::experimental and our embedded pdalboost.  I
 guess I just don't see any advantage at this point, though I'm fine with
removing the direct boost::filesystem stuff from the PCL plugins by adding
necessary hooks in FileUtils.

-- 
Andrew Bell
andrew.bell.ia@gmail.com

hobu:
> std::experimental and our embedded pdalboost. 

As I said, I don't think we should do `experimental` stuff. But we should switch over once `filesystem` is standard, and it is convenient to do so. We should trend that way.

mloskot:
On 29 February 2016 at 16:09, Howard Butler <notifications@github.com>
wrote:

> I don't think we should do experimental stuff yet. Is this generally
> available at c++14?
>

"As of C++ 14, the <filesystem> header is not yet a C++ standard, although
it is expected to be standardized in approximately its current form in the
C++ 17 time frame."
~ https://msdn.microsoft.com/en-us/library/hh874694.aspx

Hence, the patch uses the experimental only if available. So, I don't think
any new dependency is actually introduced, rather a built-in C++ feature is
enabled if possible.

PR #1166, opened 2016-02-28
Deduce plugin type from base class

mloskot:
Add PluginType metafunction that allows to deduce plugin type enumerator from base class of the plugin class.
Remove superfluous base class parameter from macros `CREATE_SHARED_PLUGIN` and `CREATE_STATIC_PLUGIN`.
The PluginType makes also a compile-time check if a plugin class is derived from one of the required bases.

### DIFF
--- a/examples/writing-filter/MyFilter.cpp
+++ b/examples/writing-filter/MyFilter.cpp
@@ -15,7 +15,7 @@ static PluginInfo const s_info =
     PluginInfo("filters.name", "My awesome filter",
                "http://link/to/documentation");
 
-CREATE_STATIC_PLUGIN(1, 0, MyFilter, Filter, s_info)
+CREATE_STATIC_PLUGIN(1, 0, MyFilter, s_info)
 
 std::string MyFilter::getName() const
 {
--- a/examples/writing-kernel/MyKernel.cpp
+++ b/examples/writing-kernel/MyKernel.cpp
@@ -23,7 +23,7 @@ namespace pdal {
     "http://link/to/documentation"
   };
 
-  CREATE_SHARED_PLUGIN(1, 0, MyKernel, Kernel, s_info);
+  CREATE_SHARED_PLUGIN(1, 0, MyKernel, s_info);
 
   std::string MyKernel::getName() const { return s_info.name; }
 
--- a/examples/writing-reader/MyReader.cpp
+++ b/examples/writing-reader/MyReader.cpp
@@ -10,7 +10,7 @@ namespace pdal
     "My Awesome Reader",
     "http://link/to/documentation" );
 
-  CREATE_SHARED_PLUGIN(1, 0, MyReader, Reader, s_info)
+  CREATE_SHARED_PLUGIN(1, 0, MyReader, s_info)
 
   std::string MyReader::getName() const { return s_info.name; }
 
--- a/examples/writing-writer/MyWriter.cpp
+++ b/examples/writing-writer/MyWriter.cpp
@@ -14,7 +14,7 @@ namespace pdal
     "My Awesome Writer",
     "http://path/to/documentation" );
 
-  CREATE_SHARED_PLUGIN(1, 0, MyWriter, Writer, s_info);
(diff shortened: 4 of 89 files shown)


hobu:
Not that I'm against this, but renaming breaks a number of plugins that are not actually in the PDAL repository. 

Can we do a rename (plus any other refactoring of this that makes sense -- you had mentioned getting rid of s_info too) and keep the `CREATE_SHARED_PLUGIN` and `CREATE_STATIC_PLUGIN` macros around for a release cycle or two?

mloskot:
I thought about it, but I was not sure about the policy. I mean, it's loose proposal of a slight simplification, no strong attachment. Whatever is the least intrusive option, it can be accepted (refined or not) or it can dies, no problem.

hobu:
I would like to 

1. Keep the hold macros around, even if underneath they do something different, so that existing plugins out in the wild can register using them.
2. Refactor to something better with a different name. Update all drivers/plugins inside of PDAL that we control to use this new approach.
3. Kill the old macros about PDAL 1.6 or 1.7 release timeframe.

mloskot:
@hobu OK, I can apply your suggestions.

Meanwhile, I'd appreciate review of the core proposal (`PluginType` metafunction).
For example, it assumes a plugin class can only derived from single class: `Filter`, `Kernel`, `Reader` or `Writer` but never multi-inherit. 

Regarding the new macros, what about name like `DEFINE_{SHARED|STATIC}_PLUGIN`?

abellgithub:
I guess I don't care.  This doesn't seem like much of an advantage to me, but it also doesn't harm anything.

abellgithub:
Could probably just change things so that you do "ADD_READER", "ADD_WRITER", "ADD_FILTER", "ADD_KERNEL" and dispense with the plugin type altogether.

mloskot:
On 29 February 2016 at 15:51, Andrew Bell <notifications@github.com> wrote:

> This doesn't seem like much of an advantage to me
>

AFAICT, currently, one may create a Reader subclass, but register it as a
filter, for example.
If such errors are possible, then the proposal makes it a bit more error
prone.

abellgithub:
Yes.  I see that.

I'm really non-committal on this.  Is there any advantage in current code?  I don't think so.  Perhaps we should shelve this until we re-look at the plugin stuff in general?  I'd just like not to have to ever edit StageFactory, for instance.  If you want error-prone, that's it.

mloskot:
On 29 February 2016 at 16:11, Andrew Bell <notifications@github.com> wrote:

> Is there any advantage in current code?


Just code removal and interface simplification.


>  Perhaps we should shelve this until we re-look at the plugin stuff in
> general?


Fine by me.


> I'd just like not to have to ever edit StageFactory, for instance.
>

This PR does not touch the StageFactory.

mloskot:
@abellgithub 
> Could probably just change things so that you do "ADD_READER", "ADD_WRITER", "ADD_FILTER", "ADD_KERNEL" and dispense with the plugin type altogether.

To my sense, this does not improve anything I aim with this PR. The four macros would still mean the knowledge of "what kind of a plugin is X" is encoded in at least two places: 1) what is a base class of X and 2) what macro is used to declare X as a plugin

Since it drifts away from my original intentions, I'm dropping the PR. It may be easier to make new one while reworking the `StageFactory`.

abellgithub:
On Mon, Feb 29, 2016 at 11:44 AM, Mateusz Łoskot <notifications@github.com>
wrote:

> @abellgithub <https://github.com/abellgithub>
>
> Could probably just change things so that you do "ADD_READER",
> "ADD_WRITER", "ADD_FILTER", "ADD_KERNEL" and dispense with the plugin type
> altogether.
>
> To my sense, this does not improve anything I aim with this PR. The four
> macros would still mean the knowledge of "what kind of a plugin is X" is
> encoded in at least two places: 1) what is a base class of X and 2) what
> macro is used to declare X as a plugin
>

Yes, I was agreeing in an earlier comment.  Sorry not to be clear.

-- 
Andrew Bell
andrew.bell.ia@gmail.com

PR #1173, opened 2016-03-03
Add two include guards for libgeotiff

gadomski:
If installing w/o libgeotiff, can't build w/o these two guards. I'm not
sure how to fix the "no optional components" test to run w/o geotiff,
since this Docker stuff is a bit opaque to me ATM.

### DIFF
--- a/io/las/LasReader.cpp
+++ b/io/las/LasReader.cpp
@@ -46,7 +46,9 @@
 #include <pdal/util/IStream.hpp>
 #include <pdal/pdal_macros.hpp>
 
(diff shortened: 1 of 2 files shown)


abellgithub:
Seems like the guards should go inside the header file.

gadomski:
Including the file `GeotiffSupport.hpp` when we don't have geotiff seems wonky to me. Also, there's other guards sprinkled throughout that .cpp, which I took as tacit permission to guard up the .cpp.

But that's just my opinion, happy to move them to the header if that makes more sense to you. Either way.

PR #1175, opened 2016-03-04
g++-5 cleanup

gadomski:
I just built out PDAL with g++-5 (Homebrew gcc 5.3.0 --without-multilib) on my mac, and a couple of errors and warnings popped up. This pull request fixes all of them.

99a60df fixes errors so is critical; the other two fix warnings and could be kept or left depending on your preferences.


abellgithub:
You OK with that @gadomski ?

gadomski:
:+1: much better to handle than ignore.

PR #1176, opened 2016-03-09
ACFTB and AIMIDB defaults added to NitfWriter

marops:


marops:
The ACFTB and AIMIDB Tagged Record extension are required in the NITF file in accordance with the specifications. Added appropriate default values, therefore XML template and/or command line arguments are simplified.

Will update documentation with example for pipeline XML file and command line options.

abellgithub:
I'm not sure how setField() works.  Will it overwrite a previous field value if the key is the same?  If so, this seems fine.  Perhaps we should add a test for this?

marops:
setField will overwrite. Therefore it creates the TRE  (ex. acftbTre) then populate it with the defaults using setField. Then it loops through the options (ex. m_acftb) and sets the corresponding option field in the TRE (ex. acftbTre). So if the option field would overwrite the default field.

I have been testing and validating this on our LIDAR data. I concur that a test script should added/updated.

marops:
The some of the defaults in the TRE that are provided by the NITRO NITF lib when instantiating the TRE class are not appropriate defaults in accordance with the NITF specification. The defaults here are appropriate for NITF wrapped LIDAR. It reduces the complexity of the end users pipeline template file and/or command line provided options. 

This means I can use pipeline with one XML file and I only have to provide the readers.nitf.filename, writers.nitf.filenameacquisition date/time on the command line. Before, for each file I would programmaticly generate the xml file then call PDAL pipeline

EXAMPLE test.xml:
<?xml version="1.0" encoding="utf-8"?>
<Pipeline version="1.0">
  <Writer type="writers.nitf">
    
    <Option name="compression">laszip</Options>
    <Option name="forward">all</Option>
    <Option name="acftb">SENSOR_ID:BUCKEY,SENSOR_ID_TYPE:LILN</Option>

    <Reader type="readers.nitf">
      
    </Reader>
  </Writer>
</Pipeline>

EXAMPLE Call:
pdal pipeline -i test.xml --readers.nitf.filename=test.las.ntf --writers.nitf.filename=test.laz.ntf

marops:
EXAMPLE test.xml:
`<?xml version="1.0" encoding="utf-8"?>
<Pipeline version="1.0">
  <Writer type="writers.nitf">
    
    <Option name="compression">laszip</Options>
    <Option name="forward">all</Option>
    <Option name="acftb">SENSOR_ID:BUCKEY,SENSOR_ID_TYPE:LILN</Option>

    <Reader type="readers.nitf">
      
    </Reader>
  </Writer>
</Pipeline>`

PR #1183, opened 2016-03-19
Load mod_spatialite.so on Linux too.

sebastic:
The loadable SQLite3 module is named `mod_spatialite` since 4.2.0, see the related discussion in the [SpatiaLite 4.2.1 RC0 release announcement](https://groups.google.com/d/msg/spatialite-users/jknlgTY9zgU/3p_ESVIeyoIJ).

This change allows the SQLite plugin to build with `libsqlite3-dev` & `libsqlite3-mod-spatialite` on Debian unstable.

### DIFF
--- a/cmake/spatialite.cmake
+++ b/cmake/spatialite.cmake
@@ -1,0 +1,10 @@
+#
+# SpatiaLite cmake configuration
+#
+
+include(FindPkgConfig)
+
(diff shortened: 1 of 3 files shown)


sebastic:
Based on the travis failure, the `lib_extension` should be made conditional to only use `mod_` for SpatiaLite >= 4.2.0.

Issue #1188, opened 2016-03-21
Inconsistent reader behavior - and QuickInfo behavior issue


connormanning:
Here's an oddity where two identically-created Stages can behave differently based on the presence of an "unrelated" third stage.  In particular, I ran into this issue performing these steps:

 - Create a `Reader` and get its `QuickInfo`.  The file for this reader has an SRS issue, so an error message is logged.

- Create a `ReprojectionFilter`, and `prepare` it.  Behind the scenes, this triggers a call to `GlobalEnvironment::initializeGDAL`, which causes the issue in the next step.

- Create another `Reader`, identical to the first, and get its `QuickInfo`.  Instead of the error message being logged, now it will be thrown.

Here's some code and its output: https://gist.github.com/connormanning/2f666606cabd51be5e74
Here's a sample file: https://s3.amazonaws.com/iowa-lidar/IA_LAZlib/02004736.laz

It seems like the second behavior (throwing) is the desired behavior, however I think that QuickInfo specifically should still succeed even with bad SRS info in the file.  For example, in my application I am specifically using externally-supplied SRS info for my projections and ignoring the file's SRS contents completely - so I still need the point count and native bounds even if the SRS can't be determined (empty string would suit my needs just fine).  This aspect is possibly related to https://github.com/PDAL/PDAL/pull/1043.

hobu:
This issue was fixed by #1222

PR #1191, opened 2016-03-22
For inspectable Readers, try-catch around SRS init.

connormanning:
This allows `inspect` to return information other than SRS even if the file has a bad SRS.  See the latter half of #1188.

### DIFF
--- a/io/bpf/BpfReader.cpp
+++ b/io/bpf/BpfReader.cpp
@@ -120,8 +120,16 @@ void BpfReader::initialize()
         code = "EPSG:326" + Utils::toString(zone);
     else
         code = "EPSG:327" + Utils::toString(zone);
-    SpatialReference srs(code);
-    setSpatialReference(srs);
+
+    try
+    {
+        SpatialReference srs(code);
+        setSpatialReference(srs);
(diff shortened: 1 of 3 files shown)


abellgithub:
So there are two things going on during initializeGDAL().

1. We're registering all the drivers.  It makes sense that we'd want to delay this unless you really need it.

2. We're setting up an error handler.  This seems to be something we should do immediately since it doesn't cost anything and there are apparently calls we make that don't depend on drivers at all

However, this would generally mean that when a GDAL error occurs, we're going to throw an exception.  What is being requested is that an exception not be thrown.  This is kind of a general issue -- when do you throw an exception and when don't you.  I don't have any strong feelings on this.  Should we not throw exceptions when doing spatial reference transformations?  I don't know.  This really has nothing to do with preview(), except that the ticket is asking for non-standard behavior (error message instead of exception) for that function.

hobu:
GDAL error handling was tweaked in #1192

connormanning:
Fixed via the combination of ca96bbe62933c129025448149b2f5ab6818d61a9 and #1192.

PR #1197, opened 2016-03-25
portable_endian: Fix BSD->Linux macro translations.

rakuco:
The definitions for `{b,l}etoh{16,32}` were inverted: PDAL uses the Linux
names, but it was the BSD function names that were being redefined.

### DIFF
--- a/include/pdal/util/portable_endian.hpp
+++ b/include/pdal/util/portable_endian.hpp
@@ -49,14 +49,14 @@
                
 #   include <sys/endian.h>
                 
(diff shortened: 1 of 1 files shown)


abellgithub:
This doesn't seem right.  The rest of the file is defining the same values.  What usage in the code of the functions raised a flag?

rakuco:
Oops, you're right, Linux also uses the same calls. The problem is that `betoh16` don't actually exist, so the definitions do not work. I'll send another pull request just removing the macros.

PR #1209, opened 2016-04-01
PipelineKernel should pass verbose/debug settings to PipelineManager …

chambbj:
…(and on down to pipeline readers)

### DIFF
--- a/include/pdal/PipelineManager.hpp
+++ b/include/pdal/PipelineManager.hpp
@@ -60,8 +60,10 @@ public:
             m_progressFd(progressFd)
         {}
 
-    void readPipeline(std::istream& input);
-    void readPipeline(const std::string& filename);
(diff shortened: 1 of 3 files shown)


abellgithub:
Can't we do this a nicer way?  If we're specifying global debug options, can't we stick them in some globalish place rather than treat them as special and have to pass them around?

chambbj:
Well, we've got `setCommonOptions` already, which is what the other kernels use, but it operates on `Options` being defined within the kernels. The `PipelineKernel` is unique in the sense that it doesn't set options within the kernel, but within the `PipelineManager`. So, we could do something different, I suppose, but it's unclear to me who else this would benefit. This feels good enough to me, but I'm open to alternatives.

hobu:
Let's do this for now until we come up with something better.

PR #1211, opened 2016-04-01
Link to system libjsoncpp if available.

sebastic:
The static linking of libjsoncpp triggered a [lintian embedded-library error](https://lintian.debian.org/tags/embedded-library.html) for the PDAL 1.2.0 Debian package build.

The system libjsoncpp should be used when its available, the greyhound plugin already uses it too.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -101,8 +101,17 @@ set(PDAL_BASE_CPP
   ${PDAL_XML_SRC}
   ${PDAL_LAZPERF_SRC}
   ${DB_DRIVER_SRCS}
-  ${PROJECT_SOURCE_DIR}/vendor/jsoncpp-1.6.2/dist/jsoncpp.cpp
 )
+
+find_package(JSONCPP)
+if(JSONCPP_FOUND)
+  include_directories(${JSONCPP_INCLUDE_DIR})
+else(JSONCPP_FOUND)
+  set(PDAL_BASE_CPP
+    ${PDAL_BASE_CPP}
+    ${PROJECT_SOURCE_DIR}/vendor/jsoncpp-1.6.2/dist/jsoncpp.cpp
+  )
+endif(JSONCPP_FOUND)
 
 list (APPEND PDAL_CPP ${PDAL_BASE_CPP} )
 list (APPEND PDAL_HPP ${PDAL_BASE_HPP} )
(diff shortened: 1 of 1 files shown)


hobu:
Older jsoncpp packages cause test failures. jsoncpp-1.6.2+ is known to work. Our testing image uses Ubuntu 15.04, which has 0.6.0~rc2-3.1ubuntu1 installed, which is too old.

Not sure what to do here. Namespace/rename our own copy of jsoncpp? What conflicts were you seeing?

sebastic:
Using the embedded copy when the system provided jsoncpp is too old makes sense. 

jsoncpp-1.6.2+  is available in Debian jessie-backports and later, and in Ubuntu wily & xenial at least.

Testing the available version with `pkg-config` and only using the system provided jsoncpp when it's new enough seems like the way to go.

Embedded copies are bad and should be avoided in a distribution because bug fixes need to be applied to all projects embedding a copy instead of only the shared library itself.

The extended description of the [lintian embedded-library](https://lintian.debian.org/tags/embedded-library.html) tag documents the following:
> The given ELF object appears to have been statically linked to a library. Doing this is strongly discouraged due to the extra work needed by the security team to fix all the extra embedded copies or trigger the package rebuilds, as appropriate.
> 
> If the package uses a modified version of the given library it is highly recommended to coordinate with the library's maintainer to include the changes on the system version of the library.
> 
> Refer to [Debian Policy Manual](https://www.debian.org/doc/debian-policy/) section 4.13 ([Convenience copies of code](https://www.debian.org/doc/debian-policy/ch-source.html#s-embeddedfiles)) for details.

And Debian Policy section 4.13:
>  Some software packages include in their distribution convenience copies of code from other software packages, generally so that users compiling from source don't have to download multiple packages. Debian packages should not make use of these convenience copies unless the included package is explicitly intended to be used in this way.[[29](https://www.debian.org/doc/debian-policy/footnotes.html#f29)] If the included code is already in the Debian archive in the form of a library, the Debian packaging should ensure that binary packages reference the libraries already in Debian and the convenience copy is not used. If the included code is not already in Debian, it should be packaged separately as a prerequisite if possible. [[30](https://www.debian.org/doc/debian-policy/footnotes.html#f30)] 

Besides concerns for security support and code duplication, the addition of ~18 MB of Boost and the several hours of license & copyright review it caused was not fun for example. The jsoncpp codebase is not as large fortunately.

sebastic:
Travis is much happier now that the system jsoncpp is only used when it's at least version 1.6.2.

hobu:
@sebastic I have merged a more complete solution to this in master just recently. If you have opportunity to test it that would be great, but do know that we are now using external libjsoncpp on Travis for all of our builds so the scenario should be more completely tested going foward.

sebastic:
I've updated the Debian package with git snapshot cbe9b08, the external libjsoncpp is detected but configure fails because Nitro is not available. I'll patch out that requirement to test a bit more.

sebastic:
With the additional changes for #1235, #1236 & #1237 I'm quite happy with the external jsoncpp changes that resolve this issue.

PR #1213, opened 2016-04-02
Bump Python module version to 1.2.0.

sebastic:
Please update PDAL on PyPi to 1.2.0 too, version 1.1.0 fails to build with PDAL 1.2.0.


hobu:
I had neglected to add this to the release docs. It is now pushed to PyPI, and I have updated the release doc so we do it with future releases.

PR #1218, opened 2016-04-06
Docker update

gadomski:
The `/PDAL` and `/PRC` build directories weren't being cleaned off of our Docker image; the first commit fixes that problem. I also update the rivlib Dockerfile to use `pdal/pdal`, since `pdal/master` doesn't exist anymore.

### DIFF
--- a/scripts/docker/Dockerfile
+++ b/scripts/docker/Dockerfile
@@ -42,7 +42,7 @@ RUN git clone --depth=1 https://github.com/PDAL/PDAL \
 		.. \
 	&& make -j4 \
 	&& make install \
-    && rm -rf PDAL
+    && rm -rf /PDAL
 
 #RUN pip install packaging \
 #    && pip install PDAL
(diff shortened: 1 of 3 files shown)


chambbj:
While you are at it, IIRC, the dependencies Dockerfile has a similar issue. First of all, the repos end up being cloned into one anothers' build directories, which is kind of weird, but hasn't been an issue. Furthermore, all of the `rm` commands are relative to the build directories (like PDAL and PRC), so I doubt anything is actually ever being removed.

gadomski:
I think all those dependencies get removed [here](https://github.com/PDAL/PDAL/blob/master/scripts/docker/dependencies/Dockerfile#L254), though that's separated from the (ineffectual) `rm` commands in the initial RUNs. I think I'll remove the big `rm` line and make sure each RUN command cleans up after itself.

I agree that the cloning thing is a bit weird, but it seems like it's not causing problems so my instinct is to let it be until it does :-).

PR #1229, opened 2016-04-17
Move stage creation to pipeline manager

### LINKED ISSUE #1221: PipelineReaderJSON unable to parse pipelines created during translate
```console
$ pdal translate 1.2-with-color.las output.laz range --filters.range.limits=Z[400:500] -p foo.json
$ pdal pipeline foo.json
PDAL: PipelineReaderJSON: unable to process pipeline file "foo.json". JSON is invalid.
```

abellgithub:
Stage creation was in Kernel, which made it difficult to handle things consistently in code that wasn't derived from Kernel.  PipelineManager can be instantiated at will and can handle stage creation and options more consistently.

This also fixes #1221 and removes the passing around of debug/log/verbose flags.

### DIFF
--- a/include/pdal/Kernel.hpp
+++ b/include/pdal/Kernel.hpp
@@ -41,7 +41,7 @@
 #include <string>
 #include <vector>
 
-#include <pdal/StageFactory.hpp>
+#include <pdal/PipelineManager.hpp>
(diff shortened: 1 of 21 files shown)


hobu:
Is this one still alive?

PR #1236, opened 2016-04-30
Fix websocketpp detection.

sebastic:
Was removed along with jsoncpp in bd9832cfbd92c1170f31f147cefa619ce1517403.


hobu:
New unreleased version of Greyhound no longer supports websockets, so we probably won't need this going forward.

PR #1241, opened 2016-05-09
Fix #1240, misleading error with no points

### LINKED ISSUE #1240: Zero point file throws misleading error
A file that reads no points throws the following message:

```
PDAL: unable to construct OGR geometry from wkt!
```

It can't make bounding geometries because there's no points. We should state that instead of throwing this misleading error.

hobu:


filters/stats/StatsFilter.cpp
```diff
@@ -173,9 +173,14 @@ void StatsFilter::extractMetadata(PointTableRef table)
 {
     uint32_t position(0);
 
+    point_count_t numPoints(0);
     for (auto di = m_stats.begin(); di != m_stats.end(); ++di)
     {
         const Summary& s = di->second;
+        // Cache the point count of one of the stats. If we
+        // don't have one we're not going to compute some stuff
+        if (!numPoints)
```
abellgithub: It might be a little more clear just to add the point counts.  Then the variable's name would be intuitive.

PR #1244, opened 2016-05-11
Simpler osx linking

gadomski:
Removes some APPLE-specific link path logic, to hopefully make things simpler for PDAL and for downstreams.


abellgithub:
Can you explain this a little bit?

gadomski:
Right now in **master** we [explicitly set `INSTALL_NAME_DIR` on OSX](https://github.com/PDAL/PDAL/blob/master/cmake/macros.cmake#L72-L75) to `@executable_path/../lib`. This produces an installed library with linkages that look like:

```
$ otool -L ~/local/lib/libpdalcpp.dylib
/Users/gadomski/local/lib/libpdalcpp.dylib:
        @executable_path/../lib/libpdalcpp.2.dylib (compatibility version 0.0.0, current version 0.0.0)
        @executable_path/../lib/libpdal_util.2.dylib (compatibility version 0.0.0, current version 0.0.0)
... snip ...
```

There's no problem with this from PDAL's perspective. However, when a downstream project (e.g. entwine) that uses the [default rpath settings](https://cmake.org/Wiki/CMake_RPATH_handling#Default_RPATH_settings) links against PDAL, this creates double links in the resultant libraries:

```
$ otool -L ~/local/lib/libentwine.dylib
/Users/gadomski/local/lib/libentwine.dylib:
        @rpath/libentwine.dylib (compatibility version 0.0.0, current version 0.0.0)
        @executable_path/../lib/libpdalcpp.2.dylib (compatibility version 0.0.0, current version 0.0.0)
        @executable_path/../lib/libpdal_util.2.dylib (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libpdal_util.2.dylib (compatibility version 0.0.0, current version 0.0.0)
... snip ...
```

This doesn't always break things for the downstream, since the two paths usually point to the same place. But it's arguably A Bad Thing just because it's an unnecessary duplication?

I know, a lot of noise for not much fury...

abellgithub:
Thanks for the explanation!

PR #1249, opened 2016-05-13
Fix #1248 -- remove NaN and Inf from JSON output

### LINKED ISSUE #1248: NaN's in JSON output from pdal info
Running `pdal info` on [this file](https://github.com/PDAL/data/blob/master/isprs/samp71-utm.laz), I noticed that it's quite possible to have `nan`'s in the recently added skewness and kurtosis values.

The resulting JSON subsequently doesn't validate. JSON (AFAICT) doesn't address NaN/Inf directly, and it's not treated as a string because it's not quoted.

Perhaps the best solution would be to treat some of these statistics fields as optional fields, and to omit them if they are empty?

hobu:


chambbj:
👍

abellgithub:
I just wonder if this might be confusing.  "Why is there no standard deviation"?  I don't have an immediately better answer unless you want to stick "NaN" in as a string or some such.

hobu:
> unless you want to stick "NaN" in as a string or some such.

Some quick googling wasn't clear about how the different JSON parsers might react to that, which is why I chose to simply remove it.

abellgithub:
That's not a JSON issue.  It's perfectly legal.  There is no notion of "field must be a number."  But I'm not trying to make a deal, either.

PR #1252, opened 2016-05-18
New and modified filters

chambbj:
PDAL's optional dependency on PCL has been a blocker (or at least a hassle) for
a number of users. This commit aims to provide an alternative for this category
of users, implementing a number of high-value PCL filters natively in PDAL. We
now have pure PDAL implementations of:

* radius outlier removal
* statistical outlier removal
* height above ground attribution
* progressive morphological filter (ground segmentation)
* Poisson subsampling

For now, we are keeping the optional PCL plugins around. Should PCL development
gain momentum again, or should a user wish to quickly incorporate some existing
PCL methods, it will be good to have the plumbing in place for PDAL and PCL to
continue to play nicely together.

We are also adding a number of new filters. These include:

* approximate coplanar test
* eigenvalue decomposition
* normal estimation
* rank estimation

In conjunction with the newly added filters, we have added some helper methods
to compute the following metrics on a subset of points (presumably a
neighborhood):

* centroid
* covariance
* rank

Finally, in support of the rank estimation, we have upgraded the vendored
version of Eigen to v3.2.8.


hobu:
Awesome!

PR #1254, opened 2016-06-06
Use real scoped enums

connormanning:
These changes replace all pseudo-scoped enums (where a helper enumeration called `Enum` is needed, and a surrounding namespace is co-opted for the type name) with real ones.  This means that:

 - The noisy `::Enum` attached to many types is no longer needed.  A dimension ID is simply `Dimension::Id`, rather than `Dimension::Id::Enum`.
 - Implicit conversion to native types (like an int) isn't supported.  This disallows you from accidentally comparing a `Dimension::Type` with a `Dimension::Id`, for example.  To convert to native, you can use the `Utils::toNative` helper function.


io/las/LasUtils.cpp
```diff
@@ -43,16 +43,20 @@
...
+namespace
+{
+    using DT = Dimension::Type;
+    const Dimension::Type lastypes[] = {
+        DT::None, DT::Unsigned8, DT::Signed8, DT::Unsigned16, DT::Signed16,
+        DT::Unsigned32, DT::Signed32, DT::Unsigned64, DT::Signed64,
+        DT::Float, DT::Double
+    };
+}
+
```
abellgithub: Don't understand why this was hoisted out of function.  Is it used elsewhere?

abellgithub:
I have no problem with this, of course.  However, I'm not sure that we get a lot of real benefit (other than prettiness) and it changes an interface that others may already be using.

hobu:
The options changes are also a significant change to an existing interface. Let's clean these things up this release and chill out for a while on interface changes after that.

io/las/LasUtils.cpp
connormanning: It's used in both `asType()` and `setType`, and since the `using` is no longer allowed (making it a little noisier) I figured I'd consolidate the init.

hobu:
👍 to merging this so we can catch up external plugins with both Options/Args changes and this at the same time.

chambbj:
👍

abellgithub:
Commited with efa24855cf2bc72b86ee4d418b549ef5aec60b10.

PR #1257, opened 2016-06-22
rxp plugin cleanup (pointers, mostly)

gadomski:
This commit does not bring things up to speed, since PDAL now doesn't
like me providing my own "filename" option (see line 86 of `RxpReader.cpp`). @abellgithub can you advise? I need to store that fileArg for later checking in `initialize`...

### DIFF
--- a/plugins/rxp/test/RxpReaderTest.cpp
+++ b/plugins/rxp/test/RxpReaderTest.cpp
@@ -47,8 +47,7 @@ using namespace pdal;
 Options defaultRxpReaderOptions()
 {
     Options options;
-    Option filename("filename",
-        testDataPath() + "130501_232206_cut.rxp", "");
+    Option filename("filename", testDataPath() + "130501_232206_cut.rxp");
     options.add(filename);
     return options;
 }
@@ -131,7 +130,7 @@ TEST(RxpReaderTest, testRead)
 TEST(RxpReaderTest, testNoPpsSync)
 {
     Options options = defaultRxpReaderOptions();
-    Option syncToPps("sync_to_pps", "false", "");
+    Option syncToPps("sync_to_pps", false);
     options.add(syncToPps);
     RxpReader reader;
     reader.setOptions(options);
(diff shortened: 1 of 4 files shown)


abellgithub:
You can just drop it.  It's done in the base reader code.  I can mod that so that there's a pointer to the filename arg available to subclasses.

Is that what you're wanting?

abellgithub:
I added m_filenameArg and m_countArg.  Let me know if that doesn't give you what you're wanting.

gadomski:
Yeah, `m_filenameArg` should do the trick. I'll bring this branch up-to-date then merge. Thanks!

plugins/rxp/test/RxpReaderTest.cpp
```diff
@@ -131,7 +130,7 @@ TEST(RxpReaderTest, testRead)
 TEST(RxpReaderTest, testNoPpsSync)
 {
     Options options = defaultRxpReaderOptions();
-    Option syncToPps("sync_to_pps", "false", "");
+    Option syncToPps("sync_to_pps", false);
```
gadomski: @abellgithub is this still the accepted way of setting options on a `Reader`? My `sync_to_pps` argument (setting [this option](https://github.com/PDAL/PDAL/pull/1257/files#diff-f95fa921d6512b222740204d336a20faR87)) isn't getting picked up by my reader in this test, and I haven't been able to track down at what point in the option processing this option's value is getting discarded. If this is still the accepted way of setting options, then the problem must be between my keyboard and chair.

abellgithub: It's still the right way.  Booleans are weird.  There could be a bug.  I'll write a test.

abellgithub: The code is broken.  I have to think about how to deal with it.

gadomski: 👌

abellgithub: Try again...

gadomski: It's happy, thanks. I'll merge.

PR #1263, opened 2016-07-01
Update arbiter.  Simplify some of the utils code that uses it.

connormanning:


src/PDALUtils.cpp
```diff
@@ -244,15 +244,8 @@ class ArbiterInStream : public std::ifstream
...
-
-    // If we error, assume that we're local.
-    try
-    {
-        remote = arbiter::Arbiter().isRemote(path);
-    }
-    catch (arbiter::ArbiterError)
-    {}
+    arbiter::Arbiter a;
+    const bool remote(a.hasDriver(path) && a.isRemote(path));
```
abellgithub: If the point of doing things this way is to avoid the exception, then it seems like the better solution is to make a single call that does that.

abellgithub:
Merged with ad3d75f685f0d1ae6ac1c18fa13010db8306bb52

PR #1269, opened 2016-07-17
Update Dockerfile for xenial

msmitherdc:
ubuntu 15.04 is not getting security updates, change pdal/dependencies to use xenial


hobu:
Did it in #1283

PR #1272, opened 2016-07-23
Explicitly assign greyhound prefix, allowing HTTP reads.

connormanning:
Use `greyhound://` to signify the greyhound reader, inferring `http://` paths from their actual file extension.


hobu:
We were going to allow multiple drivers to be able to register they cared about specific prefix/suffix so that something that reads `.txt` files could test multiple drivers. I don't particularly like mucking the URLs for Greyhound (though I understand the hack).

Issue #1273, opened 2016-07-29
Override inferred driver type


hobu:
It is useful to be able to override drivers with a `--driver=readers.text` option in situations where inferDriver might otherwise pick up the wrong driver name.

PR #1274, opened 2016-07-29
allow override of inferred drivers in kernels (specify with --driver argument) #1273

hobu:


abellgithub:
Isn't this kind of a problem if you're dealing with more than one file at a time?  It seems like you need a way to specify a driver for each file.

hobu:
Yes it is still a problem in that scenario, but this puts back functionality that was removed as part of the ProgramArgs refactor.

abellgithub:
Hmmm.  Not sure that was ever there.  I'm pretty sure I didn't remove code like that.  But I have no problem with it.  Could do something positional to make it more robust, I guess.

hobu:
It was this thing in 1.2 
https://github.com/PDAL/PDAL/blob/1.2-maintenance/kernels/info/InfoKernel.cpp#L281

abellgithub:
The change that you made to introduce that was only put on 1.2 maintenance -- never went into master.

hobu:
> The change that you made to introduce that was only put on 1.2 maintenance -- never went into master.

D'oh.

PR #1275, opened 2016-08-05
doc fixes for pipeline and ground

visr:


abellgithub:
Thanks very much for this.  I had noticed it about a week ago but then forgot to fix.

PR #1292, opened 2016-08-28
Fix HDF5 include directory with CMake >= 3.6.0.

sebastic:
Santiago Vila and Lucas Nussbaum reported that PDAL failed to build in [Debian Bug #835668](https://bugs.debian.org/835668) & [#835723](https://bugs.debian.org/835723) respectively.

It turns out that the recent update to CMake 3.6.x caused the failure because the `HDF5_INCLUDE_DIR` variable is no longer set by `FindHDF5.cmake`.

`HDF5_INCLUDE_DIR` has been deprecated some time ago, only `HDF5_INCLUDE_DIRS` is set for CMake >= 3.6.0.

This PR supports both by setting `HDF5_INCLUDE_DIR` to the value of `HDF5_INCLUDE_DIRS` if the former is not defined but the latter is.


hobu:
pick'd it back to the `1.3-maintenance` branch too.

PR #1296, opened 2016-08-30
Handle single-digit UTM zones properly.

jasonoverland:


hobu:
It seems we never had any single-digit zones!

Thanks for the patch.

abellgithub:
This changes the meaning of the '0' entry of the coordinate space type.  The latest documentation I have shows '0' as unknown.  Has the government changed the meaning of this field without telling us?Can you point me at a recent document?

jasonoverland:
What documentation are you looking at?  It's possible I'm reading too much into what the specs mean by "cartesian".  I'm looking at v1.1 of the BPF spec published by NGA and it simply refers to coordinate space type '0' as "cartesian".  I don't know exactly what "cartesian" means but it sounded something along the lines of WGS84?

"The Coordinate Space Type defines the coordinate system and the interpretation of the Coordinate Space ID is type specific. BPF3 supports four coordinate systems: Cartesian, Universal Transverse Mercator (UTM), Terrestrial Centered Rotational (TCR), and East North Up (ENU)."

The BpfReader::initialize code used to turn coordinate space type '0' into EPSG:327 (the first part of a UTM string).  I picked WGS84 as something that seemed closer than that.

jasonoverland:
Here's the docs I'm looking at:
https://nsgreg.nga.mil/doc/view?i=4220&month=8&day=30&year=2016

abellgithub:
That's newer than what I had.  Somehow we fell out of the loop.  I'll have to look through to see if there are other changes that we weren't aware of.

jasonoverland:
Were you looking at the slightly older v1.0 of that doc, or something else entirely?

I did notice that document is prefaced with a note that: "... documents a coherent version of the BPF that clarifies the myriad existing BPF formats."  Myriad isn't a word I like to hear in this context.

hobu:
> Myriad isn't a word I like to hear in this contex

You have no idea...

abellgithub:
If you can fix the cast issue, I can merge this.

jasonoverland:
Ok, I fixed the casting issue.  Sadly I'm woefully unfamiliar with git/github (been using hg for a long time) -- can you squash my commits into one when merging them, or is that something I'll need to do?

abellgithub:
It's fine. Thank you!

PR #1301, opened 2016-09-07
Protect las reader from empty geotiff VLRs

gadomski:
Empty geotiff VLRs would cause invalid memory access (segfault). Includes test case.


abellgithub:
This looks fine to me.

Issue #1305, opened 2016-09-09
tindex broken on windows


kbevers:
I experience a few different issues with tindex on Windows.

### 1. Filespec is not recognized by the tindex kernel

```
C:\dev\pdal-build\bin>pdal tindex index.shp C:\data\colors\*.laz
PDAL: Couldn't find files to index: C:\data\colors\*.laz.

C:\dev\pdal-build\bin>dir /s /b C:\data\colors\
C:\data\colors\1km_6088_554.laz
C:\data\colors\1km_6088_555.laz
```
So, the files do exists but tindex does not find them. Looking at the [source code](https://github.com/PDAL/PDAL/blob/6b1017b06d0ecddcfd7758581703ac26a9a5e80f/kernels/tindex/TIndexKernel.cpp#L185) it makes perfect sense, since the glob-function is only implemented for non-windows systems.

### 2. --stdin option does not add files to index

As a work-around for the problem above instead it should be possible to build the tindex by reading a file-list from stdin:
```
C:\dev\pdal-build\bin>dir /s /b C:\data\colors | pdal tindex index.shp bogus-filespec -s
Warning 6: Field modified create as date field, though DateTime requested.
Warning 6: Field created create as date field, though DateTime requested.
```
A warning is shown, but returns okay otherwise. The index is unfortunately not build:
```
C:\dev\pdal-build\bin>ogrinfo index.shp index
INFO: Open of `index.shp'
      using driver `ESRI Shapefile' successful.

Layer name: index
Geometry: Polygon
Feature Count: 0
Extent: (0.000000, 0.000000) - (0.000000, 0.000000)
Layer SRS WKT:
GEOGCS["GCS_WGS_1984",
    DATUM["WGS_1984",
        SPHEROID["WGS_84",6378137,298.257223563]],
    PRIMEM["Greenwich",0],
    UNIT["Degree",0.017453292519943295]]
``` 
Also, the positional argument ```filespec``` is required but ignored when reading from stdin. At least so it seems from the command line help (have not checked the source).

### Setup
I've built PDAL with the following configuration:
```
set OSGEO4W_ROOT=C:\OSGeo4W64
set BUILD_TYPE=Release

cmake -G "Visual Studio 14 2015 Win64" ^
    -DBUILD_PLUGIN_CPD=OFF ^
    -DBUILD_PLUGIN_GEOWAVE=OFF ^
    -DBUILD_PLUGIN_GREYHOUND=OFF ^
[...]

PR #1311, opened 2016-09-14
Globbing for tindex on Windows

kbevers:
Solves part 1 of https://github.com/PDAL/PDAL/issues/1305.

### DIFF
--- a/kernels/tindex/TIndexKernel.cpp
+++ b/kernels/tindex/TIndexKernel.cpp
@@ -37,6 +37,8 @@
 #ifndef WIN32
 #include <glob.h>
 #include <time.h>
+#else
+#include <windows.h>
 #endif
 
 #include <memory>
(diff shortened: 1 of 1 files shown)


kernels/tindex/TIndexKernel.cpp
```diff
@@ -194,6 +196,30 @@ StringList TIndexKernel::glob(std::string& path)
...
+    if (INVALID_HANDLE_VALUE == handle)
+        return filenames;
+
+    size_t found = path.find_last_of("/\\");
+    std::string dir = path.substr(0, found);
+
+    if (m_absPath)
+    {
+        TCHAR full_dir[4192] = TEXT("");
+        GetFullPathName(dir.c_str(), dir.length(), full_dir, NULL);
```
abellgithub: This doesn't look right.  The second argument is supposed to be the size of the output buffer, which in this case is 4192.

kbevers: Nice catch. I definitely got things mixed up here. I'll fix it right away...

abellgithub:
Thanks for this.

PR #1316, opened 2016-09-28
support passing filters as a JSON string array to pdal translate

hobu:


chambbj:
Haven't tried it yet, but I love the idea.

A nice corollary would be to output a filters-only JSON file from a fully specified `pdal translate` command.

hobu:
> A nice corollary would be to output a filters-only JSON file from a fully specified pdal translate command.

`--pipeline` is available now, which does have the readers/writer on it.

chambbj:
Yes, I was just thinking a filters-only variant on that idea.

hobu:
Currently works like so:

```
pdal translate input.las output.las --json filters.json
```

filters.json
```
[
    {
      "type":"filters.crop",
      "bounds":"([63600,638500],[0,853400],[0,1000000])"
    }
]
```

abellgithub:
I don't understand the point.  Isn't this already supported?

```
pdal pipeline filters.json
```

filters.json
```
[
    input.las
    {
      "type":"filters.crop",
      "bounds":"([63600,638500],[0,853400],[0,1000000])"
    }
    output.las
]

hobu:
yes, it is pretty close to the same thing as `pdal pipeline` minus the full pipeline. Given this PR, there'd now be two ways to achieve the scenario of "here's my filters, substitute appropriately over a bunch of files". 

The first is `pdal pipeline --readers.las.filename=input.las --writers.las.filename=output.las" pipeline.json`. This method has an advantage in that it is very specific. It is a little laborious to override the options. Your pipeline has to have `filename` options for this to work. You can only use `readers.las` and `writers.las`.

The second is to `pdal translate input.las output.las --json filters.json`. This has an advantage of being able to infer input and output drivers. You don't have to substitute. 

Same cat, same skin, different methods.

PR #1322, opened 2016-10-05
Implement filters.colorinterp

hobu:


filters/colorinterp/ColorinterpFilter.cpp
```diff
@@ -0,0 +1,221 @@
...
+void ColorinterpFilter::addArgs(ProgramArgs& args)
+{
+    std::string dimension;
+    args.add("dimension", "Dimension to interpolate", dimension, "Z");
+    args.add("minimum", "Minimum value to use for scaling", m_min);
+    args.add("maximum", "Maximum value to use for scaling", m_max);
+    args.add("ramp", "GDAL-readable color ramp image to use", m_colorramp, "pestel_shades");
+    args.add("invert", "Invert the ramp direction", m_invertRamp, false);
+    args.add("stddev", "Compute minimum/maximum given this many standard deviations", m_stdDevThreshold, 0.0);
+    m_interpDim = Dimension::id(dimension);
```
abellgithub: I don't think this is going to do what you want.  `dimension` isn't assigned this point, other than to perhaps a default.  You should put this in `prepared()` if you need it.  It's also going to silently fail if `dimension` doesn't refer to an available dimension.

filters/colorinterp/ColorinterpFilter.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+void ColorinterpFilter::addDimensions(PointLayoutPtr layout)
+{
+    layout->registerOrAssignDim("Red",
```
abellgithub: These are well-known dimensions.  Just call registerDim(Dimension::Id::Red); ...

Or

layout->registerDims( { Dimension::Id::Red, ...} );

filters/colorinterp/ColorinterpFilter.cpp
```diff
@@ -0,0 +1,221 @@
...
+    gdal::registerDrivers();
+
+    m_raster = openRamp(m_colorramp);
+    m_raster->open();
+
+    log()->get(LogLevel::Debug) << getName() << " raster connection: "
+                                             << m_raster->m_filename << std::endl;
+}
+
+bool ColorinterpFilter::processOne(PointRef& point)
```
abellgithub: `processOne()` assumes streaming mode, but you can't use streaming mode with this filter because you have to compute statistics on all the points.  Should probably rename this so that the system doesn't think you CAN stream with it.

filters/colorinterp/ColorinterpFilter.cpp
```diff
@@ -0,0 +1,221 @@
...
+{
+
+    double v = point.getFieldAs<double>(m_interpDim);
+    double red(1.0), green(1.0), blue(1.0);
```
abellgithub: Probably better just to declare these when first assigned and then you can lose the initialization to 1.

filters/colorinterp/ColorinterpFilter.cpp
```diff
@@ -0,0 +1,221 @@
...
+    return false;
+}
+
+void ColorinterpFilter::filter(PointView& view)
+{
+    if ( m_stdDevThreshold != 0.0)
+    {
+        std::vector<double> values(view.size());
+
+        pdal::stats::Summary summary(pdal::Dimension::name(m_interpDim), pdal::stats::Summary::NoEnum);
```
abellgithub: Seems like if we're going to use this outside of the stats filter, we should hoist it somewhere else.  We could just add the median thing, too.  Would make this a lot cleaner and wouldn't have to double-store the values.

filters/colorinterp/ColorinterpFilter.cpp
```diff
@@ -0,0 +1,221 @@
...
+    "filters.colorinterp",
+    "Assigns RGB colors based on a dimension and a ramp",
+    "http://pdal.io/stages/filters.colorinterp.html" );
+
+CREATE_STATIC_PLUGIN(1, 0, ColorinterpFilter, Filter, s_info)
+
+std::string ColorinterpFilter::getName() const { return s_info.name; }
+
+
+#define GETRAMP(name) \
```
hobu: I wrote a macro because the variable name for the byte arrays is the ramp name, and I'm lazy.

The VSIFILE* that is created just ends up hanging out there, but the bytes it points at of course are our byte arrays in color-ramps.hpp.

filters/colorinterp/ColorinterpFilter.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+        auto median = [](std::vector<double> vals)
+        {
+            std::nth_element(vals.begin(), vals.begin()+vals.size()/2, vals.end());
+
+            return *(vals.begin()+vals.size()/2);
+        };
+
+        double med = median(values);
+        double threshold = (m_stdDevThreshold * summary.stddev());
```
abellgithub: @chambbj: That's confusing.  Have no idea what MAD is, but the magic number on line 101 is evil.  Seems like the whole thing needs some comment if what you're doing is computing some bastardized stdev.  But I guess this is the wrong review for this :)

PR #1323, opened 2016-10-05
Gdalwriter

abellgithub:
Interpolate points to make a GDAL raster.  Replaces p2g if desired.


doc/stages/writers.gdal.rst
```diff
@@ -0,0 +1,123 @@
...
+
+    If a circle with the provided radius doesn't encompass the entire cell,
+    it is possible that some points will not be considered at all, including
+    those that may be within the bounds of the raster cell.
+
+The GDAL writer creates rasters using the data in the 'Z' dimension of the 
+points.  In a :ref:`pipeline_command` one can precede the GDAL writer with
+the ferry filter (:ref:`filters.ferry`) in order to copy data from some
+other dimension to the 'Z' dimension if desired.
+
```
hobu: Is there a reason we can't use any given dimension and default to ``Z``? For example, Intensity rasters are a very common request. `filters.ferry` is a dirty hack that we don't want to institutionalize people into using IMO.

doc/stages/writers.gdal.rst
```diff
@@ -0,0 +1,123 @@
...
+value, a square of rasters surrounding an empty cell, and the value of each
+non-empty surrounding is averaged using inverse distance weighting to provide
+a value for the subject cell.  The value provided for window_size is the
+maximum horizontal or vertical distance that a donor cell may be in order to
+contribute to the subject cell (A window_size of 1 essentially creates a 3x3
+array around the subject cell.  A window_size of 2 creates a 5x5 array, and
+so on.)
+
+Cells that have no value after interpolation are given the empty value of -9999.
+
```
hobu: Maybe a down-the-road feature, but GDALBands have NoData masking capability with their "mask bands". If `writers.gdal` were to provide a mask band, GDAL can do the right thing for NODATA for each driver type. Then we can finally default to something like -9999.

doc/stages/writers.gdal.rst
```diff
@@ -0,0 +1,123 @@
...
+.. code-block:: json
+
+    {
+      "pipeline":[
+        "pdal/test/data/las/autzen_trim.las",
+        {
+          "edge_length", 10,
+          "radius", 14.14,
+          "filename":"outputfile.tif"
```
hobu: Not valid JSON

include/pdal/GDALUtils.hpp
```diff
@@ -46,7 +47,7 @@
 #include <vector>
 
 #include <cpl_port.h>
-#include <gdal.h>
+#include <gdal_priv.h>
```
hobu: We're supposed to avoid including ``gdal_priv.h``. It may be unavoidable though.

io/gdal/GDALGrid.cpp
```diff
@@ -0,0 +1,407 @@
...
+    if (m_outputTypes & statStdDev)
+        m_stdDev.reset(new DataVec(size));
+}
+
+
+int GDALGrid::numBands() const
+{
+    int num = 0;
+
+    if (m_outputTypes & statCount)
```
hobu: Use a regular enum for statCount and friends?

io/gdal/GDALWriter.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    for (PointId idx = 0; idx < view->size(); ++idx)
+    {
+        double x = view->getFieldAs<double>(Dimension::Id::X, idx) -
+            m_bounds.minx;
+        double y = view->getFieldAs<double>(Dimension::Id::Y, idx) -
+            m_bounds.miny;
+        double z = view->getFieldAs<double>(Dimension::Id::Z, idx);
+
+        m_grid->addPoint(x, y, z);
```
hobu: As described above, it would be really handy to be able to specify any dimension here, not just Z.

src/GDALUtils.cpp
```diff
@@ -324,19 +415,151 @@ void BandReader::readBlock(int x, int y, uint8_t *data)
...
+        m_errorMsg = "Requested driver '" + m_drivername + "' does not "
+            "support file creation.";
+        return GDALError::InvalidDriver;
+    }
+
+    std::vector<const char *> opts;
+
+    for (size_t i = 0; i < options.size(); ++i)
+    {
+        if (options[i].find("INTERLEAVE") == 0)
```
hobu: Need a note about this in the documentation. Is the reason this doesn't work is you're copying directly into the array and expect a certain arrangement?

src/GDALUtils.cpp
```diff
@@ -519,24 +738,30 @@ GDALError Raster::computePDALDimensionTypes()
...
+    }
 
-    for (int i=0; i < m_band_count; ++i)
+    for (int i=0; i < m_numBands; ++i)
     {
-        GDALRasterBandH b = GDALGetRasterBand(m_ds, i+1);
+        GDALRasterBandH b = GDALGetRasterBand(m_ds, i + 1);
         if (GDALRasterIO(b, GF_Read, pixel, line, 1, 1,
-            &pix[0], 1, 1, GDT_CFloat64, 0, 0) == CE_None)
+            &pix[0], 1, 1, GDT_Float64, 0, 0) == CE_None)
```
hobu: heh. I wonder how that ever worked...

io/gdal/GDALGrid.cpp
abellgithub: Can't.  They're a bit set.

include/pdal/GDALUtils.hpp
abellgithub: That's where the definitions are.  Seemed weird to me.  That's what the examples show.

doc/stages/writers.gdal.rst
abellgithub: Thought of this as an enhancement.  I think `readers.gdal` does the same thing.

doc/stages/writers.gdal.rst
abellgithub: I don't understand this.

src/GDALUtils.cpp
abellgithub: I'm overriding it.  I'll put a note in the docs.

doc/stages/writers.gdal.rst
abellgithub: Doh.

PR #1329, opened 2016-10-12
Log a stacktrace if PDAL segfaults on Unix

connormanning:
If log-level is Debug, change segfault output to show a demangled stacktrace rather than just `Segmentation fault: 11`.  Sample output shown [here](https://gist.github.com/connormanning/5d1b2444c2eb6e05790d2f1f901970c6).

The StatsFilter tweak here avoids some extraneous logging output at level Debug.  Because `Log::get` produces immediate output even when nothing is streamed in, caching the stream in advance was producing an empty log header.

### DIFF
--- a/apps/pdal.cpp
+++ b/apps/pdal.cpp
@@ -50,7 +50,6 @@
 #include <execinfo.h>
 #include <unistd.h>
 #include <dlfcn.h>
-#include <cxxabi.h>
 #endif
 
 using namespace pdal;
@@ -241,8 +240,7 @@ int main(int argc, char* argv[])
 
             if (dladdr(buffer[i], &info))
             {
-                char* demangled = abi::__cxa_demangle(info.dli_sname, nullptr,
-                        0, &status);
+                const std::string demangled(Utils::demangle(info.dli_sname));
 
                 const std::size_t offset(
                         static_cast<char*>(buffer[i]) -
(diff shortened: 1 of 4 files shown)


apps/pdal.cpp
```diff
@@ -208,17 +215,76 @@ void App::addArgs(ProgramArgs& args)
...
+        std::vector<std::string> lines;
+
+        for (std::size_t i(0); i < size; ++i)
+        {
+            std::string symbol(symbols[i]);
+            Dl_info info;
+
+            if (dladdr(buffer[i], &info))
+            {
+                char* demangled = abi::__cxa_demangle(info.dli_sname, nullptr,
```
abellgithub: We have Utils::demangle.  Returns a string.

apps/pdal.cpp
```diff
@@ -208,17 +215,76 @@ void App::addArgs(ProgramArgs& args)
         m_showOptions);
 }
 
+namespace
+{
```
abellgithub: Why not just capture the logPtr into the sig_t function?  It doesn't look like it's directly used anyplace other than main().

apps/pdal.cpp
```diff
@@ -46,6 +46,13 @@
 #include <string>
```
abellgithub: General: is there some reason not to wait until the log level has been extracted in execute() to install the handler?

apps/pdal.cpp
connormanning: You can't convert a stateful lambda to a function pointer, so I had to hoist it to an accessible scope.  The rename to `logPtr` was to avoid possible conflicts of the name `log` with `cmath`.

apps/pdal.cpp
connormanning: No, my first instinct was just to register it right away.  If that's a more reasonable spot it can certainly be moved there.

PR #1335, opened 2016-10-20
Use GCC builds for Docker containers instead of clang

hobu:


scripts/docker/dependencies/Dockerfile
```diff
@@ -65,6 +69,7 @@ RUN apt-get update && apt-get install -y --fix-missing --no-install-recommends \
         libgdal1-dev \
     && rm -rf /var/lib/apt/lists/*
 
+RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-3.6 20 && update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-3.6 20
```
hobu: clang is used for PCL (only)

Issue #1282, opened 2016-08-23
Any `--polygon` argument should be made to take in GeoJSON


hobu:
Right now most kernels that take in `--polygon` only allow WKT. WKT is hard to generate, and we can now support GeoJSON with the new `pdal::Polygon` class.

hobu:
Added in #1339

PR #1339, opened 2016-10-25
Polygon GeoJSON ingest improvements #1282

hobu:


hobu:
Our users want to pass in both GeoJSON and WKT. This seemed like the place to transparently support both.

PR #1341, opened 2016-10-25
Allow wildcard input to JSON pipelines.

connormanning:
This adds wildcard expansion capability to JSON pipelines.  For example:
```json
{ "pipeline": [
    "~/data/something/*.las",
    "~/data/output/something-merged.las"
] }
```

The current rules for globbing default to a non-recursive match using `*`, and a recursive search can be specified with `**`.  I've put some sample output for various glob results of a nested directory structure [here](https://gist.github.com/connormanning/9220fb61691b981dbdcc8f0a4ea6908a).  Multiple glob indicators within a single wildcard path are supported for a variety of selection types.

I've also added tilde-expansion for filenames in the JSON output as an afterthought (as in the sample above) - if people don't like that it can be nixed.


abellgithub:
I'm going to merge this, but I'm probably going to change some things after to consolidate all the places we have that do this and eliminate the #ifdef stuff.

PR #1342, opened 2016-10-26
Adding writer to a flat PostgreSQL table

oscarmartinezrubi:
The writer uses pqputcopydata in binary form to make it as faster as possible (though it still is slow compared to tool las2pg in liblas)

### DIFF
--- a/doc/stages/writers.pgflatpointcloud.rst
+++ b/doc/stages/writers.pgflatpointcloud.rst
@@ -40,22 +40,10 @@ table
 schema
   Database schema to write to. [Default: **public**]
 
-column
-  Table column names for each of the dimensions specified with output_dims. [Default: same names as output_dims]
-
 overwrite
   To drop the table before writing set to 'true'. To append to the table set to 'false'. [Default: **true**]
-
-scale_x, scale_y, scale_z / offset_x, offset_y, offset_z
-  If ANY of these options are specified the X, Y and Z dimensions are adjusted
-  by subtracting the offset and then dividing the values by the specified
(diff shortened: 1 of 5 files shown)


abellgithub:
Is this a standard way of storing points?  Is it compatible with any other software?

oscarmartinezrubi:
Well, the idea is to use this writer when you want to end up creating pgpointcloud blocks in a different way than what PDAL-pgpointcloud-writer does by default.

Once you import points in a flat table with this new writer, you can create within the DB PC_POINTS:

CREATE TABLE points ( id SERIAL PRIMARY KEY,  pt PCPOINT(1));
INSERT INTO points (pt) SELECT PC_MakePoint(1, ARRAY[x,y,z,intensity]) FROM (SELECT  x,y,z,intensity FROM flat) AS values;

And then you can create PC_PATCHES in any way you want, for example:

CREATE TABLE patches (id SERIAL PRIMARY KEY, pa PCPATCH(1));
INSERT INTO patches (pa) SELECT PC_Patch(pt) FROM points GROUP BY id/10;

So, basically this writer is complementary to the existing pgpointcloud (that is why I put it in the same plugin folder)

hobu:
@oscarmartinezrubi I know you don't have any time to refactor this, but I think we should table this effort until it can be rolled into the existing `writers.pgpointcloud` rather than a separate thing to maintain. Given that people have `las2pg` as an alternative also gives people a way forward.

Thank you for providing the PR. If someone wants this bad enough to pick it up again, they can start improving in the context of the existing driver.

PR #1349, opened 2016-11-01
Add multi-thread in crop filter (via new option num_threads)

oscarmartinezrubi:
PDAL tests pass
Default is to use only 1 thread (as before) but now the user may use more threads via the option num_threads


hobu:
@oscarmartinezrubi Thanks for the pull request. We are going to table multi-threaded access for the moment until we can have a wider discussion about multi-threaded PDAL in general. Other than providing a little thread safety in some upper-level classes such as PluginManager, we haven't made any library-wide statements about how to thread with PDAL. We need to have that discussion before adding threading complexity into PDAL

PR #1352, opened 2016-11-04
Add pointViewToEigen utility function

gadomski:
This function just converts a pointView (with x, y, z) into an x, y, z
Eigen matrix. I've written this so many times I thought I'd just put it
into PDAL.


abellgithub:
Great.  But are you going to use this somewhere other than a test?

gadomski:
Nope. I created it to be a utility function as a part of the API, for downstream users.

abellgithub:
Could you add a comment to that effect?  I'd tend to purge code that seems dead/unused.

gadomski:
Done.

chambbj:
I've done something quite similar to this several other places. Definite utility! Will have to go look to see where it can be used. 👍

PR #1353, opened 2016-11-04
Add GDAL_INCLUDE_DIR to interface include dirs

gadomski:
If GDAL is installed to a non-standard location, anything that uses
`PDALUtils.hpp` will fail b/c it can't find `cpl_port.h`. Since we use
`include_directories` instead of `target_include_directories`, the CMake
library interface is not populated with upstream include directories.
This patch fixes that *for GDAL only*.

A more comprehensive population of the target include interface is
probably wise. IMO the best way to be to switch all usages of
`include_directories` to `target_include_directories`, which is some
work.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -198,6 +198,12 @@ target_link_libraries(${PDAL_BASE_LIB_NAME}
     INTERFACE
         ${PDAL_LIBDIR}
 )
+# This allows downstream projects to use GDAL_INCLUDE_DIR simply by linking
+# against PDAL_BASE_LIB_NAME.
+target_include_directories(${PDAL_BASE_LIB_NAME}
+    INTERFACE
+        ${GDAL_INCLUDE_DIR}
+)
 
 if (WITH_GEOTIFF)
     target_link_libraries(${PDAL_BASE_LIB_NAME} PUBLIC ${GEOTIFF_LIBRARY})


abellgithub:
I use a non-standard GDAL location and have no  trouble.  Not sure I understand.  Is this only relevant for other projects that are using the PDAL cmake config?

gadomski:
> Is this only relevant for other projects that are using the PDAL cmake config?

Yes.

A non-standard GDAL location is fine for PDAL itself, but if you build and install PDAL on your system w/ a non-standard location, then try to use PDAL in a downstream, the PDAL header files expect GDAL to be available. Since PDAL CMake configuration doesn't populate the `INTERFACE_INCLUDE_DIRECTORIES` properties of the `pdalcpp` target, downstreams that link against PDAL:

```cmake
find_package(PDAL REQUIRED)
add_executable(main main.cpp)
target_link_libraries(main PRIVATE pdalcpp)
```

won't get the custom `GDAL_INCLUDE_DIR` when compiling, and so won't find the GDAL headers that are included in the PDAL public headers.

abellgithub:
Will this work if changed from `PUBLIC` to `INTERFACE`, since we're also including the directory with the `include_directories` option?  Either way, I think a comment is in order, because otherwise it looks like it's being included twice.

Agree with changing to `target...()` commands in general.

gadomski:
Done and #1354.

PR #1357, opened 2016-11-07
Eigen utility enhancements

chambbj:
This pull request begins by renaming source and header files for our Eigen utility functions from Eigen to EigenUtils (as we've done with GDALUtils and GEOSUtils previously). We also introduce an `eigen` namespace for these Eigen-based utility functions.

@gadomski Does it impact you to introduce the namespace and rename the files?

Finally, `padMatrix`, `matrixOpen`, `matrixClose`, and `createDSM` had multiple implementations across some of the bare earth filters. These are general purpose enough that it makes sense to move them to EigenUtils.

Apart from Eigen, we move a commonly used clamp function to Utils and make it a template.


gadomski:
> Does it impact you to introduce the namespace and rename the files?

No problem from my end — moar organization ftw 👍 .

PR #1360, opened 2016-11-07
Protect IStream::get from invalid vector address

gadomski:
Empty buffers are not read into at all.

### DIFF
--- a/include/pdal/util/IStream.hpp
+++ b/include/pdal/util/IStream.hpp
@@ -37,6 +37,7 @@
 #include <sys/types.h>
 #include <stdint.h>
 
+#include <cassert>
 #include <fstream>
 #include <memory>
 #include <stack>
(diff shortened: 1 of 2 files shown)


abellgithub:
I'd call this user-error and not worth a check.  Can impact performance as well, though I'm sure it's teeny.

gadomski:
If a las file has an empty geotiff vlr (e.g. `test/data/las/1.2-empty-geotiff-vlrs.las`), the las reader attempts to read w/ an empty buffer: https://github.com/PDAL/PDAL/blob/master/io/las/LasHeader.cpp#L389 and https://github.com/PDAL/PDAL/blob/master/io/las/VariableLengthRecord.cpp#L80. Would you like me to bump this fix up to the las reader (don't try to get the data if the length is zero) rather than protecting in `IStream::get()`?

abellgithub:
That would be my vote.  Perhaps a assertion in get() might also make sense.

gadomski:
Both done.

PR #1366, opened 2016-11-08
Make `pdal pipeline --validate` do something useful

hobu:


kernels/pipeline/PipelineKernel.cpp
```diff
@@ -92,6 +92,16 @@ int PipelineKernel::execute()
         m_progressFd = Utils::openProgress(m_progressFile);
 
     m_manager.readPipeline(m_inputFile);
+
+    if (m_validate)
```
abellgithub: Seems like something less brute-force might be in order.  This is going to open files and stuff like that.

Issue #1332, opened 2016-10-18
Feature: PDAL JVM bindings


lossyrob:
There is currently not much support for reading point cloud data into and writing point cloud data out of JVM languages. By JVM languages I mean programming languages that compile down to JVM bytecode and execute via a Java Virtual Machine (this includes Java, Scala, Clojure, Groovy, and more). JVM bindings for PDAL would fill this gap in the open source ecosystem, and allow JVM-based projects to utilize not only it's point cloud io capabilities, but also it's extensive Pipeline architecture.

Personally, as a developer on the [GeoTrellis](https://github.com/geotrellis/geotrellis) project, I would like to take advantage of PDAL for reading point cloud data into [Apache Spark](http://spark.apache.org/) [RDDs](http://spark.apache.org/docs/latest/programming-guide.html#resilient-distributed-datasets-rdds) for distributed processing. The first step to a point-cloud-on-spark story is data ingestion, and I see utilizing PDAL to be the best path forward for that purpose.

This issue intends to track the creation of JVM bindings for PDAL.
### Proposed Architecture

There are a variety of ways for the JVM to interact with native code out there (e.g. JNI, JNA, JavaCPP). After considering the options, I believe that implementing JNI bindings is the most clear way forward. If anyone has opinions on the matter, let's hear them.

We will need to expose a PDAL pipeline, very similarly to how the python bindings do this (see [Pipeline.cpp](https://github.com/PDAL/PDAL/blob/ff4b63e24ea7aebe2adb7c20dbf024f9191d5b2a/python/pdal/Pipeline.cpp))

We will not be able to directly translate the python binding code to JNI, however, since as we can see from the [return type of the getArrays method](https://github.com/PDAL/PDAL/blob/ff4b63e24ea7aebe2adb7c20dbf024f9191d5b2a/python/pdal/Pipeline.cpp#L44), the Arrays which are referenced are [actually NumPy arrays](https://github.com/PDAL/PDAL/blob/ff4b63e24ea7aebe2adb7c20dbf024f9191d5b2a/src/plang/Array.cpp#L145). Since NumPy is not accessible to the JVM, we will have to construct something separately.
### Issue of language

[...]

PR #1371, opened 2016-11-10
JNI Binings

pomadchin:
It's a WIP PR, created on a very late stage, to have necessary comments on how to organize project and how to perform it in result. Fixes https://github.com/PDAL/PDAL/issues/1332

- [x] code with tests coverage
- [x] docker images update (https://github.com/PDAL/PDAL/pull/1405)
- [x] travis ci update
- [x] HOWTORELEASE update
- [x] deployment, right now it's deployed into my bintray repo, it makes sense to have maven central deployment
- [x] deploy to maven central (release)
- [x] deploy to maven central (snapshot)
- [x] add jni build into general build (see `cmake/macros.cmake` and and the ADD_LIBRARY macro (c))
- [x] avoid double build of jni binary in tests and use already built version in tests


.gitignore
```diff
@@ -149,3 +149,49 @@ test/data/*.json
 test/data/pipeline/*.json
 test/data/filters/*.json
 test/data/plang/*.json
+
```
gadomski: I'd probably prefer not polluting this PR with `.gitignore` changes. If you think the project's `.gitignore` should be updated, that should be done in another PR. You can use a local `.gitignore` to ignore files that relate to you and your editors (e.g. I ignore `build/` locally).

abellgithub:
Is there some reason for this to be a pull request?  It doesn't seem like it's ready for one.  That or it's past the point.  Personally, I'd prefer a small working bit that can be reviewed, then a pull request.  After the change is merged (assuming it is), we can look at updates and improvements.  It's really hard to deal with constant updates to the pull request and it's going to end up being a huge thing to review and evaluate.  Thanks.

hobu:
> Is there some reason for this to be a pull request? 

I mentioned to do it in a PR so I could track progress.

java/core/src/main/scala/com/azavea/pdal/Main.scala
```diff
@@ -0,0 +1,54 @@
+package com.azavea.pdal
+
+import collection.JavaConverters._
+
+/** Test object should be removed from this PR and replaced with tests */
+
+object Main {
+  val iter: Iterator[Int] = Set(2).toIterator
```
lossyrob: Glad to see these manual tests run, now's the time to start pulling them out into unit tests IMO

java/core/src/main/scala/com/azavea/pdal/Main.scala
```diff
@@ -0,0 +1,54 @@
+package com.azavea.pdal
```
lossyrob: We should change the namespacing from this to hobu (since the project is not azaveas).

@hobu we need to namespace, so should we do

```
package com.hobu.pdal
```

or something else?

java/data/las.json
```diff
@@ -0,0 +1,8 @@
+{
+  "pipeline":[
+    {
+      "filename":"/Users/daunnc/subversions/git/github/pomadchin/pdal-jni/data/1.2-with-color.las",
```
lossyrob: Should point to something in top level `test/data` instead?

java/core/src/main/scala/com/azavea/pdal/PointView.scala
```diff
@@ -0,0 +1,13 @@
+package com.azavea.pdal
+
+class PointView extends Native {
```
lossyrob: Needs methods to get dimensions out of packed points, e.g.

```scala
case class PointCloudData(z: Double, value: Double)

val pv: PointView = ???
val valueDim = pv.layout.findDimType("fooValue")

cfor(0)(_ < pv.length, _ + 1) { i =>
  result += {
    // Construct point feature from packed point
    val x = pv.getX(i)
    val y = pv.getY(i)
    val z = pv.getZ(i)
    val value = pv.get(i, valueDim)

    geotrellis.vector.PointFeature(Point(x, y), PointCloudData(z, value))
  }
}
```

java/core/src/main/scala/com/azavea/pdal/Pipeline.scala
```diff
@@ -0,0 +1,27 @@
+package com.azavea.pdal
+
+import ch.jodersky.jni.nativeLoader
+
+@nativeLoader("pdaljni0")
```
lossyrob: We should load outside of the JAR resources, since we won't be able to distribute in this way. How will JNI find the required object file on the system? When we were working with GDAL bindings we needed to do something in java options, a la https://github.com/geotrellis/geotrellis-gdal/blob/master/gdal/build.sbt#L18

java/README.md
```diff
@@ -0,0 +1,10 @@
+# PDAL-JNI Scala bindings
```
lossyrob: Should call the Java bindings, since they are made to be accessed through Java or any JVM language. Would be good to make mention of the fact that we are coding in Scala, but these types are easily accessible in Java.

java/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+[![Join the chat at https://gitter.im/geotrellis/geotrellis](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/geotrellis/geotrellis?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
```
lossyrob: Wrong gitter! Should be PDAL's

java/core/src/main/scala/com/azavea/pdal/Main.scala
hobu: `package io.pdal` or something like that would be fine. No need for hobu, which is just confusing I think.

java/core/src/main/scala/io/pdal/PackedPoints.scala
```diff
@@ -0,0 +1,66 @@
...
+import java.nio.ByteBuffer
+import java.util
+
+import scala.collection.JavaConversions._
+
+/**
+  * PackedPoints abstraction to work with packed point(s) in JVM memory.
+  * SizedDimType contains size and offset for a particular packed point with the current set of dims.
+  **/
+case class PackedPoints(bytes: Array[Byte], dimTypes: util.Map[String, SizedDimType], metadata: String = "", schema: String = "") {
```
lossyrob: @pomadchin  Does it make sense to call this class "PointCloud"? It seems like the main interface into how we work with PointClouds on the JVM, and the fact that they are backed by a byte array of "packed" points feels like an implementation detail that might confuse users (or give less of an eyebrow raise if it was called something that's more in the domain vernacular).

jamesmcclain: Yes, this class appears to be doing roughly the same thing.  The code that I have should work with this class with little modification.

lossyrob: Great, perhaps it makes sense to try and push as much relevant functionality into class so others could use it. That functionality move could maybe happen as separate PRs after we have a working DEM conversion, to test on our side and not complicate development; but if we rename this PointCloud we'll have a naming conflict, which we can get around using the extension methods, or absorbing some of more simple methods into here.

jamesmcclain: If the maintainers are okay with that, then it sounds good to me.

java/core/src/main/scala/io/pdal/Pipeline.scala
```diff
@@ -0,0 +1,23 @@
+package io.pdal
+
+import ch.jodersky.jni.nativeLoader
+
+class Pipeline(val json: String) extends Native {
+  Pipeline // instead of java static
```
lossyrob: "instead of java static" is a little confusion, perhaps

```scala
// reference the object so that the nativeLoader will load the JNI native libraries
```

java/core/src/main/scala/io/pdal/PackedPoints.scala
```diff
@@ -0,0 +1,66 @@
...
+        if (t > bytes.length) bytes.length else t
+      }
+
+      bytes.slice(from, to)
+    }
+  }
+
+  def get(idx: Int, dim: SizedDimType): ByteBuffer = get(idx, dim.dimType.id)
+  def get(idx: Int, dim: DimType): ByteBuffer = get(idx, dim.id)
+  def get(idx: Int, dim: String): ByteBuffer = ByteBuffer.wrap(get(get(idx), dim))
```
lossyrob: We should have versions like `getInt`, `getByte` etc to do the ByteBuffer => Value conversions for the user.

java/core/src/main/scala/io/pdal/PointView.scala
```diff
@@ -0,0 +1,77 @@
+package io.pdal
+
+import java.nio.ByteBuffer
+
+class PointView extends Native {
+  def getPackedPointWithMetadata(idx: Long, metadata: String, schema: String): PackedPoints =
+    getPackedPoint(idx).copy(metadata = metadata, schema = schema)
+  def getPackedPoint(idx: Long): PackedPoints = getPackedPoint(idx, layout.dimTypes())
```
lossyrob: I'm a bit confused about what the difference between a PointView and PackedPoints is. Isn't this "PackedPoints" going to just wrap a single point? The fact that you get a class named `PackedPoints` (or potentially PointCloud) is strange here because you're getting a single point.

pomadchin: `PackedPoints` is a generic name for `Array[Byte]` and all metadata around it, to perform operations on point(s) in JVM memory, so it can be a point with a certain dims number.

The main difference between `PointView` and `PackedPoints` is that `PointView` works as a "proxy class" and works using C++ memory and functions.

java/core/src/main/scala/io/pdal/PackedPoints.scala
pomadchin: Probably to rename it to `PointCloud` makes sense, as the current name may cause semantics confusion.

lossyrob:
@pomadchin those changes address the comments that I had. What are the next steps for this PR?

pomadchin:
@lossyrob if @jamesmcclain wants to make additions he is welcome, after that:
* we need to wait #1405 to be merged in
* after that travis tests settings needs to be updated (to run jni tests)
* HOWTORELEASE file update
* and probably it makes sense to update settings to publish it on maven central

java/core/src/main/scala/io/pdal/PointCloud.scala
```diff
@@ -0,0 +1,101 @@
...
+import java.nio.ByteBuffer
+import java.util
+
+import scala.collection.JavaConversions._
+
+/**
+  * PointCloud abstraction to work with packed point(s) in JVM memory.
+  * SizedDimType contains size and offset for a particular packed point with the current set of dims.
+  **/
+case class PointCloud(bytes: Array[Byte], dimTypes: util.Map[String, SizedDimType], metadata: String = "", schema: String = "") {
```
lossyrob: `metadata` and `schema` shouldn't be on here. The class doesn't use it and it's really a separate concern from this class. Seems like it's here to substitute for a composition, (metadata, schema, pointCloud)

java/core/src/main/scala/io/pdal/Pipeline.scala
```diff
@@ -0,0 +1,23 @@
+package io.pdal
+
+import ch.jodersky.jni.nativeLoader
+
+class Pipeline(val json: String) extends Native {
+  Pipeline // reference the object so that the nativeLoader will load the JNI native libraries
+
+  @native def initialise(): Unit
+  @native def execute(): Unit
+  @native def pointViews(): PointViewIterator
```
lossyrob: should be `getPointViews` for consistency

java/core/src/main/scala/io/pdal/PointView.scala
```diff
@@ -0,0 +1,106 @@
...
+  def getPointCloud(idx: Long): PointCloud = getPointCloud(idx, layout.dimTypes())
+  def getPointCloud(idx: Long, dims: Array[DimType]): PointCloud =
+    PointCloud(
+      bytes       = getPackedPoint(idx, dims),
+      dimTypes    = layout.toSizedDimTypes(dims)
+    )
+
+  def getPointCloud(metadata: String, schema: String): PointCloud =
+    getPointCloud.copy(metadata = metadata, schema = schema)
+  def getPointCloud: PointCloud = getPointCloud(layout.dimTypes())
```
lossyrob: Can't have parameterless methods because of java.

java/core/src/main/scala/io/pdal/PointCloud.scala
```diff
@@ -0,0 +1,101 @@
...
+  val pointSize: Long = dimTypes.values.map(_.size).sum
```
lossyrob: Why is this a long? If one point is more bytes than a long, we got a problem...

pomadchin: @lossyrob native PDAL pointSize method has `std::size_t` return type, made so for consistency.

lossyrob: I think it makes sense to cast it down on the translation, or else clients will always have to translate it themselves (and I can't see wanting this to be a long)

java/core/src/main/scala/io/pdal/PointViewIterator.scala
```diff
@@ -0,0 +1,7 @@
+package io.pdal
+
+class PointViewIterator extends Iterator[PointView] with Native {
```
lossyrob: is `Iterator` a scala type?

hobu:
Master was recently reorganized in a significant way. Hopefully, the file moving will not cause too challenging of a merge, but be prepared for some movement.

lossyrob:
@hobu in terms of planning - is there a timeline for the 1.4 release? We want to get this in for that, but there's some movement going on at GeoTrellis that's keeping our attention from wrapping this up completely. We'll probably have the focus to try to wrap this up second half of December.

We were talking about future iterations of this as well - moving some of the code we find useful into a pdal-scala project under this directory, and also potentially creating a plugin so that GeoTrellis work can be directly accessed in PDAL pipelines (e.g. our Delaunay triangulation DEM creation, still working out the kinks, but it is pretty quality AFAIK).

@pomadchin and I are headed down to Daytona Beach for the OSGeo code sprint, with the current idea that we would hack on the plugin aspects there. Will you be there?

hobu:
> Delaunay triangulation DEM creation

In Java, I presume?

1.4 release is planned for next week.  I will be in Daytona, but I don't know if @abellgithub or @connormanning will be.

lossyrob:
Ah, so we won't make it 1.4 it looks like. 1.5, here we come...

The TIN algorithm is in Scala, yeah. It runs at a decent clip for the JVM though, and comparing the results against las2dem, it looks good. As for timing, I think a rough estimate is 5 million points to create a 2500x4000 raster in about 8 minutes (single thread, on my macbook)

pomadchin:
@lossyrob we can prepare current version to be published with PDAL 1.4 (it's stable and has only basic things), and to make further updates in a separate PR, what do you think?

lossyrob:
@pomadchin - yeah I guess the last thing to do is to publish to maven central. Unclear if there's any other work to do...if not, and it's cool to try to get this merged in, then as long as it doesn't mess with any hot and heavy deadlines ;) I'd say go.

For maven central, you'll have to sign up with sonatype.

Steps for that, follow: http://central.sonatype.org/pages/ossrh-guide.html

Sign up with a jira account that is generic, so we can pass of the credentials to the project maintainers (e.g. pdaldev as user name). Create a jira ticket and say that you need a repo for groupId "io.pdal". They will ask you if you own that domain, @hobu if you could give permission to use that groupId (e.g. "I own the pdal.io domain and give permission for the upcoming sonatype jira ticket to grant access to that groupid), @pomadchin you could link them to the comment in the description. They have a pretty quick turnaround, so might happen tomorrow or Monday. Let me know if you need help with that process.

@hobu also any thoughts on trying to merge this before 1.4 release? We've been testing it a lot on our side, and it seems to be working like a charm. The API will need some working over, and like I said there's more to add, but perhaps getting an initial out the door would be great.

hobu:
> @hobu also any thoughts on trying to merge this before 1.4 release? We've been testing it a lot on our side, and it seems to be working like a charm. The API will need some working over, and like I said there's more to add, but perhaps getting an initial out the door would be great.

Happy to merge whenever you 👍 it.

lossyrob:
Awesome. I'll take some time to do a thorough review once I get a bit, and @pomadchin if you could kick off the sonatype process in the meantime, gracias.

pomadchin:
@hobu can you comment [here](https://issues.sonatype.org/browse/OSSRH-26851) too?  It's a topic to publish PDAL to maven central; +that would give you possibility to control `io.pdal` package your own (if that would be necessary). + @lossyrob

hobu:
done. https://issues.sonatype.org/browse/OSSRH-26851?focusedCommentId=384450&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-384450

lossyrob:
+1 for this. Looks good, we tested it a lot - the API is a little rough, but I think is great for a v1.

PR #1373, opened 2016-11-10
Add a Radial Density filter

chambbj:
This filter creates a new dimension called RadialDensity that records the
density of points within a sphere of specified radius and centered at the
current point.


abellgithub:
Is this really useful on its own?  A density for every point?  Would expanding GDALWriter to voxels (3D) make more sense?

hobu:
> Is this really useful on its own? A density for every point? Would expanding GDALWriter to voxels (3D) make more sense?

This is different than voxelizing the data, and you might use this density factor as an input to some other processing technique further on down the line.

chambbj:
> This is different than voxelizing the data, and you might use this density factor as an input to some other processing technique further on down the line.

What he said.

abellgithub:
I get that.  Seems better as a function instead of a filter.

hobu:
>  Seems better as a function instead of a filter.

I disagree. After computing this with the filter and storing it in output data, a non-PDAL process could certainly take this value and act on its value.  In combination with `filters.range`, this would be useful for throwing out data that is too high or too low of density. If it were buried in a function, we couldn't do those kinds of activities.

PR #1375, opened 2016-11-10
Add a filter to compute the range from the sensor to each return

chambbj:
This only applies to data having frame and pixel numbers, and sensor coords
marked by pixel number equal to -5.


filters/computerange/ComputeRangeFilter.cpp
```diff
@@ -0,0 +1,113 @@
...
+{
+    using namespace Dimension;
+
+    const PointLayoutPtr layout(table.layout());
+    if (layout->findDim("Frame Number") == Id::Unknown)
+        throw pdal_error("ComputeRangeFilter: missing Frame Number dimension in input PointView");
+    if (layout->findDim("Pixel Number") == Id::Unknown)
+        throw pdal_error("ComputeRangeFilter: missing Pixel Number dimension in input PointView");
+
+    m_pixelNumber = layout->registerOrAssignDim("Pixel Number", Type::Double);
```
chambbj: Not writing to these. Just making sure they exist, and then reading from them later.

These have spaces in the names in the BPF dimension labels. If we want to coerce these so that they have no spaces, it seems that should be a separate change to the BPF reader.

filters/computerange/ComputeRangeFilter.cpp
```diff
@@ -0,0 +1,113 @@
...
+    m_frameNumber = layout->registerOrAssignDim("Frame Number", Type::Double);
+}
+
+void ComputeRangeFilter::filter(PointView& view)
+{
+    using namespace Dimension;
+  
+    // Sensor coordinates are provided for each frame. The pixel number -5 is
```
chambbj: I'm not sure how many others have adopted the practice. There is at least one sensor that does this, but the practice may have bled into some of the related programs.

hobu: I think there's at least two classes of sensors that produce data in this fashion.

filters/computerange/ComputeRangeFilter.cpp
abellgithub: Don't understand the registerOrAssign bit.

How about:

```
m_pixelNumber = layout->findDim("Pixel Number");
if (m_pixelNumber == Id:Unknown)
    throw...

etc.

chambbj:
I fear I dorked things up when updating to master. Closing this PR and reopening a clean one.

PR #1381, opened 2016-11-11
Add kernel to compute Hausdorff distance between two point clouds

chambbj:
Just provides a measure of similarity between point clouds.

https://en.wikipedia.org/wiki/Hausdorff_distance

### DIFF
--- a/include/pdal/PDALUtils.hpp
+++ b/include/pdal/PDALUtils.hpp
@@ -54,6 +54,9 @@
 namespace pdal
 {
 class Options;
+class PointView;
+
+typedef std::shared_ptr<PointView> PointViewPtr;
 
 namespace Utils
 {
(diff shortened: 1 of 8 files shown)


abellgithub:
Tests?

hobu:
This can't be used from pipelines, right? Is there a scenario where it could be? Seems like more of a utility function than a filter because of that...

chambbj:
Well I've got it coded as a kernel not a filter anyway. I did start to pull
it out as a utility function to make it easier to test per Andrew's
comment, but I had no plans to use in a pipeline regardless, I'd just call
that distance function from the kernel.

hobu:
d'oh. I didn't even read :)

chambbj:
@abellgithub Tests added.

PR #1361, opened 2016-11-07
Fixes for compiling with Visual Studio 2013.  This addresses

cffk:
(1) lack of support for constexpr
(2) absence of {} as an initializer
(3) need to have consistent return types

Fixing these doesn't detract too much from the expressive power of
C++11.  I recommend adding Visual Studio 2013 and Visusal Studio 2015
compilers to the continuous integration suite!

PR #1383, opened 2016-11-11
Windows 2015 warnings rollup

hobu:
Related to #1361, I got a MSVC 2015 build system going and attempted to clean up all of the warnings it was complaining about so we have a better chance of seeing changes in the output as development continues.


src/SpatialReference.cpp
```diff
@@ -362,7 +362,7 @@ int SpatialReference::calculateZone(double lon, double lat)
     // Everywhere else.
     else
     {
-        zone = floor((lon + 180.0) / 6) + 1;
+        zone = (int) floor((lon + 180.0) / 6) + 1;
```
abellgithub: It complains about THIS?  So silly.  Makes me shudder for thoughts of lint.

filters/transformation/TransformationFilter.cpp
```diff
@@ -54,7 +54,7 @@ std::string TransformationFilter::getName() const { return s_info.name; }
 TransformationMatrix transformationMatrixFromString(const std::string& s)
 {
     std::istringstream iss(s);
-    TransformationMatrix matrix;
+	TransformationMatrix matrix{ 0 };
```
abellgithub: Sometimes the programmer does know better :)

include/pdal/PDALUtils.hpp
```diff
@@ -106,7 +106,7 @@ inline Everything extractDim(Extractor& ext, Dimension::Type type)
 {
     using Type = Dimension::Type;
 
-    Everything e;
+	Everything e{ 0 };
```
abellgithub: This is a bit of performance sensitive code.  The optimizer may eliminate the unnecessary initialization, but I'm not sure.  Certainly this is an unnecessary initialization, though.

abellgithub: Can we just suppress the warning?

filters/transformation/TransformationFilter.cpp
hobu: backed out in @ac3e4fb

Issue #1387, opened 2016-11-14
Add spherical cropping to filters.crop


hobu:
We need to be able to crop with a 2D or 3D point/radius in addition to a bbox.

PR #1388, opened 2016-11-15
support point/radius and sphere/radius cropping #1387

hobu:


filters/crop/CropFilter.cpp
```diff
@@ -68,7 +69,11 @@ void CropFilter::addArgs(ProgramArgs& args)
     args.add("outside", "Whether we keep points inside or outside of the "
         "bounding region", m_cropOutside);
     args.add("a_srs", "Spatial reference for bounding region", m_assignedSrs);
-    args.add("bounds", "Bounds box for cropped points", m_bounds);
+    args.add("bounds", "Point box for cropped points", m_bounds);
+    args.add("point", "Crop within 'distance' from a 2D or 3D point", m_points).
+        setErrorText("Invalid point specification must be in the form \"(1.00, 1.00)\""
```
abellgithub: Might be better if you could combine the coordinate/distance into a single entity.  It would allow for multiples and make the options more clear.

filters/crop/CropFilter.cpp
```diff
@@ -154,6 +154,14 @@ PointViewSet CropFilter::run(PointViewPtr view)
         crop(box.to2d(), *view, *outView);
         viewSet.insert(outView);
     }
+
+    for (auto& point: m_points)
```
abellgithub: `Points` is confusing.  How about `spheres` or something?  Could they be merged into geoms?

filters/crop/CropFilter.cpp
```diff
@@ -199,6 +207,42 @@ void CropFilter::crop(const GeomPkg& g, PointView& input, PointView& output)
     }
 }
 
+void CropFilter::crop(const cropfilter::Point& point, double distance, PointView& input, PointView& output)
+{
+    bool bIs3D = point.is3d();
+
+    for (auto& point: m_points)
```
abellgithub: Aren't you already looping `m_points` in the caller?

filters/crop/Point.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+    if (istr.good())
+    {
+        point.x = x;
+        point.y = y;
+    }
+    return istr;
+}
+
+std::istream& operator>>(std::istream& istr, cropfilter::Point3D& point)
```
abellgithub: I think it would be nice for user if you could tell them what's wrong if you detect an error as you're parsing and you know the problem.

filters/crop/CropFilter.cpp
hobu: indeed.

filters/crop/CropFilter.cpp
hobu: Do you mean having some sort of specification for a 2D/3D point *and* a radius in a single option?

filters/crop/CropFilter.cpp
hobu: I will rename. They can't be merged into geoms because they aren't the same thing.

filters/crop/Point.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+Point3D Point::to3d() const
+{
+    if (m_point.x == LOWEST && m_point.y == LOWEST)
+        return Point3D();
+    return m_point;
+}
+
+
+std::istream& operator>>(std::istream& istr, cropfilter::Point2D& point)
```
abellgithub: Can't we do this all with the well-known text stuff like we do for polygons?

filters/crop/CropFilter.cpp
abellgithub: Yes.  Then you do multiples without order dependencies.  Just think it's cleaner, too.

hobu: If we do the WKT/GeoJSON thing as proposed below, we can't really do the combined thing proposed here. There were no ordering issues because the points are all buffered by the same distance.

filters/crop/Point.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+Point::Point(const std::string& wkt_or_json, SpatialReference ref)
+    : Geometry(wkt_or_json, ref)
+{
+
+}
+
+void Point::update(const std::string& wkt_or_json, SpatialReference ref)
+{
+    bool isJson = wkt_or_json.find("{") != wkt_or_json.npos ||
```
connormanning: `npos` is a [static value](http://en.cppreference.com/w/cpp/string/basic_string/npos), a bit strange to see it called from an instance.

PR #1392, opened 2016-11-16
XML->JSON updates to filter documentation

kbevers:
A few minor details I  stumpled upon today


doc/stages/filters.decimation.rst
```diff
@@ -10,18 +10,22 @@ Example
 
 .. code-block:: xml
```
connormanning: Code block type should probably be `json` as well.

PR #1397, opened 2016-11-20
Experimental python support on windows

kbevers:
A first step on the way to getting a working PDAL python module on windows. So far this is untested but it *SHOULD* work with python3 and msvc14.

This PR adds a windows specific ```pdal-config.bat``` and modifies the python setup-script so it handles the usual windows quirks. 

Unfortunately it seems impossible to support python2 because per default it wants to build extenstions with msvc9 which is not compatible with PDAL. Building with mingw is not an option either, since python's distutils is not compatible with newer versions of mingw (which is needed because C++11).

I am more or less forced to work inside the OSGeo4W64 environment, which is why I haven't been able to test this just yet. The setup script seems to work though, and only fails because of compiler issues. Hopefully OSGeo4W will adopt python3 at some point in the not so distant future (QGIS3.0 will most likely motivate this).

I realize this PR should probably not be merged just yet. I mainly share this because I made some progres towards having PDAL python-bindings on windows. I am hoping that someone has a windows setup with python3 that can test this for me.


kbevers:
Living on the edge, huh @hobu ? :)

hobu:
Why not? (there wasn't anything that was really non-Windows here...)

kbevers:
Untested stuff. It would not be unreasonable to leave it until someone had confirmed that it works. But sure, it shouldn't break anything. There is a good chance it doesn't work either though. At least the python setup. ```pdal-config.bat``` works.

When I have better time I'll make sure to set up a system where I can test it properly.

PR #1398, opened 2016-11-20
Support setting output dimension type and bounds for writers.gdal

hobu:


abellgithub:
I don't think this will work at all.  There's a lot more that has to be done to support various types.  We build the internal grid based on type.  Do tests pass?

abellgithub:
I can work on this, but I don't think it's simple, which is why it wasn't done.

hobu:
no, the tests do not pass.

The Float64 rasters end up being too huge and mostly not needed, but I suppose there are a few situations where you might need 64 bits.  If we could support additionally support Float32, that is probably good enough (don't need all of the other types).

abellgithub:
Most of the raster formats seem to take only things like bytes and ints.  We can make this work better, but it would be nice if it were easy -- if we could tell the supported type from the file/driver, for example.  Alas, GDAL doesn't support that knowledge :(  Still, we can make this better.

PR #1400, opened 2016-11-22
Update PluginManager.cpp

lbartoletti:
Use real name for FreeBSD. Add DragonFly too.


sebastic:
This commit broke support for GNU/kFreeBSD added in #1187 & #1189.

The `__FreeBSD_kernel__` define is specific to the GNU/kFreeBSD flavor, please leave those in place in the future and add `__FreeBSD__` for plain FreeBSD support.

Issue #1146, opened 2016-02-23
writers.derivative GDAL output types


hobu:
I would like to specify any GDAL writeable output format type as an option in `writers.derivative`

abellgithub:
@chambbj: Can this be closed?

chambbj:
With #1402, yes. Did you want to take a look at that? Or should I just merge?

abellgithub:
#1402 is fine, but does that do types for the derivative writer?

chambbj:
Yes, it now passes a user-provided driver type through to GDALUtils to
write the raster - the default is GTiff. No checks or guarantees on whether
or not the driver type supports the output as we briefly discussed on Slack.

abellgithub:
OK.

PR #1402, opened 2016-11-23
DerivativeWriter refactor

chambbj:
All of the derivatives, topographic attributes, etc. have been pulled out of the `DerivativeWriter` and pushed into `EigenUtils`.

As a result of this refactor, the flow of `DerivativeWriter` should be easier to follow -- they all start with a max elevation raster that is "cleaned" and padded at the edges, before the actual attributes are computed for 3x3 blocks of data within the cleaned and padded DSM.

We also have some tests for the individual calculations now, which should be quite easy to expand upon.

Finally, catchment area has been removed as a primitive type as it had never been thoroughly tested. Expect it to return at some point in the future.


chambbj:
Ready for review.

@abellgithub I still haven't bothered to rename any of the matrix/raster algorithms (e.g., from `EigenUtils` to `RasterAlgs` or whatever). I'm just still not where to draw the line between "utils" vs "algorithms". Seems a bit arbitrary in some cases.

chambbj:
Should be pretty easy to drop in a fix for #1146 at the same time.

This would simply be a matter of updating `eigen::writeMatrix` to accept the desired driver name (and subsequently adding the logic to `writers.derivative` to parse the desired driver and pass it downstream to `eigen::writeMatrix`.

io/derivative/DerivativeWriter.cpp
```diff
@@ -138,1322 +119,73 @@ std::string DerivativeWriter::generateFilename(const std::string& primName,
...
-    log()->clearFloat();
+    // Determine the number of rows and columns at the given cell size.
+    size_t cols = ((bounds.maxx - bounds.minx) / m_edgeLength) + 1;
+    size_t rows = ((bounds.maxy - bounds.miny) / m_edgeLength) + 1;
 
-    BOX2D& extent = getBounds();
-    double yMax = extent.miny + m_GRID_SIZE_Y * m_GRID_DIST_Y;
-    log()->get(LogLevel::Debug4) << yMax << ", " << extent.maxy << std::endl;
+    // Begin by creating a DSM of max elevations per XY cell.
+    MatrixXd DSM = createMaxMatrix(*data.get(), rows, cols, m_edgeLength,
```
abellgithub: It's pretty weird to name variables with all caps.

hobu: that's from p2g

PR #1411, opened 2016-11-29
CMake missing PCD writer sources

### LINKED ISSUE #1410: writers.pcd won't initialize
I have a fresh install with `gcc 4.9.3` and cannot get `writers.pcd` to initialize using `pdal pipeline` or `pdal translate`. From the log, the plugin loads (`src/PluginManager.cpp: 418`) but "Failed to initialize" (`src/PluginManager.cpp: 429`).

`readers.pcd` works, so I don't think it is a problem with linking to PCL (1.8.0).

`WITH_BUILD_PCL=ON`
`CMAKE_CXX_FLAGS=-std=c++11`
`BUILD_SHARED_LIBRARIES=ON`

```
pdal translate -i test.las -o test_out.pcd --writers.pcd.xyz=true
(pdal translate Error: 0): Failed to initialize plugin function 
  for plugin '/path/to/lib/libpdal_plugin_writer_pcd.so'.
PDAL: Couldn't create writer stage of type 'writers.pcd'.
```

```
pdal translate -i test.pcd -o test_out.las 
# ... works fine
```

I tried a rolledback version from when I last modified `writers.pcd` and it works fine.

Can anyone confirm this bug?

chambbj:
Fix #1410

### DIFF
--- a/plugins/pcl/CMakeLists.txt
+++ b/plugins/pcl/CMakeLists.txt
@@ -76,32 +76,13 @@ endif()
 #
 # PCD Reader
 #
-set(srcs
-    io/PcdCommon.cpp
-    io/PcdReader.cpp
-)
-
-set(incs
-    io/PcdCommon.hpp
-    io/PcdReader.hpp
-    io/point_types.hpp
-    PCLConversions.hpp
-)
-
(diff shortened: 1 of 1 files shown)


abellgithub:
This says we need some basic tests. Sorry this got missed.

PR #1419, opened 2016-12-06
Blanket include libxml2 include dir

gadomski:
We blanket link against the library a few lines below, so I follow that
lead. IMO it might be more explicit to check against PDAL_HAVE_LIBXML2
first?

cc @abellgithub


abellgithub:
Don't think there's any reason to check.  That variable will be empty if we can't find LIBXML2.

PR #1422, opened 2016-12-07
Improve BPF streaming performance for dimension-major format.

connormanning:
The dimension-major streaming interface was quite slow due to seeking `<number of dimensions>` times per point.  This PR removes the seeks by using a separate stream for each dimension.  This method is approximately 16 times faster.

### DIFF
--- a/io/BpfReader.cpp
+++ b/io/BpfReader.cpp
@@ -269,8 +269,9 @@ void BpfReader::ready(PointTableRef)
 
 void BpfReader::done(PointTableRef)
 {
-     delete m_stream.popStream();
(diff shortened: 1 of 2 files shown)


io/BpfReader.cpp
```diff
@@ -269,8 +269,9 @@ void BpfReader::ready(PointTableRef)
 
 void BpfReader::done(PointTableRef)
 {
-     delete m_stream.popStream();
-     m_stream.close();
+    if (auto s = m_stream.popStream())
```
abellgithub: Why?

PR #1426, opened 2016-12-09
PMF updates

chambbj:
* Revert back to a 2D box search (vice radius) as in the original PCL
  implementation.
* Do not store map of neighbors, which can grow quite large and blow up
  available memory, just take the hit and redo the search in both the
  erosion and dilation steps.

### DIFF
--- a/filters/PMFFilter.cpp
+++ b/filters/PMFFilter.cpp
@@ -35,8 +35,8 @@
 #include "PMFFilter.hpp"
 
 #include <pdal/EigenUtils.hpp>
-#include <pdal/KDIndex.hpp>
 #include <pdal/pdal_macros.hpp>
+#include <pdal/QuadIndex.hpp>
 #include <pdal/util/ProgramArgs.hpp>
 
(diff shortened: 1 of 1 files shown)


filters/PMFFilter.cpp
```diff
@@ -105,7 +105,11 @@ std::vector<double> PMFFilter::morphOpen(PointViewPtr view, float radius)
     // dilate
     for (PointId i = 0; i < np; ++i)
```
abellgithub: Confused by this.  Why do you go through the points twice?  Can't you compute both the local min and local max on one pass through the loop and one fetch of the neighbors?  Maybe I'm missing something.

chambbj:
Yes, just missed the remaining references to `neighborMap` - will fix.

filters/PMFFilter.cpp
chambbj: In the second pass through the points, you are comparing the current elevation to the eroded elevation of the neighbors, which is unknown until you are done with the first pass.

PR #1444, opened 2016-12-15
Add tests for native JSON options.  Convert GreyhoundReader.

connormanning:
#1435 added the capability to parse JSON that was nested as a pipeline option into a string.  This also inadvertently added the ability to natively back a `pdal::Option` with a reference to an underlying `Json::Value`, and it would "just work" even though we weren't really trying to support it.  This removes the need for manual parsing within the Stage, which was the original aim of #1435.  It also makes the Stage code more clear, since you can semantically convey an option as being JSON by its type.

I propose that we standardize this behavior since it is a more natural way for a Stage to accept JSON options.  I've added some tests for the behavior and converted the GreyhoundReader's `filter` option, which previously used the stringified method.  There's no PDAL code change here, so that method still works as well.

For reviewers, it ends up working because the `<<` and `>>` stream operators, which are used for type coercion by ProgramArgs (see the unspecialized versions of `Utils::fromString` and `Utils::toString`), are implemented to serialize/deserialize `Json::Value` objects.

### DIFF
--- a/plugins/greyhound/io/GreyhoundReader.cpp
+++ b/plugins/greyhound/io/GreyhoundReader.cpp
@@ -276,35 +276,24 @@ void GreyhoundReader::initialize(PointTableRef table)
 
     m_queryBounds = toBounds(m_queryBox).intersection(m_fullBounds);
 
-    if (m_filterArg.size() && !parse(m_filterArg).isNull())
-    {
-        m_filterString = m_filterArg;
-    }
-
     if (m_pathsArg.size())
     {
-        Json::Value json;
-        if (m_filterString.size()) json = parse(m_filterString);
-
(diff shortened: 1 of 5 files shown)


abellgithub:
Thanks for this.  I'm not sure where things are with Appveyor.

chambbj:
I think this is fine for your code. I may adopt it in the PCL plugin as well.

What do you mean when you say you'd like to standardize the behavior though? Are you proposing a change to PDAL?

connormanning:
'Standardize' was a poor word choice.  From the wording of #1435 it didn't seem like this behavior was expected, but I was hoping to implement.  Since it seems like an implementation-detail-coincidence that it already works, I just wanted to add some tests and hopefully let it be a supported way of doing things going forward.

Issue #1427, opened 2016-12-11
Allow 'user_data' JSON object on any PDAL Stage


hobu:
JSON doesn't have comments, but we need a way for upstream and downstream applications to communicate textual comments and maybe other JSON objects across pipeline invocations. A generic, dump-whatever-you-want object that can be attached to any/every node should do the trick.

msmitherdc:
👍 
possible to get this in 1.4?

abellgithub:
Don't understand this.  Can someone provide a use case?

msmitherdc:
Yes, Chris Irwin provided some good examples where it would be useful to document in the pipeline why particular params are used especialy when the pipeline is kept around for repeated use.

kbevers:
Also a 👍  from me. I keep a bunch of general use pipeline-files around, some of them fairly complicated. Right now it is only really possible to describe what they do via the filename. A comment at the start of a file would help a lot in that regard. @msmitherdc also has a good point about documenting why certain parameters are used.

hobu:
There's two parts here. The first are simple comments, and the second is extension facility for other PDAL pipeline consumers/producers.

# Comments

The following pipeline translates some RXP files to LAS, and it does so with a few transformation filters. The order of the transformations matter, and it would be very helpful to be able to communicate what each matrix represents. 

```
{
  "pipeline":[
      {
          "type":"readers.rxp",
          "filename": "/data/scan.rxp"
      },
    {
      "type":"filters.transformation",
      "matrix":" 0.886021478 -0.463609330 -0.005685887 1001.767853725 0.463642504 0.885984748 0.008164234 927.201546884 0.001252594 -0.009869906 0.999950507 514.513223273 0.000000000 0.000000000 0.000000000 1.000000000",
      "comment":"Scanner's Own Coordinate System. Measured from blah blah and blah."
    },
    {
      "type":"filters.transformation",
      "matrix":"1.000000000 0.000000000 0.000000000 536000.000000000 0.000000000 1.000000000 0.000000000 7356000.000000000 0.000000000 0.000000000 1.000000000 0.000000000 0.000000000 0.000000000 0.000000000 1.000000000",
      "comment":"Geo Coordinate System. Measured from blah blah and blah."
    },
    {
      "type":"filters.colorinterp",
      "ramp":"pestel_shades",
      "minimum":0,
      "maximum":300
    } ,
    {
      "type":"writers.las",
      "a_srs":"EPSG:32624",
[...]

PR #1451, opened 2016-12-16
Allow 'application' JSON option to be place on any stage and support serialization #1427 -- #needstests

hobu:


pdal/Options.cpp
```diff
@@ -50,7 +50,12 @@ std::string Option::toArg() const
 
 void Option::toMetadata(MetadataNode& parent) const
 {
-    parent.add(getName(), getValue());
+    // 'application' nodes on Stages are JSON
+    if (!Utils::iequals(getName(), "application"))
```
abellgithub: Names are limited to lowercase (and in any case, aren't matched in a case sensitive way), so you can just do getName() == "application"

PR #1453, opened 2016-12-21
Add filter to locate max value in a given dimension

chambbj:
If many points share the max value, this simply gives you the first one.


chambbj:
@abellgithub Good thought. Done.

PR #1460, opened 2016-12-29
How to use of proj strings in filters.reprojection

kbevers:
Using Proj.4 to do coordinate transformation was previously undocumented. I have added an example that explains how to do a vertical gridshift with Proj.4.


hobu:
Actually any SRS string that GDAL's SetFromUserInput method can consume can be put in the out_srs (or any PDAL spatial reference entry).

Thanks for the docs!

kbevers:
Thanks for merging this.

PR #1471, opened 2017-01-12
Add 'metadata', 'schema', and 'spatialreference' variables to filters.programmable and filters.predicate Python filter functions so they can take advantage of this kind of information in their processing chains.

hobu:


pdal/plang/Environment.cpp
```diff
@@ -279,31 +279,38 @@ std::string readPythonString(PyObject* list, Py_ssize_t index)
...
         return;
+    }
 
-    for (Py_ssize_t i = 0; i < PyList_Size(list); ++i)
-    {
-        PyObject *tuple = PyList_GetItem(list, i);
-        if (!PyTuple_Check(tuple) || PyTuple_Size(tuple) != 5)
-            continue;
+    if (!PyDict_Check(dict) )
+        throw pdal::pdal_error("'metadata' member must be a dictionary!");
```
abellgithub: I've been trying to make sure that all errors have a stage name attached if possible:

getName() + ": blah"

Perhaps we should standardize.

pdal/plang/Invocation.cpp
```diff
@@ -268,22 +265,145 @@ bool Invocation::execute()
...
     if (numArgs > 1)
         PyTuple_SetItem(m_scriptArgs, 1, m_varsOut);
-    if (numArgs > 2)
-        PyTuple_SetItem(m_scriptArgs, 2, m_metaIn);
-    if (numArgs > 3)
-        PyTuple_SetItem(m_scriptArgs, 3, m_metaOut);
 
-    m_scriptResult = PyObject_CallObject(m_function, m_scriptArgs);
+
+    int success = PyModule_AddObject(m_module, "metadata", m_metadata_PyObject);
```
abellgithub: These strike me as kind of pointless error handling.  A user can't do anything with the info and a developer is going to get a crash to examine pretty quickly.  If this kind of thing fails, something is really wrong that has nothing to do with metadata.

hobu: This is Python's pattern though. If this PyModule_AddObject method returns NULL, an exception will be set on the thread's exception stack, and PDAL goes and does some work to emit that back to the user.

PR #1483, opened 2017-01-31
Strip whitespace from dimension names in text reader

gadomski:
Includes regression test.

I briefly grepped for a `rstrip` function in the PDAL codebase and didn't find one (not saying there isn't one), so I used this dumb-as-nails as a quick solution. @abellgithub if you have a preferred whitespace-stripping method let me know and I'll change to that.


abellgithub:
What was the problem, exactly?

gadomski:
A header like `X,Y,Z,Intensity` with dos line endings would leave the trailing `\r` on the intensity field, making the dimension name `Intensity\r`.

abellgithub:
Won't adding "false" to the openFile (text mode) fix the problem?
https://github.com/PDAL/PDAL/blob/master/io/TextReader.cpp#L57

gadomski:
It didn't for me.

/me force-pushes a change to the branch to include a `.gitattributes` file that *hopefully* will keep the crlf line endings in the git repo.

abellgithub:
Well, I think the fix is with the line endings, not general dimension names.  I'll work on it and make sure your regression is OK.

gadomski:
Ok. Try testing my regression test w/o my `TextReader.cpp` fix to make sure the line endings come through the git ok.

abellgithub:
If you put DOS text files on UNIX, you're likely to have problems, right?  Setting the file open mode to text should fix on Windows, but I'm not sure we need to do anything if you have a DOS file on Unix -- seems like you should convert the file when bringing it over (ASCII mode FTP or whatever).

gadomski:
In the wild, there's going to be text files with dos line endings on people's unix-y systems. They could get them by downloading them off of the internet, as the output of a poorly-written tool, or whatever. And not all downstream users will have the chops to easily recognize and repair line ending issues, possibly leading to bug spam.

A warning/error when a dimension name has a newline-y character in it seems like a reasonable solution to me as well, since (again, IMO) it's unlikely a user wants to have newlines in a dimension name. Heck, we could explicitly forbid it 😈.

abellgithub:
I just wrote some code to check for _invalid_ dimension names and it blew up because there are lots of invalid dimension names (don't conform to *my* notion of reasonable) even in our tests.  So I guess I'm not sure why we should restrict.  This is a weird case.  Is the warning the best thing?  Special case for text reader where we just strip \r off the end if it appears?

I don't know.  At some point I think a user has to take responsibility.  It's not like it's hard to figure out how to change /r/n to /n with a quick Google.  At this point, I'll defer to you.  Let me know and I'll fix.

gadomski:
This patch is my "good enough" solution, so if you're not against it go ahead and merge this PR.

kbevers:
I'm afraid the changes to .`gitattributes` causes git to act up a bit now. I just did a clean clone of the repo and then tried to check out an earlier commit:

```
$ git clone https://github.com/PDAL/PDAL.git pdal
Cloning into 'pdal'...
remote: Counting objects: 86940, done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 86940 (delta 0), reused 0 (delta 0), pack-reused 86938
Receiving objects: 100% (86940/86940), 98.70 MiB | 21.95 MiB/s, done.
Resolving deltas: 100% (53277/53277), done.
Checking out files: 100% (5015/5015), done.

$ cd pdal/

$ git checkout 1.4.0
error: Your local changes to the following files would be overwritten by checkout:
        test/data/text/intensity.txt
Please, commit your changes or stash them before you can switch branches.
Aborting

$ git status
# On branch master
# Changes not staged for commit:
#   (use "git add <file>..." to update what will be committed)
#   (use "git checkout -- <file>..." to discard changes in working directory)
#
#       modified:   test/data/text/intensity.txt
#
no changes added to commit (use "git add" and/or "git commit -a")

$ git diff test/data/text/intensity.txt
diff --git a/test/data/text/intensity.txt b/test/data/text/intensity.txt
index 29f4d5c..0f641c7 100644
--- a/test/data/text/intensity.txt
+++ b/test/data/text/intensity.txt
@@ -1,4 +1,4 @@
-X,Y,Z,Intensity
+X,Y,Z,Intensity
 289814.15,4320978.61,170.76,0
 289814.64,4320978.84,170.76,1
 289815.12,4320979.06,170.75,2
```
Stashing the local changes doesn't work, so effectively I am locked out from going back in time. This is on a linux computer. On windows it works perfectly.

I can change `.gitattributes` locally to fix the problem, I guess. I'm guessing there's a better way. Ideas?

PR #1486, opened 2017-02-01
Reader/writer short names in translate kernel

gadomski:
In the translate kernel, filters can already be specified by their short name, e.g. `-f decimation`. This adds the same support for readers and writers:

```
build/bin/pdal translate -i test/data/las/simple.las -o /dev/null -r las -w las -f decimation --filters.decimation.step=2
```


abellgithub:
Seems fine.  BTW, there is writers.null if you want that.

PR #1488, opened 2017-02-03
Improve quality of derivative writer outputs

chambbj:
First, we had a bug in how we computed slope for the D8 method. We were
dividing the slope in the cardinal directions by sqrt(2) instead of the
diagonals.

Second, we offer a switch to output slopes in degrees, not just percent. Note,
this is really just a convenience as analysts should have been able to convert
in the GIS tool of their choice.

Finally, our DSM generation approach had a tendency to leave voids in the max
elevation matrix. We switched to something very similar to what the GDAL writer
does, that is, specifying a radius and considering all points within the
cylinder centered at the current grid cell. At the same time, we remove the DSM
cleaning step.

Future work should consider unifying the behavior of GDALGrid/GDALWriter with
DerivativeWriter such that we can leverage the existing implementation and hole
filling capabilities. We may also be able to convert this to support streaming.

### DIFF
--- a/io/DerivativeWriter.cpp
+++ b/io/DerivativeWriter.cpp
@@ -163,13 +163,13 @@ void DerivativeWriter::write(const PointViewPtr data)
                 {
                     out(r-1, c-1) = computeSlopeD8(block, m_edgeLength);
                     if (Utils::iequals(m_slopeUnit, "degrees"))
-                        out(r-1, c-1) = std::atan(out(r-1, c-1) / 100.0) * 180.0 / 3.14159;
+                        out(r-1, c-1) = percentSlopeToDegrees(out(r-1, c-1));
                 }
                 if (to.m_type == SLOPE_FD)
                 {
                     out(r-1, c-1) = computeSlopeFD(block, m_edgeLength);
                     if (Utils::iequals(m_slopeUnit, "degrees"))
-                        out(r-1, c-1) = std::atan(out(r-1, c-1) / 100.0) * 180.0 / 3.14159;
+                        out(r-1, c-1) = percentSlopeToDegrees(out(r-1, c-1));
                 }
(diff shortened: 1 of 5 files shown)


io/DerivativeWriter.cpp
```diff
@@ -161,9 +160,17 @@ void DerivativeWriter::write(const PointViewPtr data)
                     continue;
                 Matrix3d block = paddedDSM.block(r-1, c-1, 3, 3);
                 if (to.m_type == SLOPE_D8)
+                {
```
abellgithub: Maybe a comment to explain the math so one doesn't have to work it out?

abellgithub: Also, I think there's a predefined PI somewhere.

chambbj: I'll push a comment. Couldn't find the predefined PI though. Maybe I'm missing something. We could use this in a handful of places, if it indeed exists.

abellgithub: May not be portable, but:
http://stackoverflow.com/questions/1727881/how-to-use-the-pi-constant-in-c

abellgithub: Not a big deal, though.  It's pretty obvious what it is.

chambbj: I guess I thought you meant that we had it defined within PDAL already. I also always forget about the `atan` and `acos` "tricks" to obtaining PI.

I'll probably just make a percent slope to degrees slope function to document what's going on.

chambbj: @abellgithub Take a look now. Any better? Or just different.

abellgithub: Nice!  Thanks.

PR #1492, opened 2017-02-11
PDAL scala

pomadchin:
PDAL Scala support

Adds Pipeline DSL with some hand written types for JSON generation.
DSL Coverage is not complete, for that purposes was introduced a special type which allows to write in a raw JSON.

- [x] Pipeline DSL
- [x] ~Pipeline DSL Generation~
- [x] ~Exclude generated files from index~
- [x] ~More Scala code on top of all jni types~
- [x] Travis CI
- [x] Snapshot
- [ ] Publishment 
  - [x] 1.5.0-RC1
  - [x] 1.5.0-RC2
  - [x] 1.5.0-RC3
  - [ ] 1.5.0
  - [ ] 1.6.0

PR #1502, opened 2017-02-24
[WIP] Plugins params description expansion

pomadchin:
For DSL generation (required here: https://github.com/PDAL/PDAL/pull/1492) it is required to have additional information about params: `type` and is field `required` or not.

The idea to make the following json document: 

```json
{
	"plugin" :
	[
		{
			"description" : "<description>",
			"name" : "<field name>",
			"default" : "<default value>",
			"required" : "{true | false}",
			"type" : "<string name of a type>"
		}
	]
}
```

_NOTE: There can be a small confusion about required field: the field can be required if another field is missing. In terms of DSL generation it means that such fields are _optional_. Probably it's enough to say some words about this dependency in docs or in a description field._

- [ ] Dirty version implementation
- [ ] Refactor according to reviews

So there we have a problem in identifying the type of argument: we can try it to convert basing on args type during compilation: https://github.com/pomadchin/PDAL/blob/feature/richer-showjson/pdal/util/TypeName.hpp or we can make additional functions overload with possibility to specify custom / hardcoded type which we want to be visible in the output json / strings.
Both variants are possible at the same time. 
Some questions:
+ can I find somewhere a list of required fields for all plugins?
+ what approach should be used to modify current code? as multiple overloads of all constructors of all types looks like not the best idea

### DIFF
--- a/pdal/Reader.cpp
+++ b/pdal/Reader.cpp
@@ -40,7 +40,7 @@ namespace pdal
 
 void Reader::readerAddArgs(ProgramArgs& args)
 {
-    m_filenameArg = &args.add("filename", "Name of file to read", m_filename);
+    m_filenameArg = &args.addRequired("filename", "Name of file to read", m_filename);
     m_countArg = &args.add("count", "Maximum number of points read", m_count,
         std::numeric_limits<point_count_t>::max());
 }
--- a/pdal/util/ProgramArgs.hpp
+++ b/pdal/util/ProgramArgs.hpp
@@ -25,8 +25,10 @@
 #include <map>
 #include <memory>
 #include <vector>
+#include <typeinfo>
 
 #include <pdal/util/Utils.hpp>
+#include <pdal/util/TypeName.hpp>
 
 namespace pdal
 {
@@ -152,9 +154,15 @@ protected:
       \param description  Argument description.
     */
     Arg(const std::string& longname, const std::string& shortname,
-        const std::string& description) : m_longname(longname),
+        const std::string& description, const bool required) : m_longname(longname),
         m_shortname(shortname), m_description(description), m_set(false),
-        m_hidden(false), m_positional(PosType::None)
+        m_hidden(false), m_positional(PosType::None), m_required(required), m_valType(TypeName<std::string>::get())
+    {}
+
+    Arg(const std::string& longname, const std::string& shortname,
+        const std::string& description, const bool required, const std::string& valType) : m_longname(longname),
+        m_shortname(shortname), m_description(description), m_set(false),
+        m_hidden(false), m_positional(PosType::None), m_required(required), m_valType(valType)
     {}
 
 public:
@@ -299,6 +307,22 @@ public:
         { return m_longname; }
 
     /**
+      Return whether argument is required.
+
+      \return  By default arguments are not required.
+    */
+    bool isRequired() const
+        { return m_required; }
+
+    /**
+      Return whether argument type in a string.
+
+      \return  By default returns string.
+    */
+    std::string valType() const
+        { return m_valType; }
+
+    /**
       Returns text indicating the longname and shortname of the option
       suitable for displaying in help information.
       \note  Not intended to be called from user code.
@@ -334,6 +358,8 @@ protected:
     bool m_set;
     bool m_hidden;
     PosType m_positional;
+    bool m_required;
+    std::string m_valType;
     std::string m_error;
 };
 
@@ -356,10 +382,11 @@ public:
       \param variable  Variable to which the value of the argument should
         be bound.
       \param def  Default value to be assigned to the bound variable.
+      \param required  Is field required or not.
     */
     TArg(const std::string& longname, const std::string& shortname,
-        const std::string& description, T& variable, T def) :
-        Arg(longname, shortname, description), m_var(variable),
+         const std::string& description, T& variable, T def, const bool required) :
+        Arg(longname, shortname, description, required, TypeName<T>::get()), m_var(variable),
         m_defaultVal(def), m_defaultProvided(true)
     { m_var = m_defaultVal; }
 
@@ -373,10 +400,11 @@ public:
       \param description  Argument description.
       \param variable  Variable to which the value of the argument should
         be bound.
+      \param required  Is field required or not.
     */
     TArg(const std::string& longname, const std::string& shortname,
-        const std::string& description, T& variable) :
-        Arg(longname, shortname, description), m_var(variable),
+         const std::string& description, T& variable, const bool required) :
+        Arg(longname, shortname, description, required, TypeName<T>::get()), m_var(variable),
         m_defaultVal(T()), m_defaultProvided(false)
     { m_var = m_defaultVal; }
 
@@ -426,6 +454,8 @@ public:
     virtual void reset()
     {
         m_var = m_defaultVal;
+        m_valType = TypeName<T>::get();
+        m_required = false;
         m_set = false;
         m_hidden = false;
     }
@@ -499,11 +529,12 @@ public:
       \param variable  bool variable to which the value of the argument should
         be bound.
       \param def  Default value to be assigned to the bound variable.
+      \param required  Is field required or not.
     */
     TArg(const std::string& longname, const std::string& shortname,
-        const std::string& description, bool& variable, bool def) :
-        Arg(longname, shortname, description), m_val(variable),
-        m_defaultVal(def), m_defaultProvided(true)
+         const std::string& description, bool& variable, bool def, const bool required) :
+         Arg(longname, shortname, description, required, TypeName<bool>::get()), m_val(variable),
+         m_defaultVal(def), m_defaultProvided(true)
     { m_val = m_defaultVal; }
 
     /**
@@ -516,11 +547,12 @@ public:
       \param description  Argument description.
       \param variable  bool variable to which the value of the argument should
         be bound.
+      \param required  Is field required or not.
     */
     TArg(const std::string& longname, const std::string& shortname,
-        const std::string& description, bool& variable) :
-        Arg(longname, shortname, description), m_val(variable),
-        m_defaultVal(false), m_defaultProvided(false)
+         const std::string& description, bool& variable, const bool required) :
+         Arg(longname, shortname, description, required, TypeName<bool>::get()), m_val(variable),
+         m_defaultVal(false), m_defaultProvided(false)
     { m_val = m_defaultVal; }
 
     /**
@@ -568,6 +600,8 @@ public:
     virtual void reset()
     {
         m_val = m_defaultVal;
+        m_valType = TypeName<bool>::get();
+        m_required = false;
         m_set = false;
         m_hidden = false;
     }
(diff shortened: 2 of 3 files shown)


hobu:
> can I find somewhere a list of required fields for all plugins?

There is no annotation by the drivers as to which arguments are required and which are not.

pomadchin:
@hobu )`: than ok, it's the first try to introduce this annotation + each argument try; I'll walk through docs, looks like the best idea.

pdal/Reader.cpp
```diff
@@ -40,7 +40,7 @@ namespace pdal
 
 void Reader::readerAddArgs(ProgramArgs& args)
 {
-    m_filenameArg = &args.add("filename", "Name of file to read", m_filename);
+    m_filenameArg = &args.addRequired("filename", "Name of file to read", m_filename);
```
pomadchin: Just an example of usage, though we can just add params to the current add function.

pdal/util/ProgramArgs.hpp
```diff
@@ -516,11 +547,12 @@ class TArg<bool> : public Arg
...
         be bound.
+      \param required  Is field required or not.
     */
     TArg(const std::string& longname, const std::string& shortname,
-        const std::string& description, bool& variable) :
-        Arg(longname, shortname, description), m_val(variable),
-        m_defaultVal(false), m_defaultProvided(false)
+         const std::string& description, bool& variable, const bool required) :
+         Arg(longname, shortname, description, required, TypeName<bool>::get()), m_val(variable),
+         m_defaultVal(false), m_defaultProvided(false)
```
pomadchin: It doesn't look like a good idea to change every constructor, but if it's a good approach than ok.

pdal/util/ProgramArgs.hpp
```diff
@@ -356,10 +382,11 @@ class TArg : public Arg
       \param variable  Variable to which the value of the argument should
         be bound.
       \param def  Default value to be assigned to the bound variable.
+      \param required  Is field required or not.
     */
     TArg(const std::string& longname, const std::string& shortname,
-        const std::string& description, T& variable, T def) :
-        Arg(longname, shortname, description), m_var(variable),
+         const std::string& description, T& variable, T def, const bool required) :
+        Arg(longname, shortname, description, required, TypeName<T>::get()), m_var(variable),
```
pomadchin: May be, we need to add an overload which allows to add a custom handwritten type here.

hobu:
You can get structured output of options

`pdal --drivers --showjson`

```
[
	{
		"description" : "SBET Writer",
		"link" : "http://pdal.io/stages/writers.sbet.html",
		"name" : "writers.sbet"
	},
	{
		"description" : "Write data to SQLite3 database files.",
		"link" : "",
		"name" : "writers.sqlite"
	},
	{
		"description" : "Text Writer",
		"link" : "http://pdal.io/stages/writers.text.html",
		"name" : "writers.text"
	}
]
```

And then issuing

`pdal --showjson --options writers.las`

```
{
	"writers.las" :
	[
		{
			"description" : "User JSON",
			"name" : "user_data"
		},
		{
			"description" : "Debug output filename",
			"name" : "log"
		},
		{
			"description" : "Output filename",
			"name" : "filename"
		},
		{
			"description" : "Spatial reference to use to write output",
			"name" : "a_srs"
		},
		{
			"default" : "None",
			"description" : "Compression to use for output ('LASZIP' or 'LAZPERF')",
			"name" : "compression"
		},
		{
			"description" : "Discard points with out-of-spec return numbers.",
			"name" : "discard_high_return_numbers"
		},
		{
			"description" : "Dimensions to write above those in point format",
			"name" : "extra_dims"
		},
		{
			"description" : "Dimensions to forward from LAS reader",
			"name" : "forward"
		},
		{
			"default" : "\u0001",
			"description" : "LAS major version",
			"name" : "major_version"
		},
		{
			"default" : "\u0002",
			"description" : "LAS minor version",
			"name" : "minor_version"
		},
		{
			"default" : "\u0003",
			"description" : "Point format",
			"name" : "dataformat_id"
		},
		{
			"default" : "\u0003",
			"description" : "Point format",
			"name" : "format"
		},
		{
			"description" : "Global encoding byte",
			"name" : "global_encoding"
		},
		{
			"description" : "Project ID",
			"name" : "project_id"
		},
		{
			"default" : "PDAL",
			"description" : "System ID",
			"name" : "system_id"
		},
		{
			"default" : "PDAL 1.4.0 (b09ecb)",
			"description" : "Software ID",
			"name" : "software_id"
		},
		{
			"default" : "54",
			"description" : "Creation day of year",
			"name" : "creation_doy"
		},
		{
			"default" : "2017",
			"description" : "Creation year",
			"name" : "creation_year"
		},
		{
			"default" : ".01",
			"description" : "X scale factor",
			"name" : "scale_x"
		},
		{
			"default" : ".01",
			"description" : "Y scale factor",
			"name" : "scale_y"
		},
		{
			"default" : ".01",
			"description" : "Z scale factor",
			"name" : "scale_z"
		},
		{
			"description" : "X offset",
			"name" : "offset_x"
		},
		{
			"description" : "Y offset",
			"name" : "offset_y"
		},
		{
			"description" : "Z offset",
			"name" : "offset_z"
		}
	]
```

pomadchin:
@hobu yes, that's what I am doing here: https://github.com/PDAL/PDAL/pull/1492/files#diff-7af39184b47f5fdb301d31be18511673R14 

I tried to make fields optional in generated dsl, if there is a default value specified, but that would be not the full coverage of optional fields.

hobu:
Whether or not something is optional is often context sensitive. For example, setting an optional argument might cause another to now be required.

Of the choices, I probably like `addRequired` the most. It would allow us to casually go back and annotate things without being too disruptive.

pomadchin:
Yes, there can be a confusion about required field: the field can be required if another field is missing. In terms of DSL generation it means that such fields are optional. Probably it's enough to say some words about this dependency in docs or in a description field. O:

abellgithub:
I really don't understand what you're trying to do here.  You can make a command-line argument required by calling "setPositional()" on the argument:

args.add("filename", "The filename", m_filename).setPositional();

pomadchin:
@abellgithub Cool! Thank you, did not notice that, and booleans are always optional and false by default if there is no default value provided. I'll refactor the code. 

But type is still required, mb to add a function `setValType(std::string type)` which can be used in some cases where we can't derive proper string? Or it is better to solve it just by a more clever [TypeName](https://github.com/pomadchin/PDAL/blob/af1f8108f110b7617df80e8913b27a734ad0a04a/pdal/util/TypeName.hpp)?

abellgithub:
What are you going to do with the option information?

abellgithub:
I don't see any reason to require people to specify types.  The whole point of command-line arguments is to allow specification as strings.  There are no limits to values that are necessarily known from the INTERNAL type of a variable.  For instance, you can specify the verbose level (--verbose) as either a name or a number ("6" or "Debug" both work).  That information is usually in the description.  Similarly, --verbose only takes values 0 - 8 (I think).  Knowing that something is an int doesn't tell you much and internally it's stored as an enumerated type.

pomadchin:
Does it mean that json pipeline allows everything as a string? 

So the whole idea is to create a PDAL-Scala project, where pipeline can be built using internal DSL and not just by writing JSON strings explicitly:

```scala
val pipeline = 
  LasReader("filename", "crs", "tag1") ~ OtherReader("filename", "tag2") ~ Filter(..) ~ Merge(...) toPipeline

pipeline.execute()
```

The problem that there can be a lot of hand written and hardly parsable machinery: https://github.com/locationtech/geotrellis/blob/milestone/pointcloud/pointcloud/src/main/scala/geotrellis/pointcloud/pipeline/PipelineExpressions.scala#L31

And more over, arguments for each plugin can change, and Howard suggested to generate DSL, and for these purposes it is possible to use `pdal options ... --showjson`, like I did [here](https://github.com/PDAL/PDAL/pull/1492/files#diff-7af39184b47f5fdb301d31be18511673R14).

Right now the problem is in generating _correct_ DSL, as it's unknown which fields are optional and what type each field has (except cases where default value is specified): [example of not absolutely correct generated DSL class](https://github.com/PDAL/PDAL/pull/1492/files#diff-aa56f9442473edde49dcb4fc84994f12R91).

pomadchin:
If it doesn't make any sense, I'll close PR, probably were can make a DSL with all fields 'optional' (in DSL terms), it's more confusing for the final user, but doesn't require pdal source code changes.

abellgithub:
So you're trying to write your own pipeline language in Scala?  Not sure I get it.  Who would use that besides programmers and why wouldn't they be fine with JSON?

But there's nothing stopping you from doing this -- you simply need to translate your Scala thingee to JSON or the C++ API.  I don't know why you need to know options/types at all.  You can always specify the option names as strings and the option values as strings.  The internal types are irrelevant.

I'm probably not getting this at all.

pomadchin:
Because it's inconvenient to write in raw json, and that's a try to make a convinient library :) ): mb I don't completely understand your point, sorry for that.

Options and types a required for semantics - the wider it is the less problems can happen because of a developer fault, ofc everything can be erased and castet to string, but some times such semantics would help you to catch bug in compile time :0 and JSON has types too, I'd like to use them (ofc if it makes sense in context of pdal)

abellgithub:
Well, I still don't get it.  I don't know why PROGRAMMERS would be writing pipelines and if they are, it doesn't seem like a big deal to figure out the proper options.  But it's up to you.

There is no way to catch the kind of error that I think you want to catch as there is no notion of valid value for options known outside of running the code.  There is no compile-time ability to know the valid option values because they're not determined statically -- they run some C++ code at runtime to check.  We could enhance the `--showjson` option to display whether an option is required or not, but we can't provide valid values because they aren't known at compile time.

From the standpoint of a user (and you're a user in this case), all options take string values.

pomadchin:
Ok I see that you don't see why there should be a DSL for PDAL on top of JSON, but that can be easier to write all JSON related code and not to think about it at all, only one time implemented logic, as in theory, this DSL wth typed params would be self descriptive (not completely but sometimes it saves time as you don't need to go to pdal site and to look at json params). But I'll try to show you another advantages of such DSL.

For PDAL yes, all types you get in PDAL runtime, with the current PR I can see the following: 

```json
{
	"writers.las" : 
	[
		{
			"name" : "filename",
			"required" : "true",
			"type" : "string"
		},
		{
			"description" : "Spatial reference to use to write output",
			"name" : "a_srs",
			"required" : "false",
			"type" : "string"
		},
		{
			"...": "..."             
		},
		{
			"description" : "Z offset",
			"name" : "offset_z",
			"required" : "false",
			"type" : "double"
		}
	]
}
```

That would allow me to generate DSL with enough types to prevent in Scala compile time  _PDAL runtime exception_ (exmpl. we can guarantee that offset_z would only be double). 

And yes, we write pipelines, currently it's the only way to [interact](https://github.com/PDAL/PDAL/blob/master/java/core/src/main/scala/io/pdal/Pipeline.scala) with PDAL from JVM.

pomadchin:
P.S. Guys, excuse me for being annoying :sweat:

pomadchin:
Hey guys, we talked about pdal-scala DSL life and it was decided not to implement this DSL generation functionality (at least now), as the current description of fields is not enough. At first some basic things would be implemented as DSL expressions by hand, and there would be no expressions for every plugin PDAL supports. However clients of this library would have two options to use any plugin DSL doesn't support: 1. to implement a DSL expression 2. to use a raw JSON / JSON builders. I'll create a new PR instead of https://github.com/PDAL/PDAL/pull/1492 in order to keep it in history.

PR #1514, opened 2017-03-07
[GDALWriter] clip points to bounds

gadomski:
Previously, if points were outside of bounds, they were still added as
negative numbers.

### DIFF
--- a/io/GDALWriter.cpp
+++ b/io/GDALWriter.cpp
@@ -194,13 +194,14 @@ void GDALWriter::writeView(const PointViewPtr view)
 
 bool GDALWriter::processOne(PointRef& point)
 {
-    double x = point.getFieldAs<double>(Dimension::Id::X) -
-        m_curBounds.minx;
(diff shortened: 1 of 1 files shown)


abellgithub:
Did you have a problem, because addPoint() should be checking that things are in bounds and not add points outside of the valid grid:

`    // First quadrant;
    int i = iOrigin + 1;
    int j = jOrigin;
    while (i < (int)m_width && j >= 0)
    {
        double d = distance(i, j, x, y);
        if (d < m_radius)
        {
            update(i, j, z, d);
            i++;
        }
        else
        {
            if (i == iOrigin + 1)
                break;
            i = iOrigin + 1;
            j--;
        }
    }`

gadomski:
You are correct. I can't reproduce the error I thought that I created in the past. Closing as notanissue. Sorry for the noise.

abellgithub:
No problem!

PR #1522, opened 2017-03-11
Move GeoTIFF and PROJ.4 to required dependencies

mloskot:
Both libraries are actually required, or successful build is not feasible without GDAL with PROJ.4 enabled.


hobu:
That entire doc needs a once-over. I've been meaning to do it for a few releases, but it has been easier to point people at the Dockerfile.

mloskot:
Dockerfile as example of `cmake` command is fine, just that it's not necessarily clear if those options/deps are minimal or recommended set.

PR #1540, opened 2017-03-24
Fix PgWriter DROP TABLE statement

elemoine:
When a schema is specified the following statement is currently issued:

```sql
    DROP TABLE IF EXISTS "public.table_name"
```

which is not correct. With this patch the following statement is created:

```sql
    DROP TABLE IF EXISTS "public"."table_name"
```

I can add tests if people think it would be useful.


hobu:
A test addition would be nice, but isn't required.

PR #1550, opened 2017-04-03
Fix LASZIP_LASZIP_DEFINES typo

sebastic:
Replace LASZIP_LASZIP_DEFINES with LASZIP_DEFINES.

Fixes #1549.


hobu:
@sebastic The `ready-release` branch is current where I have been collecting release fixes. I will manually keep that current and cut the next RC from that when you're ready.

I was confused on SOVERSION (as always 😄  ), so please take a look at that too.

sebastic:
There were quite a few removed symbols, so incrementing `PDAL_API_VERSION` seems appropriate.

Specifically `AttributeFilter_{Exit,Init}Func` & `DerivativeWriter_{Exit,Init}Func` stand out.

PR #1554, opened 2017-04-06
Add Extended Local Maximum (ELM) filter

chambbj:
Based off the work of Chen, et al. In the original work, the authors used ELM
to generate grid minimums that were resistent to low outliers in ground
segmentation and DTM generation. We implement here as a standalone filter that
simply marks the low points as noise. Downstream filtering stages can ignore,
extract, or handle these noise points as they see fit.

### DIFF
--- a/doc/references.rst
+++ b/doc/references.rst
@@ -7,6 +7,8 @@ References
 .. index:: References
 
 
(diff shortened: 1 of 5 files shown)


hobu:
Happy to stick this into 1.5 if you want to get it in under the wire. It shouldn't have any release impact.

chambbj:
Sounds good to me.

PR #1555, opened 2017-04-06
LAS extra_dims handling for standard PDAL dimensions

connormanning:
Previously, if a standard PDAL dimension (but which was not a standard LAS dimension) was written to a LAS file with the `extra_dims` option, it would appear as a proprietary dimension when reading the file.  This PR allows the registered value to be used.

This breaks a previously allowed behavior - where you could have an unrelated proprietary dimension with the same name as a standard dimension in the same file.  Was anyone relying on this behavior?  Tests have been updated accordingly.


hobu:
> Was anyone relying on this behavior? 

I doubt it, and this change makes things behave as you would expect anyway.

PR #1567, opened 2017-04-17
Resolve some dangling references to deprecated extract and classify options

chambbj:
`extract` and `classify` have been used in several filters, namely outlier and PMF. With the release of PDAL v1.5, we removed these options, opting to always classify points, and to force users to ignore or extract them as they see fit in their pipelines.

The outlier filter's extract option can be replicated by adding a range filter stage as shown following the outlier filter:

    {
      "type":"filters.range",
      "limits":"Classification![7:7]"
    }

The PMF filter's extract option can be replicated by adding the following range filter:

    {
      "type":"filters.range",
      "limits":"Classification[2:2]"
    }

The intention here was to remove any ambiguity as to how individual filters implement `extract`, and to instead be explicit in the pipeline specification as to how points are extracted.

### DIFF
--- a/doc/apps/translate.rst
+++ b/doc/apps/translate.rst
@@ -80,15 +80,16 @@ Example 2:
 --------------------------------------------------------------------------------
 
 Given these tools, we can now construct a custom pipeline on-the-fly. The
-example below uses a simple LAS reader and writer, but stages a PCL-based
-voxel grid filter, followed by the PMF filter. We can even set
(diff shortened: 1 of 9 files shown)


chambbj:
@hobu Best to merge to master and then cherry-pick over to 1.5-maintenance for the doc refresh?

hobu:
> Best to merge to master and then cherry-pick over to 1.5-maintenance for the doc refresh

Yup.

PR #1569, opened 2017-04-17
Add a buffer option to the splitter

gadomski:
This buffer grows each square by the specified amount, allowing for an
overlap between tiles. This is particularily useful for tiled change
detection and other such algorithms.


gadomski:
@abellgithub patch updated with your suggestions, thanks.

PR #1577, opened 2017-04-25
Add ICP and CPD filters, remove CPD kernel

gadomski:
This filter applies Iterative Closest Point (ICP) and/or Coherent Point Drift (CPD) to two (and only two) inputs. The output is the transformed points, and the metadata contains useful information e.g. the transformation matrix.

The PCL plugin (obviously) includes PCL-finding CMake logic, but since this is a separate plugin that may or may not actually depend on PCL, I did my own PCL-finding (nothing complicated). I do use one of the PCL-plugin utility functions, which is a bit of cross-plugin pollution.

I did a bit of main-tree modification to make `Metadata` aware of Eigen matrices. It's not perfect, so @abellgithub and/or @chambbj may want to review those bits (~725dad16~ 45d4385cad27b22d0b0d003ea1b5c20e70ded792).

This plugin replaces the old CPD kernel.


chambbj:
@gadomski Haven't had a chance to look closely at this, but I don't have a problem with the Eigen portion. What's not clear to me is the name and description. I see how this feeds into change detection, but it doesn't seem like a true change detection filter itself. What am I missing?

gadomski:
If a change detection filter == compute the change between two datasets, then this is that.

This filters runs ICP/CPD on two datasets, outputs the resulting (changed) points, and puts the change detection parameters (e.g. transformation matrix) into the filter's metadata. That's how it works in my head at least, but that might not be intuitive to others (I've been in these weeds for a bit). Suggestions on naming/flow welcome.

chambbj:
There will surely be an audience that hears "change detection" and expects to see highlighted changes, where things have appeared or disappeared from one collect to another.

Part of me thinks that `filters.icp` and `filters.cpd` would be nice things to have and should clear up any ambiguity as to their purpose. Another part of me wonders if this is better as a kernel.

I struggle with the same things though when working with outlier detection, ground segmentation, etc. Should I create a single filter for a particular type of operation, and let users control the method/algorithm through the stage options, or just break it into many different filters? I guess I've been doing more of the latter recently.

In the end, I suppose it's up to you. Hate to bikeshed it too much. Just my two cents.

gadomski:
> There will surely be an audience that hears "change detection" and expects to see highlighted changes, where things have appeared or disappeared from one collect to another.

Great point. I'm not sure what to call it then. "Motion detection"?

> Part of me thinks that `filters.icp` and `filters.cpd` would be nice things to have and should clear up any ambiguity as to their purpose.

I've liked the straightforwardness of the way you've split out the ground algorithms. I had it in my head that I wanted to mush these into one filter, but that was more of my preconceived notion rather than a well-thought-out decision. Splitting them makes sense to me (and will simplify the code), I'll change to that.

> Another part of me wonders if this is better as a kernel.

The two (and only two) input nature of these algorithms does map better to a kernel than a filter, that is true. I guess an argument for being a filter is if you wanted to do something more with the output, e.g. downsampling or gridding. Maybe both (wrap the filter in a kernel)?

gadomski:
@hobu I've got this passing Travis. Since AppVeyor's not working, this can be considered good to merge from my perspective.

PR #1617, opened 2017-06-29
Simply the ChipperFilter code

elemoine:
The ChipperFilter code currently includes a `finalSplit` function that calculates values (namely `left1`, `left2`, `right1` and `right2`) that are not used when emitting results. This PR removes the `finalSplit` function and makes the `split` function itself do the emissions when there are two partitions left.

Please tell me if I am missing something here.

### DIFF
--- a/filters/ChipperFilter.cpp
+++ b/filters/ChipperFilter.cpp
@@ -57,13 +57,9 @@ of the spare array then becomes the active array for the narrow direction.
 This avoids resorting of the arrays, which are already sorted.
 
 This procedure is then recursively applied to the created blocks until
-they contains only one or two partitions.  In the case of one partition,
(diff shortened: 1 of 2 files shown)


abellgithub:
No, you're not missing anything.  When the code was first written, we provided the bounds for each "cell".  This code would make sure the bounds were available when each cell was created.  Now that we don't write the bounds, this code is superfluous.

Thanks for catching.

elemoine:
Thanks for the review and merging this.

elemoine:
> When the code was first written, we provided the bounds for each "cell". This code would make sure the bounds were available when each cell was created. Now that we don't write the bounds, this code is superfluous.

You've made me realize there's more unneeded code in the Chipper filter. I'll create a new PR with more clean-up.

Issue #1638, opened 2017-08-03
Normal flipping


chambbj:
Need to provide a mechanism within the Normal filter to flip normals to a consistent viewpoint.

tomvantilburg:
Makes sense to do it according to a viewpoint, though for my personal use of airborne lidar an 'always up' flag would also do.

chambbj:
Should always up be the default? And users can just override with a
specific viewpoint?

tomvantilburg:
My guess is that airborne lidar will be the main use case for PDAL (I may be wrong) and that 'always up' makes sense as a default.  Would there be a penalty in speed when calculating against the coordinates of a viewpoint? If not, the default might be some miles above the center of the extent I guess. If so, a check for the normalz sign might be faster.

chambbj:
@tomvantilburg See #1639

tomvantilburg:
Did you check for any speed penalty? As far as I can see there are 2 extra calculations involved that can be skipped when you already know that it's only about the normalZ :
- [This block](https://github.com/PDAL/PDAL/pull/1639/files#diff-08cec75aadf8f2e081e144fdac96235cR113) that compares the eigevector and
- view.calculateBounds(bounds); 

I guess the penalty is small with these simple substractions but would be nice to know how small.

chambbj:
@tomvantilburg A quick test on a 38000 point test file, shows the normal flipping method to be about 3-4% slower. On a file that size (which is admittedly small), it's a matter of milliseconds. Only flipping on negative normalZ, obviously falls somewhere in between.

I'd be open to your first proposal, to offer options both for orienting towards the viewpoint and just blindly assuming always up. In the former case, I think there should be no default viewpoint coordinates. The user either provides all three coordinates, or none at all. Viewpoint and always up options will be mutually exclusive.

tomvantilburg:
@chambbj I agree with this, it sounds logical to be mutually exclusive and put the default on always up.

chambbj:
Okay, the PR has been updated. `always_up` is now the default behavior. Users can optionally specify a `viewpoint` as a WKT or GeoJSON point - if they do, `always_up` is ignored. The only way to prevent any flipping is to set `always_up=false` and do not provide a `viewpoint`.

chambbj:
@tomvantilburg Just merged this into master. Closing the issue. Let me know if you have any other issues with the normal filter.

PR #1639, opened 2017-08-03
Provide for normal flipping

chambbj:
User can always set a preferred viewpoint by providing X, Y, and Z coordinates.
If a viewpoint coordinate is not provided, it will be replaced as follows.

* X and Y will default to the midpoint of the X/Y extents
* Z will default to the maximum Z plus 1000

Normals are flipped when the dot product of the normal and the vector between
the current point and the viewpoint is less than zero.


filters/NormalFilter.cpp
```diff
@@ -58,33 +58,41 @@ std::string NormalFilter::getName() const
...
     m_knn = 8;
     filter(view);
 }
 
-
 void NormalFilter::filter(PointView& view)
 {
     KD3Index& kdi = view.build3dIndex();
 
+    BOX3D bounds;
```
chambbj: We are discussing alternatives in #1638, so this will all be refactored in a bit.

PR #1656, opened 2017-08-29
Fix for windows linking in PDALTargets.cmake.in

jasonbeverage:
This may not be the correct way to fix this, but I had issues linking applications against PDAL (at least on Windows) using the PDAL CMake scripts because not all the pdal libraries are actually included in the PDAL_LIBRARIES list.  It's only including pdalcpp.

This patch adds all of the pdal libraries that might need to be linked to.

If there is a better way to get around this I'd be interested in that as well.

Thanks!

Jason

### DIFF
--- a/PDALConfig.cmake.in
+++ b/PDALConfig.cmake.in
@@ -17,6 +17,11 @@ endforeach(_dir)
 
 include("${CMAKE_CURRENT_LIST_DIR}/PDALTargets.cmake")
 
-set(PDAL_LIBRARIES "@PDAL_LIB_NAME@" "pdal_util" "pdal_jsoncpp" "pdal_arbiter" "pdal_boost")
+if (WIN32)
+# On windows link in pdal_util as well
+  set(PDAL_LIBRARIES "@PDAL_LIB_NAME@" "pdal_util")
+else (WIN32)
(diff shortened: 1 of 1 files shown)


abellgithub:
arbiter, jsoncpp and boost should all be statically linked into pdalcpp, so I don't think you should need to include them.  Are the symbols from those libraries not in pdalcpp?

pdal_util is different.  On Unix, we do special stuff to make it so that you only have to link pdalcpp (pdal_util gets included by magic).  I'm not sure that we've ever looked into this on Windows.  I could do some research if you're willing to test.

No matter, I think that this needs if (WINDOWS) around it.

jasonbeverage:
I'll check and make sure, maybe only pdal_util is needed.  I see to
remember having issues with jsoncpp though as well.

jasonbeverage:
I just checked and you're correct, I only needed pdal_util as an extra dependency.  I updated the branch and also wrapped it with an IF (WIN32) block so it only includes pdal_util on windows.

abellgithub:
Can you change "pdal_util" to "@PDAL_UTIL_LIB_NAME@"?  At that point we're good until I can make time to research a single-library link on Windows.  Thanks!

jasonbeverage:
Done, thanks again!

PR #1658, opened 2017-08-30
Check for execinfo/backtrace

chambbj:


abellgithub:
Seems like the cmake file could set the value for PDAL_HAVE_EXECINFO_H for Windows and then we could dispense with the check in the files.

chambbj:
I've actually got Alpine in mind. I agree that checking both `_WIN32` and `PDAL_HAVE_EXECINFO_H` is redundant though. Did you have something else in mind?

Issue #1651, opened 2017-08-25
Allow optional suppression of name/id -specified VLRs for readers.las


hobu:
The Merrick VLRs are obnoxious. There's 100s-1000s of them sometimes, and stuff can blow up.

```
{
    "type":"readers.las",
    "ignore_vlrs":"Merrick",
    "ignore_vlrs":"pdal/12"
    "filename":"input.las"
}
```

This specification would remove all ``Merrick`` VLRs, regardless of ``record_id``, and it would only remove the ``pdal`` VLR with a ``record_id`` of ``12``.

PR #1664, opened 2017-09-07
add ignore_vlr option to readers.las #1651

hobu:


abellgithub:
Is there any reason not to add VLRs as they're read, rather than remove them after the fact?  Seems like that would keep code centralized in the LASHeader streaming stuff.  Also, it means you don't need the remove... functions.

hobu:
> Is there any reason not to add VLRs as they're read, rather than remove them after the fact?

I was just trying to keep with the flow.

PR #1667, opened 2017-09-12
Greyhound reader updates

connormanning:
Simplify Greyhound reader options - allow a fully-specified `read` URL as the sole option to the reader.  With this capability, also allow this reader to be specified as a string instead of an object in a pipeline, e.g.
```
{ "pipeline": [
  "greyhound://data.greyhound.io/resource/autzen/read?bounds=[10,10,10,20,20,20]",
  ...
] }
```

The internals have also been simplified - there is no more client-side query traversal.  Docs/tests are updated.


abellgithub:
I'd get chastised where I used to work for submitting a change set this large.

connormanning:
Sounds stifling.  Anyway, it's pretty much a complete rewrite since the original was overly complicated for little purpose - so I'd just look at the GreyhoundReader files directly rather than the diff.  They're pretty small files.

PR #1676, opened 2017-09-20
Remove 2D and output filename options from delta kernel

chambbj:
While parsed (and even required in the case of output filename), neither of
these have any impact on the code as written.


abellgithub:
Beat me to it :)  Need to remove "[output]" from command line in doc as well.

chambbj:
@abellgithub I think I got the `[output]` you are referring to. If I missed one, let me know.

abellgithub:
👍

PR #1679, opened 2017-09-20
Remove KernelFactory and move all kernel plugin loading to StageFactory

chambbj:
Long ago, KernelFactory did more than just load kernel plugins. What's left is
duplicated within StageFactory. Though not functionally any different, kernel
plugin loading can safely be moved to StageFactory to simplify code somewhat.


abellgithub:
Looks good.

PR #1693, opened 2017-10-02
Add ability to groupby list of DimRanges

chambbj:
Inspired by recent mailing list question, this is at least one possible solution to being able to split into multiple PointViews such as `ReturnNumber == 1` and `ReturnNumber != 1` natively within PDAL versus subsetting in Python.

### DIFF
--- a/doc/stages/filters.groupby.rst
+++ b/doc/stages/filters.groupby.rst
@@ -3,13 +3,17 @@
 filters.groupby
 ===============================================================================
 
-The groupby filter takes a single PointView as its input and creates a PointView
-for each category in the named ``dimension`` as its output.
+The groupby filter takes a single PointView as its input and creates an output
+PointView for each category in the named ``dimension`` or each range in
+``ranges``.
 
 .. embed::
 
-Example
--------
+Example #1
+----------
+
+The first example pipeline will create a new output PointView for each unique
+classification value.
 
 .. code-block:: json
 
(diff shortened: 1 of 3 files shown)


abellgithub:
I haven't looked over the code, but I'm wondering at this point how this fits with range filter.  Are we duplicating code?  Should we merge something?

chambbj:
The only real duplication is in the range parsing. In the end, they both use `valuePasses` in DimRange to check if a value is within the provided range.

You can, in fact, create a simple (single DimRange) range filter out of this. But with multiple DimRanges, the behavior is different. The groupby option imposes no boolean logic on the list of ranges like the range filter does. Each is treated independently, and creates its own output PointView.

chambbj:
As @abellgithub has pointed out, the behavior with `ranges` specified is slightly different than with `dimension`, in a perhaps unexpected/not-so-obvious way, in that `ranges` can result in PointViews that are not mutually exclusive. For example,

```
{
  "type":"filters.groupby",
  "ranges":"Classification[1:6],Classification[4:10]"
}
```

I'm beginning to think that this still has utility, but perhaps a name change would be in order. Other tools/packages support the concept of slicing an array, which is basically an identical concept. We are slicing our PointView by some condition to create multiple PointViews. So, I'm thinking `filters.slice` would be the way to go.

hobu:
reopen this if you want to resolve it in 1.8

PR #1703, opened 2017-10-13
Remove '..' in configured temp path for unit tests

chambbj:
musl has changed it's handling of '.' and '..' when globbing [1]. These only
recently made their way into the alpine:edge image that we use in Travis and
for Docker automated builds [2]. For now, we simply configure the temp path
directly, rather than relying on '..'.

[1] https://github.com/alpinelinux/aports/blob/afdba0ba0b8c0e92c376e7f1bfb151c07bf654b6/main/musl/0068-fix-glob-descent-into-.-and-.-with-GLOB_PERIOD.patch
[2] http://www.openwall.com/lists/musl/2017/01/12/5

### DIFF
--- a/test/unit/Support.cpp
+++ b/test/unit/Support.cpp
@@ -72,7 +72,7 @@ std::string Support::configuredpath(const std::string& file)
 
 std::string Support::temppath()
 {
-    return TestConfig::dataPath() + "../temp/";
+    return TestConfig::tempPath();
 }
 
 std::string Support::temppath(const std::string& file)
--- a/test/unit/TestConfig.hpp.in
+++ b/test/unit/TestConfig.hpp.in
@@ -36,6 +36,7 @@
 #define UNITTEST_TESTCONFIG_INCLUDED
 
 #define UNITTEST_TESTCONFIG_DATA_PATH "@CMAKE_SOURCE_DIR@/test/data/"
+#define UNITTEST_TESTCONFIG_TEMP_PATH "@CMAKE_SOURCE_DIR@/test/temp/"
 #define UNITTEST_TESTCONFIG_CONFIGURED_PATH "@CMAKE_BINARY_DIR@/test/data/"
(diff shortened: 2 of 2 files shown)


abellgithub:
This seems like a bug in their code somewhere, as we aren't asking for or receiving "." or ".." as a pattern.  This isn't what GLOB_PERIOD is doing.

Changing this does fix things for us, but it revealed an OS issue (rather than our bug), so I'm a little iffy on this change.  Perhaps we do "#ifdef alpine" for now or something like that.

Can we tie to an older release that doesn't have this problem?  Should we report a bug?

chambbj:
> we aren't asking for or receiving "." or ".." as a pattern

We do though...here https://github.com/PDAL/PDAL/blob/master/test/unit/FileUtilsTest.cpp#L237. The original temp path had "..", and that is passed to the glob function.

> Perhaps we do "#ifdef alpine" for now or something like that.

I feel like this change is safe on its own, but if we feel strongly about the globbing patterns not operating the way we'd like, then we should maybe expand on that in a dedicated test.

> Can we tie to an older release that doesn't have this problem?

I tried with little success, as the system was wanting to automatically upgrade all of the libraries. I plan on continuing to look for a way to truly pin it.

> Should we report a bug?

Given some of the comments in the commits referenced above, this appears to be a purposeful decision.

Do we already have the capability to simply resolve anything we send to our FileUtils glob to an absolute path prior to sending to system glob? Thus sanitizing the paths of any "." or ".."?

abellgithub:
>> we aren't asking for or receiving "." or ".." as a pattern

>We do though...

There is a ".." in the path, but that's not what GLOB_PERIOD is doing.  And I can't think of any reason why glob shouldn't work with a leading "." or ".." in the path.  That seems broken.

chambbj:
I see what you are saying. I'll continue to dig into what is actually going on under the hood with this particular test/path and submit a report.

PR #1711, opened 2017-10-25
Add option to remove empty output files

gadomski:
So I've been using PDAL for some batch processing, and I ran into a case where a certain range filter would produce zero-point output files, scattered among non-zero-point files. I thought it might be nice if PDAL could clean these up for me. This patch is a stab at the idea, but the mechanisms are totally up for debate. Here's how I did it for this pass:

- `Stage`s can advertise that they support "removal" via a `canRemove` method, i.e. they have an
output that can be removed. For now, `FlexWriter` supports removal,
since it has a filename. `Stage`s also get a `remove` method that throws an error by default, since by default stages aren't removable.
- `PipelineManager` has a `m_removeEmpty` option, that checks the total
point count and if (a) the point count is zero, (b)
`m_removeEmpty` is true, and (c) the last stage supports removal,
removes the output.
- Both the pipeline and translate kernels get a `--remove-empty` option.

### DIFF
--- a/kernels/PipelineKernel.cpp
+++ b/kernels/PipelineKernel.cpp
@@ -89,6 +89,7 @@ void PipelineKernel::addSwitches(ProgramArgs& args)
     args.add("stdin,s", "Read pipeline from standard input", m_usestdin);
     args.add("stream", "Attempt to run pipeline in streaming mode.", m_stream);
     args.add("metadata", "Metadata filename", m_metadataFile);
+    args.add("remove-empty", "Remove the output file if it is empty and the output writer supports removal.", m_removeEmpty);
 }
 
 
@@ -119,7 +120,10 @@ int PipelineKernel::execute()
         m_manager.executeStream(table);
     }
     else
+    {
+        m_manager.setRemoveEmpty(m_removeEmpty);
         m_manager.execute();
+    }
 
     if (m_metadataFile.size())
     {
--- a/kernels/PipelineKernel.hpp
+++ b/kernels/PipelineKernel.hpp
@@ -69,6 +69,7 @@ private:
     int m_progressFd;
     bool m_usestdin;
     bool m_stream;
+    bool m_removeEmpty;
 };
 
 } // pdal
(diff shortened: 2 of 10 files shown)


abellgithub:
This is only an issue with FlexWriter, correct?

gadomski:
> This is only an issue with FlexWriter, correct?

Maybe? `FlexWriter::readyFile` is the "culprit", as it were, because it usually creates a file even if the eventual file will have zero points.

That's why I added the `canRemove` method, because I wasn't sure if other types of writers could be "removed".

hobu:
Can we polish this up and merge it?

gadomski:
/me will try to find time to polish today or tomorrow, apologies.

abellgithub:
I can do this.  I'd rather just drop empty point views.  It's cleaner. No creating files we don't need in the first place. Thinking it would meet the request.

gadomski:
@abellgithub sounds good. Will you also be able to address the `FlexWriter` issue, i.e. the `FlexWriter` creating a file even if it ends up not having any point views to work with?

abellgithub:
@gadomski : Want to see if this does what you what? https://github.com/PDAL/PDAL/tree/issue-1711

gadomski:
@abellgithub this does *not* do what I was requesting. E.g.:

```
build/bin/pdal translate \
  -i test/data/las/autzen_trim.las \
  -o no-points.las \
  -f range --filters.range.limits=Classification[42:42]
```

creates the `no-points.las` file with zero points.

abellgithub:
I was only expecting that you wanted to eliminate files created from grouping/splitting into point views containing 0 points. In this case, you're explicitly asking to create the file.  Seems wrong that we wouldn't write it.

gadomski:
> In this case, you're explicitly asking to create the file.

I guess that's what my original option, `--remove-empty`, was getting at. I agree that you're *generally* explicitly asking to create a file, but the `--remove-empty` option changes the parameters of that ask.

I understand if this is over-featuring and would be fine to close w/o merging if you think this is too much added complexity :-).

Issue #1716, opened 2017-11-06
Automatically set UTM zone when writing BPF data for UTM coordinate systems


hobu:
# Problem

GDAL has the `OSRGetUTMZone` method ([link](http://www.gdal.org/classOGRSpatialReference.html#a4b17d85f5db53978c749780b214c4237)) that will return a single number that is sufficient for the current `writers.bpf` `coord_id` value. We can automatically set the `coord_id` with that value if the specified coordinate system is actually UTM.

# Proposal

* If the user sets the `autoutm` option, the `writers.bpf` will attempt to compute one if the active coordinate system is UTM. If the coordinate system is not UTM, no zone value will be computed.
* If OSR does not determine a UTM zone ID, the default `coord_id` user is used (current behavior).

abellgithub:
-1 is a valid UTM zone.  Also, there is no default `coord_id`.  The default is to specify that the SRS type is cartesian.

abellgithub:
Done in https://github.com/PDAL/PDAL/pull/1723

PR #1717, opened 2017-11-06
Attempt to auto-assign UTM zones for BPF output that have UTM coordinate systems #1716

hobu:


abellgithub:
I'd rather see a special value used for `coord_id`. That would prevent having to figure out what to do if you set a value for `coord_id` and `autoutm`. The type could be changed to a string if that would be more clear.

hobu:
Please adapt as you see fit. The current pull request and tests exhibit the behavior we would like to achieve.

PR #1722, opened 2017-11-08
Compression

abellgithub:
Add new compression types. Standardize compression API.


chambbj:
In

https://github.com/PDAL/PDAL/blob/62f751cc4bd700944e3df2b2cd0b7b246b56688c/scripts/docker/master/ubuntu/Dockerfile#L19

the Ubuntu master Dockerfile also needs to install at least `zstd-dev`, probably something for `xz` too.

The dependencies Dockerfile also only installs `zstd-dev`, nothing for `xz`.

PR #1743, opened 2017-12-15
Explicitly cast m_curl as CURL* in arbiter.cpp

nickvaughn:
Change arbiter.cpp to avoid GCC 4.9.2 errors during compile on a
Centos 7 system. GCC threw errors complaining that the curl_easy_*
functions could not take void* arguments. Explicitly cast these
m_curl in each of these calls to fix problem.

### DIFF
--- a/vendor/arbiter/arbiter.cpp
+++ b/vendor/arbiter/arbiter.cpp
@@ -2790,7 +2790,7 @@ Curl::Curl(const Json::Value& json)
 Curl::~Curl()
 {
 #ifdef ARBITER_CURL
-    curl_easy_cleanup(m_curl);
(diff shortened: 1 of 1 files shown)


connormanning:
I think [this](https://github.com/connormanning/arbiter/commit/e3fead2f855dc2d3193cd3bb784f7d90b5d5daff) might be a cleaner fix - the `void` usage was a lazy way to get things compiling when libcurl headers were not available.  Arbiter fixes should preferably be made in that repository to avoid changes being hammered over when someone rebundles the library into PDAL.

hobu:
I will refresh arbiter in PDAL and consider this one dead.

PR #1748, opened 2017-12-21
Access stages of a pipeline from PipelineManager

### LINKED ISSUE #1747: Accessor for Stages in PipelineManager
While writing a Kernel I have the need to access to each stage in the PipelineManager prior to execution.  In this case it is to perform Stage dependency injection prior to execution starting.

Of the current accessors for `Stage*` in the PipelineManager interface, I dont see any that provide access to the full chain.  The root and leaf methods are a subset and the getStage is just the back?

What am I missing?  Or how about a [PR to add an accessor](https://github.com/jw3/PDAL/blob/pipeline_stage_access/pdal/PipelineManager.hpp#L136-L137)?

jw3:
resolves #1747

### DIFF
--- a/pdal/PipelineManager.hpp
+++ b/pdal/PipelineManager.hpp
@@ -133,6 +133,9 @@ public:
     std::vector<Stage *> leaves() const;
     void replace(Stage *sOld, Stage *sNew);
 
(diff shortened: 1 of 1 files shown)


abellgithub:
Close #1747

Issue #1760, opened 2017-12-27
Doxygen documentation


cvvergara:
For developers it would be nice to have.
I made this temporarly:
https://cvvergara.github.io/pdal/

abellgithub:
My personal opinion on this is that it doesn't provide much value if it isn't done carefully.  Generating text for random classes seems more confusing than clarifying.  What advantage do you think this provides over header files?

cvvergara:
I just made the PR that @hobu asked me to do

PR #1761, opened 2017-12-27
Generating doxygen

cvvergara:
Adding basic doxygen files
Currently use:
```
cd doxygen
doxygen 
#or to avoid the screen messages (18000+ lines)
doxygen > tmp.txt
cd ..
```
See the doxygen documentation errors
```
more doxygen/LOG
```
The generated files will be in
```
ls doxygen/documentation
```


cvvergara:
Fixes #1760

hobu:
I'm merging this, but we're not building the doxygen output for the website. It's too noise and sucks up all of our google juice for other more relevant bits of documentation we have there. This patch is useful if someone wants to build the doxygen output for themselves, however.

PR #1762, opened 2017-12-28
Removed extra ";"

cvvergara:
I temporarily added the `-pedantic flag` and removed almost all the warnings of extra `;` it showed, and removed the flag.
I didn't fix the ones here:
https://travis-ci.org/Georepublic/PDAL/builds/322294088#L1076
because is almost modifying the whole file. When the real issue is on the way the macro is defined with a semicolon at the end.


hobu:
> When the real issue is on the way the macro is defined with a semicolon at the end.

We don't control the macro, so we'll just have to take them off of our usage. See #1765 for a fix.

PR #1766, opened 2017-12-28
Implement -Werror and -Wpedantic

hobu:


abellgithub:
Does this really buy us anything? Unconvinced. And the #define final seems wrong.

hobu:
> Does this really buy us anything? 

For our existing codebase? I don't think so. For future changes and additions, maybe it does.

PR #1769, opened 2018-01-02
Improve C++ code on Text R & W

cvvergara:
Improving C++11 code of io/TextReader.* and io/TextReader.*
- Using range loops
- Using auto


abellgithub:
I see nothing in this PR that needs to be done.  It isn't usually helpful to change code for no functional reason.  It can introduce bugs and confuses someone looking at the code history.

cvvergara:
#1772 # Range for loop (faster)

Code differences:
- by number of lines 29 lines with range loop VS 35 lines with the "normal" for loop
- This example is with a -O2 flag

### with range loop
```
int foo(std::vector<int> &num) {
    for (const auto& n : num) {
        std::cout << n;
    }
}
```
generated code:
```
foo(std::vector<int, std::allocator<int> >&):
  push rbp
  push rbx
  sub rsp, 8
  mov rbx, QWORD PTR [rdi]
  mov rbp, QWORD PTR [rdi+8]
  cmp rbx, rbp
  je .L2
.L3:
  mov esi, DWORD PTR [rbx]
  mov edi, OFFSET FLAT:std::cout
  add rbx, 4
  call std::basic_ostream<char, std::char_traits<char> >::operator<<(int)
  cmp rbx, rbp
  jne .L3
.L2:
  add rsp, 8
  pop rbx
  pop rbp
  ret
_GLOBAL__sub_I__Z3fooRSt6vectorIiSaIiEE:
  sub rsp, 8
  mov edi, OFFSET FLAT:std::__ioinit
  call std::ios_base::Init::Init()
  mov edx, OFFSET FLAT:__dso_handle
  mov esi, OFFSET FLAT:std::__ioinit
  mov edi, OFFSET FLAT:std::ios_base::Init::~Init()
  add rsp, 8
  jmp __cxa_atexit
```

## without range loop
```
int foo(std::vector<int> &num) {
    for (int i = 0; i < num.size(); ++i) {
        std::cout << num[i];
    }
}
```

```
foo(std::vector<int, std::allocator<int> >&):
  mov rdx, QWORD PTR [rdi]
  cmp QWORD PTR [rdi+8], rdx
  je .L6
  push rbp
  push rbx
  mov rbp, rdi
  xor ebx, ebx
  sub rsp, 8
.L3:
  mov esi, DWORD PTR [rdx+rbx*4]
  mov edi, OFFSET FLAT:std::cout
  add rbx, 1
  call std::basic_ostream<char, std::char_traits<char> >::operator<<(int)
  mov rdx, QWORD PTR [rbp+0]
  mov rax, QWORD PTR [rbp+8]
  sub rax, rdx
  sar rax, 2
  cmp rax, rbx
  ja .L3
  add rsp, 8
  pop rbx
  pop rbp
  ret
.L6:
  rep ret
_GLOBAL__sub_I__Z3fooRSt6vectorIiSaIiEE:
  sub rsp, 8
  mov edi, OFFSET FLAT:std::__ioinit
  call std::ios_base::Init::Init()
  mov edx, OFFSET FLAT:__dso_handle
  mov esi, OFFSET FLAT:std::__ioinit
  mov edi, OFFSET FLAT:std::ios_base::Init::~Init()
  add rsp, 8
  jmp __cxa_atexit
```

## adding const

Using const on functions that are not intended to modify the object, allows programmers to do better code.
Unless of course that by design for example, the TextWriter can modify the data.

## Use delete on functions that are not allowed
```
TextWriter(const TextWriter&) = delete;
```
Avoiding implicit generation of the copy constructor or of the assignment operator.


## First include the header
I agree that might be a matter of style
https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

abellgithub:
If this was a contribution of new functionality, I'd have no problem with any of the changes you're proposing here (I'm not in love with `auto` when a simple type declaration conveys more information, but whatever...)  The issue is that these changes aren't in the context of an enhancement that provides any benefit to users and potentially introduces bugs (there is one in this PR).

I don't want to discourage you from contributing to PDAL and improving the code base, but I also don't want to get into arguments about what constitutes best practices or fights about style.  We'd love to see code that adds stages or enhances the capability of existing stages. We're also interested in performance improvements that users will notice.

Here are some general rules that we try to follow.

- Don't change code that works if you don't have to.  We don't do formal code review and any change can introduce bugs. We have OK, but incomplete test coverage. Improved test coverage is desirable.
- When you're modifying code in an existing file, follow along with what exists, updating where appropriate.  Don't modify all the code just because you're making a small change.
- Don't optimize at the expense of clarity unless you have a really good reason.
- Try to keep lines to 80 characters.

Sometimes we break these guidelines, but it's a starting point.

Send me an email if you want to talk more about this kind of thing. (andrew@hobu.co).

io/TextReader.cpp
```diff
@@ -158,7 +159,7 @@ void TextReader::ready(PointTableRef table)
 
 point_count_t TextReader::read(PointViewPtr view, point_count_t numPts)
 {
-    PointId idx = view->size();
+    auto idx = view->size();
```
mloskot: I'm not sure if that is an improvement.

io/TextReader.cpp
```diff
@@ -197,15 +198,8 @@ bool TextReader::processOne(PointRef& point)
 
 bool TextReader::fillFields()
 {
-    while (true)
+    for(std::string buf; std::getline(*m_istream, buf) && m_istream->good(); ++m_line)
```
mloskot: I'm not sure if that is an improvement wrt code readability.

io/TextWriter.cpp
```diff
@@ -117,10 +117,10 @@ void TextWriter::ready(PointTableRef table)
     // Yes, this isn't efficient when, but it's simple.
     if (m_dimOrder.empty() || m_writeAllDims)
     {
-        Dimension::IdList all = table.layout()->dims();
-        for (auto di = all.begin(); di != all.end(); ++di)
-            if (!Utils::contains(m_dims, *di))
-                m_dims.push_back(*di);
+        for (const auto d : table.layout()->dims()) {
```
mloskot: `const auto d : table.layout()->dims()`
should read
`const auto& d : table.layout()->dims()`

Generally, I'd suggest careful review when replacing iterator-based loops to range-based loops to avoid unnecessary copying of potentially heavy objects.

io/TextWriter.cpp
```diff
@@ -141,57 +141,54 @@ void TextWriter::writeHeader(PointTableRef table)
...
-void TextWriter::writeFooter()
+void TextWriter::writeFooter() const
 {
     if (m_outputType == "GEOJSON")
     {
         *m_stream << "]}";
         if (m_callback.size())
             *m_stream  <<")";
     }
-    m_stream.reset();
```
mloskot: How this is an improvement of C++ Code

PR #1773, opened 2018-01-03
Fix compilation with gcc > 7

pblottiere:
Hi,

I had to add some gcc flags as well as adding a header to successfully compile with gcc 7.2.0. 

BTW, the `-Werror` flag is currently duplicated so I removed the second one.

Let me know if you're interested in this PR and if I have to modify something.

And thank you for your work!


abellgithub:
Will such warnings be ignored by older compilers?

pblottiere:
> Will such warnings be ignored by older compilers?

I'm not sure... 

Otherwise, these flags may be added for gcc > 7 only. Would you prefer that?

abellgithub:
Yes.  These warnings are unknown by earlier compilers and cause them to complain.  Thanks.

pblottiere:
> Yes. These warnings are unknown by earlier compilers and cause them to complain.

I updated the cmake file accordingly. Let me know if it suits you.

PR #1796, opened 2018-01-29
Config

abellgithub:
Simplify build system/compilation dependencies.


jlaxson:
@abellgithub I think this broke detecting laszip with PDAL_HAVE_LASZIP?  There is no place it gets defined now.

abellgithub:
Yes. Thanks. Fixed.

PR #1799, opened 2018-02-05
Update Alpine version to 3.7 for Travis

chambbj:
We have been pushing Docker images built on Alpine v3.7, but our Travis config
has been stuck back on v3.6. These should match.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,10 @@ sudo: required
 services: docker
 
 before_install:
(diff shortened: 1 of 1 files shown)


abellgithub:
Are you OK with me merging this?  Looks like a no-brainer.

chambbj:
Yes, I just wanted to see it pass Travis first.

PR #1807, opened 2018-02-12
tindex kernel fixes

hobu:
* initialized()
* no boundaries found


hobu:
> I don't understand the changes in TIndexKernel::slowBoundary. Unless I misread, the new code is functionally equivalent to what was there.

Previously, we would try to findChild("error") to the variable `m`, and then we would do `m.findChild` again. This would cause us to not get the corect data back.

PR #1808, opened 2018-02-12
Pyext

abellgithub:


abellgithub:
Some of these changes are differences from what was on PyPi.  Not sure what's what.

python/setup.py
```diff
@@ -189,7 +190,6 @@ def get_pdal_config(option):
         'Topic :: Scientific/Engineering :: GIS',
     ],
     cmdclass           = {},
-    install_requires   = ['numpy', 'packaging'],
```
hobu: Was there a reason to remove these?

Issue #1563, opened 2017-04-14
GDALGrid can fail if point doesn't fall in bounds


abellgithub:
If bounds are specified for a grid and a point doesn't fall in the boundary, there's a case that wasn't checked for OOB.

PR #1824, opened 2018-03-01
Fix segfault by checking bounds when adding a point to a GDALGrid.

andrew-hardin:
This pull request fixes out-of-bounds reads and writes when adding a point to a `GDALGrid`. It's related to #1563, but it handles all of the remaining edge cases. Here's an example of how to duplicate one of those edge cases:

1. Construct a grid with `width = 10; height = 10; edgeLength = 1; radius = 5`.
2. Call `GDALGrid::addPoint(-1000, 1, 0)`.
3. The first iteration for the first quadrant calls `GDALGrid::update` with a grid index of `-999,1`.
4. Out of bounds memory access; `-999, 1` doesn't fall within the bounds of the grid.

In keeping with the perceived intent of #1563, points that fall outside of the grid are silently ignored.

### DIFF
--- a/io/GDALGrid.cpp
+++ b/io/GDALGrid.cpp
@@ -281,9 +281,7 @@ void GDALGrid::addPoint(double x, double y, double z)
     // This is a questionable case.  If a point is in a cell, shouldn't
     // it just be counted?
     double d = distance(iOrigin, jOrigin, x, y);
-    if (d < m_radius &&
-        iOrigin >= 0 && jOrigin >= 0 &&
-        iOrigin < (int)m_width && jOrigin <= (int)m_height)
+    if (d < m_radius)
         update(iOrigin, jOrigin, z, d);
 }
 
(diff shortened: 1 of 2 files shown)


abellgithub:
What system are you using? I'm not sure how negative values would get past the distance check before update is called, as they're converted to huge unsigned values.  I can't recreate a failure with the data you provided.  That said, I think it can be done better.

I'm working on an adjustment that shouldn't require bounds checking of every addition. Maybe you could look it over?

abellgithub:
Here's the branch: https://github.com/PDAL/PDAL/compare/issue-1824

andrew-hardin:
Thank you for taking the time to review the issue and craft an alternative patch.
I pulled + tested your branch, but it still has invalid memory read + writes.
This commit fixed it so that your branch passed my tests:

abellgithub:
Yeah, I saw that one earlier yersterday, changed it, and then started some stuff over and it forgot to change it.  Thanks for catching.

abellgithub:
I'll add some tests for OOB situations.

abellgithub:
Closing in favor of #1830.

PR #1827, opened 2018-03-05
Plugin directory

abellgithub:
Simplify some plugin loading. Do less directory scanning.


chambbj:
I like the direction you've headed here. So no need to touch StageFactory at all when writing new stages (except for the extensions)?

abellgithub:
Not at all when I'm done. I'm going to read them from a json file. At least
that's the plan.

There's no boilerplate any more other than the CREATE macros. I'd still
like to remove the template stuff from the PluginManager, but I need to
find a good way to do it without duplicating too much code if I want to
keep the interface based on static methods. Thoughts appreciated.


On Mar 8, 2018 8:56 PM, "chambbj" <notifications@github.com> wrote:

I like the direction you've headed here. So no need to touch StageFactory
at all when writing new stages (except for the extensions)?

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/PDAL/PDAL/pull/1827#issuecomment-371687814>, or mute
the thread

PR #1828, opened 2018-03-05
output more detail for readers.bpf when the coordinate system configuration is not a combo we support

hobu:


io/BpfReader.cpp
```diff
@@ -135,6 +135,21 @@ void BpfReader::initialize()
...
+        {
+            std::ostringstream oss;
+            oss << "BPF has ECEF/TCR coordinate type defined, but the ID of '"
+                <<  m_header.m_coordId << "' is invalid";
+            throwError(oss.str());
+        }
+    }
     else
     {
        // BPF supports something called Terrestrial Centered Rotational
```
abellgithub: Comment needs fixing.

PR #1834, opened 2018-03-09
support skipping header lines with an option in readers.text

hobu:


hobu:
Maybe not committed?

io/TextReader.hpp
```diff
@@ -143,6 +143,7 @@ class PDAL_DLL TextReader : public Reader, public Streamable
     size_t m_line;
     std::string m_headerOverride;
     std::string m_headerInsert;
+    uint32_t m_headerSkip;
```
abellgithub: Should probably be size_t, though it doesn't much matter.

io/TextReader.cpp
```diff
@@ -131,7 +131,12 @@ void TextReader::initialize(PointTableRef table)
     if (!m_istream)
         throwError("Unable to open text file '" + m_filename + "'.");
 
+
     std::string header;
+
+    // Skip any lines requested
+    for (uint32_t i = 0; i < m_headerSkip; ++i)
```
abellgithub: This logic needs to be repeated in ready(), or some other accommodation needs to be made.

PR #1853, opened 2018-03-19
Add returns filter

chambbj:
Users can create PointViews that contain first, intermediate, last, and only returns (based on ReturnNumber and NumberOfReturns).

### DIFF
--- a/filters/ReturnsFilter.cpp
+++ b/filters/ReturnsFilter.cpp
@@ -118,7 +118,7 @@ PointViewSet ReturnsFilter::run(PointViewPtr inView)
             viewSet.insert(firstView);
         else
             log()->get(LogLevel::Warning)
-                << "Requested first returns set it empty\n";
+                << "Requested returns group 'first' is empty\n";
     }
 
     if (m_outputTypes & returnIntermediate)
@@ -127,7 +127,7 @@ PointViewSet ReturnsFilter::run(PointViewPtr inView)
             viewSet.insert(intermediateView);
         else
             log()->get(LogLevel::Warning)
-                << "Requested intermediate returns set is empty\n";
+                << "Requested returns group 'intermediate' is empty\n";
     }
 
     if (m_outputTypes & returnLast)
@@ -136,7 +136,7 @@ PointViewSet ReturnsFilter::run(PointViewPtr inView)
             viewSet.insert(lastView);
         else
             log()->get(LogLevel::Warning)
-                << "Requested last returns set is empty\n";
+                << "Requested returns group 'last' is empty\n";
     }
 
     if (m_outputTypes & returnOnly)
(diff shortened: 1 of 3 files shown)


filters/ReturnsFilter.cpp
```diff
@@ -0,0 +1,126 @@
...
+        if ((m_outputTypes & returnIntermediate) && (rn > 1) && (rn < nr) &&
+            (nr > 2))
+            intermediateView->appendPoint(*inView.get(), idx);
+        if ((m_outputTypes & returnLast) && (rn == nr) && (nr > 1))
+            lastView->appendPoint(*inView.get(), idx);
+        if ((m_outputTypes & returnOnly) && (nr == 1))
+            onlyView->appendPoint(*inView.get(), idx);
+    }
+
+    if (firstView->size())
```
abellgithub: It seems like if you've asked for certain sets, you should get them, regardless of whether they're empty or not.  I think this will omit the intermediate view, even if you've asked for it, as an example.

filters/ReturnsFilter.hpp
```diff
@@ -0,0 +1,78 @@
...
+ ****************************************************************************/
+
+#pragma once
+
+#include <pdal/Filter.hpp>
+
+#include <map>
+#include <string>
+
+extern "C" int32_t ReturnsFilter_ExitFunc();
```
abellgithub: This boilerplate is no longer necessary.

filters/ReturnsFilter.hpp
```diff
@@ -0,0 +1,78 @@
...
+{
+public:
+    ReturnsFilter() {}
+
+    static const int returnFirst = 1;
+    static const int returnIntermediate = 2;
+    static const int returnLast = 4;
+    static const int returnOnly = 8;
+
+    static void* create();
```
abellgithub: This create(), destroy() boilerplate is no longer necessary.

filters/ReturnsFilter.hpp
```diff
@@ -0,0 +1,78 @@
...
+private:
+    std::map<uint64_t, PointViewPtr> m_viewMap;
+    StringList m_returnsString;
+    int m_outputTypes;
+
+    virtual void addArgs(ProgramArgs& args);
+    virtual void prepared(PointTableRef table);
+    virtual PointViewSet run(PointViewPtr view);
+
+    ReturnsFilter& operator=(const ReturnsFilter&); // not implemented
```
abellgithub: You can probably explictly delete these now. I think Windows is OK with it.

filters/ReturnsFilter.cpp
chambbj: I'm happy to make that change. I'm struggling to come up with a scenario in which an empty PointView would be useful. Would it be more useful to log a warning, at least acknowledging the request and explaining why it was not fulfilled?

PR #1861, opened 2018-03-21
Invalidate aux. point view products

### LINKED ISSUE #1858: Rebuilding KDTree index upon adding/removing points
The KDTree index should be invalidated and (partly?) rebuild when removing or adding points in the pipeline process. Currently the index is propagated in the pipeline but there seems no way of invalidating it when changes on the pointset occur.

This would be of value when for instance filtering out points based on certain normal attributes  like only keeping the points that are on a horizontal plane. The index would have been build for calculating the normals and still exist after discarding points.

N.B. Afaik there are no filters currently adding points but possible future use cases for adding points could be synthetic points to fill up holes like removed buildings or water.

abellgithub:
Close #1858

### DIFF
--- a/filters/TransformationFilter.cpp
+++ b/filters/TransformationFilter.cpp
@@ -122,6 +122,7 @@ void TransformationFilter::filter(PointView& view)
         point.setPointId(idx);
         processOne(point);
     }
+    view.invalidateProducts();
 }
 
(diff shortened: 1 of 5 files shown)


hobu:
Invalidate what?

PR #1875, opened 2018-03-28
Add enumerate option to info command

ahuarte47:
This adds the "enumerate" option to "info" command in order to set the Dimensions whose values should be enumerated by StatsFilter.


ahuarte47:
Thanks! @abellgithub

PR #1886, opened 2018-04-03
change name of fileDimensions to avoid confliction with SbetCommon

### LINKED ISSUE #1885: Conflict definitions of pdal::fileDimensions in SbetCommon and PcdCommon
When translating from a PCD file that contains intensity information, the output data have a set of fields that are mostly unrelated to the input, and the intensity field is missing:

>"GpsTime","Y","X","Z","XVelocity","YVelocity","ZVelocity","Roll","Pitch","Azimuth","WanderAngle","XBodyAccel","YBodyAccel","ZBodyAccel","XBodyAngRate","YBodyAngRate","ZBodyAngRate"

The reason is that the registered dimensions in the PcdReader are defined in`plugins/pcl/io/PcdReader.cpp`, which calls `pdal::fileDimensions()`. However, this method is defined in `io/SbetCommon.cpp` and `plugins/pcl/io/PcdCommon.cpp` at the same time. And the PcdReader links to the wrong version of implementation.

Maybe a simply rename of the method in the PCL plugin could solve the problem?
I don't know why the linker dose not complain about the duplicate definitions.

ambling:
Fixes #1885

### DIFF
--- a/plugins/pcl/io/PcdCommon.cpp
+++ b/plugins/pcl/io/PcdCommon.cpp
@@ -37,7 +37,7 @@
 namespace pdal
 {
 
-Dimension::IdList fileDimensions()
+Dimension::IdList pcdDimensions()
 {
     Dimension::IdList ids;
 
--- a/plugins/pcl/io/PcdCommon.hpp
+++ b/plugins/pcl/io/PcdCommon.hpp
@@ -39,6 +39,6 @@
 namespace pdal
 {
 
-PDAL_DLL Dimension::IdList fileDimensions();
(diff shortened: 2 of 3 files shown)


abellgithub:
Thanks for finding this!

I think I'd rather the symbols be made local so as to avoid any possibility of a future conflict. Would you like to do this or should I?

ambling:
@abellgithub Yes I agree with the local symbol method. 

However, I am not very much familiar with PDAL right now. Please fix the issue if you have any time, and close this pr after that.

ambling:
close due to another better method #1888

PR #1888, opened 2018-04-03
Issue 1885

abellgithub:
Make fileDimensions() local to the translation units in which they're included.


abellgithub:
@ambling: If you want to look this over or test, that would be great.

ambling:
@abellgithub It seems that the problem has been solved. Thanks!

PR #1892, opened 2018-04-04
Execute info command as streamable

ahuarte47:
This change allows to execute info command as a streamable stage

### DIFF
--- a/kernels/InfoKernel.cpp
+++ b/kernels/InfoKernel.cpp
@@ -349,7 +349,17 @@ MetadataNode InfoKernel::run(const std::string& filename)
     else
     {
         if (m_needPoints || m_showMetadata)
-            m_manager.execute();
+        {
+            if (m_manager.pipelineStreamable())
+            {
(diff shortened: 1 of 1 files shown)


abellgithub:
There are problems with this given the current design of the info kernel. It can cause crashes. Sorry.

I'll open a ticket to do something, but it's more complicated than this.

ahuarte47:
Ok @abellgithub, thanks!

abellgithub:
If you look at the history, you'll see this change was there and then pulled. The issue is that some options assume non-streaming mode (assume a point view exists). The existing code probably needs an overhaul to do this right. If you want to tackle that, let me know, but it's a bit ugly. Thanks for your interest in this!

ahuarte47:
I see it, I am only using "--dimensions" and "--enumerate" options with very large LAS files (~3gb). If you agree and I could detect what options force to "info" to run as standard mode in order to safely set the streamable mode. Of course, if you prefer to do yourself, I do nothing now.

PR #1924, opened 2018-04-12
CMake modification

chambbj:
This PR does the following:
- removes unused macros `DISSECT_VERSION`, `PDAL_ADD_EXECUTABLE`, `PDAL_ADD_KERNEL`, and `PDAL_ADD_DRIVER`
- converts two macros to CMake include files that are called once during setup (`GET_OS_INFO` and `SET_INSTALL_DIRS`)
- parses version numbers automatically from `project()` (**bumps CMake required version to 3.0+**)
- specifies C+11 for all targets and disables extensions (**bumps CMake required version to 3.1+**)


cmake/osinfo.cmake
```diff
@@ -0,0 +1,12 @@
+# Get the operating system information. Generally, CMake does a good job of
+# this. Sometimes, though, it doesn't give enough information. These steps will
+# distinguish between the UNIX variants. Otherwise, use the CMake variables
+# such as WIN32 and APPLE and CYGWIN.
+# Sets OS_IS_64BIT if the operating system is 64-bit.
+# Sets LINUX if the operating system is Linux.
+string(REGEX MATCH "Linux" OS_IS_LINUX ${CMAKE_SYSTEM_NAME})
```
abellgithub: I just searched and can't find this stuff being used anywhere.  Perhaps it can all be ditched.

hobu: I think it is old junk from early CMake 2.x days.

chambbj: Interesting. I wasn't sure of the origin. Just this particular line @abellgithub or the whole thing?

chambbj: Forget it. I see it's really just checking that one thing.

CMakeLists.txt
```diff
@@ -15,6 +14,10 @@ include(${ROOT_DIR}/cmake/common.cmake NO_POLICY_SCOPE)
 # internal cmake settings
 #------------------------------------------------------------------------------
 
+set(CMAKE_CXX_STANDARD 11)
```
abellgithub: It seems conceivable to me that we inherit bits of code that need to be compiled as C or not as C++11.  In changes I'm making for #1922, I stuck this in the function pdal_target_compile_settings().  Is that better than global? Just wondering.

chambbj: I'll take it out here, and let you define it on the targets where we know we want it.

cmake/osinfo.cmake
abellgithub: I meant that entire block of code that you were moving.

CMakeLists.txt
chambbj: I don't really know how these play together. I wonder if it is still worthwhile to disable extensions, even while leaving the standard undefined. Maybe that's a per target thing as well.

hobu:
I don't want to merge this to 1.7 because it increases the minimum CMake version. I'll add a note to the release notes to that effect.

chambbj:
Reopening as #2693

PR #1925, opened 2018-04-13
CMake will automatically append lib to the library name on Windows

chambbj:


chambbj:
Closing. I forgot that this **is** the expected naming structure on both Windows and *nix variants.

PR #1928, opened 2018-04-13
PointTableRef accessor for PointView

hobu:


jw3:
Can this be picked into 1.7?

hobu:
Yes.

PR #1946, opened 2018-04-20
Use CMAKE_DL_LIBS

chambbj:
I believe we inadvertently removed `CMAKE_DL_LIBS` [here](https://github.com/PDAL/PDAL/commit/afbfccab6b2a06dab22b8b5e2e9d3e2bdd9e384b#diff-9df7bbdfc0579f4e154ab1226d9ffd7f), and then replaced it with `dl` [here](https://github.com/PDAL/PDAL/commit/95407c152bdc2afcc401578b1297d2bc2915d801#diff-9df7bbdfc0579f4e154ab1226d9ffd7f).

`CMAKE_DL_LIBS` is already used in the following locations:

https://github.com/PDAL/PDAL/blob/b6474df9cc58599509ec93e6d97d1ad4b6284541/dimbuilder/CMakeLists.txt#L40

https://github.com/PDAL/PDAL/blob/c709b5d118edd13f2871deb93620630241655d4d/plugins/python/filters/CMakeLists.txt#L12

https://github.com/PDAL/PDAL/blob/f22bffd723c530af57d89eb97609fcdda253bc4f/plugins/python/io/CMakeLists.txt#L9

We should probably be consistent and use it across the board (although clearly it hasn't seemed to cause any issues to date).

### DIFF
--- a/dimbuilder/CMakeLists.txt
+++ b/dimbuilder/CMakeLists.txt
@@ -31,11 +31,10 @@ target_include_directories(dimbuilder PRIVATE
 PDAL_TARGET_COMPILE_SETTINGS(dimbuilder)
 target_link_libraries(dimbuilder
     PRIVATE
-        ${EXECINFO_LIBRARY})
+        ${EXECINFO_LIBRARY}
+        ${CMAKE_DL_LIBS}
+)
 if (PDAL_HAVE_JSONCPP)
     target_link_libraries(dimbuilder ${JSON_CPP_LINK_TYPE}
         ${PDAL_JSONCPP_LIB_NAME})
-endif()
-if (UNIX AND NOT APPLE)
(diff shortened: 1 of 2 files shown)


abellgithub:
I'm almost certain this doesn't need to guarded with if (APPLE) or whatever.  I think the point of it is to avoid it.  I'm also not sure why we're not linking pdalcpp with it, because it clearly needs it for plugins.

chambbj:
Yes, I had wondered about the guards, but kept with what we'd done
elsewhere. I can verify locally, and perhaps replace all guarded instances
in this PR.

hobu:
Cherry-picked to `1.7-maintenance` @b76cf5081f852c4cc5b55cdfb00dadc0e7601070

PR #1951, opened 2018-04-21
Pass parameters by const reference (cppcheck)

rouault:
This fixes all cppcheck warnings
```(performance) Function parameter 'XXX' should be passed by const reference```
(except in vendor/ subdirectory)

### DIFF
--- a/filters/StatsFilter.hpp
+++ b/filters/StatsFilter.hpp
@@ -57,7 +57,7 @@ typedef std::map<double, point_count_t> EnumMap;
 typedef std::vector<double> DataVector;
 
 public:
-    Summary(std::string name, EnumType enumerate) :
+    Summary(const std::string& name, EnumType enumerate) :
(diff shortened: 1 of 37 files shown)


abellgithub:
I'd really rather not do this. It seems totally unnecessary.

abellgithub:
With the exception of the vectors being passed to the Eigen calls, I see
nothing here that matters at all. None of the functions are called in a way
where the change would be consequential from a performance standpoint.  I
simply prefer not to introduce changes that don't have any consequence for
users or to make things structurally better for developers.  Changing a
bunch of value calls to const refs calls where the calls aren't used in any
volume simply don't matter, other than to introduce another change in the
file history.

Someone else is welcome to offer another opinion.  In the end it doesn't
matter much, which is kind of my point.

abellgithub:
I've committed the Eigen change.  I don't think there's anything else here worth pushing as a block change.  This can be reopened if necessary.

PR #1953, opened 2018-04-21
XMLSchema::load(): fix cppcheck warning 'Statements following return, break, continue, goto or throw will never be executed'

rouault:


rouault:
Surrounding code also uses this pattern : ```std::cerr << "....."; return false; ```

abellgithub:
OK.  All of this needs an overhaul.  The error messages have no context, but I understand that's not the issue you're addressing.

PR #1954, opened 2018-04-21
[cppcheck] Remove a number of unused variables or add FIXME where they should likely be used

rouault:


pdal/SpatialReference.cpp
```diff
@@ -196,7 +196,7 @@ std::string SpatialReference::getVerticalUnits() const
 
     std::string wkt = getVertical();
     const char* poWKT = wkt.c_str();
-    OGRScopedSpatialReference poSRS = ogrCreateSrs(m_wkt);
+    OGRScopedSpatialReference poSRS = ogrCreateSrs(poWKT);
```
hobu: This would likely be an intentional or unintentional behavior change.

rouault: wkt / poWKT was unused before. It makes more sense for a getVerticalUnits() method to use the wkt of getVertical() right ?

plugins/hexbin/filters/HexBin.cpp
```diff
@@ -192,7 +192,6 @@ void HexBin::done(PointTableRef table)
     {
         // Compute a UTM polygon
         BOX3D box = p.bounds();
-        int zone = SpatialReference::calculateZone(box.minx, box.miny);
```
abellgithub: Was there something wrong here? Whether this line comes before or after the definition of the function would seem to make no difference.

rouault: No there wasn't anything wrong. cppcheck has false positives in its analysis (it is not a full blown C++ compiler). That's a workaround to avoid one

abellgithub:
The important stuff here has been fixed.  cppcheck tells you LOTS of things that might be wrong. If we're going to use it formally, we should probably discuss it.

PR #1958, opened 2018-04-23
Add condition option to assign filter

### LINKED ISSUE #1956: Conditional assignment of values to a DimRange
I would like to use a condition in one dimension to drive assignment of values to another dimension. For example,

- Assign classification of 2 (ground) to all points with height above ground less than 0.3 m
- Assign classification of 6 (building) to all points with estimated rank of 2

In our current setup, my only option using **core stages** is to implement a sequence of range/assign filters to achieve the same end goal (it would clearly be possible to do this with `filters.matlab` or `filters.python` as well). Using the second example above, assuming we have already estimated the rank, the pipeline required to make the assignment would be

```
[
    {
        "type":"filters.range",
        "inputs":["precomputed-rank"],
        "limits":"Rank[2:2]",
        "tag":"rank2"
    },
    {
        "type":"filters.assign",
        "assignment":"Classification[:]=6",
        "inputs":["rank2"]
    }
]
```

If we also wanted to apply a vegetation label (e.g., classification 5) to points whose estimated rank is 3

```
[
    {
        "type":"filters.range",
        "inputs":["precomputed-rank"],
        "limits":"Rank[3:3]",
        "tag":"rank3"
    },
    {
        "type":"filters.assign",
        "assignment":"Classification[:]=5",
        "inputs":["rank3"]
    }
]
```

Because we have reused the `precomputed-rank` input stage, we now have multiple paths through the pipeline that must be computed (each time we branch, we actually create two separate pipelines under the hood).

Another option would be to code flexibility in assignments into the stages themselves. We do have some limited precedent. For example, several filters (e.g., `filters.elm`) allow us to specify the value that will be assigned to the Classification dimension, but not the dimension itself.

A more straightforward modification of the existing `filters.assign` would allow us to take a single pass through the pipeline, assigning values based on the specified condition.

Reusing the above example, the combined assignment of classifications based on estimated rank becomes:

```
[
    {
        "type":"filters.assign",
        "condition":"Rank[2:2]",
        "assignment":"Classification[:]=6"
    },
    {
        "type":"filters.assign",
        "condition":"Rank[3:3]",
        "assignment":"Classification[:]=5"
    }
]
```

chambbj:
Values will only be assigned when the condition (specified as a DimRange) is
met.

Closes #1956

### DIFF
--- a/filters/AssignFilter.cpp
+++ b/filters/AssignFilter.cpp
@@ -123,12 +123,19 @@ void AssignFilter::addArgs(ProgramArgs& args)
 {
     args.add("assignment", "Values to assign to dimensions based on range.",
         m_assignments);
+    args.add("condition", "Condition for assignment based on range.",
+        m_condition);
 }
 
 
 void AssignFilter::prepared(PointTableRef table)
 {
     PointLayoutPtr layout(table.layout());
+
+    m_condition.m_id = layout->findDim(m_condition.m_name);
+    if (m_condition.m_id == Dimension::Id::Unknown)
+        throwError("Invalid dimension name in 'condition' option: '" +
+            m_condition.m_name + "'.");
 
     for (auto& r : m_assignments)
     {
@@ -143,7 +150,8 @@ void AssignFilter::prepared(PointTableRef table)
 bool AssignFilter::processOne(PointRef& point)
 {
     for (AssignRange& r : m_assignments)
-        if (r.valuePasses(point.getFieldAs<double>(r.m_id)))
+        if (r.valuePasses(point.getFieldAs<double>(r.m_id)) &&
+            m_condition.valuePasses(point.getFieldAs<double>(m_condition.m_id)))
             point.setField(r.m_id, r.m_value);
     return true;
 }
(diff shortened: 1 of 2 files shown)


filters/AssignFilter.cpp
```diff
@@ -123,13 +123,20 @@ void AssignFilter::addArgs(ProgramArgs& args)
...
+        m_condition);
 }
 
 
 void AssignFilter::prepared(PointTableRef table)
 {
     PointLayoutPtr layout(table.layout());
 
+    m_condition.m_id = layout->findDim(m_condition.m_name);
+    if (m_condition.m_id == Dimension::Id::Unknown)
```
chambbj: If no condition is provided, then `filters.assign` should continue to function as it always has, applying the assignment to **all points**, instead of throwing an error.

chambbj:
We should continue to discuss in #1956 whether or not this is the desired way forward.

chambbj:
I will close this for now to keep our PR queue tidy. We can reopen if needed pending discussion in #1956.

PR #1972, opened 2018-04-25
Allow underscores in stage names.

### LINKED ISSUE #1923: PDAL 1.7 plugins with underscores in name
We have plugins that have underscores in their names, eg. a filter named `foo_bar_baz`, its shared lib is `libpdal_plugin_filter_foo_bar_baz.so`.

When upgrading to 1.7 a library name in this format does not appear to be detected.  A quick look shows maybe [this block in PluginDirectory](https://github.com/PDAL/PDAL/blob/c9a06fd42f48782e505071043ab4ac95c176edf2/pdal/PluginDirectory.cpp#L92-L96) is limiting it.

[This change](https://github.com/jw3/PDAL/commit/29ac2227f98853e1bf270abf266857a5acb611a8) might be a quick "fix" for it.  Are there any other concerns within the plugin system after changing the behavior of `PluginDirectory`?

abellgithub:
Harden check for extraction of valid plugins.
Add tests for valid plugin names.
Close #1923

### DIFF
--- a/pdal/PluginDirectory.cpp
+++ b/pdal/PluginDirectory.cpp
@@ -43,13 +43,13 @@ namespace
 {
 
 #if defined(__APPLE__) && defined(__MACH__)
-    static const std::string dynamicLibraryExtension("dylib");
+    static const std::string dynamicLibraryExtension(".dylib");
     static const char pathSeparator(':');
 #elif defined(__linux__) || defined(__FreeBSD__) || defined(__DragonFly__) || defined(__FreeBSD_kernel__) || defined(__GNU__)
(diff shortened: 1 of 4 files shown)


abellgithub:
The previous code required a lowercase letter to be the first character of the name.  If you're wanting it to be open to any letter, this looks fine to me, though a test case for that would be nice.

hobu:
Pushed, with test addition.

hobu:
Cherry picked to `1.7-maintenance` @6d86711c0d47102763bf790cec63cfcbff86f754

PR #1975, opened 2018-04-26
Fix warning about unused formal parameter in Streamable.hpp

mloskot:
Follows convention already used in the file.

----

OS: Windows
Compiler: `cl.exe /W4` from VS2017

```
streamable.hpp(102): warning C4100: 'srs': unreferenced formal parameter 
```


mloskot:
@abellgithub I've updated the issue

PR #1999, opened 2018-05-07
Make southern UTM zones negative.

### LINKED ISSUE #1989: SpatialReference::getUTMZone() in southern hemisphere
We're using the new `SpatialReference::getUTMZone()` accessor in 1.7, and getting a return value of 0 for a point cloud in UTM zone 58S.

The problem seems to be the [hemisphere logic](https://github.com/PDAL/PDAL/blob/d76922bd893e99f1426d5d6adab2f50b3f088d14/pdal/SpatialReference.cpp#L397-L399):

    int north(0);
    int zone = OSRGetUTMZone(current.get(), &north);
    return north*zone;

This would make sense if `north` were set to  `-1` for the southern hemisphere, but according to the [GDAL documentation](http://www.gdal.org/classOGRSpatialReference.html#a137a1f15ba0d18b84e7fc4104c74cdb9), it's actually set to `false` for the C/C++ bindings, so the multiplication results in a 0 return value.

abellgithub:
Close #1989

### DIFF
--- a/test/unit/SpatialReferenceTest.cpp
+++ b/test/unit/SpatialReferenceTest.cpp
@@ -392,6 +392,7 @@ TEST(SpatialReferenceTest, test_bounds)
 
 }
 
+// Make sure we get positive, negative and 0 back for UTM zones.
 TEST(SpatialReferenceTest, issue_1989)
 {
     SpatialReference srs;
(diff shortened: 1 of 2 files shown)


test/unit/SpatialReferenceTest.cpp
```diff
@@ -391,3 +391,16 @@ TEST(SpatialReferenceTest, test_bounds)
     EXPECT_FLOAT_EQ(static_cast<float>(b2.maxy), 39.01261687f);
 
 }
+
+TEST(SpatialReferenceTest, issue_1989)
+{
+    SpatialReference srs;
```
hobu: Maybe add a test case for `EPSG:32732`, which is a common style of UTM zone that many use to ensure it always passes. Also, please add a comment in the test in reference to what issue 1989 is. If our ticket system were to migrate or move away, knowing a tiny bit of context so that people can get by without having to dereference to the ticket system is helpful.

PR #2021, opened 2018-05-16
Handle repeated options.

### LINKED ISSUE #1941: Repeated options in pipelines don't work
Repeating options in pipelines doesn't work:

```
{
  "type": "filters.assign",
  "assignment": "Classification[:]=0",
  "assignment": "Red[:]=0
}
```

Our JSON parser throws away all but the last value for the repeated option.

abellgithub:
Close #1941

### DIFF
--- a/doc/pipeline.rst
+++ b/doc/pipeline.rst
@@ -122,7 +122,7 @@ exist.
 Each of these output sets is called a point view.  Point views are carried
 through a PDAL pipeline individually.  Some writers can produce separate
 output for each input point view.  These writers use a placeholder character
-(#) in the output filename which is replaced by a incrementing integer for
+(#) in the output filename which is replaced by an incrementing integer for
 each input point view.
 
 The following pipeline provides an example of writing multiple output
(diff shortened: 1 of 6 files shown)


doc/pipeline.rst
```diff
@@ -113,6 +113,39 @@ with the :ref:`writers.gdal` writer:
...
+Point Views and Multiple Outputs
+................................................................................
+
+Some filters produce sets of points as output.  filters.splitter, for example,
+creates a point set for each tile (rectangular area) in which input points
+exist.
+Each of these output sets is called a point view.  Point views are carried
+through a PDAL pipeline individually.  Some writers can produce separate
+output for each input point view.  These writers use a placeholder character
+(#) in the output filename which is replaced by a incrementing integer for
```
chambbj: Should read `replaced by an incrementing integer`.

abellgithub:
If an option is bound to vector<string>, it can be parsed by ProgramArgs as a comma-separated list.

Theoretically, you can handle any argument type that takes a vector<type> as you like by adding appopriate code.  But with this change you can let the JSON parser break the options up for you and then add them one at a time.  There are a bunch of places where we take vector<std::string> as options, and these can be handled either way.  I changed the colorization example just to provide an example using the syntax that's now supported.

chambbj:
LGTM

PR #2027, opened 2018-05-19
Simplify os-specific #ifdefs

landryb:
Instead of manually listing all non-macos unix platforms, make it the fallback
case, this way all BSDs are included.

As a side effect, fixes OpenBSD and NetBSD for free :)


abellgithub:
Thanks for this.

PR #2034, opened 2018-05-23
Improve filters.stats

### LINKED ISSUE #1878: filters.stats is expensive
We should make many of the advanced statistics in `filters.stats` optional, or we should push them into another filter that is allowed to be slow.

A simple `pdal info` of boring data shows lots of the time being spent computing statistics that we don't need most of the time. 

screen shot 2018-03-30 at 8 16 22 am

abellgithub:
Eliminate some unnecessary math.
Put skewness and kurtosis under an advanced option.
Improve tests.
Close #1878


abellgithub:
I think the default argument is there on line 60. Are we talking about two different things?

hobu:
Oh I missed it on the diff. All good.

PR #2045, opened 2018-06-05
Fix segmentation of last returns

chambbj:
segmentLastReturns now defines last returns as those points whose ReturnNumber
equals the NumberOfReturns when the NumberOfReturns is greater than 1.

This segmentation is used in both the SMRF and PMF ground segmentation filters,
so ground segmentation results may differ from what was previously returned.

### DIFF
--- a/pdal/Segmentation.cpp
+++ b/pdal/Segmentation.cpp
@@ -130,12 +130,11 @@ void ignoreDimRange(DimRange dr, PointViewPtr input, PointViewPtr keep,
 void segmentLastReturns(PointViewPtr input, PointViewPtr last,
                         PointViewPtr other)
 {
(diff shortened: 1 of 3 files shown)


pdal/Segmentation.cpp
```diff
@@ -130,12 +130,11 @@ void ignoreDimRange(DimRange dr, PointViewPtr input, PointViewPtr keep,
 void segmentLastReturns(PointViewPtr input, PointViewPtr last,
                         PointViewPtr other)
 {
-    PointRef point(*input, 0);
     for (PointId i = 0; i < input->size(); ++i)
     {
-        point.setPointId(i);
```
abellgithub: Why did this line go away?

PR #2054, opened 2018-06-11
Add prefiltering options to ground kernel

### LINKED ISSUE #1579: Ground kernel pre-filtering
The kernel could probably be improved by adding the ability to reset existing classifications and mark noise points.

chambbj:
- Add a switch to optionally reset classifications prior to segmenting ground returns
- Add a switch to optionally denoise prior to segmenting ground returns

Fixes #1579

### DIFF
--- a/kernels/GroundKernel.cpp
+++ b/kernels/GroundKernel.cpp
@@ -65,7 +65,7 @@ std::string GroundKernel::getName() const
 GroundKernel::GroundKernel()
     : Kernel(), m_inputFile(""), m_outputFile(""), m_maxWindowSize(33),
       m_slope(1), m_maxDistance(2.5), m_initialDistance(0.15), m_cellSize(1),
-      m_extract(false)
+      m_extract(false), m_reset(false), m_denoise(false)
 {
 }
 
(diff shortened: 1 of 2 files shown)


kernels/GroundKernel.cpp
```diff
@@ -78,25 +78,42 @@ void GroundKernel::addSwitches(ProgramArgs& args)
...
     Stage& readerStage(makeReader(m_inputFile, ""));
-    Stage& groundStage = makeFilter("filters.pmf", readerStage, groundOptions);
+
+    Stage* assignStage = &readerStage;
+    if (m_reset)
+        assignStage = &makeFilter("filters.assign", readerStage, assignOptions);
+
+    Stage* outlierStage = assignStage;
+    if (m_denoise)
+        outlierStage = &makeFilter("filters.outlier", *assignStage, outlierOptions);
```
abellgithub: I wonder the name of the option.  Are we hiding `filters.outlier` by calling it `denoise`?  Would it be more obvious what was going on if we just called it `outlier`?

chambbj:
Maybe. Although the purpose of the ground kernel and options is to provide
an easy button of sorts, and to hide implementation details. Denoise should
I think trigger whatever state of the art filter (or filters) sets the
ground algorithm up for success. We could just as easily insert
`filters.elm` under the hood. People who want to control more directly have
methods of doing that without the kernel.

PR #2056, opened 2018-06-12
Rewrite quickstart docs with Conda focus, dropping Docker

chambbj:
Still need to generate some new screenshots, but this is a first look at the rewritten Quickstart featuring Conda.


chambbj:
@hobu Okay to let go of the images and just display the expected outputs as literal code blocks?

PR #2076, opened 2018-07-05
Add avg option to NNdistance filter.

### LINKED ISSUE #2071: Add average distance support for nearest neighbor queries.

abellgithub:
Close #2071

### DIFF
--- a/filters/NNDistanceFilter.cpp
+++ b/filters/NNDistanceFilter.cpp
@@ -72,8 +72,7 @@ std::istream& operator>>(std::istream& in, NNDistanceFilter::Mode& mode)
     else if (s == "avg")
         mode = NNDistanceFilter::Mode::Average;
     else
-        throw pdal_error("filters.nndistance: Invalid 'mode' option '" + s +
-            "'. Valid options are 'kth' and 'avg'.");
+        in.setstate(std::ios_base::failbit);
     return in;
 }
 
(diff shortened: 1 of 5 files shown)


filters/NNDistanceFilter.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+
+void NNDistanceFilter::filter(PointView& view)
+{
+    using namespace Dimension;
+
+    // Build the 3D KD-tree.
+    log()->get(LogLevel::Debug) << "Building 3D KD-tree...\n";
+    KD3Index& index = view.build3dIndex();
```
chambbj: Is this new? I don't remember seeing this before.

filters/NNDistanceFilter.cpp
```diff
@@ -0,0 +1,142 @@
...
+        double val;
+        if (m_mode == Mode::Kth)
+            val = std::sqrt(sqr_dists[k - 1]);
+        else // m_mode == Mode::Average
+        {
+            val = 0;
+
+            // We start at 1 since index 0 is the test point.
+            for (size_t i = 1; i < k; ++i)
+                val += sqr_dists[i];
```
chambbj: It looks like you are actually returning the average squared distance. Am I missing something?

abellgithub: Nope. I am ;)

filters/NNDistanceFilter.cpp
abellgithub: Which bit? The log message?  The function call isn't new, but I used it instead of what was there before - if a KDtree already exists, this will just return it.

Issue #2077, opened 2018-07-05
Add definition of pdal::Config::lazPerfEnabled()


jimmysoda:
The `pdal::Config::lazPerfEnabled()` function declared in pdal/pdal_config.hpp is not defined in pdal/pdal_config.cpp or elsewhere in PDAL.

jimmysoda:
The `pdal::Config::lazPerfEnabled` declaration should be removed instead. See #2078 and #2080.

PR #2078, opened 2018-07-06
Add definition of pdal::Config::lazPerfEnabled()

jimmysoda:
Addresses #2077. The `pdal::Config::lazPerfEnabled()` function declared in pdal/pdal_config.hpp was not defined in pdal/pdal_config.cpp or elsewhere in PDAL. This pull request adds its definition to pdal/pdal_config.cpp.

### DIFF
--- a/pdal/pdal_config.cpp
+++ b/pdal/pdal_config.cpp
@@ -112,6 +112,12 @@ bool hasFeature(Feature f)
 }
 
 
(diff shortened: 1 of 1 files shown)


abellgithub:
The function signature should have been removed when the config code was overhauled.  Please use pdal::Config::hasFeature(LAZPERF) instead.

PR #2096, opened 2018-07-19
make readers.gdal streamable

hobu:


io/GDALReader.cpp
```diff
@@ -126,81 +131,60 @@ void GDALReader::ready(PointTableRef table)
...
+    m_raster->pixelToCoord(m_col, m_row, coords);
+    double x = coords[0];
+    double y = coords[1];
+    point.setField(Dimension::Id::X, x);
+    point.setField(Dimension::Id::Y, y);
 
-    std::vector<uint8_t> band;
+    static std::vector<double> data;
 
     for (int b = 0; b < m_raster->bandCount(); ++b)
```
abellgithub: I don't think you want to do this.  You're calling m_raster->read() for each band, but it returns data for all bands.  You want the read() outside of the band loop.

io/GDALReader.cpp
```diff
@@ -126,81 +131,60 @@ void GDALReader::ready(PointTableRef table)
...
+        m_row ++;
     }
+    m_raster->pixelToCoord(m_col, m_row, coords);
+    double x = coords[0];
+    double y = coords[1];
+    point.setField(Dimension::Id::X, x);
+    point.setField(Dimension::Id::Y, y);
 
-    std::vector<uint8_t> band;
+    static std::vector<double> data;
```
abellgithub: Making this static precludes threading.  Better just to drop the static IMO.

io/GDALReader.cpp
```diff
@@ -81,6 +83,8 @@ void GDALReader::initialize()
     m_count = m_raster->width() * m_raster->height();
     m_bandTypes = m_raster->getPDALDimensionTypes();
     m_raster->close();
+    m_row = 0;
```
abellgithub: These belong in ready().  That allows execute() to be called more than once.  Yes, this probably never happens.

hobu: The rules of when to do which aren't so clear.

Issue #1965, opened 2018-04-24
Add points using filters.python


clucas111:
The point cloud we are working with regularly has holes on the roofs of houses, which is a consequence of puddles of water at the time of scanning. These holes in the data result in a reduced performance of our algorithm. To fix this we want to add synthetic points at these locations. Since puddles of water only appear on flat surfaces we can fill these holes with points at the same height as the surrounding points.

I made an algorithm in python that finds the holes in the data and fills them with these synthetic points. Now I want to add this python script in a PDAL pipeline, but I have found no way of adding these points to the point cloud, to be used in subsequent steps in a PDAL pipeline. If I set the 'outs' 'X', 'Y' and 'Z' parameters to equal the arrays of the x, y and z of the points with the synthetic points included:

outs['X'] = points_filled[:, 0]
outs['Y'] = points_filled[:, 1]
outs['Z'] = points_filled[:, 2]

and subsequently run the PDAL pipeline and save it to a file it will only write the original points. Is there a way for me to add these points?

hobu:
This is not possible today, but it would be possible to implement. `filters.python` pre-dates the PointView concept in PDAL, and that is why it uses its special `Mask` dimension. `filters.python` could also be made to add or remove points from its PointView(s), but modifying the schema of the data in-place would never be allowed. 

The approach would require some special consideration of PointId for points, but that limitation shouldn't be too restrictive.

hobu:
Added for 1.8 in f1b695b874a8c3dc43e1a061ecb96089d282acef

PR #2120, opened 2018-08-02
support adding and deleting points to filters.numpy 'out' dictionary #1965

hobu:
If you want to delete points, all dimensions must have the same point count. See the `PythonFilterTest_modify` in PythonFilterTest.cpp for an example.

### DIFF
--- a/plugins/python/plang/Invocation.cpp
+++ b/plugins/python/plang/Invocation.cpp
@@ -163,7 +163,7 @@ void Invocation::insertArgument(std::string const& name, uint8_t* data,
 
 
 void *Invocation::extractResult(std::string const& name,
-    Dimension::Type t)
+    Dimension::Type t, size_t& num_elements)
 {
     PyObject* xarr = PyDict_GetItemString(m_varsOut, name.c_str());
     if (!xarr)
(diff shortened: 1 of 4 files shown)


plugins/python/plang/Invocation.cpp
```diff
@@ -416,13 +424,22 @@ void Invocation::end(PointView& view, MetadataNode m)
...
+        void *data = extractResult(name, dd->type(), arrSize);
         char *p = (char *)data;
         for (PointId idx = 0; idx < view.size(); ++idx)
         {
             view.setField(d, dd->type(), idx, (void *)p);
             p += size;
         }
+
+        // Set fields for any extra points we've added
+        for (PointId idx = view.size(); idx < arrSize; ++idx)
```
hobu: this is dumb. We should just iterate to `arrSize` so users can add/remove as necessary

abellgithub:
I looked some more at this and I'm not sure that it can work in a general manner.  The mask part should work, since you're not changing any data, just grabbing a subset of points.  But the add/delete thing won't.  In order for the add/delete to work, all existing unchanged points need to be copied to a new view.  But you don't know which points have been added/deleted.  Right now it's also an issue because we're only copying back the dimensions that are in the "outs" list, which will make the out-of-sync with the existing data.  I'm not sure the best way to handle this.

PR #2126, opened 2018-08-08
Don't use KDtree.

### LINKED ISSUE #2124: VoxelCenterNearestNeighborFilter can be much faster
The filter uses a KDTree but doesn't need to.  Large KDTrees are slow.

abellgithub:
By my test case, this is about a 10x improvement.
Close #2124


chambbj:
It's not obvious to me that this would give the same answer, and of course I never committed a unit test for the original filter. I'll have to take a closer look.

abellgithub:
Test fails.  Now I have to figure out why :)

abellgithub:
There might be a test issue based on order.

I also thought of a difference in behavior.  It's possible to have points outside a voxel that are closer to the center of the voxel than all the points IN the voxel.  In the old code, we'd pick the point outside of the voxel over the one in the voxel.  In the new code, the intent was to choose the closest point inside the voxel itself.  I didn't consider that case when I rewrote things, though I'm not sure that one way is better or worse than the other.

abellgithub:
So, both things were happening.  I had tuples stored C/R/D and the old code was R/C/D, so comparisons were bad.  Fixed that and found the case that I described above.

I'd appreciate thoughts on the behavior difference.  It was obviously unintuitive to me that a voxel's value would come from a point outside of that voxel, especially when we create voxels based upon the fact that they contain points, rather than they have a point within some distance of influence.  If speed is any consideration, this PR is much, much faster.

chambbj:
I agree. A point outside the voxel sounds like a bug for sure. If you've
posted the tests (I'm AFK), I can take another look. I certainly like the
speed improvement!

PR #2133, opened 2018-08-14
Configure targets to link libxml2 only if found

mloskot:
Although according to the docs libxml2 is an optional dependency, main target unconditionally uses `LIBXML2_INCLUDE_DIR` and `LIBXML2_LIBRARIES` which are flagged as `-NOTFOUND`. That is not valid value for `target_*`   commands.
This makes libxml2 a required dependency leading to CMake configuration errors, if not found.
Remove Ilvis2Reader.cpp from sources if libxml2 if not found.

-----

CMake will fail if libxml2 is not found:

```
1> -- Could NOT find LibXml2 (missing: LIBXML2_LIBRARY LIBXML2_INCLUDE_DIR) 
...
1> CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
1> -- Configuring incomplete, errors occurred!
1> See also "D:/PDAL/_build/x64-Debug/CMakeFiles/CMakeOutput.log".
1> See also "D:/PDAL/_build/x64-Debug/CMakeFiles/CMakeError.log".
1> Please set them or make sure they are set and tested correctly in the CMake files:
1> LIBXML2_LIBRARY (ADVANCED)
1>     linked by target "pdalcpp" in directory D:/PDAL
```

-----

Possibly, there may be more similar issues for other optional dependencies. I haven't checked them all.
This is just a very minimal patch that indicates the problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,6 +216,7 @@ if (NOT PDAL_HAVE_LIBXML2)
     file(GLOB XML_SRCS
         io/Ilvis2MetadataReader.cpp
         io/Ilvis2Metadata.cpp
+        io/Ilvis2Reader.cpp
         ${PDAL_SRC_DIR}/DbWriter.cpp
         ${PDAL_SRC_DIR}/DbReader.cpp
         ${PDAL_SRC_DIR}/XMLSchema.cpp)
--- a/cmake/libxml2.cmake
+++ b/cmake/libxml2.cmake
@@ -12,6 +12,11 @@ mark_as_advanced(CLEAR LIBXML2_INCLUDE_DIR)
 mark_as_advanced(CLEAR LIBXML2_LIBRARIES)
 if (LIBXML2_FOUND)
     set(PDAL_HAVE_LIBXML2 1)
+else()
+    unset(LIBXML2_INCLUDE_DIR CACHE)
+    unset(LIBXML2_LIBRARY CACHE)
+    unset(LIBXML2_XMLLINT_EXECUTABLE CACHE)
+    unset(LIBXML2_LIBRARIES) # Find-module output variable, not cache variable
 endif()
 
 set_property(GLOBAL PROPERTY _LIBXML2_INCLUDED TRUE)


abellgithub:
This used to work fine.  Perhaps it's a cmake change that's causing an error?

If LIBXML2 is not found, I believe that LIBXML2_INCLUDE_DIR and LIBXML2_LIBRARIES are empty, so should cause no harm.  I have done this a bunch of places, so if cmake is going to whine about it, I'll have to rework things.

mloskot:
Yes, CMake is getting stricter, apparently.

> If LIBXML2 is not found, I believe that LIBXML2_INCLUDE_DIR and LIBXML2_LIBRARIES are empty, so should cause no harm.

They are not empty, but they are `-NOTFOUND`.
Using such varaible in `set()` command is fine, but that is different for `target_*` and other commands.

abellgithub:
Whoever thought to put a literal "NOTFOUND" into variables with special meaning should be given a talking to.

mloskot:
The `NOTFOUND` flag on variables have been there since CMake 2.6 or even earlier.

abellgithub:
I know it's not new.  It's still awful.

mloskot:
CMake is not all milk, honey and beauty, but that is irrelevant to this issue, isn't it.

abellgithub:
This used to work and has been tested.  Breaking existing behavior is bad.  If we weren't so invested...

mloskot:
> This used to work and has been tested

That what is puzzling me - how possibly AppVeyor is green. Another issue I've just been hitting is:

```
d:\pdal\vendor\gtest\include\gtest\gtest-printers.h(497): error C2220: warning treated as error - no 'object' file generated 
d:\pdal\vendor\gtest\include\gtest\gtest-printers.h(497): warning C4996: 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED. You can define _SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING to acknowledge that you have received this warning. 
```

Apparently, PDAL forces `/W4 /WX` for all, including the third-party libraries.
Is that expected?

abellgithub:
I don't like this PR because it doesn't hide the issue.  If there is no way to disable the warning/error message, I would suggest un-setting the variables in libxml2.cmake if the package isn't found, which should restore the behavior from earlier cmake versions.

mloskot:
Sooner or later, you will need to switch to CMake 3.x and accept the changed behaviours. It is quite impractical to support CMake 2.8 and 3.x lines.

mloskot:
@abellgithub Although I don't like this kind of plumbing, I understand your reasoning and applied the fix as you suggested.

@hobu I'd love to help, but currently I'm balls-deep with new CMake for GEOS, struggling with some CMake for GDAL, and have not yet managed to get into PROJ.4 tests and CMake update. So, no promises I'm afraid.

abellgithub:
The find module ends up setting LIBXML2_LIBRARY rather than LIBXML2_LIBRARIES.  I understand that this isn't the canonical method, but without changing the find module...  See the error in the first comment.

mloskot:
Thanks for catching my sloppiness, sorry. It should be fixed now.

> The find module ends up setting LIBXML2_LIBRARY rather than LIBXML2_LIBRARIES.

That is correct, as per https://github.com/Kitware/CMake/blob/master/Modules/FindLibXml2.cmake#L92
In CMake,  `FOO_LIBRARY` and `FOO_INCLUDE_DIR` are Find-module **cache** variables, while `FOO_LIBRARIES` is just output/result non-cache variable. (Though, variables are deprecated anyway and exported targets are now preferred.)
Typically, the FOO package is considered to be found if both `FOO_LIBRARY` and `FOO_INCLUDE_DIR` are valid.

mloskot:
I should have written exported/imported targets.

The latest version of the module does define `LibXml2::LibXml2` target, https://github.com/Kitware/CMake/blob/master/Modules/FindLibXml2.cmake#L10-L14 so, PDAL can do:

```
target_link_libraries(${PDAL_BASE_LIB_NAME} PRIVATE LibXml2::LibXml2)
```

That will set all the LibXml2 usage requirements (eg. include directories too).

https://gitlab.kitware.com/cmake/community/wikis/doc/tutorials/Exporting-and-Importing-Targets and in the CMake manual for `EXPORT` and `IMPORTED` and `target_*` commands.

hobu:
CMake 3.5+ required as of #2137

abellgithub:
Not sure why that matters.

hobu:
> Not sure why that matters.
I thought we don't like warnings.

abellgithub:
I guess I'm lost on this.

mloskot:
So, what's the final resolution
 - Shall I close as dropped?
 - Shall I apply any additional changes?

abellgithub:
It's fine by me, but as I said, there are probably a bunch of other places where this is an issue.

mloskot:
I fixed actually experienced issue. I may along fix issues obvious to spot. I don't aim to fix any  presumably existing issues.

Why holding back fix, even if turns partial in big picture, for confirmed issue.

abellgithub:
@hobu: Was there something you had an interest in re cmake versions?

mloskot:
Thanks!

PR #2135, opened 2018-08-14
Enable Unicode support for paths on Windows

mloskot:
This is a minimal changeset that enables PDAL to recognise filesystem locations that require Unicode encoded paths (e.g. Japanese paths).

On Windows, every path in ASCII or UTF-8 is converted to **UTF-16** and passed to Win32 API functions or C++ file streams (_Microsoft extension_) which expect paths encoded in `wchar_t`-character string.

## Tasks

- [ ] Test if compilation on Linux has not broke

-----

Successfully tested using pipelines with paths like, for example:

```
{
  "pipeline": [
    {
      "type": "readers.las",
      "filename": "f:\\平仮名ひらがなラドクリフ\\trees.las"
    },
    {
      "type": "writers.text",
      "filename": "f:\\平仮名ひらがなラドクリフ\\trees.txt"
    }
  ]
}
```


hobu:
I've wanted to do this for a while. Thanks.

hobu:
Can you add a test to the suite?

mloskot:
I've added basic test case to `FileUtilsTest.cpp`. If that works for Travis, I'll add a test with `pipeline.json` that uses Unicode paths and eg. translates .las to .txt.

Issue #2024, opened 2018-05-17
issues with ~/paths on a mac


calvinmetcalf:
I'm at @hobu 's talk on pdal and on the [clipping exersize](https://pdal.io/workshop/exercises/analysis/clipping/clipping.html#workshop-clipping) doing `~/projects/pdal/exercises/analysis/clipping/autzen.laz` gets a 

```
 Unable to open stream for '~/projects/pdal/exercises/analysis/clipping/autzen.laz' with error 'No such file or directory'
```

but replacing the ~ with `/Users/me` in the vrt and the json and it works

abellgithub:
I don't understand.  There is no file reference in the document you linked with such a path.

hobu:
> I don't understand.

Confirmed.

A pipeline with 
```
    {
      "type": "readers.text",
      "skip": 2,
      "filename": "/Users/hobu/dev/git/pdal/test/data/text/utm17_1.txt"
    },
```

and one with
```
    {
      "type": "readers.text",
      "filename": "~/dev/git/pdal/test/data/text/utm17_1.txt"
    },
``` 

does not.

abellgithub:
This has never been supported.  Tilde expansion is a function of the shell.

abellgithub:
Perhaps interesting, g++ on OSX doesn't do tilde expansion:

master $ g++ -std=c++11 F.cpp -I~/pdalinstall/include -L~/pdalinstall/lib -lpdalcpp -dynamiclib -o libpdal_plugin_filter_f.dylib
F.cpp:1:10: fatal error: 'pdal/Filter.hpp' file not found
#include <pdal/Filter.hpp>
         ^~~~~~~~~~~~~~~~~
master $ g++ -std=c++11 F.cpp -I ~/pdalinstall/include -L ~/pdalinstall/lib -lpdalcpp -dynamiclib -o libpdal_plugin_filter_f.dylib

hobu:
Let's address this issue by detecting the following conditions

* We're on windows
* We detect a `~` in the given filename

When we have both of these, we should error out and say we can't open the filename before giving it to arbiter or the OS to attempt to open.

PR #2138, opened 2018-08-15
throw an error if we try to open a file with ~ in the file name because we do not do shell expansion. Throw an error for globs with ~ in the file name on windows #2024

hobu:


hobu:
complaining that it wasn't used on unix.

PR #2140, opened 2018-08-20
Get the dimension Name even its Id has not a predefined value

ahuarte47:
This fix allows correctly trace the name of dimensions with no predefined values


abellgithub:
Thanks.  Goes to show that this isn't much used!

PR #2141, opened 2018-08-20
Reset cells member to use Delaunay::set_vertices more than one time

ahuarte47:
`Delaunay::nb_cells()` returns a wrong value (it returns the last value assigned by a previous call) when you call `Delaunay::set_vertices` with vertices don't define a valid mesh and the method returns without success. 

[Example](https://github.com/PDAL/PDAL/blob/master/plugins/delaunay/filters/Delaunay_psm.cpp#L20560)

### DIFF
--- a/plugins/delaunay/filters/Delaunay_psm.cpp
+++ b/plugins/delaunay/filters/Delaunay_psm.cpp
@@ -20554,6 +20554,7 @@ namespace GEO {
     void Delaunay::set_vertices(
         index_t nb_vertices, const double* vertices
     ) {
+        nb_cells_ = 0;
         nb_vertices_ = nb_vertices;
         vertices_ = vertices;
         if(nb_vertices_ < index_t(dimension()) + 1) {


abellgithub:
This code comes from [Geogram](http://alice.loria.fr/software/geogram/doc/html/index.html).  I'd suggest requesting the change there and refreshing.

However, I'm not sure that I see how this would be an issue the way the filter is run, as `triangulation` is created new right before `set_vertices` is called in `DelaunayFilter::run()`

ahuarte47:
Hi @abellgithub, thanks for reviewing. I am writing my own Outlier Filter that creates several meshes (with this Delaunay class) grouped by the `classification` dimension. I share an unique Delaunay instance to improve the performance. I know this is a personal use case but I thought to inform you about this very hidden bug that it could raise in the future. Any way you can freely close this pull!

abellgithub:
I haven't used this code at all, and I doubt your change harms anything.  There is a lot of stuff in the Geogram code that we really should pitch because it duplicates functionality we already have.  It will take some work.  I'll merge this if it helps you out, but be aware that at some point we may grab newer versions of the Geogram code (at least in part) and this may well be lost.

ahuarte47:
Thank you very much @abellgithub !

PR #2142, opened 2018-08-20
It allows you share the Python script between Point Views

ahuarte47:
You could have a script with objects or configurations you want to use repeatedly (False by default).

### DIFF
--- a/doc/stages/filters.python.rst
+++ b/doc/stages/filters.python.rst
@@ -271,5 +271,9 @@ pdalargs
   A JSON dictionary of items you wish to pass into the modules globals as the
   ``pdalargs`` object.
 
+sharescript
+  It allows you share the Python script between Point Views. You could have a
+  script with objects or configurations you want to use repeatedly.
+
 .. _Python: http://python.org/
 .. _NumPy: http://www.numpy.org/
--- a/plugins/python/filters/PythonFilter.cpp
+++ b/plugins/python/filters/PythonFilter.cpp
@@ -53,6 +53,17 @@ CREATE_SHARED_STAGE(PythonFilter, s_info)
 
 std::string PythonFilter::getName() const { return s_info.name; }
 
+PythonFilter::~PythonFilter()
+{
+    if (m_script)
+    {
+        delete m_pythonMethod;
+        m_pythonMethod = NULL;
+        delete m_script;
+        m_script = NULL;
+    }
+}
+
 void PythonFilter::addArgs(ProgramArgs& args)
 {
     args.add("source", "Python script to run", m_source);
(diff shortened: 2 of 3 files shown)


hobu:
> I don't understand this.

It is so you can use a PDAL pipeline to wake up data and then reuse it. I wonder if we shouldn't make a more holistic solution, however. I have a branch that allows `readers.numpy` to be used from the Python extension, and @abellgithub's refactor of the PythonPointTable will make everything perform better as well. In short, there is activity on this front. I suggest we hold off on merging this until the dust settles on all of these things.

ahuarte47:
Hi @abellgithub, thanks for reviewing. I am writing a filter that creates an instance of PythonFilter to process many input files. The python script needs to create and configure an object that takes a very long time to initialize. With this new parameter I can share the same python instance to execute for these many input files.

abellgithub:
You're making one filter from another?  Perhaps, then, the infrastructure needs to be changed, because I don't understand the use case in a normal workflow.  Or maybe the filter you're constructing should call the plang functions directly?  It's hard to tell without seeing...

hobu:
@ahuarte47 I think your model might be better if you were completely in Python and could invoke PDAL on arbitrary numpy arrays. Then you can control the lifetimes of things without having to be constrained by PDAL. As I mentioned, I have a branch that starts in on that, but it is not yet complete. I am hoping to have it for the 1.8.0 release, which will include a few other Python improvements.

ahuarte47:
> You're making one filter from another?

Ok, I am going to try explain me with code:
```
pdal::Stage *stage = m_stageFactory.createStage("filters.python");
pythonOb = dynamic_cast<pdal::Filter*>(stage);

pdal::Options options;
options.add("function", "main");
options.add("module", "anything");
options.add("script", script);
options.add("pdalargs", scriptArgs);
options.add("sharescript", true);
pythonOb->setOptions(options);
...

for (int i = 0; i < 1000; ++i)
{
   table = .... from a LAS file
   pythonOb->prepare(table);
   pythonOb->execute(table);
}
...
```

ahuarte47:
> @ahuarte47 I think your model might be better if you were completely in Python and could invoke PDAL on arbitrary numpy arrays. Then you can control the lifetimes of things without having to be constrained by PDAL. As I mentioned, I have a branch that starts in on that, but it is not yet complete. I am hoping to have it for the 1.8.0 release, which will include a few other Python improvements.

ok, I agree, then I will close this! Thank you very much and congratulations for this amazing software!

hobu:
> ok, I agree, then I will close this! 

We'll leave it open and tag it 1.8 so we remember to deal with it before release. Thanks for the contribution!

ahuarte47:
Ok, thanks @hobu, I will adapt my code to this approach.

hobu:
Please report back if you have any trouble or we need to modify it to better suit your needs.

PR #2147, opened 2018-08-22
Remove pgpointcloud GHT compression

elemoine:
Remove pgpointcloud GHT compression, as the support for GHT compression has been removed in pgpointcloud v1.2.0. See https://github.com/pgpointcloud/pointcloud/pull/214.

### DIFF
--- a/doc/stages/writers.pgpointcloud.rst
+++ b/doc/stages/writers.pgpointcloud.rst
@@ -70,7 +70,6 @@ compression
 
   * **none** applies no compression
   * **dimensional** applies dynamic compression to each dimension separately
-  * **ght** applies a "geohash tree" compression by sorting the points into a prefix tree
   * **lazperf** applies a "laz" compression (using the `laz-perf`_ library in PostgreSQL Pointcloud)
 
 overwrite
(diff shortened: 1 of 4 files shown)


elemoine:
> I can't imagine it matters, but we should probably leave the values of the CompressionType the same since they were specified explicitly.

Yes, I was hesitating. And don't we think we should keep the Ght compression type in the enum class? It would be weird to have this:

```c++
enum class CompressionType
{
    None = 0,
    Dimensional = 2,
    Lazperf = 3,
    Unknown = 256
}
```

The absence of value `1` would be hard to to understand, if even possible.

elemoine:
I've added `Ght` back to the `CompressionType` enum class.

abellgithub:
I'd probably do something like:

```c++
enum class CompressionType
{
    None = 0,
//  Ght = 1,   -- Removed compression type
    Dimensional = 2,
    Lazperf = 3,
    Unknown = 256
}
```

elemoine:
Works for me too. Done.

abellgithub:
Thanks so much!

abellgithub:
Close #2148

PR #2158, opened 2018-08-29
Properly handle negative values as arguments in vectors

### LINKED ISSUE #2155: Passing negative numeric list arguments to filter
Related to discussion from #2113 

On master we are seeing a failure when passing negative values as a vector parameter to a filter.  This failure occurs from both a pipeline json and from command line arguments.

data defined as
```
      struct Args
      {
         std::vector<int> ivec;
      };
```
then
```
{
  "pipeline": [
    {
      "ivec": [-1],
      "type": "filters.vec"
    }
  ]
}
```
or
```
pdal info --driver filters.vec --filters.vec.ivec=-1 no-points.las
```
Produces error

> PDAL: filters.vec: Argument 'ivec' needs a value and none was provided.

abellgithub:
Close #2155


jw3:
Works for me :ok_hand:

PR #2160, opened 2018-09-06
Fix FixUtils::glob to substr Win32 Unicode path at correct position

mloskot:
This corrects change applied in #2135.
Add minor refinements to fromNative and toNative functions.


mloskot:
@hobu Thank you for merging

PR #2176, opened 2018-09-22
I3S and SLPK readers

kylemann16:
Adds two new readers as plugins: An I3S reader which uses an "i3s://" protocol prefix as well as an SLPK reader which uses a ".slpk" suffix.


plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,435 @@
...
+            std::string localUrl;
+
+            localUrl = m_filename + "/nodes/" + std::to_string(nodes[i]);
+
+            p.add([localUrl, this, &view]()
+            {
+                createView(localUrl, view);
+            });
+        }
+        const std::size_t pointSize(view->layout()->pointSize());
```
abellgithub: This looks like a no-op.

plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,435 @@
...
+            b += rymax.vec()[1] * (i & 2 ? 1.0 : -1.0);
+            c += rymax.vec()[2] * (i & 2 ? 1.0 : -1.0);
+
+            a += rzmax.vec()[0] * (i & 4 ? 1.0 : -1.0);
+            b += rzmax.vec()[1] * (i & 4 ? 1.0 : -1.0);
+            c += rzmax.vec()[2] * (i & 4 ? 1.0 : -1.0);
+
+            OCTTransform(m_toNativeTransform, 1, &a, &b, &c);
+
+            minx = std::min(minx, a);
```
abellgithub: This can all be replaced with a BOX3D and BOX3D.grow().

cmake/options.cmake
```diff
@@ -22,6 +22,11 @@ option(BUILD_PLUGIN_GEOWAVE
 add_feature_info("GeoWave plugin" BUILD_PLUGIN_GEOWAVE
     "Read and Write data using GeoWave")
 
+option(BUILD_PLUGIN_I3S
+    "Choose if I3S support should be built" FALSE)
+add_feature_info("I3S plugin" BUILD_PLUGIN_I3S
+    "Read from a i3s server")
```
connormanning: Description should mention SLPK support as well.

doc/stages/readers.i3s.rst
```diff
@@ -0,0 +1,43 @@
...
+readers.i3s
+===========
+
+`Indexed 3d Scene Layer (I3S)`_ is a specification created by Esri as a format for their 3D Scene Layer and scene services. The I3S reader handles RESTful webservices in an I3S file structure/format.
+
+Example
+--------------------------------------------------------------------------------
+This example will download the Autzen dataset from the arcgis scene server, reproject it to another spatial reference, and output it to a las file. This is done through PDAL's command line interface or through the pipeline.
+
+.. code-block:: json
```
connormanning: Not sure it's useful to do reprojection stuff in this example since I3S is the focus.  Also, fix JSON - missing a comma.

doc/stages/readers.slpk.rst
```diff
@@ -0,0 +1,43 @@
...
+--------------------------------------------------------------------------------
+This example will unarchive the slpk file, store it in a temp directory, and traverse it. The data will be reprojected to another spatial reference, and output to a las file. This is done through PDAL's command line interface or through the pipeline.
+
+.. code-block:: json
+    {
+        "pipeline":[
+            {
+                "type": "readers.slpk",
+                "filename": "PDAL/test/data/i3s/SMALL_AUTZEN_LAS_All.slpk"
+                "bounds": "([-123.075542,-123.06196],[44.049719,44.06278])"
```
connormanning: Same as I3S doc comment.  Can these two docs be combined or one refer to the other?  Seems error-prone to have copy-pasted docs.

kyle-config.sy
```diff
@@ -0,0 +1,23 @@
+cd ubuild
```
connormanning: Remove this file.

pdal/StageExtensions.cpp
```diff
@@ -80,7 +80,9 @@ R"PDALEXTENSIONS(
     "writers.sqlite" : "sqlite, gpkg",
     "readers.mrsid" : "sid",
     "readers.rxp" : "rxp",
-    "readers.fbx" : "fbx"
+    "readers.fbx" : "fbx",
+    "readers.i3s" : "i3s",
```
connormanning: This isn't right, is it?  I thought I3S was supported only by protocol `i3s://`, and that only SLPK was supported by extension.  Is a `.i3s` path handled properly?

plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,435 @@
...
+            }
+            else
+            {
+                const auto data = fetchBinary(
+                        attrUrl, std::to_string(key), ".bin.gz");
+
+                std::lock_guard<std::mutex> lock(m_mutex);
+
+                // TODO Correct types.
+                if(dataType == "Uint8")
```
abellgithub: TODO?

pdal/util/FileUtils.hpp
```diff
@@ -85,6 +85,14 @@ namespace FileUtils
     */
     PDAL_DLL bool createDirectory(const std::string& dirname);
 
+    /**
+      Create a set of directories.
+
+      \param dirname  Directory name.
+      \return  Whether the directory was created.
+    */
+    PDAL_DLL bool createDirectories(const std::string& dirname);
```
connormanning: Document this better.  The arg is a string, but this represents multiple directories - how is this string formatted?  To me a `std::vector<std::string>` or a `set` would be a better input.

plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,435 @@
...
+        {
+            throw pdal_error(std::string("Failed to fetch info: ") + e.what());
+        }
+
+        //create pdal Bounds
+        m_bounds = createBounds();
+        //find number of nodes per nodepage
+        m_nodeCap = m_info["store"]["index"]["nodesPerPage"].asInt();
+
+        log()->get(LogLevel::Debug) << "\n\nFileName: "
```
connormanning: Styling consistency - put the `<<` at the end of the previous line instead of at the start of lines.  Also remove the `\n` characters from these logged strings - we're already using `std::endl` so skip the extra whitespace.

plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,435 @@
...
+        layout->registerDim(Dimension::Id::Y);
+        layout->registerDim(Dimension::Id::Z);
+        for (Json::ArrayIndex i = 0; i < attributes.size(); i++)
+        {
+            dimData data;
+            std::string readName = attributes[i]["name"].asString();
+            int key = std::stoi(attributes[i]["key"].asString());
+            data.key = key;
+
+            if(attributes[i].isMember("valueType"))
```
connormanning: `if (`

plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,435 @@
...
+            {
+                std::string type = attributes[i]["valueType"].asString();
+                data.dataType = type;
+            }
+
```
connormanning: Remove whitespace.

plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,435 @@
...
+        */
+
+        //Build the node list that will tell us which nodes overlap with bounds
+        std::vector<int> nodes;
+        buildNodeList(nodes, 0);
+
+        //Create view with overlapping leaf nodes
+        //Will create a thread pool on the createview class and iterate
+        //through the node list for the nodes to be pulled.
+        std::cout << "Fetching binaries" << std::endl;
```
connormanning: Use proper log throughout.  No `cout` anywhere.

plugins/i3s/io/EsriReader.cpp
abellgithub: I think you can dispense with all this if you were to store the PDAL data type that corresponds to your data in  dimType when you create the dimension.  Then you should be able to call `setField(Dimension::Id, Dimension::Type, PointId, void *)` in a loop, like you do in `setAs()` and eliminate this case-y thing.

plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,435 @@
...
+        pxmax.w() = 0;
+        pymax.w() = 0;
+        pzmax.w() = 0;
+        pxmax.vec() = vxmax;
+        pymax.vec() = vymax;
+        pzmax.vec() = vzmax;
+
+        // Rotate all the individual vectors
+        // gives us offset for the of the new x/y/zmax/min planes
+        Eigen::Quaterniond rxmax = q * pxmax * q.inverse();
```
connormanning: These should be named better.  They aren't "max" values, they are XYZ vector of OBB-half-size in a transformed coordinate space.  Maybe `xPrime` or `xTransformed`, for example?

plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,435 @@
...
+                }
+            }
+            else
+            {
+                const auto data = fetchBinary(
+                        attrUrl, std::to_string(key), ".bin.gz");
+
+                std::lock_guard<std::mutex> lock(m_mutex);
+
+                // TODO Correct types.
```
connormanning: Address this TODO.  Presumably the solution here is the floating point types which seem ambiguous here.  Is there a list of these types in the spec somewhere?

plugins/i3s/io/I3SReader.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+        int pageSize = nodeIndexJson["nodes"].size();
+        int initialNode = nodeIndexJson["nodes"][0]["resourceId"].asInt();
+
+        for (int i = 0; i < pageSize; i++)
+        {
+            BOX3D nodeBox = parseBox(nodeIndexJson["nodes"][i]);
+            int cCount = nodeIndexJson["nodes"][i]["childCount"].asInt();
+            bool overlap = m_bounds.overlaps(nodeBox);
+            int name = nodeIndexJson["nodes"][i]["resourceId"].asInt();
```
abellgithub: Should be moved inside if()

plugins/i3s/io/EsriReader.hpp
```diff
@@ -0,0 +1,165 @@
...
+    TransformPtr m_toEcefTransform;
+    TransformPtr m_toNativeTransform;
+
+    struct dimData
+    {
+        int key;
+        std::string dataType;
+    };
+    std::map<Dimension::Id, dimData> m_dimMap;
+    std::map<std::string, Dimension::Id> m_dimensions {
```
connormanning: I don't think `m_dimensions` needs to be a member, does it?  I believe it could be a `const` in the unnamed namespace in the CPP file.

plugins/i3s/io/SlpkExtractor.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    ILeStream in(m_filename);
+
+    zheader h;
+    std::string name;
+    std::vector<char> extra;
+    int32_t magic;
+
+    in >> magic;
+    while (magic == 0x04034b50)
```
abellgithub: This is incorrect on a big-endian machine.  I will fix.

plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,435 @@
...
+* OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+* OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY
+* OF SUCH DAMAGE.
+****************************************************************************/
+
+#include "EsriReader.hpp"
+#include "../lepcc/src/include/lepcc_c_api.h"
+#include "../lepcc/src/include/lepcc_types.h"
+#include "pool.hpp"
+#include "SlpkExtractor.hpp"
```
abellgithub: Don't think you need this here.

plugins/i3s/io/I3SReader.cpp
```diff
@@ -0,0 +1,117 @@
...
+* OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+* OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY
+* OF SUCH DAMAGE.
+****************************************************************************/
+
+#include "EsriReader.hpp"
+#include "../lepcc/src/include/lepcc_c_api.h"
+#include "../lepcc/src/include/lepcc_types.h"
+#include "pool.hpp"
+#include "SlpkExtractor.hpp"
```
abellgithub: Don't think you need this here.

plugins/i3s/io/EsriReceiver.hpp
```diff
@@ -0,0 +1,102 @@
+/******************************************************************************
```
connormanning: I don't think that EsriReceiver is a clear description of the contents of this file.  EsriUtils?  EsriSupport?

plugins/i3s/io/I3SReader.cpp
```diff
@@ -0,0 +1,117 @@
...
+    //the tree and test if it overlaps with the bounds created by user.
+    //If it's a leaf node(the highest resolution) and it overlaps, add
+    //it to the list of nodes to be pulled later.
+    void I3SReader::buildNodeList(std::vector<int>& nodes, int pageIndex)
+    {
+        Json::Value nodeIndexJson;
+        std::string nodeUrl = m_filename + "/nodepages/"
+            + std::to_string(pageIndex);
+
+        // TODO Avoid doing this by using the max child node to figure out when
```
connormanning: Do the TODO.

plugins/i3s/io/SlpkReader.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+        int pageSize = nodeIndexJson["nodes"].size();
+        int initialNode = nodeIndexJson["nodes"][0]["resourceId"].asInt();
+
+        for (int i = 0; i < pageSize; i++)
+        {
+            BOX3D nodeBox = parseBox(nodeIndexJson["nodes"][i]);
+            int cCount = nodeIndexJson["nodes"][i]["childCount"].asInt();
+            bool overlap = m_bounds.overlaps(nodeBox);
+            int name = nodeIndexJson["nodes"][i]["resourceId"].asInt();
```
abellgithub: Move `name` declaration inside `if`.

plugins/i3s/io/EsriReader.hpp
```diff
@@ -0,0 +1,165 @@
...
+    std::unique_ptr<ILeStream> m_stream;
+    std::unique_ptr<arbiter::Arbiter> m_arbiter;
+
+    I3SArgs m_args;
+    Json::Value m_info;
+    std::mutex m_mutex;
+    BOX3D m_bounds;
+    int m_nodeCap;
+
+    gzip::Decompressor m_decomp;
```
abellgithub: Is this coming from arbiter?  Somewhere else?

connormanning: Yes.  Arbiter probably needs to namespace it accordingly.

connormanning:
Another style tweak throughout to maintain consistency with the rest of PDAL - don't indent due to a top-level namespace.  Many of your files have all of the content indented by one.  Do this in its own commit since the diff will be noisy.

pdal/util/FileUtils.hpp
kylemann16: I'm not sure that I changed this...

pdal/StageExtensions.cpp
kylemann16: Stagefactory portion that looks for protocols adds the .i3s extension to the file.

plugins/i3s/io/EsriReader.hpp
```diff
@@ -0,0 +1,165 @@
...
+* BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+* OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+* AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+* OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+* OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY
+* OF SUCH DAMAGE.
+****************************************************************************/
+
+#pragma once
+#define ARBITER_ZLIB
```
connormanning: Need to get this configured at the cmake level - this line should be removed.

pdal/util/FileUtils.hpp
abellgithub: This is something I did for SlpkExtractor().  It just makes all the directories necessary to complete a path.  I'll make the description better.

plugins/i3s/io/EsriReader.cpp
abellgithub: So nitpicky.

plugins/i3s/io/EsriReader.cpp
```diff
@@ -112,42 +111,83 @@ namespace pdal
...
+                    data.dimType = Dimension::Type::Signed16;
+                else if(data.dataType == "Int32")
+                    data.dimType = Dimension::Type::Signed32;
+                else if(data.dataType == "Int64")
+                    data.dimType = Dimension::Type::Signed64;
+                else if(data.dataType == "Float64")
+                    data.dimType = Dimension::Type::Double;
+                else if(data.dataType == "Float32")
+                    data.dimType = Dimension::Type::Float;
+                /*
```
abellgithub: Dead code.

plugins/i3s/io/EsriReader.cpp
```diff
@@ -112,42 +111,83 @@ namespace pdal
...
                 m_dimMap[Dimension::Id::Red] = data;
             }
+            //Available dimension types can be found at https://git.io/fAbxS
             else if (m_dimensions.find(readName) != m_dimensions.end())
             {
-                layout->registerDim(m_dimensions.at(readName));
-                m_dimMap[m_dimensions.at(readName)] = data;
-
+                data.dimType = Dimension::Type::None;
+                if(data.dataType == "UInt8")
```
abellgithub: IMO this kind of thing belongs in a function.  Perhaps just in dimType?
```
using namespace Dimension;
std::map<std::string, Dimension::Type> dimmap { {"Uint8", Type::Unsigned8}, {"Uint16", Type::Unsigned16}, ...

auto di = dimmap.find(name);
return (di != dimmap.end() ? di->second : Type::Unknown);
```
To me, this is just cleaner and doesn't clutter up your code.  You can find() in a vector, too, but in practical terms, it makes no difference.

plugins/i3s/io/EsriReader.hpp
connormanning: Arbiter needs an update for this - the code here is correct for the current arbiter, which pollutes the global namespace with a `gzip` namespace in its current iteration.  I will fix this one.

plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,438 @@
...
+                m_dimMap[m_dimensions.at(readName)] = data;
+            }
+            else
+            {
+                log()->get(LogLevel::Warning) <<
+                    "Could not find Dimension type for " <<
+                    readName << std::endl;
+            }
+        }
+        else if (attributes[i].isMember("attributeValues"))
```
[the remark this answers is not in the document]
kylemann16: There are some objects that don't have attributeValues, like elevation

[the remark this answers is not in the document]
kylemann16: And there is no binary data stored for elevation, so I don't want to add it to the map.

plugins/i3s/io/EsriReader.cpp
```diff
@@ -0,0 +1,438 @@
...
+    for (std::size_t i = 0; i < nodes.size(); i++)
+    {
+        log()->get(LogLevel::Debug) << "\r" << i << "/" << nodes.size();
+        std::string localUrl;
+
+        localUrl = m_filename + "/nodes/" + std::to_string(nodes[i]);
+
+        p.add([localUrl, this, &view]()
+        {
+            createView(localUrl, view);
```
connormanning: Not thread-safe - pass the `PointView&` by reference.

abellgithub: It's a shaued pointer. Why is this bad?

connormanning: The `shared_ptr` is passed by value (see prototype for `createView`), so each of these calls is reference-counting, which needs to be locked.

connormanning: Actually from the spec I'm not sure about this.

"If multiple threads of execution access the same shared_ptr without synchronization and any of those accesses uses a non-const member function of shared_ptr then a data race will occur".  The copy constructor is not a non-const member so maybe it's ok...

plugins/i3s/io/I3SReader.cpp
```diff
@@ -90,9 +90,17 @@ void I3SReader::buildNodeList(std::vector<int>& nodes, int pageIndex)
...
+        //density calculated as (number of points in node) / (lod threshold)
+        int pCount =  nodeIndexJson["nodes"][i]["vertexCount"].asInt();
+        double lodThreshold =
+            nodeIndexJson["nodes"][i]["lodThreshold"].asDouble();
+        double density = (double)pCount / lodThreshold;
         bool overlap = m_bounds.overlaps(nodeBox);
-        if (cCount == 0 && overlap)
+        if (density > m_args.lod && density < m_args.lod + 0.5 && overlap)
         {
+            std::cout << "lod: " << lodThreshold << std::endl;
```
hobu: * Remove debugging lint
* Complex single-line conditional with no parens
* what does this magic `0.5` number do?

plugins/i3s/io/EsriReader.hpp
```diff
@@ -109,7 +109,7 @@ class PDAL_DLL EsriReader : public Reader
       Bounds bounds;
       uint16_t threads = 8;
       std::vector<std::string> dimensions;
-      uint64_t lod;
+      double lod;
```
hobu: Add test to SLPK tests of `lod` option

plugins/i3s/io/I3SReader.cpp
kylemann16: 0.5 doesn't really mean anything. I just picked an arbitrary threshold to limit it. 0.5 seemed like a good cutoff. Could have them enter a range for the lod.

kylemann16: Otherwise you're getting a lot of overlapping resolutions if you pick greater than or less than one single number.

plugins/i3s/io/EsriReader.hpp
kylemann16: I wrote one for i3s, but the slpk file we currently have only has one node and wouldn't show any valid pruning based on the lod.

PR #2177, opened 2018-09-24
support fetching bounds information for readers.gdal QuickInfo

hobu:


pdal/GDALUtils.hpp
```diff
@@ -771,13 +771,21 @@ class PDAL_DLL Raster
...
+        BOX2D box2 = bounds();
 
         double minimum; double maximum;
         double mean; double stddev;
-        statistics(1, &minimum, &maximum, &mean, &stddev);
+        statistics(nBand, &minimum, &maximum, &mean, &stddev);
 
-        BOX3D output(minx, miny, minimum,
-                     maxx, maxy, maximum);
+        BOX3D output(box2.minx, box2.miny, minimum,
```
abellgithub: Might be nice to have the following:

    explicit BOX3D(const BOX2D& box, double minz, double maxz) :
        BOX2D(box), minz(minz), maxz(maxz)
    {}

test/unit/io/GDALReaderTest.cpp
```diff
@@ -69,9 +70,9 @@ TEST(GDALReaderTest, simple)
     PointViewSet s = gr.execute(t);
     PointViewPtr v = *s.begin();
     PointLayoutPtr l = t.layout();
-    Dimension::Id id1 = l->findDim("band-1");
-    Dimension::Id id2 = l->findDim("band-2");
-    Dimension::Id id3 = l->findDim("band-3");
+    Dimension::Id id1 = l->findDim("Intensity");
```
abellgithub: Probably good to have two tests -- one with header and one without.

io/GDALReader.cpp
```diff
@@ -112,15 +121,41 @@ void GDALReader::addDimensions(PointLayoutPtr layout)
 {
     layout->registerDim(pdal::Dimension::Id::X);
     layout->registerDim(pdal::Dimension::Id::Y);
-    for (int i = 0; i < m_raster->bandCount(); ++i)
+
+    if (m_header.size())
```
abellgithub: The cerr stuff needs to go.

Also, IMO would be cleaner if you just assigned band-X... to dimNames if it was empty.  Then you only need one loop to assign dims.

io/GDALReader.cpp
```diff
@@ -100,7 +100,16 @@ QuickInfo GDALReader::inspect()
         throwError("Couldn't open raster file '" + m_filename + "'.");
 
     qi.m_pointCount = m_raster->width() * m_raster->height();
-    qi.m_bounds = m_raster->bounds();
+
+    auto p = std::find(m_bandIds.begin(), m_bandIds.end(), Dimension::Id::Z);
+
+    int nBand(1);
```
abellgithub: Probably should document the fact that the first band is taken to be Z if there is no header specified or if Z isn't present.

test/unit/io/GDALReaderTest.cpp
hobu: The other tests continue to use the "band-#" nomenclature.

io/GDALReader.cpp
hobu: > the first band is taken to be Z

only for the QuickInfo. I wasn't sure what to do otherwise.

io/GDALReader.cpp
hobu: Done in d0f7d0646c04baf68249a2ba60dc51971f21c97b

PR #2180, opened 2018-09-25
Add auto-identify horizontal/vertical EPSG methods to SpatialReference.

connormanning:


abellgithub:
What is this for?  The tests don't seem too test-y.  Seems like the point is to provide some obnoxious WKT and generate an EPSG code from it, right?

connormanning:
Correct.

hobu:
> Seems like the point is to provide some obnoxious WKT and generate an EPSG code from it, right?

Yep. If we can extract an authority code for the SRS, it makes many things simpler for users. We're just reflecting what GDAL provides anyway.

abellgithub:
I guess my question was also about where/when this was going to get used.

hobu:
> I guess my question was also about where/when this was going to get used.

* We can try to output a single EPSG code in metadata.
* PDAL API users can do the same thing (Entwine in this case).

PR #2183, opened 2018-09-26
Allow a JSON pipeline to be an array

connormanning:
The specification for a JSON pipeline to be an object with key `pipeline` may be useful in some contexts but it's almost always just extra noise.  This PR allows a pipeline to be specified simply as an array while maintaining the current functionality as well.  For example: `[ "in.las", "out.laz" ]`.

### DIFF
--- a/doc/pipeline.rst
+++ b/doc/pipeline.rst
@@ -33,10 +33,12 @@ a pipeline provides useful advantages for more complex things:
 Introduction
 --------------------------------------------------------------------------------
 
+A PDAL processing pipeline is represented in JSON.  The structure may either:
 
-A JSON object represents a PDAL processing pipeline.  The structure is always a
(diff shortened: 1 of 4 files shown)


hobu:
I would point out that PDAL should only ever output `pipeline` object-style pipelines, not simple arrays, for situations where writePipeline is called.

abellgithub:
I have to look this over. I have no problem with it in general, but I think
we should wait until next release.

hobu:
> but I think we should wait until next release.

Given that this is allowing an additional form and not taking anything away, I don't have the concern about it going in the 1.8 release. The 1.9 release is 4-6 mo. away, and that's a bunch more time to wait for a feature that is nice-to-have and not too consequential. Release manager's call...

PR #2189, opened 2018-09-27
Report actual non-streamable stage in error.

### LINKED ISSUE #2186: Streamable error reports incorrect stage.
If an attempt is made to stream a pipeline containing a non-streamable stage, the error that is returned doesn't reference the stage that can't be streamed.  This can be confusing.

abellgithub:
Close #2186

### DIFF
--- a/pdal/Stage.hpp
+++ b/pdal/Stage.hpp
@@ -138,6 +138,15 @@ public:
     { return false; }
 
     /**
+      Determine if a pipeline with this stage as a sink is streamable.
+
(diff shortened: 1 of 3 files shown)


connormanning:
I have no problems with the code but I find the function name and the return value confusing: `s.checkStreamable()` returning a truthy value for  *non*-streamable pipelines and `null` if the check *passed*.  Perhaps `getNonStreamable()` instead?  Something like that seems more self-documenting.

abellgithub:
Agree. I'll try to come up with a better name.

PR #2190, opened 2018-09-28
add 'memorycopy' option to readers.gdal

hobu:
We sometimes need to allow user to create a GDAL MEM driver copy of the given raster for situations where the format is degenerate for whatever reason. This is kind of yucky, but otherwise the user has to manually convert the data into something more friendly to our reading strategy.

### DIFF
--- a/io/GDALReader.cpp
+++ b/io/GDALReader.cpp
@@ -64,6 +64,11 @@ GDALReader::GDALReader()
     : m_index(0)
 {}
 
+GDALReader::~GDALReader()
+{
+    m_raster.reset();
+}
+
 
 void GDALReader::initialize()
 {
@@ -71,6 +76,10 @@ void GDALReader::initialize()
     m_raster.reset(new gdal::Raster(m_filename));
 
     m_raster->open();
+
+    if (m_useMemoryCopy)
+        m_raster.reset(m_raster->memoryCopy());
+
     try
     {
         setSpatialReference(m_raster->getSpatialRef());
@@ -82,7 +91,6 @@ void GDALReader::initialize()
 
     m_count = m_raster->width() * m_raster->height();
     m_bandTypes = m_raster->getPDALDimensionTypes();
-    m_raster->close();
 
 }
 
@@ -142,16 +150,18 @@ void GDALReader::addDimensions(PointLayoutPtr layout)
 void GDALReader::addArgs(ProgramArgs& args)
 {
     args.add("header", "A comma-separated list of dimension IDs to map raster bands to dimension id", m_header);
+    args.add("memorycopy", "Load the given raster file entirely to memory", m_useMemoryCopy, false);
 }
 
 
 void GDALReader::ready(PointTableRef table)
 {
     m_index = 0;
-    if (m_raster->open() == gdal::GDALError::CantOpen)
+    if (m_raster->open() == gdal::GDALError::CantOpen && ! m_useMemoryCopy)
         throwError("Couldn't open raster file '" + m_filename + "'.");
     m_row = 0;
     m_col = 0;
+
 }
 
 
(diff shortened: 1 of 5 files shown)


abellgithub:
It's not clear what the point of this is.

hobu:
PDAL reads pixel-at-a-time. Depending on the underlying format, this can be really slow. The `memorycopy` option for `readers.gdal` is to allow the entire raster to be slurped into memory using a [GDAL MEM driver](https://www.gdal.org/frmt_mem.html) so that read performance is much better at the cost of memory.

connormanning:
I have no comments on the implementation itself or the methods used, but as to the point of the PR I have taken some timing data.  I have a directory of 6MB tiff files and a script that creates `n` GDAL readers pointing at some of these files, and records the time to `execute` them each in their own thread with a `FixedPointTable`.  Here is the behavior with the default options vs with the `memorycopy` option set - times are in milliseconds:

threads | default | memorycopy
--------|--------|--------------
1  | 2.026 | 1.417
2 | 8.781 | 1.411
3 | 49.027 | 1.461
4 | 72.900 | 1.474
5 | 93.557 | 1.791
6 | 119.224 | 2.134

abellgithub:
What does memory use look like?  This kind of argues for just scrapping the normal GDAL interface.  I think we can do things better in the GDALReader, but why bother if we just always load rasters into memory this way?

connormanning:
> if we just always load rasters into memory this way?

Not sure if I'd go that far - these are relatively small rasters (~2.5M points) so the memory footprint is reasonable.  I don't do much tiff stuff - but from my understanding people commonly use tiffs that are many orders of magnitude larger than this, in which case loading everything into memory might be prohibitive.

hobu:
> This kind of argues for just scrapping the normal GDAL interface.

We don't always load the rasters into memory, only if the user sets the `memorycopy` option. That's the point of this pull request – provide a relief valve if someone is willing to burn some memory. GDAL's normal interface is just fine if we were trying to read a remote raster resource over the internet, or read a TIFF that was too big to reasonably hold in memory. I took this hacky approach because @connormanning is probably unique (tons of threading of smallish files), and I didn't want to disrupt the interface and approach that was already working.

If we were to refactor the `readers.gdal` code to read block-at-a-time instead of pixel-at-a-time, it would go a long way toward easing the issue, but GDAL's block cache is still going to get in the way eventually. A longer term option might be for GDAL to implement a thread local block cache, and I requested that @rouault give us some options for that approach.

abellgithub:
I know we don't load the rasters, but I was wondering if we should.  If the rasters are small, we shouldn't be running into block cache issues (the block cache isn't a count thing, it's a fraction of physical memory, and can be adjusted).  So arguing that the rasters are small really doesn't explain a thing.  If they're small, then why are we running into an issue that we think is the block cache?

But something is going on.  It would be nice to understand things better before we add this "yucky" stuff.

rouault:
> PDAL reads pixel-at-a-time

Huh, do you mean GDALRasterIO( ..., x, y, 1, 1, ...) calls ? That's terrible as you hit a lot of layers within GDAL just to get a single pixel. You should consider fetching w x h pixels at a time.

abellgithub:
But the access code is the same with mem driver or tiff.  Does this speak to a specific issue with tiff driver?

rouault:
> But the access code is the same with mem driver or tiff. Does this speak to a specific issue with tiff driver?

The TIFF driver will always delegate to the block cache, and may suffer for lock contention when accessing to it. On the contrary, the MEM driver has optimizations in common cases ( windows size == buffer size ) that completely avoid the block cache as the pixel values of the whole raster are immediately accessible and caching is useless for MEM.

abellgithub:
How about we just use the mem driver if the raster size is less than 100meg or something like that?

hobu:
> How about we just use the mem driver if the raster size is less than 100meg or something like that?

I lean towards no. I think GDAL's drivers might be doing something like that underneath us anyway depending on which driver is used. Keeping this as a super-secret option for now doesn't impose the pathway on anyone unless they explicitly ask for it, and when GDAL's threading situation for this scenario improves, we can remove it without disrupting anything.

PR #2192, opened 2018-10-01
Use Utils::openFile in the BpfReader so it can read remote files.

### LINKED ISSUE #2191: BPF reader fails to read remote files
The error is also a bit confusing:
```
$ ./bin/pdal info https://entwine.io/data/red-rocks.bpf
(pdal info readers.bpf Error) Unsupported BPF version = 0.
{
  "filename": "https:\/\/entwine.io\/data\/red-rocks.bpf",
  "pdal_version": "1.7.1 (git-version: 69b73a)",
  "stats":
  {
  }
}
```

connormanning:
Fixes #2191.

### DIFF
--- a/io/BpfReader.cpp
+++ b/io/BpfReader.cpp
@@ -200,6 +200,7 @@ void BpfReader::initialize()
     if (pos > m_header.m_len)
         throwError("BPF Header length exceeded that reported by file.");
     m_stream.close();
+    Utils::closeFile(m_istreamPtr);
 }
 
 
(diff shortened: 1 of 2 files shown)


abellgithub:
I don't understand what's going on here.  What I think I see as an error is that the stream isn't closed at the end of initialize().

connormanning:
As for "what's going on here", this PR uses `Utils::openFile` to create its `istream` since that utility function handles file downloading if the file is remote.

The lack of stream closing was a quick attempt not to double-download files between `initialize` and `read`.  I've just fixed it (at the expense of a potential double-download).

Issue #2205, opened 2018-10-04
Skipped points cannot be determined during StreamPointTable::reset


connormanning:
Currently the list of skipped points during a streaming execution is local to `Streamable::execute`, so if a derived `StreamPointTable` is doing batch-processing on the points in that table during `reset`, there's no way to determine which points have been filtered out.

PR #2206, opened 2018-10-04
Move skip list functionality into StreamPointTable virtuals

connormanning:
See #2205.  Not sure if the default behavior here is desired - maybe it's preferable to make them pure virtual?


connormanning:
I have an alternate implementation [here](https://github.com/PDAL/PDAL/tree/streamable-skips-native) that I prefer to this one, but it's a bit more intrusive.

connormanning:
Closing in favor of #2224.

Issue #2209, opened 2018-10-04
Bad SRS in input files can cause crash even when overridden


abellgithub:
SpatialReference::set() will throw pdal_error if the SRS can't be understood.  This might be fine in some cases, but in others a user may explicitly override the SRS in the pipeline/command line.  The failure to translate an SRS should be re-examined in the locations where it might throw.

Issue #2195, opened 2018-10-01
Reader "spatialreference if one is not found" option


connormanning:
Related to (and depends on the resolution of) #2194 - it would be convenient if the `spatialreference` option of a reader could be set to be used as a default but not override the file's SRS if one is found.

Currently this behavior is inconsistent between reader types, so the presumption here is that #2194 makes the behavior consistent one way or the other - probably defaulting to overriding the file's SRS with the value of the `spatialreference` option.

In this case, an option like `force_spatialreference`, defaulting to `true`, could instead be set to `false` to provide a default SRS only if one is not found in a file.

PR #2210, opened 2018-10-04
Default/override SRS options for all readers

connormanning:
Solves #2195.  Adds `override_srs` and `default_srs` options to all readers, with `override_srs` being an alias to the existing `spatialreference` option.  `setSpatialReference` is now a no-op if called by a Reader whose `override_srs` option was set, so Readers don't have to do anything special.

A few readers had try/catches around their spatial-reference calls - these were removed to make the reader behavior consistent and will be handled separately in #2209.


test/unit/SpatialReferenceTest.cpp
```diff
@@ -316,6 +316,59 @@ TEST(SpatialReferenceTest, test_vertical_and_horizontal)
...
+
+    {
+        Options o;
+        o.add("filename", Support::datapath("las/test_utm17.las"));
+        LasReader r;
+        r.setOptions(o);
+
+        PointTable t;
+        r.prepare(t);
+        native = r.getSpatialReference().getWKT();
```
abellgithub: This depends to some extent on the code that you've added working as you expect.  It seems worthwhile to check this against the known SRS of the file.

test/unit/SpatialReferenceTest.cpp
```diff
@@ -316,6 +316,59 @@ TEST(SpatialReferenceTest, test_vertical_and_horizontal)
 
 }
 
+TEST(SpatialReferenceTest, readerOptions)
+{
+    const std::string utm16(SpatialReference("EPSG:26916").getWKT());
```
abellgithub: All this getWKT stuff is unnecessary.  SRSs compare just fine.

pdal/Stage.cpp
```diff
@@ -336,9 +330,17 @@ void Stage::setSpatialReference(const SpatialReference& spatialRef)
 void Stage::setSpatialReference(MetadataNode& m,
     const SpatialReference& spatialRef)
 {
+    if (!m_spatialReference.empty() && !m_overrideSrsArg.empty())
```
abellgithub: valid() works here too

test/unit/SpatialReferenceTest.cpp
connormanning: Updated this to use the QuickInfo output.

pdal/Reader.cpp
```diff
@@ -43,7 +43,26 @@ void Reader::readerAddArgs(ProgramArgs& args)
...
+    args.addSynonym("override_srs", "spatialreference");
+
+    args.add("default_srs",
+            "Spatial reference to apply to data if one cannot be inferred",
+            m_defaultSrsArg);
+}
+
+void Reader::readerInitialize(PointTableRef)
+{
+    if (!m_overrideSrsArg.empty() && !m_defaultSrsArg.empty())
```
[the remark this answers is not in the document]
connormanning: I tend to agree, initially I wanted to add `SpatialReference::exists` which was the negation of empty for these checks and then I noticed `valid()`.  But since the implementation of `valid()` isn't quite equivalent to `!empty()` I was playing it safe with the current method.  I will switch them.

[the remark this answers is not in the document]
connormanning: Would this suggested behavior still work properly if an existing but non-valid `default_srs` or `override_srs` was supplied?

PR #2213, opened 2018-10-05
Support stream mode for info kernel

### LINKED ISSUE #1894: Add streaming for info kernel
It would be nice if kernels.info would support streaming, but it's a little messy as some options currently require a PointView in order to work properly.  It's all a little complicated because sometimes we only read the file header, other times we could stream and some options require standard mode. In general, the code could be improved to make it less fiddly.

abellgithub:
Close #1894


hobu:
This branch works as expected for me. I'd like `--all` output to include `--boundary` now that `filters.hexbin` is embedded in the PDAL tree (I think this is why it was optional before).

PR #2219, opened 2018-10-08
Fix CMake dependencies for Riegl RDB Plugin

norbertwenzel:
When I tried to build PDAL with RDB plugin on a relatively minimal Ubuntu (Bionic) Docker container the build failed due to missing Eigen and JSONCPP.
In a pristine Ubuntu container I installed only the following packages
- build-essential
- cmake
- libgdal-dev
- RIEGL RDB packages
and ran the following CMake commands:
```cmake
cmake -HPDAL -Bbuild -DBUILD_PLUGIN_RDBLIB=ON; cmake --build build
```
The RDB plugin was missing Eigen and JSONCPP libraries to build.
To fix these errors I
- Added missing include directory for JSONCPP: I made sure this change still works regardless of the JSONCPP library used (system install or PDAL vendor).
- Always use the Eigen library from PDAL vendor. That's what the other plugins using Eigen do.
- (And I removed a redundant include directory for RDB.)

@hobu Since from the history of these files you seemed to have troubles building with the RDB plugin, could you please confirm I did not break your build with these changes. I only tested on different Ubuntu flavors.

### DIFF
--- a/plugins/rdb/CMakeLists.txt
+++ b/plugins/rdb/CMakeLists.txt
@@ -13,7 +13,6 @@ PDAL_ADD_PLUGIN(libname reader rdb
     ${PDAL_JSONCPP_LIB_NAME}
     rdbcpp)
 
-target_include_directories(${libname} PUBLIC ${RDB_INCLUDE_CPP_DIR})
 target_include_directories(${libname} PUBLIC ${EIGEN_INCLUDE_DIRS})
 
 
@@ -29,7 +28,6 @@ if (BUILD_RDBLIB_TESTS)
     PDAL_ADD_TEST(${RDB_TEST_NAME}
         FILES test/RdbReaderTest.cpp
         LINK_WITH ${libname})
-    target_include_directories(${RDB_TEST_NAME} PUBLIC ${RDB_INCLUDE_CPP_DIR})
     target_include_directories(${RDB_TEST_NAME} PRIVATE
         ${PROJECT_BINARY_DIR}/plugins/rdb/test
         ${PROJECT_SOURCE_DIR}/plugins/rdb/io)
--- a/plugins/rdb/CMakeLists.txt
+++ b/plugins/rdb/CMakeLists.txt
@@ -14,6 +14,7 @@ PDAL_ADD_PLUGIN(libname reader rdb
     rdbcpp)
 
 target_include_directories(${libname} PUBLIC "${PDAL_VENDOR_DIR}/eigen")
+target_include_directories(${libname} PRIVATE ${PDAL_JSONCPP_INCLUDE_DIR})
 
 
 
--- a/plugins/rdb/CMakeLists.txt
+++ b/plugins/rdb/CMakeLists.txt
@@ -13,7 +13,7 @@ PDAL_ADD_PLUGIN(libname reader rdb
     ${PDAL_JSONCPP_LIB_NAME}
     rdbcpp)
 
-target_include_directories(${libname} PUBLIC ${EIGEN_INCLUDE_DIRS})
+target_include_directories(${libname} PUBLIC "${PDAL_VENDOR_DIR}/eigen")
 
 
 


hobu:
> Always use the Eigen library from PDAL vendor. That's what the other plugins using Eigen do.

Maybe we need a `PDAL_EIGEN_INCLUDE_DIR` for too. I suppose include paths could get mucked or mixed incorrectly depending on what outside code is doing.

norbertwenzel:
> Maybe we need a `PDAL_EIGEN_INCLUDE_DIR` for too. I suppose include paths could get mucked or mixed incorrectly depending on what outside code is doing.

I thought so too. But since all other plugins I've checked seemed to use PDAL's vendor Eigen and I am not sure whether Eigen is part of some PDAL interface I figured using the very same version as PDAL might be the safest option.

But if Eigen was configurable the same way as JSONCPP users could choose their preferred Eigen version, as long as all plugins and PDAL still use the same version (of course assuming Eigen is actually part of some PDAL interface). It shouldn't be too hard to adapt `json.cmake` and create something similar for Eigen. I'd give it a try if you're ok with that.

hobu:
> I'd give it a try if you're ok with that.

Yes, that would be greatly appreciated. My note was to myself to implement that otherwise.

abellgithub:
Is there a good reason not to always link eigen that's provided with pdal?

hobu:
> Is there a good reason not to always link eigen that's provided with pdal?

If some other application is linking and using a system Eigen and PDAL build/links its own, aren't we going to have a clash?

norbertwenzel:
> Is there a good reason not to always link eigen that's provided with pdal?

The only reason I could think of would be to use a newer version if available. But then again, since all sources for PDAL are available one could just update the vendor drop manually. It's a little more cumbersome, but not really hard to do.

abellgithub:
I don't understand the removal of RDB_INCLUDE_CPP_DIR.  But I don't see where it got set in the first place :)  Is the assumption that we'll always find <reigl/rdb.hpp> in some standard location?

norbertwenzel:
> I don't understand the removal of RDB_INCLUDE_CPP_DIR. But I don't see where it got set in the first place :) Is the assumption that we'll always find <reigl/rdb.hpp> in some standard location?

@abellgithub The path is set using `target_include_directories(rdbcpp PUBLIC ${RDB_INCLUDE_CPP_DIR})` in `rdb-config.cmake` (provided by the RIEGL RDB package). So by using the `rdbcpp` target in `LINK_WITH` the include directory for RDB was already set (propagated by CMake) which made this setting redundant. We had missed this redundancy when we initially released the plugin.

abellgithub:
Not a fan of magic modern cmake.  How would one know that it was being imported?  Oh well...  I'll stick a comment in there.

norbertwenzel:
By saying `LINK_WITH` (or in plain CMake `target_link_libraries`) one already knows it is imported. If one wants to use a certain library instance the path to that specific instance can be set upfront, before calling `find_library`. Afterwards the required include/source directories are fixed and need to match the library found and setting these fixed paths is imo just unnecessary manual work and therefore a source for errors. 
Case in point: (vendor) JSONCPP, which required the include path to be set manually, although it had explicitly been used by `LINK_WITH`. (see c9f000d464333d3656e7c71e333ab28f1f7fb6b9, which is part of this PR)

I honestly fail to see the magic here, but I will revert 8e9dbdeef9a1bebdd75edb52b591e1e8327f8b28 if required.

abellgithub:
Wasn't saying that you've done a thing wrong. I'm just not fond of a
reference to, essentially, another project by name inferring libraries. And
it seems plain wrong to automatically infer header file location etc.. But
I get that this is how it works now. It's just highly dependent on
everything being done the "modern cmake" way, which I don't care for.
There's less described in the build file and more tied to cmake
rules/procedures which IMO makes things more opaque, especially when things
don't work.

Just a difference of opinion.

norbertwenzel:
I didn't understand it that way. But PDAL is not my project so if there is a question of style my personal preference does not count. Anyway, thanks for your clarification, I'll leave the PR unchanged.

Issue #2226, opened 2018-10-12
readers.rxp: No output


sedot42:
Hi, I encountered a problem using readers.rxp compiled with rivlib-2_5_7-x86_64-linux-gcc55. Despite the input file being 1.2 GB, the output .las-file is 227 bytes. I'm using the following pipeline executed via `pdal pipeline pipe.json`
```
{
    "pipeline": [
        {
            "type": "readers.rxp",
            "filename": "180418_110027.rxp",
            "sync_to_pps": "true",
            "reflectance_as_intensity": "true"
        },
        {
            "type": "writers.las",
            "filename": "outputfile.las",
            "discard_high_return_numbers": "true"
        }
    ]
}
```

Since the output file is only 227 bytes, I can post it in its entirety:
```
$ cat outputfile.las
LASFWaPDALPDAL 1.7.2 (Releas)���"{�G�z�?{�G�z�?{�G�z�?���������������������������������������������%
```
I'll be happy to provide any more information.

abellgithub:
What is the output from `pdal info` on the input file?

sedot42:
```
$ stat 180418_110027.rxp
  File: 180418_110027.rxp
  Size: 1244612692	Blocks: 2430888    IO Block: 4096   regular file
Device: 822h/2082d	Inode: 4068        Links: 2
Access: (0777/-rwxrwxrwx)  Uid: ( xxxxxx/  xxxxxx)   Gid: (  100/   users)
$ pdal info 
{
  "filename": "180418_110027.rxp",
  "pdal_version": "1.7.2 (git-version: Release)",
  "stats":
  {
    "statistic":
    [
      {
        "average": 0,
        "count": 0,
        "maximum": -1.797693135e+308,
        "minimum": 1.797693135e+308,
        "name": "X",
        "position": 0,
        "stddev": 0,
        "variance": 0
      },
      {
        "average": 0,
        "count": 0,
        "maximum": -1.797693135e+308,
        "minimum": 1.797693135e+308,
        "name": "Y",
        "position": 1,
        "stddev": 0,
        "variance": 0
      },
      [... exact same result for Z, Intensity, Amplitude, Reflectance, ReturnNumber, NumberOfReturns, GpsTime, IsPpsLocked, Deviation, BackgroundRadiation, EchoRange except for "position" ...]
    ]
  }
}

```

hobu:
[...]

PR #2228, opened 2018-10-16
Fix rxp reader

gadomski:
This fixes the broken rxp reader. The break was due to an improper usage of Riegl's scanlib API that just happened to work for a while, then broke at least in RiVLib version 2.5.7. The fix is in 6ef5582  and there are some additional housekeeping commits.

There were two issues raised in #2226:
1. Unable to read rxp data
2. Broken `rxptest`

This for sure fixes the second problem and it might fix the first. The improper API usage was causing overallocation of views in the reader that might have been exceeding OP's available memory.

### DIFF
--- a/plugins/rxp/io/RxpPointcloud.cpp
+++ b/plugins/rxp/io/RxpPointcloud.cpp
@@ -127,8 +127,9 @@ void RxpPointcloud::on_echo_transformed(echo_type echo)
     point_count_t idx = m_view->size();
     unsigned int returnNumber = 1;
     Id timeId = getTimeDimensionId(m_syncToPps);
-    for (const auto& t : targets)
+    for (scanlib::pointcloud::target_count_type i = 0; i < target_count; ++i)
     {
+        auto t = targets[i];
         m_view->setField(Id::X, idx, t.vertex[0]);
         m_view->setField(Id::Y, idx, t.vertex[1]);
         m_view->setField(Id::Z, idx, t.vertex[2]);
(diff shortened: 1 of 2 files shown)


abellgithub:
For posterity, can you explain a bit about how the API works?

gadomski:
For sure. We use `scanlib`, the C++ interface provided as part of [Riegl's RiVLib](http://www.riegl.com/index.php?id=224). You can't just "open a file and read `n` points from it" with `scanlib`, you have to use a subclass-based callback system and some API-provided bits and bobs. This is a minimal example:

```cpp
#include <iostream>
#include <riegl/scanlib.hpp>

class Pointcloud : public scanlib::pointcloud {
    public:
        Pointcloud(bool sync_to_pps) : scanlib::pointcloud(sync_to_pps) {}

    protected:
        void on_echo_transformed(scanlib::pointcloud::echo_type echo) {
            std::cout << "do a thing with an echo aka point" << std::endl;
        }
};

int main(int argc, char** argv) {
    std::string filename(argv[1]);
    bool sync_to_pps(true);
    std::shared_ptr<scanlib::basic_rconnection> connection = scanlib::basic_rconnection::create(filename);
    scanlib::decoder_rxpmarker decoder(connection);
    Pointcloud pointcloud(sync_to_pps);
    scanlib::buffer buffer;
    connection->open();
    for (decoder.get(buffer); !decoder.eoi(); decoder.get(buffer)) {
        pointcloud.dispatch(buffer.begin(), buffer.end());
    }
    connection->close();
}
```

Thought I have the `on_echo_transformed` callback that fires every time a new echo (aka point) is read, I won't know how many echos were returned for a single pulse until I've read all of the echo (rxp is a streaming format). This means that I can't set `NumberOfReturns` for a given point (for las) until I've read all of the echos.

abellgithub:
Close #2226

PR #2229, opened 2018-10-17
Add the minimal option as documented.

### LINKED ISSUE #2225: readers.rxp: Unexpected argument: "minimal"
Hi and thanks for PDAL. I want to use it to handle .rxp-files. When using the argument `minimal` as [documented](https://pdal.io/stages/readers.rxp.html), I get an error `Unexpected argument: "minimal"`. I'm using the following JSON pipeline from within Python.
```
pipeline_json =
"""{
   "pipeline":[
      {
         "type":"readers.rxp",
         "filename":"180417_090506.rxp",
         "reflectance_as_intensity":true,
         "min_reflectance":20,
         "minimal":true,
         "tag":"icp_reference"
      },
      {
         "type":"readers.rxp",
         "filename":"180417_131456.rxp",
         "reflectance_as_intensity":true,
         "min_reflectance":20,
         "minimal":true,
         "tag":"icp_moving"
      }
   ]
}"""
pipeline = pdal.Pipeline(pipeline_json)
pipeline.validate()
```

abellgithub:
Close #2225

### DIFF
--- a/plugins/rxp/io/RxpReader.cpp
+++ b/plugins/rxp/io/RxpReader.cpp
@@ -91,6 +91,8 @@ void RxpReader::addArgs(ProgramArgs& args)
     args.add("reflectance_as_intensity", "Reflectance as intensity", m_reflectanceAsIntensity, DEFAULT_REFLECTANCE_AS_INTENSITY);
     args.add("min_reflectance", "Minimum reflectance", m_minReflectance, DEFAULT_MIN_REFLECTANCE);
     args.add("max_reflectance", "Maximum reflectance", m_maxReflectance, DEFAULT_MAX_REFLECTANCE);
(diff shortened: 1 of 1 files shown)


gadomski:
Is there a filter that can remove dimensions, i.e. the opposite of `filters.ferry`? If so, we should remove the `minimal` option and instead recommend using that filter.

gadomski:
Dumb question: is streaming "enabled by default" i.e. does `readers.rxp` get streaming powers just by existing?

abellgithub:
Not a dumb question.

In order to stream, the class must inherit from Streamable and is then required to implement processOne().

abellgithub:
It would be good to be streamable, but I'd like to commit this for the 1.8 release if it works for you in order to make things consistent.  Making `readers.rxp` streamable can be another ticket.

PR #2237, opened 2018-10-19
Switch Delaunay filter to use delaunator-cpp and move to core stages

### LINKED ISSUE #1969: Delaunay should be promoted from plugin status to one of our core stages
It requires no additional third-party dependencies.

@plimkilde Thoughts? I assume it started this way because you were finding your geometry package separately and later decided to just embed it.

plimkilde:
Fixes #1969.

This should also fix the complications of the current (optional) Delaunay filter, which is based on Geogram.

Instead, the filter now uses the recent [delaunator-cpp](https://github.com/delfrrr/delaunator-cpp) library (MIT-licensed). It is a single .hpp file and should build without any weird requirements.

The primary downside of switching is likely the fact that delaunator-cpp does not use exact geometric predicates, unlike Geogram. Also, it does not implement 3D Delaunay (which we currently don't use anyway). The performance of delaunator-cpp seems competitive with Geogram and other Delaunay libraries.

Apart from the library switch, this PR is basically the same as #1985 and includes @chambbj's unit test.


hobu:
@abellgithub I would much rather this PR goes in the release than the current delaunator implementation that is currently staged for release. No one has used the delaunay code either way, so this is quite isolated. Releasing the current release candidate stuff instead of this is going to create some support headaches when people try to build and use it.

PS the windows tests need to pass before we can merge it though (not currently compiling there due to NOMINMAX macro situation on windows)

plimkilde:
I've pushed a commit that wraps all min/max calls in parentheses. Sorry if I missed something in the contributing guidelines regarding CI - the [testing page](https://pdal.io/development/testing.html) linked seems to be missing.

hobu:
You haven't missed anything in the contributing guidelines, and thanks again for the contribution. 

That page has moved https://pdal.io/project/testing.html Where did you see it linked? We should clean up the pointer.

plimkilde:
Ok, great :-) The link in question is in CONTRIBUTING.md ("Run _all_ the tests") at the root of the repo. (GitHub's interface popped up a message box when I opened the PR saying those guidelines had changed since my last contribution, which left me wondering a bit.)

abellgithub:
The test that's here doesn't really test that you get correct/reasonable output.  Perhaps we could have a test with a simple set of input points and compare the output to what would be expected?

plimkilde:
Finally got around to writing a unit test that actually checks the resulting triangulation. Also, I noticed that the filter (with the Delaunator library) wrote vertices in clockwise order, which I've now flipped.

plimkilde:
Great. It just occurred to me that the test can rule out wrong triangulations simply by adding a check for the number of triangles, so I've just made a small commit for that.

PR #2240, opened 2018-10-23
Support cast of double -> float of NaN.

### LINKED ISSUE #2239: Numeric cast of NaN from double to float is broken
Casting NaN from double to float should yeild NaN.  It currently returns an error.

abellgithub:
Close #2239

### DIFF
--- a/test/unit/PointViewTest.cpp
+++ b/test/unit/PointViewTest.cpp
@@ -343,6 +343,17 @@ TEST(PointViewTest, issue1264)
     EXPECT_THROW(v.setField(foo, 0, d), pdal_error);
 }
 
+TEST(PointViewTest, getFloatNan)
+{
+    PointTable table;
+    PointLayoutPtr layout(table.layout());
+    layout->registerDim(Dimension::Id::ScanAngleRank, Dimension::Type::Float);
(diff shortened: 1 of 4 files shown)


test/unit/PointViewTest.cpp
```diff
@@ -343,6 +343,17 @@ TEST(PointViewTest, issue1264)
...
 
+TEST(PointViewTest, getFloatNan)
+{
+    PointTable table;
+    PointLayoutPtr layout(table.layout());
+    layout->registerDim(Dimension::Id::ScanAngleRank, Dimension::Type::Float);
+    PointViewPtr view(new PointView(table));
+    const float scanAngleRank = std::numeric_limits<float>::quiet_NaN();
+    view->setField(Dimension::Id::ScanAngleRank, 0, scanAngleRank);
+    float retrieved = view->getFieldAs<float>(Dimension::Id::ScanAngleRank, 0);
```
connormanning: Is this testing that the last line doesn't throw?  If so, `EXPECT_NO_THROW` would make it more explicit.

abellgithub: An oversight.  Thanks.

PR #2247, opened 2018-10-25
Don't use a destructed value in GDALReader::inspect.

connormanning:
`inspect` currently segfaults since `m_raster` is reset at the end of `initialize` but still accessed afterward in `inspect`.

### DIFF
--- a/io/GDALReader.hpp
+++ b/io/GDALReader.hpp
@@ -35,7 +35,7 @@
 #pragma once
 
 #include <string>
-
+#include <vector>
 
 #include <pdal/Dimension.hpp>
 #include <pdal/Reader.hpp>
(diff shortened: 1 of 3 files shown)


abellgithub:
The raster needs to be closed in initialize().  We should save any information required -- most of the information is already stored.  Think we just need bounds.

io/GDALReader.hpp
```diff
@@ -77,6 +77,9 @@ class PDAL_DLL GDALReader : public Reader , public Streamable
     point_count_t m_index;
     int m_row;
     int m_col;
+
+    BOX3D m_bounds;
+    std::vector<std::string> m_dimNames;
```
abellgithub: Not terribly important, but std::vector<std::string> is StringList.

PR #2252, opened 2018-10-25
Update EptReader for EPT formatting changes before EPT release.

connormanning:
The relevant EPT changes are listed here: https://github.com/connormanning/entwine/pull/139


io/EptReader.cpp
```diff
@@ -381,64 +418,74 @@ void EptReader::readLaszip(PointView& dst, const Key& key) const
...
+        {
+            if (m_remoteLayout->dims()[i] == id)
+            {
+                return i;
+            }
+        }
+        throw pdal_error("Invalid dimIndex lookup");
+    };
+
+    using D = Dimension::Id;
```
abellgithub: Seems silly to do this mapping for every point when it never changes.

io/EptReader.cpp
```diff
@@ -241,23 +265,45 @@ void EptReader::addDimensions(PointLayoutPtr layout)
...
+
+    m_dimTypes = m_remoteLayout->dimTypes();
+    for (DimType& dt : m_dimTypes)
+    {
+        // If the data is stored as Laszip, then PDAL's LasReader will unscale
+        // XYZ for us.
+        if (m_info->dataType() == EptInfo::DataType::Laszip &&
+                (dt.m_id == D::X || dt.m_id == D::Y || dt.m_id == D::Z))
```
abellgithub: Then wouldn't it be an error to define a scale/offset for XYZ for LASzip source?  Seems some sort of a check/notice is in order.

io/EptReader.cpp
```diff
@@ -381,64 +418,74 @@ void EptReader::readLaszip(PointView& dst, const Key& key) const
...
+    const double x = pr.getFieldAs<double>(Dimension::Id::X) *
+        xforms[0].m_scale.m_val + xforms[0].m_offset.m_val;
+    const double y = pr.getFieldAs<double>(Dimension::Id::Y) *
+        xforms[1].m_scale.m_val + xforms[1].m_offset.m_val;
+    const double z = pr.getFieldAs<double>(Dimension::Id::Z) *
+        xforms[2].m_scale.m_val + xforms[2].m_offset.m_val;
+
+    const uint64_t o = pr.getFieldAs<uint64_t>(Dimension::Id::OriginId);
 
     const bool selected = !m_queryOriginId || o == *m_queryOriginId;
```
abellgithub: This is an offset into the sources, right?  You could set it to -1 rather than mess with the pointer thing, right?

io/EptReader.cpp
```diff
@@ -381,64 +418,74 @@ void EptReader::readLaszip(PointView& dst, const Key& key) const
...
     if (selected && m_queryBounds.contains(x, y, z))
     {
-        dst.setField(Dimension::Id::X, pointId, x);
-        dst.setField(Dimension::Id::Y, pointId, y);
-        dst.setField(Dimension::Id::Z, pointId, z);
-
-        char* pos(dst.getPoint(pointId));
-
-        for (const DimType& dim : m_dimTypes)
+        for (const DimType& dt : m_dimTypes)
```
abellgithub: Aren't you fetching and scaling for XYZ twice here?

io/EptReader.cpp
```diff
@@ -241,23 +265,45 @@ void EptReader::addDimensions(PointLayoutPtr layout)
...
+        // If the data is stored as Laszip, then PDAL's LasReader will unscale
+        // XYZ for us.
+        if (m_info->dataType() == EptInfo::DataType::Laszip &&
+                (dt.m_id == D::X || dt.m_id == D::Y || dt.m_id == D::Z))
+        {
+            continue;
+        }
+
+        const Json::Value dim(m_info->dim(m_remoteLayout->dimName(dt.m_id)));
+        if (dim.isNull())
```
abellgithub: There's a bunch of code in EptInfo to create m_transforms and then it's not used.  There's this stuff instead.

io/EptReader.cpp
```diff
@@ -121,32 +161,30 @@ void EptReader::initialize()
...
-    log()->get(LogLevel::Debug) << "Scale: " <<
-        scale[0] << ", " << scale[1] << ", " << scale[2] << std::endl;
-    log()->get(LogLevel::Debug) << "Offset: " <<
-        offset[0] << ", " << offset[1] << ", " << offset[2] << std::endl;
     log()->get(LogLevel::Debug) << "Threads: " << m_pool->size() << std::endl;
 }
 
 void EptReader::handleOriginQuery()
 {
     if (m_args.origin().empty()) return;
```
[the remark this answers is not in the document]
connormanning: Yes that's correct.  You're right, that's missing from the "spec" (I hesitate to call it a spec since it's just a pretty informal description) at this point.

io/EptReader.cpp
```diff
@@ -121,32 +161,30 @@ void EptReader::initialize()
...
 
+    if (m_info->sources() == 0)
```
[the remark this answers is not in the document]
connormanning: The idea behind the count for `sources` was so that it could be optional to include this information in an EPT set.  I am thinking now that this key should be removed and that a valid EPT set must have a `ept-sources/list.json` which may be an empty array to omit the source information.

io/EptReader.cpp
connormanning: The scale/offset are still specified if laszip is used as EPT storage, so this would not be an error.  Having laszip storage without scaled XYZ is invalid - that should be added to the spec but is already implicitly enforced via the nature of laszip.

io/EptReader.cpp
connormanning: Yes.

io/EptReader.cpp
connormanning: Yes, it seemed simpler logically to keep the same flow for all dimensions after this check but you're correct.  I will change it to use the pre-calculated XYZ values.

io/EptReader.cpp
abellgithub: Maybe I wasn't clear.  Couldn't you have scale/offset IN the laszip file and have that be different from what's specified in the schema?  That seems problematic unless you enforce it somehow.

io/EptReader.cpp
```diff
@@ -180,33 +218,19 @@ void EptReader::handleOriginQuery()
...
-        q.miny -= std::fabs(scale[1]);
-        q.minz -= std::fabs(scale[2]);
-        q.maxx += std::fabs(scale[0]);
-        q.maxy += std::fabs(scale[1]);
-        q.maxz += std::fabs(scale[2]);
-    }
-
     // Clip the bounds to the queried origin bounds.  Don't just overwrite it -
     // it's possible that both a bounds and an origin are specified.
     m_queryBounds.clip(q);
```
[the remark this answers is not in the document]
connormanning: - Assume arg `origin` selects a file with bounds `[0, 0, 20, 20]`.
- Assume arg `bounds` is `[0, 0, 10, 10]`.

Rather than resetting the query bounds to `[0, 0, 20, 20]`, instead we *clip* the existing query bounds to the intersection of the two bounds.  The thing I'm "not overwriting" is `m_queryBounds`, meaning I'm *not* doing `m_queryBounds = q`.

io/EptReader.cpp
```diff
@@ -159,11 +197,11 @@ void EptReader::handleOriginQuery()
...
         // by a basename or a tile ID rather than a full path is convenient).
-        // Find it within the files list, and make sure it's specified uniquely
-        // enough to select only one filt.
-        for (Json::ArrayIndex i(0); i < files.size(); ++i)
+        // Find it within the sources list, and make sure it's specified
+        // uniquely enough to select only one filt.
+        for (Json::ArrayIndex i(0); i < sources.size(); ++i)
         {
-            const Json::Value& entry(files[i]);
+            const Json::Value& entry(sources[i]);
```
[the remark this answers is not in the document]
connormanning: The throw is after the loop, here: https://github.com/PDAL/PDAL/blob/20043d2a6ce55a6c279de0b20350168ead388ee9/io/EptReader.cpp#L218

io/EptReader.cpp
connormanning: Fixed in 5a752aadedd66938174b734acde7c1a1c884087b.

io/EptReader.cpp
connormanning: Fixed in 5a752aadedd66938174b734acde7c1a1c884087b.

io/EptReader.cpp
connormanning: Fixed in 5a752aadedd66938174b734acde7c1a1c884087b.

connormanning:
@abellgithub I've addressed these findings, please take another look.  Regarding above post:

- An `Endpoint` doesn't represent an open handle, just a couple of strings representing a driver/path combo (e.g. `s3://` and a bucket or just a filesystem path string).
- EPT itself does have versioning info in the format (`version` key of format `major.minor.patch`).
- A validation program would be nice, this will be something to consider making prior to EPT 1.0's release.  There isn't one yet.

Issue #2305, opened 2018-12-12
filters.crop doesn't properly filter outside bounds when not in stream mode.


dbrookes96:
Found this when integrating PDAL C++ into a project. When using the CropFilter with bounds and the "outside" option to true, the filter will still only keep the inside of the bounds. This behavior only happens when not using streaming. 
The following pipeline can be used to observe the the results with the test data las\interesting.las. The expected result is 39 points inside and 1026 points outside.

```
{
	"pipeline":
	[
		{
		"type": "readers.las",
		"filename": "interesting.las"
		},
		{
		"type": "filters.crop",
		"bounds": "([638000, 640000], [849000, 849500])",
		"outside": true
		},
		{
		"type": "writers.las",
		"filename": "interesting_out.las"
		}
	]
}
```
Having a look in CropFilter.cpp the m_args->m_cropOutside bool which flips the logic of the crop functions is applied twice for ```void CropFilter::crop(const BOX2D& box, PointView& input, PointView& output)``` as it calls ```bool CropFilter::crop(const PointRef& point, const BOX2D& box)``` which already flips the logic. i.e. The logic is flipped twice and therefore when outside is true, the inside is still kept. This doesn't happen in streaming mode as streaming mode only uses ```bool CropFilter::crop(const PointRef& point, const BOX2D& box)```. Seems like a small fix. PR incoming with a new test case also.

PR #2306, opened 2018-12-12
Fix filter.crop outside logic being flipped twice when not in stream mode.

dbrookes96:
Fixes issue #2305.


abellgithub:
Nicely done. Thanks.

hobu:
windows test failure is due to conda screwing up, not a real failure. merge at will.

PR #2314, opened 2018-12-14
Fix some errors in the pipeline doc.

abellgithub:


doc/pipeline.rst
```diff
@@ -354,25 +349,23 @@ then creates the DTM using the :ref:`writers.gdal`.
       "pipeline":[
           "autzen-full.las",
           {
-              "type":"ground",
-              "approximate":true,
+              "type":"filters.pmf",
```
hobu: we should change this whole block to `filters.smrf`.  We shouldn't encourage `filters.pmf` at all

doc/pipeline.rst
```diff
@@ -161,7 +156,7 @@ reduces memory requirements.
 
 When using :ref:`pdal translate<translate_command>` or
 :ref:`pdal pipeline<pipeline_command>`
-PDAL uses stream mode if possible.  If stream mode can't use used
+PDAL uses stream mode if possible.  If stream mode can't be used
```
hobu: We should enhance this section to note that one can find out if a stage is streamable or not by looking for the "Streamable Stage" header in any stage document.

abellgithub:
Didn't know we were trying to phase out pmf.  I just picked it because its arguments were the closest to the former ground filter in the example.

Will fix.

hobu:
Not trying to phase it out, but smrf definitely puts the best foot forward for most scenarios

Issue #1508, opened 2017-03-02
readers.pts doesn't deal with unexpected characters in field names


hobu:
PTS uses `readers.text`, and users can name the dimensions whatever they want. We need to strip out and normalize characters we can't consume in field names like `=`'s and `.`'s.

![dimngacaddhlmjld](https://cloud.githubusercontent.com/assets/87751/23491906/e82f10f2-fec6-11e6-9dbb-94520461ef7d.png)

![delomnjaifoimiij](https://cloud.githubusercontent.com/assets/87751/23491902/dc6ddc80-fec6-11e6-92ff-d7ea8d16cfc0.png)

Poux-Florent:
Here are the first 8 lines (second line is number of points) : 

//X;Y;Z;Rf;Gf;Bf;Intensity;curvature;Illuminance (PCV);Volume density (r=0.01);Roughness(0.01);Nx;Ny;Nz
832641
1.48558950;2.78713155;1.71654594;1;0.976471;0.988235;0.514534;0.000000;0.889831;86421136.000000;0.000164;0.842968;-0.279958;-0.459379
0.66884160;2.91032434;1.71380365;0.996078;0.996078;0.996078;0.419318;0.149188;0.851695;41061980.000000;0.001045;-0.557930;-0.818396;-0.137633
1.60763741;2.74084330;1.70552599;1;0.996078;1;0.307012;0.000000;0.614407;56579588.000000;0.000358;0.410376;-0.815962;-0.407183
0.82228744;2.85927010;1.71652365;0.870588;0.870588;0.878431;0.632944;0.000000;0.983051;51088740.000000;0.000713;0.094079;0.464780;-0.880414
1.69874430;2.57411289;1.70909595;1;1;1;0.524788;0.000000;0.745763;47746488.000000;0.000465;-0.370771;-0.556973;0.743175
1.14563203;2.78293252;1.71632600;1;1;0.980392;0.452033;0.000000;0.889831;42016908.000000;0.001126;-0.219483;-0.975079;0.032383

abellgithub:
[...]

PR #2317, opened 2018-12-18
Improvements to reader.pts

dbrookes96:
Peek the first line of the input file to determine what dimensions to add. Allows for reading PTS files which don't have Intensity, R, G or B such as the bunnyData.pts point cloud from http://www.libe57.org/data.html.

Also, limit the number of points read by the expected point count on the first line of the PTS file.

Added a unit test and bunny_8.pts test data to check the above two things. Note that bunny_8.pts actually contains 9 points but has 8 as the expected point count.

---

I'm not sure whether the original Leica specifications for the .pts format specifies if its valid for the files to have only X, Y, Z dimensions. Actually I'm not sure if the spec even exists, I wasn't able to find anything on it other than dead web links. Maybe these changes are against the file specs and aren't recommended?


abellgithub:
See #1508.  I'm not sure what a PTS file actually should look like and it doesn't sound like you had any better luck finding this out.  Where did your test file come from?

io/PtsReader.cpp
```diff
@@ -63,27 +63,34 @@ void PtsReader::initialize(PointTableRef table)
...
     m_dims.push_back(Dimension::Id::Y);
     m_dims.push_back(Dimension::Id::Z);
-    m_dims.push_back(Dimension::Id::Intensity);
-    m_dims.push_back(Dimension::Id::Red);
-    m_dims.push_back(Dimension::Id::Green);
-    m_dims.push_back(Dimension::Id::Blue);
-
-    for (auto d: m_dims)
+    std::getline(*m_istream, buf);
+    StringList fields = Utils::split2(buf, m_separator);
```
abellgithub: My concern here is that the second line doesn't always contain the fields.  If you take a look at the (supposed) PTS file in #1508, the field line looks like a comment in the first line.

I guess knowing more about why this came up (your file source), would be valuable.

abellgithub:
You can read this data with `readers.text`.  The only value that the PTS format seems to provide is a pre-defined set of dimensions, and we're not sure what those are, other than by counting and hoping?  I guess I don't care very much -- if people want to use this over `readers.text`, that's fine, but I'm not sure we should encourage it.

abellgithub:
Close #2324

Issue #2324, opened 2018-12-19
Support 7 dimension PTS files


abellgithub:
Some PTS files have seven dimensions.  Add support for those files.

PR #2326, opened 2018-12-20
Improvements to reader.pts again.

dbrookes96:
Allow for X Y Z Intensity formatted .pts files.
Check if the expected point count at the top of the file was read correctly.
Emit a warning when points read was less than the expected.

Sorry for another pull request on this. Its a continuation of PR 2317 with just things I wanted to amend to the previous PR. This now means that readers.pts expects data in the formats:
```
n
X Y Z
...
```
```
n
X Y Z Intensity
...
```
```
n
X Y Z Intensity R G B
...
```
These comply with all of the Leica Cyclone .pts files that I have seen so far and with what LAStools expects. I agree that readers.text usage is likely better, especially for the CloudCompare output which violates these to add whichever dimensions it likes to the file. Maybe the reader.pts doc should be updated to include the above format information and to recommend readers.text for files that don't fit the above format.


abellgithub:
Related to #2324

PR #2331, opened 2019-01-04
EptReader: add resolution limit

connormanning:
For example with pipeline file  `nyc.json`:
```
[{
    "type": "readers.ept",
    "filename": "http://na.entwine.io/nyc",
    "threads": 12,
    "resolution": 100
}, {
    "type": "filters.hexbin",
    "edge_size": 250,
    "threshold": 1
}]
```
Then to generate a coarse resolution hex boundary:
```
pdal pipeline nyc.json --metadata STDOUT
```


io/EptReader.cpp
```diff
@@ -149,21 +150,44 @@ void EptReader::initialize()
...
 
     // Figure out our query parameters.
     m_queryBounds = m_args.bounds();
     handleOriginQuery();
 
-    log()->get(LogLevel::Debug) << "Query bounds: " << m_queryBounds <<
-        std::endl;
-    log()->get(LogLevel::Debug) << "Threads: " << m_pool->size() << std::endl;
+    // Figure out our max depth.
+    const double querySpacing(m_args.spacing());
```
abellgithub: This is a little confusing in that spacing gets turned into querySpacing and you end up with a new meaning for spacing.

doc/stages/readers.ept.rst
```diff
@@ -39,6 +39,9 @@ filename
 bounds
     The extents of the resource to select in 2 or 3 dimensions, expressed as a string, e.g.: ``([xmin, xmax], [ymin, ymax], [zmin, zmax])``.  If omitted, the entire dataset will be selected.
 
+spacing
```
abellgithub: To me, this hard to figure out for the user.  You have to know how things are arranged internally.  I'm not even sure how your example works (how do you get from 10 to .001 pts. per meter?).  Can you do something like limit the point count for the query area and let the user do the math if they want to, say, get X points/meter?  It's going to be up to the user anyway since they know the SRS, I assume.

io/private/EptSupport.hpp
```diff
@@ -154,6 +156,7 @@ class PDAL_DLL EptInfo
     const Json::Value m_info;
     BOX3D m_bounds;
     uint64_t m_points = 0;
+    uint64_t m_span = 0;
```
abellgithub: The other items seem obvious.  This doesn't.  What is it, exactly?

connormanning: Span is the voxel span in each direction for the octree nodes.  A dataset whose octree nodes are 256x256x256 would have a span of `256`.  See https://github.com/connormanning/entwine/blob/master/doc/source/entwine-point-tile.md#span.

abellgithub: Maybe just a comment that says something like "length, width and depth of each octree node in voxels"  Or a link to the doc?

doc/stages/readers.ept.rst
connormanning: Initially I had implemented this as a `resolution2d` and `resolution3d` option, but this seemed even more confusing.  To me, spacing is simple: in my example, I want to select increasing resolutions until my points are 10m apart, and then stop.  That corresponds to `1/10^3 = .001` points/m^3 or `1/10^2 = .01` points/m^2 in terms of resolution.

Spacing corresponds pretty much 1:1 to something like `scale` values for LAS, which to me is why it seems simpler than an arg corresponding `resolution`.  A spacing of `1`, with units of meters, is simply a 1m grid, for example - trivial.  My description could be better but I think it's the most intuitive way to present such an option.

io/private/EptSupport.hpp
connormanning: Done at 604d1006e250dd183a76a16eb6c6ed98041f5077

io/EptReader.cpp
connormanning: Renamed `spacing` to `currentSpacing` in 295c5cede4102b58a3c5b2e865ad3ed2b72ece50 to avoid overloading the term `spacing` with the corresponding stage option.

doc/stages/readers.ept.rst
abellgithub: I still find this confusing.  Specifically, I shouldn't have to know anything about the internal storage of EPT in order to get a reasonable query result.  Mentioning octrees isn't really helpful to most people.  You've been doing octrees for years now, so it's second nature, but I don't think it's something to expect others to know or care about.

I guess I like the resolution notion better.  Call it scale and for all practical purposes it maps to LAS directly, I think.  `.1` gives you a grid of decimeters, `.01` centimeters.

connormanning: Removed mention of the octree, and clarified the meaning of the spacing parameter more clearly, in 5fccc4d040629992792d8a5ad5305e14c1262f25.

connormanning:
Added in e7b103a.

abellgithub:
Thanks!

PR #2333, opened 2019-01-04
Update arbiter bundle

connormanning:
There were some changes to PDAL's bundled arbiter that weren't reflected upstream.  I've ported those changes upstream (where there have been another few minor fixes) and re-bundled.  This is mostly no-op but will ensure that the changes to PDAL's bundled version won't get blown away by subsequent bundle updates.


abellgithub:
Exposing some of the classes in the DLL interface breaks the Windows build.

PR #2338, opened 2019-01-08
Implement `Streamable` for `RxpReader`

gadomski:
As discussed in #2229.

Since there isn't CI support for this plugin, I'll note here that tests pass on my personal Ubuntu box.

### DIFF
--- a/plugins/rxp/io/RxpPointcloud.cpp
+++ b/plugins/rxp/io/RxpPointcloud.cpp
@@ -52,19 +52,22 @@ Dimension::Id getTimeDimensionId(bool syncToPps)
 }
 
 
+Point::Point(scanlib::target target, unsigned int returnNumber, unsigned int numberOfReturns)
+    : target(target)
+    , returnNumber(returnNumber)
+    , numberOfReturns(numberOfReturns)
+{}
+
+
 RxpPointcloud::RxpPointcloud(
         const std::string& uri,
         bool syncToPps,
-        bool minimal,
         bool reflectanceAsIntensity,
         float minReflectance,
         float maxReflectance,
         PointTableRef table)
     : scanlib::pointcloud(syncToPps)
-    , m_view(new PointView(table))
-    , m_idx(0)
     , m_syncToPps(syncToPps)
-    , m_minimal(minimal)
     , m_reflectanceAsIntensity(reflectanceAsIntensity)
     , m_minReflectance(minReflectance)
     , m_maxReflectance(maxReflectance)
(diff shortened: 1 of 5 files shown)


plugins/rxp/io/RxpPointcloud.cpp
```diff
@@ -79,39 +83,74 @@ RxpPointcloud::~RxpPointcloud()
 }
 
 
-point_count_t RxpPointcloud::read(PointViewPtr view, point_count_t count)
+bool RxpPointcloud::readOne(PointRef& point)
 {
-    point_count_t numRead = 0;
+    if (this->readSavedPoint(point)) {
```
abellgithub: Would think `this` would be superfluous in this context.

plugins/rxp/io/RxpPointcloud.cpp
```diff
@@ -52,24 +52,28 @@ Dimension::Id getTimeDimensionId(bool syncToPps)
...
-    , m_view(new PointView(table))
-    , m_idx(0)
     , m_syncToPps(syncToPps)
-    , m_minimal(minimal)
     , m_reflectanceAsIntensity(reflectanceAsIntensity)
     , m_minReflectance(minReflectance)
     , m_maxReflectance(maxReflectance)
     , m_rc(scanlib::basic_rconnection::create(uri))
     , m_dec(m_rc)
+    , m_points()
```
abellgithub: Unnecessary.

plugins/rxp/io/RxpPointcloud.cpp
gadomski: It's not technically necessary, but I like to use `this` to disambiguate the call from a free function call. Happy to remove if you'd like -- if so, maybe worth putting into the style guide as a recommended practice ("don't use `this` for method calls")?

plugins/rxp/io/RxpPointcloud.cpp
gadomski: Habit carry-over from "you must initialize everything" languages. Same as above comment, happy to remove if you'd like, but recommend putting something in the style guide about it to help prevent future churn?

plugins/rxp/io/RxpPointcloud.cpp
abellgithub: The style guide:

"Make your code look like existing code."

Of course, people are free to interpret this as they wish.

gadomski: Code updated via force-push per both comments.

PR #2339, opened 2019-01-08
Update arbiter

connormanning:
Update arbiter bundle, which incorporates changes made to the previous PDAL bundled version directly, as well as some various other minor fixes.

### DIFF
--- a/vendor/arbiter/arbiter.hpp
+++ b/vendor/arbiter/arbiter.hpp
@@ -1,7 +1,7 @@
 /// Arbiter amalgamated header (https://github.com/connormanning/arbiter).
 /// It is intended to be used with #include "arbiter.hpp"
 
-// Git SHA: 7a497802737cfc956a8802b2efe65f6a72c29463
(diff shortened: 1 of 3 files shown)


vendor/arbiter/arbiter.hpp
```diff
@@ -3149,6 +3150,11 @@ inline std::string decompress(const char* data, std::size_t size)
 #endif
 #endif
 
+#ifdef _WIN32
```
abellgithub: I don't see the templated class, but I believe you.

This is being suppressed in a header file, so the warning should be pushed and then popped at the end of the file.

connormanning: Any variable of STL class type (of which I have many used as private member variables) throughout DLL-exported classes produces this warning.  Which seems a bit silly, as the warning is that these members will be inaccessible since they're not DLL-exported, but they're private anyways.

The offending members aren't in this file itself.  I'm following the PDAL model by shutting them off project-wide: https://github.com/PDAL/PDAL/blob/master/pdal/pdal_export.hpp#L56

PR #2345, opened 2019-01-10
Defaulted CMake config LIB_INSTALL_DIR to PDAL_LIB_INSTALL_DIR

psalvaggio:
I ran into this issue when trying to package PDAL in an RPM, that when I specified PDAL_LIB_INSTALL_DIR as lib64 (for Fedora), the CMake config was still going to /usr/lib. This makes it so it will also go to lib64 when the user specifies a PDAL_LIB_INSTALL_DIR by hand.


abellgithub:
Looks fine to me. @hobu: would you merge after tests pass and you've looked it over.

PR #2346, opened 2019-01-14
Avoid overwriting a default SRS with an empty SRS.

connormanning:
If a default SRS was set, and a reader called `setSpatialReference` with an empty SRS, it would end up clearing the SRS.  This PR makes it so that an empty SRS will not override an existing default SRS and adds a corresponding test.


abellgithub:
I don't think I like this.  It doesn't seem like it's what the default SRS should do.  I would think it should be OK to clear an SRS.  Why/how did this come up?

abellgithub:
Close #2347

PR #2352, opened 2019-01-23
Add INCLUDES to some macros

### LINKED ISSUE #1988: Add INCLUDES argument to macros
The macros take libraries to add, but not include directories. It would be nice if they did.

abellgithub:
Close #1988


chambbj:
This looks like a nice add @abellgithub!

PR #2353, opened 2019-01-25
Adding openssl support in order to activate GCS storage

nicolas-chaulet:
I also updated the Arbiter to the latest bundle. Minor changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,6 +129,7 @@ include(${PDAL_CMAKE_DIR}/json.cmake)
 include(${PDAL_CMAKE_DIR}/libxml2.cmake)
 include(${PDAL_CMAKE_DIR}/dimension.cmake)
 include(${PDAL_CMAKE_DIR}/arbiter.cmake)
+include(${PDAL_CMAKE_DIR}/openssl.cmake) # Optional
 
 #------------------------------------------------------------------------------
 # generate the pdal_features.hpp header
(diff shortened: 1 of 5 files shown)


hobu:
This needs to be upstreamed into arbiter

connormanning:
It has already - it just hasn't been re-bundled since then.

abellgithub:
IMO this stuff would be much improved if it were done at build level and we could ditch the #ifdef stuff

abellgithub:
This was reverted because it broke builds.  The issue seems to be the find module for SSL, which on some systems finds incompatible SSL libraries incompatible with headers (libraries too old and aren't version checked).  It also broke our windows build for some reason.  No time to investigate that right now.

nicolas-chaulet:
Hi @abellgithub any chance we could add openssl for UNIX only systems? I can submit a new PR with some updates (version 1.1 of openssl and not 1.0 as per this PR)

abellgithub:
I have no issue with it if it works, but if not, I'm not sure that I'm going to have much time to look into it in the short term if not.  Okay?

nicolas-chaulet:
From what I remember the Linux builds were passing, just the Windows one was breaking. Was there something else?

abellgithub:
The cmake find script that I had was having trouble dealing with both openssl 1 and openssl 1.1 (or some such) on the same system, if I recall.  Many systems come with openssl, and requiring the newer version created a conflict that I couldn't resolve in short order.

Issue #2334, opened 2019-01-06
PDAL translate creating empty LAS files


vedantbajaj96:
Trying to convert i3s service to las file. Getting empty LAS files (477kb in size).

Command I am using - 
pdal.exe translate i3s://https://tiles.arcgis.com/tiles/z2tnIkrLQ2BRzr6P/arcgis/rest/services/USGSNYC_2014_LiDAR/SceneServer D:/PDAL/output/op.las --readers.i3s.threads=64 --readers.i3s.bounds="([-73.958801,-73.938901],[40.755493,40.775493])" --readers.i3s.dimensions="ELEVATION,INTENSITY,CLASS_CODE,FLAGS,RETURNS" --writers.las.scale_x=0.000001 --writers.las.scale_y=0.000001 --writers.las.scale_z=0.000001

abellgithub:
I don't see an issue with the command.  Perhaps the bounds you've selected contain no points?  Are you sure that there are points to be fetched?

vedantbajaj96:
> I don't see an issue with the command. Perhaps the bounds you've selected contain no points? Are you sure that there are points to be fetched?

The selected bound certainly has valid points. We can duplicate this error for a lot of bounds where there is valid data.

kylemann16:
We agree there is an issue here, and are working on it. Have you tried this on any version 2.0 data sets?

abhiataero:
@kylemann16 We have not tried with v2.0 dataset, will try it out and let you know how it goes

vedantbajaj96:
@kylemann16 
We tried with a few different 2.0 datasets - 
a. Manitowoc_County_Ground_Point_Cloud https://tiles.arcgis.com/tiles/ceC3pbsIeU7iCdFg/arcgis/rest/services/Manitowoc_County_Ground_Point_Cloud/SceneServer
b. Zuid_Holland
https://tiles.arcgis.com/tiles/nSZVuSZjHpEZZbRo/arcgis/rest/services/Zuid_Holland/SceneServer
c. AUTZEN_LiDAR_2X https://tiles.arcgis.com/tiles/z2tnIkrLQ2BRzr6P/arcgis/rest/services/AUTZEN_LiDAR_2X/SceneServer
What we are trying to do is convert i3s to las by creating smallers chips of size 0.01 * 0.01.
[...]

PR #2356, opened 2019-02-01
Esri debug

kylemann16:
Changed from OBB calculations to Cube and moved srs to initialize and output messages about the data set to ready. Ticket #2334

### DIFF
--- a/plugins/i3s/io/EsriReader.cpp
+++ b/plugins/i3s/io/EsriReader.cpp
@@ -297,8 +297,6 @@ point_count_t EsriReader::read(PointViewPtr view, point_count_t count)
     const Json::Value initJson = fetchJson(m_filename + "/nodepages/0");
     log()->get(LogLevel::Debug) << "Traversing metadata" << std::endl;
     traverseTree(initJson, 0, nodes, 0, 0);
-    log()->get(LogLevel::Debug) << "Overlapping nodes: " << nodes.size() <<
-        std::endl;
 
     // Create view with overlapping nodes at desired depth
     // Will create a thread pool on the createview class and iterate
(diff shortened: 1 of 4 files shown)


plugins/i3s/io/EsriReader.cpp
```diff
@@ -144,21 +145,31 @@ void EsriReader::initialize(PointTableRef table)
...
-    log()->get(LogLevel::Debug) << "min_density: " <<
-        m_args.min_density << std::endl;
-    log()->get(LogLevel::Debug) << "max_density: " <<
-        m_args.max_density << std::endl;
-    log()->get(LogLevel::Debug) << "dimensions: " << std::endl;
-    for (std::string& dim : m_args.dimensions)
+    //create spatial reference objects
+    Json::Value spatialJson = m_info["spatialReference"];
+    std::string spatialStr = "EPSG:" + spatialJson["wkid"].asString();
+    std::cout << "SRS: " << spatialStr << std::endl;
```
connormanning: Use logger.

test/data/i3s/plineSonoma3.json
```diff
@@ -0,0 +1,15 @@
+{
```
connormanning: This file is added but tests were not changed - is this used?

kylemann16: It was something weird... I wasn't aware I changed it, but I changed it back to where it was before.

plugins/i3s/io/EsriReader.cpp
kylemann16: Just got rid of it.

abellgithub:
Is this something we really want in master?  Don't we want to get feedback from ESRI to figure out if this is what we actually want first?

hobu:
We never heard back from ESRI on this topic, but this PR does fix the issue in #2334. Merge for now for 1.9 and if something comes up we can address that in the next release.

PR #2363, opened 2019-02-08
TileDB plugin

normanb:
This plugin adds [TileDB](https://tiledb.io/) support for point clouds. For ease of testing there is a [Dockerfile](https://github.com/TileDB-Inc/PDAL/blob/pdal-tiledb/scripts/docker/master/ubuntu/Dockerfile) in a different branch.

A sample pipeline for writing 

```

{
  "pipeline":[
    {
      "type":"readers.las",
      "filename":"ot_C320000_4172000.laz"
    },
    {
      "type":"writers.tiledb",
      "config_file":"tiledb.config",
      "compression":"zstd",
      "compression_level": 75,
      "array_name":"sample_array"
    }
  ]
}
```

and a config file for TileDB

```
sm.dedup_coords true
```

To write to a S3 bucket change the TileDB config file to be;

```
sm.dedup_coords true
vfs.s3.scheme https
vfs.s3.region us-east-1
vfs.s3.endpoint_override
vfs.s3.use_virtual_addressing true
```

and the `array_name` in the pipeline to be `s3://mybucket/sample_array`

More detail on S3 and TileDB [here](https://docs.tiledb.io/en/stable/tutorials/working-with-s3.html)


plugins/tiledb/io/TileDBReader.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+    for (const auto& dim: dims)
+    {
+        layout->registerDim(Dimension::id(dim.name()));
+        m_ranges.push_back(dim.domain<double>().first);
+        m_ranges.push_back(dim.domain<double>().second);
+    }
+
+
+    for (const auto& a : attrs) {
```
abellgithub: Same thing here.  Also, you might move the attrs definition right before this.

plugins/tiledb/io/TileDBReader.cpp
```diff
@@ -0,0 +1,331 @@
...
+            subarray.push_back(kv.second.first);
+            subarray.push_back(kv.second.second);
+        }
+        query.set_subarray(subarray);
+    }
+
+    for (const auto& a : attrs) {
+        std::string attrName = a.first;
+        Dimension::Id id = Dimension::id(attrName);
+        Dimension::Type t = Dimension::defaultType(id);
```
abellgithub: defaultType() is a pretty-much useless call.  PDAL will convert from whatever your source type is, so you don't have to worry about this.  You can either a) just fetch your data as doubles and let PDAL convert it its internal storage type b) get the type of the dimension as it's stored in PDAL, which you can do by querying the layout (see dimTypes(), findDim() or findDimType()).

plugins/tiledb/io/TileDBReader.hpp
```diff
@@ -0,0 +1,71 @@
...
+private:
+    virtual void addArgs(ProgramArgs& args);
+    virtual void initialize();
+    virtual void addDimensions(PointLayoutPtr layout);
+    virtual point_count_t read(PointViewPtr view, point_count_t count);
+    virtual void done(PointTableRef table);
+
+    std::string m_arrayName;
+    std::string m_cfgFileName;
+    std::vector<double> m_ranges;
```
abellgithub: m_ranges seems to be set but never used.

plugins/tiledb/io/TileDBWriter.cpp
```diff
@@ -0,0 +1,419 @@
...
+    sp->push_back(val);
+}
+
+void writeAttributeValue(const std::string& attrName,
+        const PointViewPtr view, const PointId idx, std::map<std::string, pdalboost::any>& mp)
+{
+    Dimension::Id id = Dimension::id(attrName);
+    Dimension::Type t = view->dimType(id);
+
+    switch (t)
```
abellgithub: I'm not sure exactly what you're trying to do here.  You don't need to switch on PDAL's type and cast.  PDAL will return whatever type your writer wants, without regard to the way the data is stored in PDAL.  If you want to write your data as doubles, just fetch them as doubles with "getFieldAs<double>(...)".  You don't need to worry about the internal type at all.

plugins/tiledb/io/TileDBWriter.cpp
```diff
@@ -0,0 +1,419 @@
...
+        throw tiledb::TileDBError("Unable to parse compression type: " + name);
+}
+
+std::string TileDBWriter::getName() const { return s_info.name; }
+
+void TileDBWriter::addArgs(ProgramArgs& args)
+{
+    args.add("array_name", "TileDB array name", m_arrayName).setPositional();
+    args.add("config_file", "TileDB configuration file location", m_cfgFileName);
+    args.add("data_tile_capacity", "TileDB tile capacity", m_tile_capacity, unsigned(100000));
```
abellgithub: size_t instead of unsigned?

plugins/tiledb/io/TileDBWriter.cpp
```diff
@@ -0,0 +1,419 @@
...
+        tiledb::Array::create(m_arrayName, *m_schema);
+
+        m_array.reset(new tiledb::Array(*m_ctx, m_arrayName, TILEDB_WRITE));
+        m_query.reset(new tiledb::Query(*m_ctx, *m_array));
+        m_query->set_layout(TILEDB_UNORDERED);
+    }
+    else
+    {
+        throwError("TileDB requires the X,Y,Z dimensions to be the same type.");
+    }
```
abellgithub: You don't need this check.  Unless you have done something strange, X, Y and Z are always stored as doubles.  Furthermore, even if they aren't you can fetch the data as doubles, as you're doing below.

plugins/tiledb/io/TileDBWriter.cpp
```diff
@@ -0,0 +1,419 @@
...
+            std::string attrName = a.first;
+            writeAttributeValue(attrName, view, idx, mapAttrVals);
+        }
+    }
+
+    // set tiledb buffers
+    for (const auto& kv : mapAttrVals)
+    {
+        Dimension::Id id = Dimension::id(kv.first);
+        Dimension::Type t = view->dimType(id);
```
abellgithub: Similar comments here as applied to the reader.

plugins/tiledb/test/TileDBWriterTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+        writer.setInput(m_reader);
+
+        PointTable table;
+        writer.prepare(table);
+        writer.execute(table);
+
+        // check the sidecar exists
+        EXPECT_TRUE(pdal::Utils::fileExists(sidecar));
+    }
+}
```
abellgithub: It would be nice if the tests made sure that data was written as expected and read as expected.

plugins/tiledb/io/TileDBReader.cpp
```diff
@@ -0,0 +1,331 @@
...
+        "Read data from a TileDB array.",
+        "http://pdal.io/stages/readers.tiledb.html"
+};
+
+CREATE_SHARED_STAGE(TileDBReader, s_info)
+std::string TileDBReader::getName() const { return s_info.name; }
+
+template <typename T>
+void makeBuffer(tiledb::Query& query, std::map<std::string, pdalboost::any>& mp,  const std::string& key,
+        const unsigned bufSize)
```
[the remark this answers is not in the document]
normanb: Yes, that would be helpful, thanks. I used `any` because I needed a map of different types.

plugins/tiledb/io/TileDBReader.cpp
```diff
@@ -0,0 +1,331 @@
...
+}
+
+void TileDBReader::addDimensions(PointLayoutPtr layout)
+{
+    std::vector<tiledb::Dimension> dims = m_array->schema().domain().dimensions();
+    auto attrs = m_array->schema().attributes();
+
+    for (const auto& dim: dims)
+    {
+        layout->registerDim(Dimension::id(dim.name()));
```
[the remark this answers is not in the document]
normanb: :+1:

plugins/tiledb/io/TileDBWriter.cpp
normanb: I was aiming to write the data in the type it is stored in. I don't want to use more storage by assuming doubles.

plugins/tiledb/io/TileDBWriter.cpp
normanb: :+1:

plugins/tiledb/io/TileDBWriter.cpp
normanb: :+1:

plugins/tiledb/io/TileDBReader.hpp
normanb: :+1:

plugins/tiledb/test/TileDBWriterTest.cpp
normanb: I will add more to the tests.

cmake/modules/FindTileDB.cmake
```diff
@@ -0,0 +1,81 @@
...
+            message(FATAL_ERROR "TileDB version parsing failed for \"TILEDB_VERSION_MINOR\"")
+        endif()
+        if(NOT "${TILEDB_VERSION_PATCH}" MATCHES "^[0-9]+$")
+            message(FATAL_ERROR "TileDB version parsing failed for \"TILEDB_VERSION_PATCH\"")
+        endif()
+
+        SET(TILEDB_VERSION "${TILEDB_VERSION_MAJOR}.${TILEDB_VERSION_MINOR}.${TILEDB_VERSION_PATCH}"
+                CACHE INTERNAL "The version string for TileDB library")
+
+        IF (TILEDB_VERSION VERSION_LESS TILEDB_FIND_VERSION)
```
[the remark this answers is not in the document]
normanb: This isn't possible currently, but will be. Depending on when this PR gets merged I can add that check.

cmake/tiledb.cmake
```diff
@@ -0,0 +1,15 @@
...
+#
+find_package(TileDB QUIET 1.4.1 REQUIRED)
+set_package_properties(TileDB PROPERTIES
+        TYPE OPTIONAL
+        URL "https://www.tiledb.io"
+        PURPOSE "TileDB support")
+
+if(TILEDB_FOUND)
+    set(CMAKE_REQUIRED_LIBRARIES "${TILEDB_LIBRARIES}")
+    include_directories(${TILEDB_INCLUDE_DIRS})
```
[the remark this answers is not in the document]
normanb: :+1:

plugins/tiledb/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+#
+# TileDB Reader
+#
+PDAL_ADD_PLUGIN(reader_libname reader tiledb
+    FILES
+        io/TileDBReader.cpp
+    LINK_WITH
+        ${TILEDB_LIBRARIES}
+)
+target_include_directories(${reader_libname} PRIVATE
```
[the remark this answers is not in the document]
normanb: :+1:

plugins/tiledb/io/TileDBReader.cpp
normanb: What do you mean by move the attrs definition right before this?

normanb:
@abellgithub I have made the suggested changes. I made a couple of comments ^^. There are two remaining issues

1. A map of different types (should this is a utility outside of this driver)
1. I am intentionally mapping pdal types to tiledb types to reduce storage (rather than assuming all doubles).

cmake/modules/FindTileDB.cmake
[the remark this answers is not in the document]
abellgithub: Seems hard to do.  The FindPythonLibs module makes a good attempt.  I may try to recreate some of that for a couple of other modules.

plugins/tiledb/io/TileDBReader.cpp
abellgithub: Move line 127 right before 134.  It's a nit.  No worries.

plugins/tiledb/io/TileDBReader.cpp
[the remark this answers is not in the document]
abellgithub: I understand why you did it.  I'll merge this and then make a PR against it that you can look over.

normanb:
@abellgithub yes, the .tdb files are part of the array used in the reader test - https://github.com/PDAL/PDAL/blob/7d9261c06a4385f77e7b797fb474e9af70d90bbe/plugins/tiledb/test/TileDBReaderTest.cpp#L73

abellgithub:
I figured I was missing something.  I just didn't see them directly referenced and I know nothing about tiledb.

PR #2366, opened 2019-02-11
Eliminate PDAL-specific FindGDAL

abellgithub:


abellgithub:
Yes, there is.

hobu:
Does our minimum CMake need to increase to make sure we get a good/current enough FindGDAL.cmake? One reason we had our own is the one that was provided was inconsistent. This was even more of a problem with Boost, but we've killed that thank god.

abellgithub:
As far as I can tell, our required cmake version (3.5) will get us a FindGDAL that's "good enough" and certainly as good as what we were packaging.  There have been a few enhancements since 3.5, but I don't think anything that's going to cause issues.

PR #2369, opened 2019-02-13
Update python.cmake

abellgithub:


cmake/python.cmake
```diff
@@ -4,7 +4,7 @@
 
 # Try to find version 3.  If not, we fall back to version 2.
 find_package(PythonInterp QUIET 3)
-find_package(PythonInterp QUIET)
+find_package(PythonInterp QUIET 2.4)
```
hobu: It would be completely reasonable to set this to 2.7. Anything older is going to be problematic for a bunch of other reasons, even if it does compile.

abellgithub: It's just what we had before.

abellgithub:
FindPythonLibs will only find "compatible" libraries.  This means if you specify a minimum version of 2.4, for example, find_package() will only search for 2.4, 2.5, 2.6, 2.7.  It won't search for version 3 libraries.  Also, if PYTHON_EXECUTABLE is set, the find module will try to find libraries that match the executable (as a hint).  Also, python.cmake uses PYTHON_EXECUTABLE to get the numpy version, and that won't work if the library/executable versions don't match (you can't find a numpy installed with python 3 with a python 2 interpreter).

So much trouble with python switching versions.

PR #2370, opened 2019-02-13
Remove GEOS from PDAL

### LINKED ISSUE #2342: Remove GEOS as a dependency
We don't need it, so why use it?

abellgithub:
This removes the direct use of GEOS from PDAL.
Close #2342


filters/private/Point.cpp
```diff
@@ -44,74 +44,92 @@ namespace
...
-    if (t == -1)
-        throw pdal_error("Unable to fetch geometry point type");
-    if (t > 0)
-        throw pdal_error("Geometry type is not point!");
+    OGRwkbGeometryType t = m_geom->getGeometryType();
+    if (t != wkbPoint && t != wkbPoint25D)
+    {
+        m_geom.reset(new OGRPoint());
+        throw pdal_error("Can't set Point from string.  String doesn't "
+            "represent a string.");
```
hobu: string doesn't represent a string? Do you mean it doesn't represent a geometry?

abellgithub: Ummm, that doesn't seem a helpful message.

abellgithub: I'm not sure what you're looking at.  That's not what the current code says and I haven't changed anything.  Weird.  ???

PR #2373, opened 2019-02-14
Properly handle corner cells in filters.splitter

abellgithub:


abellgithub:
I think you wrote this code @gadomski ;)  It's good to have the check in initialize() so that a failure will occur when prepare() is called, rather than after having read a boatload of points.  I'll put a comment in so that it's a bit more clear.

gadomski:
Right you are! I had absolutely no memory of that one...apologies :-). Thanks @abellgithub.

PR #2384, opened 2019-02-24
Expression support for AssignFilter

ahuarte47:
Hi, I would like propose that AssignFilter supports arithmetic expressions to assign the dimension values of points.

For example:
```
"assignment": "X[:]=X + 30",
"assignment": "Intensity[1:3]=(Classification IN (1,2,3)) || (Red >= 90 && Red <= 120)",
"assignment": "Z[:]=0.5 * Z + (HeightAboveGround + 100)"
```

### DIFF
--- a/doc/stages/filters.assign.rst
+++ b/doc/stages/filters.assign.rst
@@ -30,11 +30,42 @@ This pipeline resets the ``Classification`` of all points with classifications
       }
   ]
 
+Example 2
+---------
+
+This pipeline sets some dimensions using arithmetic expressions even 
+fetching values from other attributes.
+
+.. code-block:: json
+
+    {
+      "pipeline":[
+        "autzen-dd.las",
+        {
+          "type":"filters.assign",
+          "assignment" : "Classification[1:1]=Classification * (Z > 1000 || (Z >= 10 && Z <= 20))",
+          "assignment" : "Red[:]=(Blue + Green)",
+          "assignment" : "Userdata[:]=1000 + (Classification * 2)",
+          "assignment" : "UserData[:]=PointSourceId IN (200,500,700)",
+          "assignment" : "Z[:]=Z + 100.0",
+        },
+        {
+          "filename":"attributed.las",
+        }
+      ]
+    }
+
+It supports the most common operators (``*,/,+,-,==,!=,<>,>,<,>=,<=,&&,||``), and ``( )`` 
+for priorizing them. It supports too ``IN`` function to check the presence of a value 
+e.g. ``"PointSourceId IN (2300,2301,2302)"``.
+
+
 Options
 -------
 
 assignment
-  A :ref:`range <ranges>` followed by an assignment of a value (see example).
+  A :ref:`range <ranges>` followed by an assignment of a value or arithmetic 
+  expression (see examples).
   Can be specified multiple times.  The assignments are applied sequentially
   to the dimension value as set when the filter began processing. [Required]
 
--- a/filters/AssignFilter.cpp
+++ b/filters/AssignFilter.cpp
@@ -38,6 +38,7 @@
 #include <pdal/util/ProgramArgs.hpp>
 
 #include "private/DimRange.hpp"
+#include "private/expr/Parser.hpp"
 
 namespace pdal
 {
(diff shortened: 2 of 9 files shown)


hobu:
Thanks for the contribution. We might have a few change requests on the syntax, but we're supportive of adding this. On vacation for the next week though, so expect some noise on this next week. We'll plan to try to include this feature in the next release.

ahuarte47:
Thanks @hobu ! every advices are welcome.

abellgithub:
I'm having a hard time seeing the utility of something like:

`"Z[:]=$( 6 + ([GpsTime] / 2.0) - (2 * [Red] + ([X] == 7)) + 1.5 * ([Red] != 1) )"`

I guess my thinking is that we should lose the equality/inequality/IN operators for now and see how it goes.  I'm not saying there's never a use for such things, but it seems of low utility.  If we're going to make a real language, then we should make a real language.

From the standpoint of syntax, I don't see any reason for the $() on the right side of the assignment and I don't see any reason for enclosing the dimension names in [].  If we're going to do this, I'd like to see something clean and familiar.

ahuarte47:
OK, you are right, the expression samples are strange, I wrote them to test the sturdiness of the code. A more real expression would be:

   ```
 {
      "type": "filters.assign",
      "assignment": "TempFlag[:]=$( [Zbridge] > 0 && (([Z]-[Zbridge]) >= -3.0) && (([Z]-[Zbridge]) <= 3.0) && [Classification] == 2 )"
    },
    {
      "type": "filters.assign",
      "assignment": "TempFlag[1:1]=Classification:32"
    },

```

You can assign the classification of a group of points according with a complex expression.

About the syntax, every changes are welcome, I added the `$()` in order to preserve the performance when the expression is a single number as until now. `[]` seemed me better than only specify the dimension name. But as I said, they were proposals. I am sure your comments will improve the readabilty of the expressions.

**Comment by abellgithub** [comment]
I offer this: https://github.com/PDAL/PDAL/compare/expr .  It's old code that I repurposed.  It has the major advantage over this PR in that is doesn't reparse the expression for each point.  It builds an expression tree that's evaluated.  This also means that the parsing happens when the assigment is encountered, allowing for error feedback at the appropriate time.

**Comment by ahuarte47** [comment]
Hello, @abellgithub, you're right, your code has important advantages, it will undoubtedly be faster than my RP. I only see one disadvantage, it does not support equality and comparison operators. At this stage, what would you like? Replace the code for yourself and/or I close my PR? Or do you prefer that I change the code adding also the missing operators?

**Comment by ahuarte47** [comment]
Hi @abellgithub; I am testing your code and I get an error in parser (Descr: `Can't prepare.  Node tree not properly parsed`). 

I am using the original expressions in the `AssignFilterTest.cpp` test file:

```
 Options fo;
 fo.add("assignment", "X[:]=27.5");
 fo.add("assignment", "Classification[:]=0");
 fo.add("assignment", "GpsTime[:]=3000");
 fo.add("assignment", "Red[:]=255");

```
Although the expressions are successfully parsed, parser classes have empty `m_nodes` and `m_tokens` members when they are going to be prepared in the `AssignFilter::prepare` method.

I think if the test has only an "assignment" item, the error dissapears.

**Comment by ahuarte47** [comment]
Hi @abellgithub, I found the error, [ProgramArgs::setValue](https://github.com/PDAL/PDAL/compare/expr#diff-13930bb68645ab1ad29c9986010773d5R768) calls to the constructor [AssignRange(const AssignRange& r)](https://github.com/PDAL/PDAL/compare/expr#diff-a54fe753ed67e4b63a6e63edd004886bR77) but it does not assign the parser.

This works for me, I remove the constructor `AssignRange(AssignRange&& r)`
and change the other one:
```
    AssignRange(AssignRange& r) : DimRange(r), m_value(r.m_value),
        m_parser(std::move(r.m_parser))
    {}

```

**Comment by abellgithub** [comment]
I'll have to look at this.  I haven't had time.  Thanks for debugging.

**Comment by ahuarte47** [comment]
You are welcome!, now I am adding the comparison and AND/OR operators to your lexer and parser classes. 

But I have found that Lparen/Rparen pair is not properly parsed. Expressions as `(3000) + 1` fail because of the parser only reads chars until the Rparen item is reached. I will find out the problem. 

Thanks.

**Comment by ahuarte47** [comment]
> But I have found that Lparen/Rparen pair is not properly parsed. Expressions as `(3000) + 1` fail because of the parser only reads chars until the Rparen item is reached. I will find out the problem.
>
I found the error, I think Rparen has to be tested in the [Lexer::getOperator](https://github.com/PDAL/PDAL/compare/expr#diff-14f02b4a7d6e359ddeb7d3c629a5a8ccR55) function.

This change works:

```
Token Lexer::getOperator(char c)
{
    Token tok(TokenType::Error);

    if (c == ')')
        tok = Token(TokenType::Rparen, m_pos, m_pos + 1);
    else if (c == '+')
        tok = Token(TokenType::Plus, m_pos, m_pos + 1);
    else if (c == '-')
        tok = Token(TokenType::Minus, m_pos, m_pos + 1);
    ...
```

Complex expressions with parenthesis run right now:

```
    fo.add("assignment", "GpsTime[:]=(3000)");
    fo.add("assignment", "Green[:]=(GpsTime * 3) + 4");
    fo.add("assignment", "Blue[:]=(GpsTime >= 3000) && (Red + 1 == GpsTime + 10 - 9)");

```

abellgithub:
@ahuarte47 - Do you want to pursue the branch that I submitted and for which you've made some fixes?  If so, I'll look it over and see if we can get it merged.

ahuarte47:
Yes, I use your code as base of this PR, I added new operators and fixes.

abellgithub:
To be completed with #2638

hobu:
Implemented in a slightly different (but maybe more complete) form in #3202

ahuarte47:
Thanks @hobu, amazing new feature!

PR #2391, opened 2019-03-04
EPT addon dimension writer

connormanning:
This PR adds an EPT addon dimension writer.

This writer lets you write additional attributes externally to the source EPT data itself, so write-access to the original data location is not needed.  Additionally to the new EPT addon writer, the EPT reader has been modified such that addon attributes may be associated with a dataset at read time.

The write-side might look something like this, which runs [SMRF ground classification](https://pdal.io/stages/filters.smrf.html) over a [remote EPT resource](http://potree.entwine.io/data/autzen.html), and writes the resulting `Classification` dimension to a local path:
```
[{
    "type": "readers.ept",
    "filename": "http://na.entwine.io/autzen/ept.json"
}, {
    "type": "filters.assign",
    "assignment": "Classification[:]=0"
}, {
    "type": "filters.smrf"
}, {
    "type": "writers.ept_addon",
    "addons": { "~/entwine/addons/autzen/smrf": "Classification" }
}]
```

The corresponding read-side might look like this, which maps the above addon location back into the `Classification` value within PDAL and writes the entire modified dataset to a single file:
```
[{
    "type": "readers.ept",
    "filename": "http://na.entwine.io/autzen/ept.json",
    "addons": { "Classification": "~/entwine/addons/autzen/smrf" }
}, 
"autzen-smrf-output.las"
]
```

Bounded queries are also supported: if an EPT addon writer is used in tandem with an EPT reader invocation which is bounded by space or resolution, then the attribute will only be applied to the queried data, and this attribute for points outside of the query will be zero-initialized on subsequent reads.

Usage note: the `addons` option is reversed between the EPT reader and writer: in each case, the right-hand side represents an assignment to the left-hand side.  In the writer, the dimension value is assigned to an addon path.  In the reader, the addon path is assigned to a dimension.

While this example discussed local paths for addons, S3/GCS/Dropbox/etc paths are also supported with appropriate configuration.

### DIFF
--- a/doc/stages/readers.ept.rst
+++ b/doc/stages/readers.ept.rst
@@ -25,6 +25,20 @@ This example downloads a small area around the the Statue of Liberty from the Ne
       "statue-of-liberty.las"
    ]
 
+Additional attributes created by the :ref:`writers.ept_addon <EPT addon writer>` can be referenced with the ``addon`` option.  Here is an example that overrides the ``Classification`` dimension with an addon dimension derived from the original dataset:
+
+.. code-block:: json
+
+   [{
+      "type": "readers.ept",
+      "filename": "http://na.entwine.io/autzen/ept.json",
+      "addons": { "Classification": "~/entwine/addons/autzen/smrf" }
+   }, {
+      "type": "writers.las",
+      "filename": "autzen-ept-smrf.las"
+   }]
+
+For more details about addon dimensions and how to produce them, see :ref:`writers.ept_addon`.
 
 Options
 --------------------------------------------------------------------------------
--- a/doc/stages/readers.ept.rst
+++ b/doc/stages/readers.ept.rst
@@ -43,7 +43,7 @@ resolution
     The resulting resolution may not be exactly this value: the minimum possible resolution that is at *least* as precise as the requested resolution will be selected.  Therefore the result may be a bit more precise than requested.
 
 addons
-    A mapping of assignments of the form ``DimensionName: AddonPath``, which assigns dimensions from the specified paths to the named dimensions.  These addon dimensions are created by the :ref:`writers.ept_addon`_.  If the dimension names already exist in the EPT `Schema`_ for the given resource, then their values will be overwritten with those from the appropriate addon.
+    A mapping of assignments of the form ``DimensionName: AddonPath``, which assigns dimensions from the specified paths to the named dimensions.  These addon dimensions are created by the :ref:`writers.ept_addon <EPT addon writer>`_.  If the dimension names already exist in the EPT `Schema`_ for the given resource, then their values will be overwritten with those from the appropriate addon.
 
     Addons may used to override well-known :ref:`dimensions`_.  For example, an addon assignment of ``"Classification": "~/addons/autzen/MyGroundDimension/"`` will override an existing EPT ``Classification`` dimension with the custom dimension.
 
--- a/doc/stages/readers.ept.rst
+++ b/doc/stages/readers.ept.rst
@@ -16,16 +16,14 @@ This example downloads a small area around the the Statue of Liberty from the Ne
 
 .. code-block:: json
 
-    {
-      "pipeline": [
-        {
-          "type": "readers.ept",
-          "filename": "http://na.entwine.io/nyc",
-          "bounds": "([-8242669, -8242529], [4966549, 4966674])"
-        },
-        "statue-of-liberty.las"
-      ]
-    }
+   [
+      {
+         "type": "readers.ept",
+         "filename": "http://na.entwine.io/nyc/ept.json",
+         "bounds": "([-8242669, -8242529], [4966549, 4966674])"
+      },
+      "statue-of-liberty.las"
+   ]
 
 
 Options
@@ -44,6 +42,11 @@ resolution
 
     The resulting resolution may not be exactly this value: the minimum possible resolution that is at *least* as precise as the requested resolution will be selected.  Therefore the result may be a bit more precise than requested.
 
+addons
+    A mapping of assignments of the form ``DimensionName: AddonPath``, which assigns dimensions from the specified paths to the named dimensions.  These addon dimensions are created by the `EPT writer`_.  If the dimension names already exist in the EPT `Schema`_ for the given resource, then their values will be overwritten with those from the appropriate addon.
+
+    Addons may used to override well-known dimensions (see `Dimensions`_).  For example, an addon assignment of ``"Classification": "~/addons/autzen/MyGroundDimension/"`` will override an existing EPT ``Classification`` dimension with the custom dimension.
+
 origin
     EPT datasets are lossless aggregations of potentially multiple source files.  The *origin* options can be used to select all points from a single source file.  This option may be specified as a string or an integral ID.
 
@@ -58,4 +61,7 @@ threads
 .. _Entwine: https://entwine.io/
 .. _Potree: http://potree.entwine.io/data/nyc.html
 .. _Plasio: http://speck.ly/?s=http%3A%2F%2Fc%5B0-7%5D.greyhound.io&r=ept%3A%2F%2Fna.entwine.io%2Fnyc&ca=-0&ce=49.06&ct=-8239196%2C4958509.308%2C337&cd=42640.943&cmd=125978.13&ps=2&pa=0.1&ze=1&c0s=remote%3A%2F%2Fimagery%3Furl%3Dhttp%3A%2F%2Fserver.arcgisonline.com%2FArcGIS%2Frest%2Fservices%2FWorld_Imagery%2FMapServer%2Ftile%2F%7B%7Bz%7D%7D%2F%7B%7By%7D%7D%2F%7B%7Bx%7D%7D.jpg
+.. _Dimensions: https://pdal.io/dimensions.html
+.. _Schema: https://entwine.io/entwine-point-tile.html#schema
+.. _EPT writer: https://pdal.io/stages/writers.ept.html
 
--- a/doc/stages/readers.ept.rst
+++ b/doc/stages/readers.ept.rst
@@ -43,9 +43,9 @@ resolution
     The resulting resolution may not be exactly this value: the minimum possible resolution that is at *least* as precise as the requested resolution will be selected.  Therefore the result may be a bit more precise than requested.
 
 addons
-    A mapping of assignments of the form ``DimensionName: AddonPath``, which assigns dimensions from the specified paths to the named dimensions.  These addon dimensions are created by the `EPT writer`_.  If the dimension names already exist in the EPT `Schema`_ for the given resource, then their values will be overwritten with those from the appropriate addon.
+    A mapping of assignments of the form ``DimensionName: AddonPath``, which assigns dimensions from the specified paths to the named dimensions.  These addon dimensions are created by the :ref:`writers.ept_addon`_.  If the dimension names already exist in the EPT `Schema`_ for the given resource, then their values will be overwritten with those from the appropriate addon.
 
-    Addons may used to override well-known dimensions (see `Dimensions`_).  For example, an addon assignment of ``"Classification": "~/addons/autzen/MyGroundDimension/"`` will override an existing EPT ``Classification`` dimension with the custom dimension.
+    Addons may used to override well-known :ref:`dimensions`_.  For example, an addon assignment of ``"Classification": "~/addons/autzen/MyGroundDimension/"`` will override an existing EPT ``Classification`` dimension with the custom dimension.
 
 origin
     EPT datasets are lossless aggregations of potentially multiple source files.  The *origin* options can be used to select all points from a single source file.  This option may be specified as a string or an integral ID.
@@ -61,7 +61,5 @@ threads
 .. _Entwine: https://entwine.io/
 .. _Potree: http://potree.entwine.io/data/nyc.html
 .. _Plasio: http://speck.ly/?s=http%3A%2F%2Fc%5B0-7%5D.greyhound.io&r=ept%3A%2F%2Fna.entwine.io%2Fnyc&ca=-0&ce=49.06&ct=-8239196%2C4958509.308%2C337&cd=42640.943&cmd=125978.13&ps=2&pa=0.1&ze=1&c0s=remote%3A%2F%2Fimagery%3Furl%3Dhttp%3A%2F%2Fserver.arcgisonline.com%2FArcGIS%2Frest%2Fservices%2FWorld_Imagery%2FMapServer%2Ftile%2F%7B%7Bz%7D%7D%2F%7B%7By%7D%7D%2F%7B%7Bx%7D%7D.jpg
-.. _Dimensions: https://pdal.io/dimensions.html
 .. _Schema: https://entwine.io/entwine-point-tile.html#schema
-.. _EPT writer: https://pdal.io/stages/writers.ept.html
 
--- a/doc/stages/writers.ept_addon.rst
+++ b/doc/stages/writers.ept_addon.rst
@@ -1,9 +1,9 @@
-.. _writers.ept:
+.. _writers.ept_addon:
 
-writers.ept
-===========
+writers.ept_addon
+=================
 
-The **EPT writer** supports writing additional dimensions to `Entwine Point Tile`_ datasets.  Thus the EPT Writer is unique among most other PDAL writers due to its limitation that it may only be used along with an `EPT reader`_.
+The **EPT addon writer** supports writing additional dimensions to `Entwine Point Tile`_ datasets.  Thus the EPT addon writer is unique among most other PDAL writers due to its limitation that it may only be used along with an `EPT reader`_ and modifies and existing dataset rather than creating a brand new one.
 
 The addon dimensions created by this writer may be stored independently from the corresponding EPT dataset, therefore write-access to the EPT resource itself is not required to create and use addon dimensions.
 
@@ -26,7 +26,7 @@ This example downloads the Autzen dataset (10M points) and runs the `SMRF filter
    }, {
       "type": "filters.smrf"
    }, {
-      "type": "writers.ept",
+      "type": "writers.ept_addon",
       "addons": { "~/entwine/addons/autzen/smrf": "Classification" },
       "threads": 12
    }]
@@ -54,7 +54,7 @@ addons
 
 .. note::
 
-   The ``addons`` option is reversed between the EPT reader and writer: in each case, the right-hand side represents an assignment to the left-hand side.  In the writer, the dimension value is assigned to an addon path.  In the reader, the addon path is assigned to a dimension.
+   The ``addons`` option is reversed between the EPT reader and addon-writer: in each case, the right-hand side represents an assignment to the left-hand side.  In the writer, the dimension value is assigned to an addon path.  In the reader, the addon path is assigned to a dimension.
 
 threads
     Number of worker threads used to write EPT addon data.  A minimum of 4 will be used no matter what value is specified.
--- a/doc/stages/writers.ept_addon.rst
+++ b/doc/stages/writers.ept_addon.rst
@@ -3,23 +3,22 @@
 writers.ept_addon
 =================
 
-The **EPT addon writer** supports writing additional dimensions to `Entwine Point Tile`_ datasets.  Thus the EPT addon writer is unique among most other PDAL writers due to its limitation that it may only be used along with an `EPT reader`_ and modifies and existing dataset rather than creating a brand new one.
+The **EPT addon writer** supports writing additional dimensions to `Entwine Point Tile`_ datasets.  Thus the EPT addon writer is unique among most other PDAL writers due to its limitation that it may only be used along with an :ref:`readers.ept <EPT reader>`_, and it creates additional attributes for an existing dataset rather than creating a brand new one.
 
-The addon dimensions created by this writer may be stored independently from the corresponding EPT dataset, therefore write-access to the EPT resource itself is not required to create and use addon dimensions.
+The addon dimensions created by this writer are stored independently from the corresponding EPT dataset, therefore write-access to the EPT resource itself is not required to create and use addon dimensions.
 
 .. embed::
 
 Example
 --------------------------------------------------------------------------------
 
-This example downloads the Autzen dataset (10M points) and runs the `SMRF filter`_, which populates the ``Classification`` dimension with ground values, and writes the resulting attribute to an EPT addon dataset on the local filesystem.
+This example downloads the Autzen dataset (10M points) and runs the :ref:`filters.smrf <SMRF filter>`_, which populates the ``Classification`` dimension with ground values, and writes the resulting attribute to an EPT addon dataset on the local filesystem.
 
 .. code-block:: json
 
    [{
       "type": "readers.ept",
-      "filename": "http://na.entwine.io/autzen/ept.json",
-      "threads": 12
+      "filename": "http://na.entwine.io/autzen/ept.json"
    }, {
       "type": "filters.assign",
       "assignment": "Classification[:]=0"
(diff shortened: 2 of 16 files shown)


connormanning:
I'm going to rename this to `writers.ept-addon` for clarity.

abellgithub:
Can't use `-`.  Underscore is OK.

connormanning:
Renamed to `writers.ept_addon` in d6544096814a5be497433c3ed520d8bb0d506c88.

doc/stages/readers.ept.rst
```diff
@@ -16,16 +16,14 @@ This example downloads a small area around the the Statue of Liberty from the Ne
 
 .. code-block:: json
 
-    {
-      "pipeline": [
```
chambbj: Did something change that PDAL is now able to parse the pipeline array directly? I think you've told me that Entwine has been able to do this for a while. Just wonder who the audience is here.

doc/stages/readers.ept.rst
```diff
@@ -44,6 +42,11 @@ resolution
 
     The resulting resolution may not be exactly this value: the minimum possible resolution that is at *least* as precise as the requested resolution will be selected.  Therefore the result may be a bit more precise than requested.
 
+addons
+    A mapping of assignments of the form ``DimensionName: AddonPath``, which assigns dimensions from the specified paths to the named dimensions.  These addon dimensions are created by the `EPT writer`_.  If the dimension names already exist in the EPT `Schema`_ for the given resource, then their values will be overwritten with those from the appropriate addon.
```
chambbj: I guess this should now read the EPT addon writer.

doc/stages/readers.ept.rst
```diff
@@ -58,4 +61,7 @@ threads
 .. _Entwine: https://entwine.io/
 .. _Potree: http://potree.entwine.io/data/nyc.html
 .. _Plasio: http://speck.ly/?s=http%3A%2F%2Fc%5B0-7%5D.greyhound.io&r=ept%3A%2F%2Fna.entwine.io%2Fnyc&ca=-0&ce=49.06&ct=-8239196%2C4958509.308%2C337&cd=42640.943&cmd=125978.13&ps=2&pa=0.1&ze=1&c0s=remote%3A%2F%2Fimagery%3Furl%3Dhttp%3A%2F%2Fserver.arcgisonline.com%2FArcGIS%2Frest%2Fservices%2FWorld_Imagery%2FMapServer%2Ftile%2F%7B%7Bz%7D%7D%2F%7B%7By%7D%7D%2F%7B%7Bx%7D%7D.jpg
+.. _Dimensions: https://pdal.io/dimensions.html
+.. _Schema: https://entwine.io/entwine-point-tile.html#schema
+.. _EPT writer: https://pdal.io/stages/writers.ept.html
```
chambbj: Now writers.ept_addon.html.

doc/stages/writers.ept_addon.rst
```diff
@@ -0,0 +1,66 @@
+.. _writers.ept_addon:
+
+writers.ept_addon
+=================
+
+The **EPT addon writer** supports writing additional dimensions to `Entwine Point Tile`_ datasets.  Thus the EPT addon writer is unique among most other PDAL writers due to its limitation that it may only be used along with an `EPT reader`_ and modifies and existing dataset rather than creating a brand new one.
```
chambbj: "...modifies ~and~ an existing dataset..."

But it doesn't really modify anything. The addon dimension stands alone, but is necessarily linked to an existing asset.

doc/stages/writers.ept_addon.rst
```diff
@@ -0,0 +1,66 @@
...
+
+The addon dimensions created by this writer may be stored independently from the corresponding EPT dataset, therefore write-access to the EPT resource itself is not required to create and use addon dimensions.
```
chambbj: Not required. Is it possible at all? If so, is it worth documenting how?

doc/stages/writers.ept_addon.rst
```diff
@@ -0,0 +1,66 @@
...
+   }]
+
+And here is a follow-up example of reading this dataset with the `EPT reader`_ - with the created addon overwriting the ``Classification`` value - and then writing the output to a single file with the `LAS writer`_.
+
+.. code-block:: json
+
+   [{
+      "type": "readers.ept",
+      "filename": "http://na.entwine.io/autzen/ept.json",
+      "addons": { "Classification": "~/entwine/addons/autzen/smrf" },
```
chambbj: It wouldn't be a horrible idea to show an example or at least mention how multiple addons can be read/written by the EPT drivers.

doc/stages/readers.ept.rst
```diff
@@ -44,6 +42,11 @@ resolution
 
     The resulting resolution may not be exactly this value: the minimum possible resolution that is at *least* as precise as the requested resolution will be selected.  Therefore the result may be a bit more precise than requested.
 
+addons
```
chambbj: You do not provide an example in the reader documentation, instead saving it for the writer. I understand it may be redundant to show it both places. Maybe a cross reference? I don't know.

doc/stages/writers.ept_addon.rst
```diff
@@ -0,0 +1,66 @@
...
+      "filename": "http://na.entwine.io/autzen/ept.json",
+      "threads": 12
+   }, {
+      "type": "filters.assign",
+      "assignment": "Classification[:]=0"
+   }, {
+      "type": "filters.smrf"
+   }, {
+      "type": "writers.ept_addon",
+      "addons": { "~/entwine/addons/autzen/smrf": "Classification" },
```
chambbj: Also, to be somewhat consistent with what other stages do (e.g., filters.assign), I wonder if it would make more sense to allow repeated specification of `addons`. For example,

```
"addons": "~/entwine/addons/autzen/smrf=Classification",
"addons": "~/entwine/addons/autzen/normal_x=NormalX"
```

Although, we already differ between filters.ferry and filters.assign, as the filters.ferry way of doing this would be

```
"addons":"~/entwine/addons/autzen/smrf<=Classification,~/entwine/addons/autzen/normal_x=NormalX"
```

io/EptAddonWriter.cpp
```diff
@@ -0,0 +1,252 @@
...
+    hierEp.put(key.toString() + ".json", h.toStyledString());
+
+    m_pool->await();
+
+    // Write the top-level addon metadata.
+    Json::Value meta;
+    meta["type"] = getTypeString(addon.type());
+    meta["size"] = static_cast<Json::UInt64>(addon.size());
+    meta["version"] = "1.0.0";
+    meta["dataType"] = "binary";
```
chambbj: I'm not very familiar with the EPT spec at this point. Is there another option?

io/EptReader.cpp
```diff
@@ -125,8 +91,8 @@ BOX3D EptReader::Args::bounds() const
     const BOX2D b(m_bounds.to2d());
     if (b.empty())
     {
-        double mn((std::numeric_limits<double>::lowest)());
-        double mx((std::numeric_limits<double>::max)());
+        const double mn((std::numeric_limits<double>::lowest)());
```
chambbj: Wouldn't this cause an issue if you had legitimate negative coordinates but did not specify bounds?

doc/stages/readers.ept.rst
hobu: :ref:`readers.ept_addon` will get you the top link to the page all the time instead of using HTML links. This will also emit links in the other help content types like man, CHM, and PDF.

doc/stages/readers.ept.rst
abellgithub: I think this went in for 1.8, but yes, you can just provide an array of stages now.

doc/stages/writers.ept_addon.rst
abellgithub: If you the argument type is vector<string>, it will automatically accept comma-separated or repeated arguments.

doc/stages/writers.ept_addon.rst
abellgithub: Really?  Sidecar files?  This seems ugh.

doc/stages/readers.ept.rst
connormanning: It was #2183.  It missed 1.8 I think but has been on master since shortly afterward.

doc/stages/writers.ept_addon.rst
hobu: The requirement is to be able to augment a tree which you do not have write access. For example, all of the data at usgs.entwine.io.

doc/stages/readers.ept.rst
connormanning: Addressed in a0f29158c966cafb1e64d886f213d3b4a2ae63e6.

doc/stages/readers.ept.rst
connormanning: Addressed in Addressed in a0f29158c966cafb1e64d886f213d3b4a2ae63e6.

doc/stages/readers.ept.rst
abellgithub: That sounds right.

io/EptAddonWriter.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+void EptAddonWriter::writeOne(const PointViewPtr view, const Addon& addon) const
+{
+    std::vector<std::vector<char>> buffers;
+    buffers.reserve(m_hierarchy.size());
+
+    // Create an addon buffer for each node we're going to write.
+    for (const auto& p : m_hierarchy)
+    {
+        buffers.emplace_back(p.second * addon.size(), 0);
```
abellgithub: Vectors are always 0-initialized, but I doubt being explicit hurts anything.

io/EptAddonWriter.cpp
```diff
@@ -0,0 +1,252 @@
...
+        if (!nodeId) continue;
+
+        nodeId -= 1;
+        pointId = pr.getFieldAs<uint64_t>(m_pointIdDim);
+
+        auto& buffer(buffers.at(nodeId));
+        assert(pointId * addon.size() + addon.size() <= buffer.size());
+        char* dst = buffer.data() + pointId * addon.size();
+        pr.getField(dst, addon.id(), addon.type());
+    }
```
abellgithub: This seems an unfortunate memory suck.  I would guess that in most instances points would be in the order read.  Can you do something like fill buffers until they hit some size and then flush them?  Can you fork a thread to flush so that you can keep processing?  Maybe that's version 2.

doc/stages/writers.ept_addon.rst
connormanning: Language fixed in c57730d.

doc/stages/writers.ept_addon.rst
connormanning: It's not really possible, as there's no EPT level tie-in to these dimensions.  However since they're independent you could place them in a subdirectory of the EPT data if you were so inclined.  I've cleaned up the language a bit to remove the implication that the data might be stored in-band in c57730d.

doc/stages/writers.ept_addon.rst
connormanning: Added colorinterp RGB example in c57730d.

doc/stages/readers.ept.rst
connormanning: Added an addon example to the reader in 59f40a0, deferring to the writer doc for details.

io/EptAddonWriter.cpp
connormanning: For EPT itself (rather than addons), `laszip` is the other supported option, which isn't appropriate for single-attribute addons.  I'll shortly be adding Draco compression support which will be an option for both EPT and for addons.  For now, the answer to your question is no - there is not another option.

io/EptAddonWriter.cpp
```diff
@@ -0,0 +1,252 @@
...
+void EptAddonWriter::writeHierarchy(Json::Value& curr, const Key& key,
+        const arbiter::Endpoint& hierEp) const
+{
+    const std::string keyName(key.toString());
+    if (!m_hierarchy.count(keyName)) return;
+
+    const uint64_t np(m_hierarchy.at(keyName));
+    if (!np) return;
+
+    if (m_hierarchyStep && key.d && key.d == m_hierarchyStep)
```
abellgithub: Seems like one of these &&'s is not needed.  Maybe the compiler is smart enough to see that.

doc/stages/writers.ept_addon.rst
connormanning: I'm strongly opposed to repeated arguments, which although technically valid JSON, have undefined parsing behavior and cannot be generated by JSON libraries in general.  I also tend to prefer the proposed native JSON mapping rather than a CSV string - I think it makes more sense from a JSON API standpoint even if asymmetrical from other filters whose interfaces were defined prior to PDAL JSON support.

io/EptReader.cpp
connormanning: Not sure I understand this one - can you explain further?  The logic here is intended to set the query boundaries to `(-Inf, -Inf, -Inf, Inf, Inf, Inf)`, if a query boundary is not supplied by the user.

chambbj:
I'm probably just confusing min and lowest again.

io/private/EptSupport.hpp
```diff
@@ -90,6 +158,127 @@ inline std::array<double, 3> toArray3(const Json::Value& json)
     return result;
 }
 
+class PDAL_DLL Key
```
abellgithub: Does this need PDAL_DLL?

io/private/EptSupport.hpp
```diff
@@ -90,6 +158,127 @@ inline std::array<double, 3> toArray3(const Json::Value& json)
...
+{
+    // An EPT key representation (see https://git.io/fAiBh).  A depth/X/Y/Z key
+    // representing a data node, as well as the bounds of the contained data.
+public:
+    Key(BOX3D b) : b(b) { }
```
abellgithub: I find this structure confusing.  Why don't you initialize b when you set x/y/z/d and vice versa?  Maybe I just find the [] operator to get at b kind of opaque.

Also, this code seems really familiar.  Is this a copy from somewhere?

io/EptReader.cpp
chambbj: I've just confused min and lowest again.

io/EptAddonWriter.hpp
```diff
@@ -0,0 +1,91 @@
...
+    virtual void addArgs(ProgramArgs& args) override;
+    virtual void addDimensions(PointLayoutPtr layout) override;
+    virtual void ready(PointTableRef table) override;
+    virtual void write(const PointViewPtr view) override;
+
+private:
+    std::unique_ptr<arbiter::Arbiter> m_arbiter;
+    std::unique_ptr<Pool> m_pool;
+
+    std::unique_ptr<Json::Value> m_addonsArg;
```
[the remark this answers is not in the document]
connormanning: Pointer was to keep it out of public headers.

io/EptAddonWriter.hpp
```diff
@@ -0,0 +1,91 @@
...
+    std::vector<std::unique_ptr<Addon>> m_addons;
```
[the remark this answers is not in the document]
connormanning: You mean the unique_ptr?  That was intended so that we could use the forward declaration since this class is defined in `private/EptSupport.hpp`.  The `Addon` class is used by both the EPT reader and the addon writer.

pdal/StageFactory.cpp
```diff
@@ -90,12 +98,12 @@ std::string StageFactory::inferWriterDriver(const std::string& filename)
...
 
-    static const std::string ghPrefix("greyhound://");
+    const std::string driverProtocol = getDriverProtocol(filename);
 
     if (filename == "STDOUT")
         ext = ".txt";
-    else if (Utils::iequals(filename.substr(0, ghPrefix.size()), ghPrefix))
-        ext = ".greyhound";      // Make it look like an extension.
+    else if (!driverProtocol.empty())
+        ext = "." + driverProtocol;
```
abellgithub: 👍

pdal/util/Utils.hpp
```diff
@@ -194,6 +194,21 @@ namespace Utils
         return (strncmp(prefix.data(), s.data(), prefix.size()) == 0);
     }
 
+    /**
+      Determine if a string ends with a particular postfix.
```
abellgithub: postfix is an awful term

pdal/util/Utils.hpp
```diff
@@ -194,6 +194,21 @@ namespace Utils
...
+
+      \param s  String to check for postfix.
+      \param postfix Postfix to search for.
+      \return  Whether the string ends with the postfix.
+    */
+    inline bool endsWith(const std::string& s, const std::string& postfix)
+    {
+        if (postfix.size() > s.size())
+            return false;
+        return (strncmp(postfix.data(), s.data() + s.size() - postfix.size(),
```
abellgithub: You can just strcmp.  The strings are guaranteed to be terminated as of C++11.

test/unit/CMakeLists.txt
```diff
@@ -119,6 +119,15 @@ PDAL_ADD_TEST(pdal_io_ept_reader_test
     INCLUDES
         ${PDAL_JSONCPP_INCLUDE_DIR}
 )
+PDAL_ADD_TEST(pdal_io_ept_addon_writer_test
+    FILES
+        io/EptAddonWriterTest.cpp
+    LINK_WITH
+        ${PDAL_JSONCPP_LIB_NAME}
+        ${GDAL_LIBRARY}
```
abellgithub: Why do you need GDAL?  Are you calling GDAL functions from the test?

io/EptAddonWriter.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+    void writeOne(const PointViewPtr view, const Addon& addon) const;
+    void writeHierarchy(Json::Value& hier, const Key& key,
+            const arbiter::Endpoint& hierEp) const;
+
+    uint64_t m_numThreads = 0;
```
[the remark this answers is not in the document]
connormanning: Changed to `size_t` and added a warning for values > 100.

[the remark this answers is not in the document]
connormanning: In ed36998.

io/private/EptSupport.hpp
```diff
@@ -90,6 +158,127 @@ inline std::array<double, 3> toArray3(const Json::Value& json)
...
+    const arbiter::Endpoint& ep() const { return m_ep; }
+    Dimension::Id id() const { return m_id; }
+    Dimension::Type type() const { return m_type; }
+    uint64_t size() const { return m_size; }
+    const std::string& name() const { return m_name; }
+
+    EptHierarchy& hierarchy() { return m_hierarchy; }
+    uint64_t count(const Key& key) const
+    {
+        if (m_hierarchy.count(key)) return m_hierarchy.at(key);
```
abellgithub: IMO this is harder to read than if you put the `return` on the next line.

abellgithub: And why can't you just return `m_hierarchy.count(key)` ?

io/private/EptSupport.hpp
```diff
@@ -90,6 +158,127 @@ inline std::array<double, 3> toArray3(const Json::Value& json)
...
+            }
+        });
+
+        for (uint8_t i(0); i < 3; ++i)
+            step(i);
+
+        return key;
+    }
+};
+
```
abellgithub: Do did the Key stuff just move in this file and the diff engine couldn't splice it?  Did something else change?

connormanning: The Key stuff probably just shifted a bit since some utils formerly in the EptReader unnamed namespace got moved here for reuse.  The only thing that changed in the `Key` class was the addition of a string constructor, used by the EptAddonWriter.

io/private/EptSupport.hpp
connormanning: `m_hierarchy.count(key)` would return 1 or 0 - `Addon::count` needs to return the number of points for that key.  I'll rename `Addon::count` to `Addon::points`, which would be less confusing given the internal `std::map` usage.

io/EptAddonWriter.cpp
```diff
@@ -0,0 +1,258 @@
...
+        { "ept_addon", "ept-addon" }
+    };
+
+    std::string getTypeString(Dimension::Type t)
+    {
+        const auto base(Dimension::base(t));
+        if (base == Dimension::BaseType::Signed) return "signed";
+        if (base == Dimension::BaseType::Unsigned) return "unsigned";
+        if (base == Dimension::BaseType::Floating) return "float";
+        throw std::runtime_error("Invalid dimension type");
```
abellgithub: Boo on throwing this.  Either throwError(), or throw something that is caught by caller and then throwError().  If someone gets this, they'll have no clue without a traceback.

io/EptAddonWriter.cpp
connormanning: Fixed at 3736533.

io/private/EptSupport.hpp
```diff
@@ -53,9 +55,75 @@
 #include <pdal/util/Bounds.hpp>
 #include <pdal/util/Utils.hpp>
 
+#include <arbiter/arbiter.hpp>
```
abellgithub: Same comment on throwing things from this class.  If it were me, I'd throw an ept_error or some such and catch/rethrow in the caller.  As an alternative, you can save the stage somewhere and then call throwError().

test/unit/CMakeLists.txt
connormanning: It's not needed - this was a copy-paste from the Reader test, which uses the CropFilter thus linked GDAL.  Removed at fce635d.

io/private/EptSupport.hpp
connormanning: Fixed at fce635d.

io/EptAddonWriter.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+void EptAddonWriter::ready(PointTableRef table)
+{
+    m_arbiter.reset(new arbiter::Arbiter());
+    m_pool.reset(new Pool(std::max<uint64_t>(m_numThreads, 4)));
+
+    MetadataNode meta(table.privateMetadata("ept"));
+
+    if (meta.findChild("info").value().empty())
+        throwError("EPT writer must descend from EPT reader");
```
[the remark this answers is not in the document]
connormanning: > Doesn't seem any different from having no info or keys, which doesn't get checked.

This is checking the `info` though - and since `info` and `keys` are added at the same time by the reader, the presence of `info` is used to indicate the presence of the EPT reader.

> Do you even need to report.

If there's no `info` and thus no `EPT reader` which is required for this writer, what behavior would be preferable to this `throw`?

io/EptAddonWriter.cpp
connormanning: Fixed in 514524e.

io/private/EptSupport.hpp
connormanning: Don't know, these DLL exports were already here - presumably they were needed.

io/private/EptSupport.hpp
connormanning: Remove bounds from constructor at 1629a3d.  Not really a copy, but the `operator[]` is used within Entwine too since it's often convenient to treat coordinates for points and bounds numerically rather than via names.

io/EptAddonWriter.cpp
connormanning: I'll leave as-is for future optimization, if it becomes an issue.  Since only buffers of a single attribute are built at a time and this stage is already non-streaming, it might never be a problem.

pdal/util/Utils.hpp
connormanning: 4e906be

io/private/EptSupport.hpp
connormanning: Error handling for exceptions in EptSupport utilities cleaned up in 4cc05e2.  The EPT reader/addon-writer now catch and re-throw the errors with `throwError`, and an `ept_error` descended from `pdal_error` was created for these exceptions (although this exception type is not currently treated specially).

io/EptAddonWriter.cpp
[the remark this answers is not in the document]
connormanning: Fixed error message in 111d9a0.

io/EptAddonWriter.cpp
```diff
@@ -0,0 +1,252 @@
...
+    m_hierarchyStep = meta.findChild("step").value<uint64_t>();
+    m_info.reset(new EptInfo(info));
+
+    for (const std::string s : keys.getMemberNames())
+    {
+        m_hierarchy[Key(s)] = keys[s].asUInt64();
+    }
+
+    const PointLayout& layout(*table.layout());
+    for (const std::string path : m_addonsArg->getMemberNames())
```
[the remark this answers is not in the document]
connormanning: Fixed in c251cad.

PR #2398, opened 2019-03-08
Allow Zstandard compression level to be configured.

connormanning:


hobu:
failure is the python bug in #2400

PR #2399, opened 2019-03-08
Use base library name to reload symbols

### LINKED ISSUE #2390: Make locating python library flexible
When a python environment is created, it reloads symbols from the python library to deal with an issue with the way python modules are built on Ubuntu.  Currently, the path used to reload the library is that used by at compile-time to find the python library.  The runtime path of the library may differ from the build-time path, so it would make more sense to set the library without a path by default and let the dynamic linker locate the library according to standard rules.

abellgithub:
Close #2390


abellgithub:
Is issue #2390 not clear?  It's the fix where we aren't using the build path to the python library when we load the symbols to deal with the python extension.

hobu:
ok. It will need to be rebased with the Invocation.cpp fixes from this morning.

abellgithub:
That's fine.  It hasn't been committed because of the failing tests, which you're addressing.

PR #2409, opened 2019-03-18
Add subcommand support for kernels

### LINKED ISSUE #2293: Make Kernel::outputHelp protected
Would it be possible to make `Kernel::outputHelp` protected to give kernel implementations access to customize the help message?

I have a Kernel that has another level of commands under it, think a docker command structure like run, ps, kill, etc. nested under pdal.  So I say

`pdal mykernel mycommand [options]`

It would be nice to have access to tweak the help message from `pdal mykernel -h`.  With the current Kernel impl I get pretty close to what is needed by parsing the command line and placing some logic in `addSwitches`, but the one thing that eludes me is the usage string

`usage: pdal mykernel [options]` instead of `usage: pdal mykernel command [options]`

Even though I'm close as is, being able to place all the help related stuff in an overridden `outputHelp` would clean things up a lot and allow for a much more native feel of this plugin.

abellgithub:
Close #2293


pdal/SubcommandKernel.cpp
```diff
@@ -0,0 +1,131 @@
...
+        // parseSimple allows us to scan for the help option without
+        // raising exception about missing arguments and so on.
+        // It also removes consumed args from the arg list, so parse a copy
+        // that will be ignored by parse() below.
+        // Also search for a subcommand.
+        ProgramArgs hargs;
+        const StringList& subs = subcommands();
+        assert(subs.size());
+        hargs.add("help,h", "Print help message", help);
+        hargs.add("subcommand", "Subcommand for application", m_subcommand).                setPositional();
```
connormanning: Formatting.

PR #2414, opened 2019-03-19
Fix up Dockerfile for ubuntu

hobu:


abellgithub:
Is `python` an alias for `python3`?  n this case you don't have python 2 installed, but what happens if you do? Seems confusing.

hobu:
It is an ubuntu-specific dockerfile, which does the python3 convention. We're ditching Python 2 and there's no reason for us to be accidentally supporting it.

Issue #2416, opened 2019-03-20
Share return filtering implementation and provide better error checks/warnings


chambbj:
The PMF, SMRF, and the Returns filters all filter on "first", "intermediate", "last", and "only" returns given that "ReturnNumber" and "NumberOfReturns" dimensions are available and valid.

PMR and SMRF both use the `segmentReturns(...)` function in filters/private/Segmentation.cpp, but the Returns filter implements this separately. These should share the same segmentation logic, which itself needs a unit test (should be able to borrow from the existing `ReturnsFilterTest`).

In the process of moving this to the `Segmentation` namespace, we need to improve our error checks. Currently, PMF and SMRF will throw an error if the segmentation produces a view with no points, which can happen if "ReturnNumber" and "NumberOfReturns" are 0. If all values are 0, then we will go ahead and process, treating this as a situation in which no return information is available, but will emit a warning to the user. If there is a mix of 0's amongst otherwise valid values, we will throw an error indicating that the return information is suspect (i.e., that 0 is undefined for other "ReturnNumber" or "NumberOfReturns"). We will keep the existing behavior of the Returns filter to warn if a returns group was requested and is empty.

chambbj:
Marking this resolved with #2417

PR #2417, opened 2019-03-20
Add warnings/errors for bad return information in PMF and SMRF

chambbj:
There is a much more efficient, or at least prettier way to do this I'm sure, but I think it gets at the problem we discussed. Partially addresses #2416.

### DIFF
--- a/filters/PMFFilter.cpp
+++ b/filters/PMFFilter.cpp
@@ -192,15 +192,8 @@ PointViewSet PMFFilter::run(PointViewPtr input)
     }
     else
     {
-        if (m_args->m_returns.size())
-        {
-            Segmentation::segmentReturns(keptView, firstView, secondView,
-                                         m_args->m_returns);
-        }
-        else
-        {
-            firstView->append(*keptView);
-        }
+        Segmentation::segmentReturns(keptView, firstView, secondView,
+                                     m_args->m_returns);
     }
 
     if (!firstView->size())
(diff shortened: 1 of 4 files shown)


abellgithub:
I'd rather not backport this to 1.8.  1.8 is done, and this isn't, technically, a bug.

I'm not sure about 1.9 because this changes behaivor -- it's not JUST a warning.

hobu:
Did this make it to 1.9-maintenance?

abellgithub:
No. My concern was that it changed behavior at the last minute and I didn't think it was that important. If you have strong feelings about it, we should talk on Thursday.

hobu:
No reason to not merge this into 2.0 now, correct @abellgithub ?

abellgithub:
Let me look.

abellgithub:
The only way that I see that m_args->returns can be empty is if we're missing the ReturnNumber of NumberOfReturns dimensions.  But we're still looping all the points to check the state of zero's, when we already know.  This seems a waste.  In general, Segmentation::segmentReturns is already looping the points and checking nr/rn, so moving this validation there seems like it would make sense.

The loops that append points are duplicated and could be replaced with `firstView->append(*keptView)`, I think.

chambbj:
@abellgithub switching from `appendPoint` to `append` is easy enough. I'd forgotten this was possible.

I'll look into whether or not it makes sense to move the validation to segmentReturns. IIRC, the need to check the actual values wasn't so much to see if they were all zero, but rather to check if any were zero. I'll look again though.

chambbj:
@abellgithub I think this addresses most of what you pointed out. The only thing I'm left wondering at this point is that this PR currently causes the PMF and SMRF filters to throw an error if we come across zeros in the returns, whereas the segmentReturns function would quietly just add them to the "second" PointView (i.e., the one that is ignored for the purposes of the ground segmentation). I think I just need to decide which is the preferable behavior. Even if we allow them to simply be ignored, warning may still be informative as it could be suspicious for these to exist in the first place. Welcome to suggestions here.

PR #2420, opened 2019-03-21
fix #2246 where image describing merge operation was backwards

### LINKED ISSUE #2246: Picture or pipeline is backwards for filters.reprojection
Hello. If I'm not mistaken, this page https://pdal.io/pipeline.html#reprojection-example shows a graphical representation which does not correspond properly to the textual pipeline shown in the example.

It's clear that merge dependes on tags [A2] and [B]. However, according to the pipeline, [A2] is filters.reprojection and [B] is B.las file, which means that the graphical representation inverted the branches.

It should be A.las -> [A2] filters.reprojection in one branch and [B] B.las in the other branch.

hobu:


hobu:
@abellgithub this needs to be backported to `1.9-maintenance` after review

PR #2427, opened 2019-03-27
Replace 'pdal ground' kernel usage of filters.pmf with filters.smrf¬

hobu:
* add 'ignore' and 'returns' option¬
 * reflect filters.smrf options


kernels/GroundKernel.cpp
```diff
@@ -81,6 +81,12 @@ void GroundKernel::addSwitches(ProgramArgs& args)
     args.add("extract", "Extract ground returns?", m_extract);
     args.add("reset", "Reset classifications prior to segmenting?", m_reset);
     args.add("denoise", "Apply statistical outlier removal prior to segmenting?", m_denoise);
+    args.add("returns", "Include last returns?", m_returns,
+             {"last", "only"});
+    args.add("scalar", "Elevation scalar?", m_scalar, 1.25);
```
chambbj: You add the `scalar` arg, but never set it as an option.

abellgithub:
needs docs modified.  Perhaps wait until I merge the 1.9 changes.

PR #2430, opened 2019-03-29
Add filters.shell that allows you to inline shell commands as part of pipelines

hobu:


abellgithub:
This seems very strange.
Imputing JSON from some braces seems bad.  An option would help.
Needs docs.

hobu:
> Imputing JSON from some braces seems bad.

I agree it is weak. It would be nice to have a `Utils::isJSON(std::string)` function that ran stuff through our json library and validated. It was stolen right from our test in pdal/Geometry.cpp.

hobu:
- [ ] Make it possible to shut off `filters.shell` for situations where input may be untrusted
- [ ] Docs

abellgithub:
JSON detection in Geometry is just there to differentiate JSON from WKT.

Don't understand the utility of an environment variable.  Doesn't the person running the program know if they want to run the pipeline as they've written it?

chambbj:
I think the environment variable was prompted by a concern I'd voiced offline about malicious code. Maybe a better approach would be to control at the project level whether or not users wanted to build, and thus expose, the filter at all.

abellgithub:
Maybe just make it an optional plugin?

hobu:
> Don't understand the utility of an environment variable. Doesn't the person running the program know if they want to run the pipeline as they've written it?

There are execution scenarios where programs take a stack of filters via the internet and apply them to data that is constrained for reading and writing. Allowing unsanitized `filters.shell` input via this mechanism would be bad. The error message as constructed is quite descriptive, and in situations that need input sanitization (like over-the-internet filtering ops), malicious actors are way deeper than PDAL if they can set their own environment variables. I think it is prudent to have users explicitly acknowledge that this thing has the potential to be dangerous to use it.

A person running from the command line can conveniently set the `PDAL_ALLOW_SHELL` environment variable, they do have control over their execution environment, and they must acknowledge that they wanted this to happen. As for @chambbj comment about build-time enabling or making it a plugin. That might be helpful, but it is quite likely that it would be built by default in our distributions like conda and so on. If it is a plugin, but practically on all the time anyway, it is much less overhead to take this approach.

abellgithub:
Allowing any execution from scripts coming from unknown sources seems bad, bad, bad.  It makes me even less excited about this than I was before.

PR #2429, opened 2019-03-29
OpenSSL support

nicolas-chaulet:
Let's see if forcing version 1.1 fixes the issues we were having

PR #2436, opened 2019-04-02
Redo of #2429 to link OpenSSL 1.1+ to PDAL

hobu:
Arbiter can use OpenSSL 1.1+. This is an enhancement of #2429 to get it working on Windows too.


nicolas-chaulet:
Great!! Thanks for taking on the windows side of things!

PR #2440, opened 2019-04-02
support streamable in tiledb r/w

normanb:
This adds streamable support for TileDB r/w.


normanb:
The writer seems logical for streamable. However from the docs it isn't clear whether `TileDBReader::read(PointViewPtr view, point_count_t count)` is called when streaming? Should I move the read ahead functionality to `processOne`?

abellgithub:
read() is not called during streaming, only processOne()

normanb:
I have made the change to remove the dependency on `read()` for tiledb streaming reads and changed the test to exercise the read ahead query.

normanb:
@abellgithub I have modified the tests to used `FixedPointTable`, sorry for the delay.

abellgithub:
No worries.  I was about to do it just to clear the PR, but thanks!

PR #2455, opened 2019-04-12
Local feature filter

nicolas-chaulet:
This filter implements the local feature descriptor introduced in 
[Weakly Supervised Segmented-Aided Classification of Urban Scenes From 3D LIDAR Point Clouds, Stéphane Guinard, Loïc Landrieu, 2017 ](https://www.int-arch-photogramm-remote-sens-spatial-inf-sci.net/XLII-1-W1/151/2017/isprs-archives-XLII-1-W1-151-2017.pdf)

It provides the following four descriptors that are based on the covariance matrix:
* linearity - higher for long thin strips
* planarity - higher for planar surfaces
* scattering - higher for complex 3d neighbourhoods
* verticality - higher for thin vertical strips

Happy to write some doc if this is something of interest.


abellgithub:
Thanks.  This looks nice.  We already have some thread pool stuff around somewhere but I don't think we've made it easily accessible.  It would be nice to use that, but perhaps that can be another PR.

chambbj:
I'm excited to take a look at this as I had the same filter in mind (as part of the Landreiu's Superpoint Graph segmentation work (https://arxiv.org/abs/1711.09869)). If I had written it, I may have called it `filters.superpoint`, but I hadn't read the paper you've referenced.

In keeping with the authors, I may advocate for defaulting `k` to 10, rather than 8. Also, with the KD searches in PDAL, we should really bump `m_knn` by one so that we include the query point and it's neighbors (also in keeping with the authors https://github.com/loicland/superpoint_graph/blob/23700fded83d6642eb150e2917216cffbdaae435/partition/ply_c/ply_c.cpp#L282).

I also really like the idea of using the thread pool. I probably have a lot of code that could benefit from this.

nicolas-chaulet:
> I'm excited to take a look at this as I had the same filter in mind (as part of the Landreiu's Superpoint Graph segmentation work (https://arxiv.org/abs/1711.09869)). If I had written it, I may have called it `filters.superpoint`, but I hadn't read the paper you've referenced.

Yes, that's where we got it from, the only thing is that the superpoint graph is a bigger beast (it also includes clustering and graph creation). Addressing your comment, good point.

hobu:
Adding to the bikeshedding discussion, how about `filters.guinard`? A little specificity might help discriminate from the next LocalFeature segmentation approach to come along. That might also compliment a `filters.landrieu` or `filters.superpoint` that comes next. Just a thought.

We will definitely need a doc page that lists the options and describes what they do. It will also need to point at the paper like we've done for PMF and SMRF and such.

Thanks for the awesome contribution! 🎊

nicolas-chaulet:
I am good with `filters.guinard`, I was also thinking that we could make that a general filter that builds on the covariance matrix to extract local feaures. We could merge it with the normal estimation one and the user could choose which local features to extract. Then something like `filters.covariancefeatures` might be more informative? Any though?

chambbj:
If you have some ideas on how best to do it (covariance feature filter), I
think that's great. I've considered it before, but there are just so many
covariance related features in the literature that you could compute that
it gets a little messy.

In general I don't like filters named after authors, though I've done this
too, as they may publish multiple papers we'd like to implement.

hobu:
> In general I don't like filters named after authors, though I've done this too, as they may publish multiple papers we'd like to implement.

True, although naming them after the authors makes sure to credit them basically forever, as that's how people have to refer to them and how they end up getting documented.

loicland:
Hi,

I am Loic Landrieu, author of the superpoint graph and the "weakly supervized ..." papers. Thanks to Nicolas for making me aware of this thread. Very excited to see our work puts to good use!

Proper credit for the dimensionality features (linearity, planarity, scattering) goes to Jérome Demantké, Clément Mallet, Nicolas David and Bruno Vallet and their paper:

DIMENSIONALITY BASED SCALE SELECTION IN 3D LIDAR POINT CLOUDS

The verticality was introduced by us in the "weakly supervized ..." paper.

If I may suggest a name, it would be `filters.dimensionality`.

loicland:
One last thing, verticality measure how vertical is the neighborhood of a point:
vertical pole = 1 (the only dimension is vertical)
vertical plane  = sqrt(2)  (one of the dimension is horizontal)
inclined strip ~ cos(angle), (the vertical component is smaller by cos(angle))
inclined angle ~ 0.4, depending on inclination (the math are a tiny bit more complicated)
anything horizontal = 0 (no vertical component)

So I think

`verticality - higher for thin vertical strips`

could be updated to

`verticality - higher for vertical structures, highest for thin vertical strips`.

My two cents!

chambbj:
Thanks for chiming in Loic!

Not to bikeshed this too much, but I'm not sure I agree (or understand anyway) with naming it `dimensionality`. I guess this is motivated by the name of the paper where they were using it to select scale based on the covariance features, but we are not doing that here. In fact quite the opposite, we are using a fixed k nearest neighbors. It seems too broad for what this is as currently written - a feature generator.

I'm also realizing just now that your published `verticality` doesn't match that of Hackel et al (2016), which is fine, it just makes it a little harder to consider this a general purpose feature generator when the verticality is specific to one particular approach.

How about going back to the idea of calling this `filters.covariancefeatures` (or similar) and essentially having an option that constitutes profiles of covariance features that could be used. One could be based on @loicland while others could be based off of similar works such as Hackel or Weinmann. We could also throw in several options for how to compute the scale (if variable) alongside fixed scales. I get that this is potentially a much bigger undertaking than you were looking to take on, but it could start with just the one profile, exactly as currently written, with room for future expansion.

PS - This is very much an aside, but I think that the cut pursuit portion of superpoint graph could be a separate filter in the future, being able to cut on a variable set of features. I think we'd all be very open to some discussion on how best to integrate directly with PyTorch, TensorFlow, etc. to feed the training/prediction of the ML models.

loicland:
I am perfectly fine with `filters.covariancefeatures`. Actually it might be better indeed.

These were initially designed for scale selection, but have been extensively used for semantic classification afterwards by Weinman, Hackel etc... I agree that it makes it a little bit confusing.

For choosing the optimal neighbourhood size, I think the best way would be this paper:

_Semantic point cloud interpretation based on optimal neighborhoods, relevant features and efficient classifiers, M Weinmann, B Jutzi, S Hinz, C Mallet._

It should be quite easy to code, but makes the all thing much slower. Unsure if it is worth it IMO.

As for adding cut pursuit to compute piecewise constant partition wrt. certain features, I would be very into it as well. Note that cut pursuit is not a deep learning algorithm, but an old fashioned functional optimization solver written in C++ No need to interface with pytorch etc... We have a brand new version that is about to be released, now fully parallel.

Let me know if you actually go through with it, I would be very interested in helping out.

chambbj:
Right, as I understand the typical workflow would be covariance features -> cut pursuit (I'm thinking a new `cluster` dimension generated by the filter to indicate membership) -> external ML work using the fully attributed point cloud.

FWIW, I have the optimal neighborhoods coded in a branch that I could commit. It is quite slow, agreed. Maybe some of the thread pool work could be used to accelerate things a bit.

chrislcs:
Having used and written about these kinds of features before, the earliest use of linearity, planarity and scatter I could find was actually:

_West, Karen F., et al. "Context-driven automated target detection in 3D data." Automatic Target Recognition XIV. Vol. 5426. International Society for Optics and Photonics, 2004._

So I think they may deserve the proper credit. 

They called them structure tensor features, since they are based on the [structure tensor](https://en.wikipedia.org/wiki/Structure_tensor) of a point. So a name like `filters.structuretensor` could also be an option. Though `filters.covariancefeatures` sounds good too.

hobu:
🚲 🏛

chambbj:
@nicolas-chaulet Are you cool with all of this? I _think_ what we are in agreement on the following:

- [ ] `filters.covariancefeatures` is an acceptable filter name
- [ ] Covariance features specific to the @loicland paper(s) can be selected by an option. This can be the default, and for now, only set of features the filter generates.
- [ ] Continue to use a static k nearest neighbors with the possibility of enabling other neighborhood selection approaches in the future.

nicolas-chaulet:
All good!

nicolas-chaulet:
> As for adding cut pursuit to compute piecewise constant partition wrt. certain features, I would be very into it as well. Note that cut pursuit is not a deep learning algorithm, but an old fashioned functional optimization solver written in C++ No need to interface with pytorch etc... We have a brand new version that is about to be released, now fully parallel.

Definitely yes when this new version comes out! It would make it really accessible for others to use and try out.
@chambbj your comments have been addressed.

doc/stages/filters.covariancefeatures.rst
```diff
@@ -0,0 +1,65 @@
+.. _filters.covariancefeatures:
+
+===============================================================================
+filters.covariancefeatures
+===============================================================================
+
+This filter implements various local feature descriptors introduced that are based on the covariance matrix of a
+point's neighborhood. The suer can pick a set of feature descriptors by
```
chambbj: `suer` should be `user`

filters/CovarianceFeaturesFilter.cpp
```diff
@@ -0,0 +1,154 @@
...
+void CovarianceFeaturesFilter::addArgs(ProgramArgs& args)
+{
+    args.add("knn", "k-Nearest neighbors", m_knn, 10);
+    args.add("threads", "Number of threads used to run this filter", m_threads, 1);
+    args.add("feature_set", "Set of features to be computed", m_featureSet, "Dimensionality");
+}
+
+void CovarianceFeaturesFilter::addDimensions(PointLayoutPtr layout)
+{
+    m_linearity = layout->registerOrAssignDim("Linearity", Dimension::Type::Float);
```
chambbj: It can wait until we actually add an alternate `feature_set`, but we won't necessarily always create these four, it could be variable.

chambbj:
Thanks again!

loicland:
Hi,

the parallel version of cut pursuit is [available](https://github.com/1a7r0ch3/parallel-cut-pursuit). It could be used to compute fast geometrically/radiometrically homogeneous partition of large point clouds. This repo is under OpenGL3.0, so if it is an issue this could be discussed with @1a7r0ch3, author of the code.

[Link to paper](https://arxiv.org/pdf/1905.02316), presented at the ICML2019 workhop on graph reasonning.

PR #2456, opened 2019-04-12
Change raster band indices to size_t to eliminate int overflows.

### LINKED ISSUE #2454: Segfault in gdal::Band due to integer overflow
Issue #2448 has not been solved by #2450.  

Upon running the same test set as in #2448 PDAL 1.90 also segfaulted.  The full BT (now that I know there is more than one  BT) [PDAL_btFull_1.9.0(std).txt](https://github.com/PDAL/PDAL/files/3067190/PDAL_btFull_1.9.0.std.txt) showed that wholeRowElts was negative (which was clearly wrong) and probably an overflow.  I added some more local parameters (so I could see other values - having no other idea how to) which reflected m_xBlockSize as xBlockSize etc.  The result was [PDAL_btFull_1.9.0(local param).txt](https://github.com/PDAL/PDAL/files/3067242/PDAL_btFull_1.9.0.local.param.txt).  

As can be seen, when you calculate wholeRowElts = 85100*((25235*1)+0) = 2147498500 (0x80003a04), the result is greater than a signed int, and hence returns the neg value seen in the BT.

Changing wholeRowElts (and partialRowElts for good measure) in GDALUtils.hpp lines 486/7 to 'long' appeared to solve the issue as below:

long wholeRowElts = (long)m_xTotalSize * ((y * m_yBlockSize) + row);
long partialRowElts = (long)m_xBlockSize * x;

I haven't done a change/pull as not sure whether this meets coding requirements and have only undertaken limited testing (though these are local variables only used with a 64-bit address so should not have any other unintended effects).  They may need to be changed to 'long long' or a platform independent type (which suitable type is beyond my knowledge) for cross-platform consistency. This change (combined with #2450) resulted in successful execution of the test command from #2448, as well as the original --nostream pipeline that triggered #2448.

If change agreed, request that this and #2450 be included in 1.8-maintenance branch (as the stable GDAL in Ubuntu 16.04 < 2.2)

abellgithub:
Close #2454

### DIFF
--- a/pdal/GDALUtils.hpp
+++ b/pdal/GDALUtils.hpp
@@ -287,6 +287,7 @@ enum class GDALError
     CantOpen,
     NoData,
     InvalidBand,
+    BadBand,
     NoTransform,
     NotInvertible,
     CantReadBlock,
@@ -299,6 +300,7 @@ enum class GDALError
 };
 
 struct InvalidBand {};
+struct BadBand {};
 struct CantReadBlock {};
 struct CantWriteBlock
 {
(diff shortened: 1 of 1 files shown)


abellgithub:
You mean to avoid a compiler warning?

connormanning:
No, I just meant that avoiding changing the signedness might have been nice in general.  I also missed the fact that the casts were of size as well as sign, so those need to be there regardless of the sign change.  Disregard the casting part of my original comment.

abellgithub:
There are checks to make sure that the ints are non-negative before the casts:

https://github.com/PDAL/PDAL/blob/59f67ecb6c1d4c3cc718142623d82d088c3ad28d/pdal/GDALUtils.hpp#L368-L370

PR #2470, opened 2019-04-19
Replace jsoncpp with nlohmann/json

### LINKED ISSUE #2343: Replace jsoncpp
We don't like jsoncpp anymore.  Use something else.

abellgithub:
Close #2343


abellgithub:
There's a lot here.  I'm not terribly confident that there aren't issues that depend on input.  Some of the data extraction is less hardened than with jsoncpp, but what you get out is going to be good, rather than an integer implicitly cast to a string or something garbagey like that.

filters/StatsFilter.cpp
```diff
@@ -295,12 +292,9 @@ void StatsFilter::extractMetadata(PointTableRef table)
...
             dddbox.add(epsg_4326_box);
 
-            Json::Reader jsonReader;
-            Json::Value json;
-            jsonReader.parse(p.json(), json);
-
+            json = p.json();
             MetadataNode ddboundary = dddbox.addWithType("boundary",
-                json.toStyledString(), "json", "GeoJSON boundary");
+                json.get<std::string>(), "json", "GeoJSON boundary");
```
connormanning: And here.

io/private/EptSupport.hpp
```diff
@@ -271,29 +263,29 @@ class PDAL_DLL EptInfo
...
-    EptInfo(Json::Value info) : m_info(info)
+    EptInfo(const NL::json& info) : m_info(info)
     {
         m_bounds = toBox3d(m_info["bounds"]);
-        m_points = m_info["points"].asUInt64();
-        m_span = m_info["span"].asUInt64();
-        m_srs = m_info["srs"]["wkt"].asString();
+        m_points = m_info["points"].get<uint64_t>();
+        m_span = m_info["span"].get<uint64_t>();
+        m_srs = m_info["srs"]["wkt"].get<std::string>();
```
connormanning: This is UB if `srs` does not have a `wkt` member, which is allowed by EPT if there is no associated SRS.  See first point of [further notes](https://github.com/nlohmann/json#further-notes).

Recommend something like
```
if (m_info.contains("srs"))
    m_srs = m_info.at("srs").value("wkt", "");
```

io/private/EptSupport.hpp
```diff
@@ -271,29 +263,29 @@ class PDAL_DLL EptInfo
...
 
         if (m_srs.empty())
```
connormanning: Similar here, maybe this check should be `if (m_srs.empty() && m_info.contains("srs"))`.

io/EptReader.cpp
```diff
@@ -290,83 +283,76 @@ QuickInfo EptReader::inspect()
...
-    const Json::Value& schema(m_info->schema());
+    const NL::json schema {m_info->schema()};
     m_remoteLayout.reset(new FixedPointLayout());
 
-    for (Json::ArrayIndex i(0); i < schema.size(); ++i)
+    for (auto& el : schema)
     {
-        const Json::Value& dim(schema[i]);
-        const std::string name(dim["name"].asString());
+        const std::string name(el["name"].get<std::string>());
```
sunapi386: https://github.com/PDAL/PDAL/issues/2495

For me, this line causes 
```
unknown file: Failure
C++ exception with description "[json.exception.type_error.305] cannot use operator[] with a string argument with array" thrown in the test body.
```

sunapi386: Adding the `&` seems to fix it. 
E.g. `const NL::json &schema {m_info->schema()};`

abellgithub: I don't understand why that would change anything.  Using dump() instead of get() makes sense, but I don't see how holding the reference rather than using a copy would change behavior.  I also don't understand why tests are failing for you but not on our platforms.  What compiler/OS version do you have?

sunapi386: Ubuntu 18.04. With holding the reference, when I debug with Clion, I can see the string values. Otherwise I error out. 

```
jason@glassbox ~/w/e/build (master)> uname -a
Linux glassbox 4.15.0-48-generic #51-Ubuntu SMP Wed Apr 3 08:28:49 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

jason@glassbox ~/w/e/build (master)> cmake --version
cmake version 3.10.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
jason@glassbox ~/w/e/build (master)> gcc --version
gcc (Ubuntu 7.4.0-1ubuntu1~18.04) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

PR #2473, opened 2019-04-24
Split Bounds class from Box classes and out of Utils

abellgithub:


hobu:
This change is going to break a lot of external code for essentially no to-the-user value. Do we have a problem with keeping a `pdal/util/Bounds.hpp` that simply includes `pdal/Bounds.hpp` for the user?

btw, I think this is the second time Bounds has moved 😦

abellgithub:
I think you're overstating the impact of this.  I can change things back, but I don't think a competent developer should have any issue with this sort of change.  If they are using Bbox, they simply change the name of the header file.  If they are using Bounds, the just need to change the namespace.

The value is that the Bounds class will support spatial references, which was the point of this change.  It can be done other ways, though.  It's just time.

PR #2475, opened 2019-04-24
Make sure temp files are unique.

abellgithub:
This avoids a lock and the random number stuff.


abellgithub:
It's probably best to truncate the temp file when opened in case a crash occurred and it wasn't deleted.  I see "binaryTruncMode", but can't tell if it's used when creating a local temp file.

PR #2482, opened 2019-04-30
add util/Bounds.hpp -> pdal/Bounds.hpp to ease 2.0 transition

hobu:


abellgithub:
This breaks the Utils independence, which is why Bounds was moved in the first place.  If this is a big deal to you, I'll change things back and do things another way.

PR #2485, opened 2019-05-01
Update filter writing docs to reflect changes to the plugin interface

chambbj:


chambbj:
It builds just fine, but no, it doesn't load as a plugin. I was trying to make the implementation match the way the narrative described (static, but easily switchable to shared). I guess I'll switch it back and massage the words, although this document could/should serve as a blueprint also for creating true static plugins.

PR #2488, opened 2019-05-02
Support bounds with spatial reference for EPT queries

### LINKED ISSUE #2295: Add spatial reference support to bounds object
This would allow bounds fetching from ept/greyhound via any supported srs rather than requiring native.

abellgithub:
This code adds the capability to add a "/" after a bounds box specification, which can then be followed by a spatial reference string.  The EPT reader will transform the bounds box from its SRS to the SRS of the EPT data.  If the EPT data has no SRS, an exception will be thrown.

Close #2295


io/EptReader.cpp
```diff
@@ -57,11 +60,64 @@ namespace
     };
 
     const std::string addonFilename { "ept-addon.json" };
+
+    Dimension::Type getType(const NL::json& dim)
+    {
+        if (dim.contains("scale"))
+            return Dimension::Type::Double;
+
+        const std::string typestring(dim["type"].get<std::string>());
```
connormanning: Using `.at()` would be safer here than `[]`.

abellgithub: How so?  `.at()` throws.

connormanning: If the key doesn't exist, `[]` potentially segfaults or reads garbage.  An exception seems safer than UB.

abellgithub: I suppose we always *should* have a type, so it's an error if it isn't present, right?

connormanning: Correct.  But for bad data where it's trivial to create an exception rather than a segfault, it seems preferable to do so.

abellgithub: Agreed.  Just wondering if I should do something better.

PR #2493, opened 2019-05-03
Compute bounds of fixed point view and apply this offset to both conversions

chambbj:


hobu:
@abellgithub we need this one in 1.9.1

Issue #2495, opened 2019-05-04
EPT Reader json parsing seems broken


sunapi386:
Clean install of Ubuntu, Anaconda (followed the forge instructions), cloned pdal, built (see build.log) and cannot use EPT reader following example from https://pdal.io/stages/readers.ept.html.
[build.log](https://github.com/PDAL/PDAL/files/3144147/build.log)

```
jason@glassbox /m/j/S/pc> pdal pipeline liberty.json --verbose 8
(PDAL Debug) Debugging...
(pdal pipeline readers.ept Debug) GDAL debug: OGRSpatialReference::Validate: No root pointer.
(pdal pipeline readers.ept Debug) GDAL debug: OGRSpatialReference::Validate: No root pointer.
(pdal pipeline readers.ept Debug) GDAL debug: OGRSpatialReference::Validate: No root pointer.
(pdal pipeline readers.ept Debug) Endpoint: http://na.entwine.io/nyc/
pdal: /home/jason/workspace/PDAL/vendor/nlohmann/nlohmann/json.hpp:16155: const value_type& nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer>::operator[](T*) const [with T = const char; ObjectType = std::map; ArrayType = std::vector; StringType = std::__cxx11::basic_string<char>; BooleanType = bool; 
[build.log](https://github.com/PDAL/PDAL/files/3144144/build.log)
NumberIntegerType = long int; NumberUnsignedType = long unsigned int; NumberFloatType = double; AllocatorType = std::allocator; JSONSerializer = nlohmann::adl_serializer; nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer>::const_reference = const nlohmann::basic_json<>&; nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer>::value_type = nlohmann::basic_json<>]: Assertion `m_value.object->find(key) != m_value.object->end()' failed.
fish: “pdal pipeline liberty.json --ve…” terminated by signal SIGABRT (Abort)
jason@glassbox /m/j/S/pc [134]> 
```

About my PDAL version
```
[...]

PR #2496, opened 2019-05-04
Use a reference & in EPT Reader

sunapi386:
https://github.com/PDAL/PDAL/issues/2495


sunapi386:
Is iterating by reference frowned upon? I don't understand why using reference poses a problem, because there seems to be no need need to make a copy of the JSON schema object.

abellgithub:
No.  This isn't what I'm saying.  I'm going to close this and just fix it.

Issue #2468, opened 2019-04-18
Using ground TIN for calculation of Z values


smgrz:
Hi,
it would be nice to have an additional option for filters.hag, to choose whether NN (Nearest Neighbor) or TIN (TIN - Triangulated Irregular Network like in LAStools http://lastools.org/download/lasheight_README.txt) is used for calculation of Z values.
In the current implementation default is NN, there are no further options, am I right? Using NN especially in hilly areas provides a jagged Z.

chambbj:
That's correct, the current implementation is quite limited. I'm sure we would happily accept a PR to expose additional options you may need, if you'd be willing to provide. Another thing we've considered is to compute against externally provided DEMs. Furthermore, I believe @gadomski also has an idea or two in the works that would make this all more robust.

gadomski:
Yeah, I've got some [work in progress](https://github.com/gadomski/PDAL/tree/hag-improved) that improves (at least hopefully) `filters.hag`. Specifically, my improved filter

1. uses multiple nearby points instead of one, interpolating a ground height using inverse distance weighting;
2. checks for "extrapolation", aka when the bounding box of the nearest neighbors does not include the sample point; and
3. includes a "maximum distance" parameter that excludes a non-ground point from HAG calculation if the nearest neighbors are too far away.

The end effect is a `filters.hag` that behaves somewhat like an "adaptive DEM" calculation by (more or less) calculating a grid cell height underneath each non-ground point.

Using a TIN would be another valid approach, but I didn't go with that simply because PDAL doesn't currently have a built-in 3D TIN, so improving the HAG as I did was the lowest-impact solution I could think of. I'd say having both approaches wouldn't be bad, if someone wanted to add the TIN one. I'm going to just implement the improved HAG for now.

abellgithub:
What does a good TIN look like?  We have some surface reconstruction algorithms, but I'm not sure that's what we want and they're really slow.  TINning trees and such seems not worthwhile.

smgrz:
@gadomski it sounds very good to have some additional interpolation and weighting for Nearest Neighbours (NN).

I wrote a python script to calculate (interpolate) Z based on TIN (created from ground points). 
The script is maybe not well readable, because I tried to write it more efficiently. For large point clouds it is not memory efficient! 
Tested with: windows 10, python 3.7, numpy 1.16.2, scipy 1.2.1, sklearn 0.20.3, pdal 1.9.0

[...]

PR #2520, opened 2019-05-16
[WiP] Add Rudimentary TIN-like Behavior to HAG Filter

jamesmcclain:
Connects https://github.com/PDAL/PDAL/issues/2468

I am new to this community, so I do not know what the expectations are in terms of speed, robustness, and coding style.  For that reason, this PR takes a straight-forward, no-additional-dependencies approach.  The branch is built on top of a rebased version of @gadomski 's [nearest neighbor branch](https://github.com/gadomski/PDAL/tree/hag-improved).

Rather than explicitly constructing a TIN, I have taken the approach of locally calculating the Delaunay fan near each point, which should be equivalent (albeit with a parameter, the neighborhood size).  I assume that the projection (x, y, z) -> (x, y) is one-to-one and use that assumption to project then triangulate neighborhoods of ground points.  Using that assumption a second time (or at least consistent with it), the height of each point is calculated by projecting it along z onto the surface.

I am eager to receive any comments that may be forthcoming.

I marked this as WiP; I do not know all that might be missing, but I imagine that tests are, at minimum.

Input:
```json
[
    "autzen.laz",
    {
        "type":"filters.decimation",
        "step": 20
    },
    {
        "type":"filters.hag",
	"count": 30,
	"delaunay_fans": true
    },
    {
      "type":"writers.gdal",
      "filename":"autzen-hag-tin.tif",
      "dimension":"HeightAboveGround",
      "data_type":"double",
      "output_type":"mean",
      "resolution":10.0
    }
]
```

Output:

Screenshot from 2019-05-15 21-15-04

### DIFF
--- a/filters/HAGFilter.cpp
+++ b/filters/HAGFilter.cpp
@@ -56,6 +56,15 @@ std::string HAGFilter::getName() const
     return s_info.name;
 }
 
+void HAGFilter::addArgs(ProgramArgs& args)
+{
+    args.add("count", "The number of points to fetch to determine the ground point [default: 1].",
+            m_count, point_count_t(1));
+    args.add("max_distance", "The maximum distance to the farthest nearest neighbor before the height above ground is not calculated [default: 0 (disabled)]", m_max_distance);
+    args.add("allow_extrapolation", "If true and count > 1, allow extrapolation [default: true].",
+            m_allow_extrapolation, true);
+}
+
 void HAGFilter::addDimensions(PointLayoutPtr layout)
 {
     layout->registerDim(Dimension::Id::HeightAboveGround);
@@ -103,9 +112,56 @@ void HAGFilter::filter(PointView& view)
     for (PointId i = 0; i < ngView->size(); ++i)
     {
         PointRef point = ngView->point(i);
+        double x0 = point.getFieldAs<double>(Dimension::Id::X);
+        double y0 = point.getFieldAs<double>(Dimension::Id::Y);
         double z0 = point.getFieldAs<double>(Dimension::Id::Z);
-        auto ids = kdi.neighbors(point, 1);
-        double z1 = gView->getFieldAs<double>(Dimension::Id::Z, ids[0]);
+        auto ids = kdi.neighbors(point, m_count);
+        double z1 = std::numeric_limits<double>::quiet_NaN();
+        assert(ids.size() > 0);
+        if (ids.size() == 1) {
+            z1 = gView->getFieldAs<double>(Dimension::Id::Z, ids[0]);
+        } else {
+            auto min_x = std::numeric_limits<double>::max();
+            auto max_x = std::numeric_limits<double>::min();
+            auto min_y = std::numeric_limits<double>::max();
+            auto max_y = std::numeric_limits<double>::min();
+            double weights = 0;
+            double z_accumulator = 0;
+            bool exact_match = false;
+            for (unsigned long j = 0; j < ids.size(); ++j) {
+                auto x = gView->getFieldAs<double>(Dimension::Id::X, ids[j]);
+                auto y = gView->getFieldAs<double>(Dimension::Id::Y, ids[j]);
+                auto z = gView->getFieldAs<double>(Dimension::Id::Z, ids[j]);
+                auto distance = std::sqrt(std::pow(x - x0, 2) + std::pow(y - y0, 2));
+                if (distance == 0) {
+                    exact_match = true;
+                    z1 = z;
+                    break;
+                }
+                if (m_max_distance > 0 && distance > m_max_distance) {
+                    break;
+                } else {
+                    auto weight = 1 / std::pow(distance, 2);
+                    weights += weight;
+                    z_accumulator += weight * z;
+                }
+                if (x > max_x) {
+                    max_x = x;
+                }
+                if (x < min_x) {
+                    min_x = x;
+                }
+                if (y > max_y) {
+                    max_y = y;
+                }
+                if (y < min_y) {
+                    min_y = y;
+                }
+            }
+            if (exact_match || m_allow_extrapolation || (x0 > min_x && x0 < max_x && y0 > min_y && y0 < max_y)) {
+                z1 = z_accumulator / weights;
+            }
+        }
         view.setField(Dimension::Id::HeightAboveGround, ngIdx[i], z0 - z1);
     }
 
--- a/filters/HAGFilter.cpp
+++ b/filters/HAGFilter.cpp
@@ -36,6 +36,8 @@
 
 #include <pdal/KDIndex.hpp>
 
+#include "private/delaunator.hpp"
+
 #include <string>
 #include <vector>
 
--- a/filters/HAGFilter.cpp
+++ b/filters/HAGFilter.cpp
@@ -196,7 +196,7 @@ void HAGFilter::filter(PointView& view)
         } else if (m_delaunay == true) { // Delaunay-based interpolation
             auto neighbors = std::vector<double>();
 
-            for(unsigned int j = 0; j < ids.size(); ++j)
+            for(size_t j = 0; j < ids.size(); ++j)
             {
                 neighbors.push_back(gView->getFieldAs<double>(Dimension::Id::X, ids[j]));
                 neighbors.push_back(gView->getFieldAs<double>(Dimension::Id::Y, ids[j]));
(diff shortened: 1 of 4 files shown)


chambbj:
@jamesmcclain Thanks for the contribution. I have yet to test this locally, but the results look very promising.

It would probably be good to have @plimkilde and @abellgithub to review the delaunay changes, as I'm less familiar with that portion of the codebase.

filters/HAGFilter.cpp
```diff
@@ -56,6 +59,17 @@ std::string HAGFilter::getName() const
...
 }
 
+void HAGFilter::addArgs(ProgramArgs& args)
+{
+    args.add("count", "The number of points to fetch to determine the ground point [default: 1].",
+            m_count, point_count_t(1));
+    args.add("max_distance", "The maximum distance to the farthest nearest neighbor before the height above ground is not calculated [default: 0 (disabled)]", m_max_distance);
+    args.add("allow_extrapolation", "If true and count > 1, allow extrapolation [default: true].",
+            m_allow_extrapolation, true);
+    args.add("delaunay_fans", "Construct local Delaunay fans and infer heights from them [default: false].",
```
chambbj: Very minor detail, but I wonder if just `delaunay` may be easier to remember (and less to type) than `delaunay_fans`.

filters/private/delaunator.hpp
```diff
@@ -261,15 +238,15 @@ Delaunator::Delaunator(std::vector<double> const& in_coords)
     ids.reserve(n);
 
     for (std::size_t i = 0; i < n; i++) {
-        const double x = coords[2 * i];
-        const double y = coords[2 * i + 1];
+      const double x = coords[2 * i];
```
chambbj: I wouldn't be opposed to a separate PR that cleans up any inconsistent code formatting, but the whitespace change in this PR detracts from the substantive changes you've made. It may be a little more clear, if you were to back these out.

plimkilde:
The `delaunator.hpp` header is currently, as far as practically possible, copied verbatim from [delaunator-cpp](https://github.com/delfrrr/delaunator-cpp). This way, it should be simple to merge fixes and optimizations from upstream. I have a feeling that this PR is effectively creating a fork of delaunator-cpp (which is fine) but at least that is a decision that should be made consciously.

If the changes to `delaunator.hpp` are generally useful, I think you should consider submitting them as a PR to the delaunator-cpp repo. If they are too specialized for that, maybe we should keep two versions of the header around (one for the Delaunay filter, one for HAG)?

Overall this looks like an exciting contribution, I'm just worried a bit about introducing less battle-tested code to the Delaunay computation. PDAL's unit testing for the Delaunay filter is currently fairly minimal, the assumption being that stress tests for the computation have been made elsewhere. For a modified `delaunator`, we would probably need more extensive test coverage in order to avoid regressions.

filters/HAGFilter.cpp
```diff
@@ -103,9 +137,89 @@ void HAGFilter::filter(PointView& view)
...
+            z1 = gView->getFieldAs<double>(Dimension::Id::Z, ids[0]);
+        } else if (m_delaunay_fans == false) { // Nearest-Neighbor-based interpolation
+            auto min_x = std::numeric_limits<double>::max();
+            auto max_x = std::numeric_limits<double>::min();
+            auto min_y = std::numeric_limits<double>::max();
+            auto max_y = std::numeric_limits<double>::min();
+            double weights = 0;
+            double z_accumulator = 0;
+            bool exact_match = false;
+            for (unsigned long j = 0; j < ids.size(); ++j) {
```
plimkilde: Why not use `size_t` for this and the other iterator variables?

filters/private/delaunator.hpp
jamesmcclain: Certainly, will address

filters/HAGFilter.cpp
jamesmcclain: Okay, will change

jamesmcclain:
> The delaunator.hpp header is currently, as far as practically possible, copied verbatim from delaunator-cpp. This way, it should be simple to merge fixes and optimizations from upstream. I have a feeling that this PR is effectively creating a fork of delaunator-cpp (which is fine, but at least that is a decision that should be made consciously).

Yes, that is a fine point

> If the changes to delaunator.hpp are generally useful, I think you should consider submitting them as a PR to the delaunator-cpp repo. If they are too specialized for that, maybe we should keep two versions of the header around (one for the Delaunay filter, one for HAG)?

The changes here were just motivated by some linking problems that I encountered.  The non-inlined
code within the header lead to "multiple definition" errors when the header was included in a second place.  If there is a cleaner way to solve this, I am more than happy to use it.

> Overall this looks like an exciting contribution, I'm just worried a bit about introducing less battle-tested code to the Delaunay computation. PDAL's unit testing for the Delaunay filter is currently fairly minimal, the assumption being that stress tests for the computation have been made elsewhere. For a modified delaunator, we would probably need more extensive test coverage in order to avoid regressions.

Yes, that makes sense.  My motivation for the changes (which amount to adding the keyword `inline` in a few places and moving the body of the constructor) was pure expedience -- they are not conceptually essential at all.

filters/HAGFilter.cpp
jamesmcclain: Certainly, will change.

filters/private/delaunator.hpp
jamesmcclain: Apropos of that, I have some general questions about the mechanics of the project (e.g. is there a mechanical way to achieve the desired style [like some set of flags to some pretty printer]?).  Where is the best place to ask those types of questions?

chambbj: Some of us will use either astyle or clang-format (I prefer the latter), both of which have style files in the root directory, but there is no standard. As much as anything, we generally like to keep a file internally consistent. Running either tool on this file _might_ introduce a ton of formatting changes. When that's done, it's typically performed separately from any functional changes for clarity.

The mailing list or Gitter would be reasonable places to ask such questions.

jamesmcclain: Okay, many thanks

filters/HAGFilter.cpp
jamesmcclain: Okay, addressed

filters/private/delaunator.hpp
jamesmcclain: Okay, I have reverted all changes to `delaunator.hpp` and instead re-declared the public interface in `filters/private/delaunator-decl.hpp`.  This is somewhat unattractive, but seems like it might be the best option for now?

filters/HAGFilter.cpp
jamesmcclain: Okay, have changed

jamesmcclain:
I think that all review comments (at time of writing) have been addressed.  The most notable change is that I have reverted all chanages to `filters/private/delaunator.hpp` and re-declared the public interface in `filters/private/delaunator-decl.hpp`.

abellgithub:
I'll take a look at this on Monday.  I really appreciate the work.

plimkilde:
> The changes here were just motivated by some linking problems that I encountered. The non-inlined
> code within the header lead to "multiple definition" errors when the header was included in a second place. If there is a cleaner way to solve this, I am more than happy to use it.
> 
Your linking problems sound like a compelling reason to split the upstream delaunator-cpp into separate header and source files. I see Andrew already created [an issue](https://github.com/delfrrr/delaunator-cpp/issues/17) for that.

filters/HAGFilter.cpp
```diff
@@ -99,13 +139,94 @@ void HAGFilter::filter(PointView& view)
...
+                double bx = gView->getFieldAs<double>(Dimension::Id::X, ids[bi]);
+                double by = gView->getFieldAs<double>(Dimension::Id::Y, ids[bi]);
+                double bz = gView->getFieldAs<double>(Dimension::Id::Z, ids[bi]);
+                double cx = gView->getFieldAs<double>(Dimension::Id::X, ids[ci]);
+                double cy = gView->getFieldAs<double>(Dimension::Id::Y, ids[ci]);
+                double cz = gView->getFieldAs<double>(Dimension::Id::Z, ids[ci]);
+
+                z1 = std::min(z1, distance_along_z(ax, ay, az, bx, by, bz, cx, cy, cz, x0, y0));
+            }
+            if (z1 == std::numeric_limits<double>::infinity())
```
abellgithub: I don't understand this.  It seems we're initializing z1 with infinity.  Then I would think the minimum of infinity and anything else would be anything else.  If I'm not misunderstanding, simpler logic would just be to assign to z1 as:

```
std::min(
    gView->getFieldAs<double>(Dimension::Id::Z, ids[0]),
    distance_along_z(ax, ay, az, bx, by, bz, cx, cy, cz, x0, y0)
);

PR #2528, opened 2019-05-17
Show if stage is streamable in pdal --options

tmontaigu:
Add the code to ouput wether a stage supports streaming in `pdal --options`. (#2480)
I did not change the outputed json by the `--showjson` switch.

And i'm not too sure about the placement of the text as it could look like an option 

streamable

### DIFF
--- a/apps/pdal.cpp
+++ b/apps/pdal.cpp
@@ -164,13 +164,15 @@ void App::outputDrivers()
         StageExtensions& extensions = PluginManager<Stage>::extensions();
         for (auto name : stages)
         {
+            Stage *s = f.createStage(name);
             std::string description = PluginManager<Stage>::description(name);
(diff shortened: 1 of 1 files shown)


abellgithub:
I agree that the placement could be confusing.  How about:

```
filters.whatever
Streamable: [yes/no]
Documentations: http://pdal.io/stages/filters.whatever.html
------------------------------------------------------------
```

hobu:
As long as it continues to be valid reStructured text so it can be conveniently be piped to HTML/PDF, I don't care.

abellgithub:
Would you be interested in moving this function to `pdal --drivers --showjson` instead of `pdal --options`?  See #2573

hobu:
That would be fine.

hobu:
Did this get merged? If so, which variant got merged?

Issue #2529, opened 2019-05-17
E57: Expose InvalidState as PDAL Omit dimension 


klassenjs:
Some E57 files have `cartesianInvalidState` or `sphericalInvalidState` dimensions which mark "invalid" points.  Exposing these as the Omit dimension allows them to be filtered out later if the user doesn't want them.

Currently, CloudCompare ignores these points while PDAL reads all the points without a way to distinguish the "invalid" ones.

PR #2530, opened 2019-05-17
E57 fixes

klassenjs:
* Map E57 InvalidState Dimension to PDAL Omit Dimension
* Create cursory E57 documentation pages.

ref: #2529

### DIFF
--- a/doc/stages/readers.e57.rst
+++ b/doc/stages/readers.e57.rst
@@ -1,0 +1,77 @@
+.. _readers.e57:
+
+readers.e57
+===========
+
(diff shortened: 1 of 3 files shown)


nicolas-chaulet:
Nice one! The support for normal information would not be totally functional though, for example if a point cloud as a transformation matrix set in the e57 file then the normals would need to be transformed accordingly.

nicolas-chaulet:
@abellgithub Not sure it is wise to merge as is. After this, normals will be read from the e57 files but they will be wrong whenever the e57 file contains pose information (rotated scan for example).

PR #2531, opened 2019-05-19
Pipeline reader parses objects properly

nicolas-chaulet:


abellgithub:
What was the actual issue here?  The test doesn't address parsing directly.

nicolas-chaulet:
I was not sure how to test it. But basically if a stage contains options that are objects then the parser breaks (tries to treat them as strings). The change I introduced fixes that I believe.

PR #2554, opened 2019-06-04
Keep processing rows/cols if we hit raster edge

### LINKED ISSUE #2545: Interpolating points to grid with writers.gdal IDW produced unexpected results
While using writers.gdal to interpolate a raster from four points, we encountered some unexpected behaviour. Together, these issues cause IDW in writers.gdal to function very differently to IDW implemented in other geospatial software (e.g. ArcGIS, QGIS, Surfer). 

1. Values were only returned for cells in columns or rows containing at least one point, resulting in holes in the raster. We expected a continuous surface across the entire area, bounded by Xmax/Xmin and Ymax/Ymin. This behaviour appears to be independent of the radius argument, as this parameter was set much higher (~10) than the actual distance between points (~1). This issue probably only appears when resolution is lower than point spacing.

2. Interpolation only actually happened for cells lying on an intersection of rows or columns containing two or more points. In columns or rows containing a single point, all cells have the exactly the same value as the point. This is not how spatial interpolation should work, and results in artificial striping of the resultant raster.

3. Implementations of IDW normally take a ‘power’ argument, to control the distance weighting. There does not appear to be an argument for power in writers.gdal. This restricts the way in which this points-to-grid function can be employed and tuned to specific applications.

_Example output raster, with four point locations shown as black dots._
screen

_Code used:_
(tested on Windows, PDAL v1.9.0)

dtm_pipeline = {
            "pipeline": [
                {
                    "type": "readers.text",
                    "filename": input_file,
                    "skip": 1,
                    "spatialreference": EPSG
                },
                {
                    "type": "writers.gdal",
                    "dimension": "Z",
                    "resolution": 0.1, 
                    "radius": 10, 
                    "output_type": "idw",
                    "gdaldriver": "GTiff",
                    "filename": dtm_outfile
                }
            ]
        }

abellgithub:
Close #2545


abellgithub:
Thanks for looking @gadomski

PR #2555, opened 2019-06-04
Don't attempt triangulation with insufficient points.

### LINKED ISSUE #2513: PDAL crashes on empty text file input
PDAL segfaults when an empty file is passed to `readers.text`.

abellgithub:
Close #2513


plimkilde:
I've now tested this on Ubuntu 18.04.2 LTS. It does fix the segfault and generates a warning instead.

However, I'm on CMake 3.10 and had to comment out CMake policy `CMP0075` (as introduced with #2548) in order to be able to build.

Perhaps we should also add a test to exercise the "too few points in triangulation" case?

plimkilde:
Seems 0f5a56e fixed the CMake issue for me, I can now build the `issue-2513` branch if I rebase it to a recent `master` commit.

PR #2559, opened 2019-06-06
Actually escape characters.

### LINKED ISSUE #2558: PDAL JSON output is not valid
JSON output from PDAL info can be invalid.  Specifically I've noticed they can't be consumed by either [jq](https://stedolan.github.io/jq/) through the CLI or PDAL's own [JSON library](https://github.com/nlohmann/json) through the C++ API.  To reproduce, assuming `jq` is installed:

```
$ pdal info --metadata https://entwine.io/data/red-rocks.laz | jq
parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 62, column 35
```

At a glance I think it's from `SpatialReference::getWKT()` ending with control characters.  This is the case on both `master` and `1.9-maintenance` branches.  I think there was some sanitization code that was removed recently because no one (including me) could remember why it was there - maybe this was it.

abellgithub:
Close #2558


connormanning:
This looks good to me and solves the issue I was seeing.  I get some of these warnings, if we want to get rid of it we could add some braces:
```
[1/320] Building CXX object dimbuilder/CMakeFiles/dimbuilder.dir/__/pdal/util/Utils.cpp.o
../pdal/util/Utils.cpp:425:7: warning: suggest braces around initialization of subobject [-Wmissing-braces]
    { "\\u0000", "\\u0001", "\\u0002", "\\u0003", "\\u0004",
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
with
```
Apple LLVM version 9.0.0 (clang-900.0.39.2)
Target: x86_64-apple-darwin16.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

PR #2562, opened 2019-06-10
Ept reader parallel

nicolas-chaulet:
EPT reader leverages mutliple threads better. It actually scales linearly on local indices. Credit to @connormanning, I only tested it...


io/EptReader.cpp
```diff
@@ -613,53 +626,69 @@ uint64_t EptReader::readLaszip(PointView& dst, const Key& key,
...
+    if (views.size() != 1)
     {
-        PointRef pr(*src);
-        for (uint64_t i(0); i < src->size(); ++i)
-        {
-            pr.setPointId(i);
-            process(dst, pr, nodeId, pointId);
-            ++pointId;
-        }
+        throwError("Unexpected view count: " + key.toString());
```
abellgithub: This error message doesn't explain what someone did wrong or how to correct.

nicolas-chaulet: I updated it. It is really a fail safe check, I don't think it would ever happen and I don't know how such a thing could happen.

nicolas-chaulet:
I don't understand why windows tests are failing, they fail in ept reader and writer but nothing changed in the writer...

abellgithub:
I would guess the writer test uses the reader.

connormanning:
I'll take a look into this test failure and your Zstandard PR today.

nicolas-chaulet:
I am actually wondering if the other simpler implementation  #2566  might be a simpler candidate for supporting streaming and easier to maintain on the long. Let me know what you think.

connormanning:
If this PR isn't noticeably better to you than #2566, then I agree with you to close this in favor of that one.  It's less intrusive and is an obvious improvement to the current code.

connormanning:
Closing in favor of #2566.

nicolas-chaulet:
> If this PR isn't noticeably better to you than #2566, then I agree with you to close this in favor of that one. It's less intrusive and is an obvious improvement to the current code.

I think it is better but introduces a more complex logic which could become troublesome moving forward. Feel free to merge #2566  and thanks for taking a look!

PR #2563, opened 2019-06-10
Initial commit of native ICP

chambbj:
Here is my initial port of ICP from PCL to native PDAL. There are certainly some variants and bells and whistles that did not come along for the ride, but I think this captures most of what the prior filter provided. Still needs a little documentation love and parameter naming work.

It also contains some changes to PointView and EigenUtils. Notably, EigenUtils now has demeanPointView and transform methods to support filters like ICP that need to do such things. It also moves calculateBounds from PointView to EigenUtils. This follows from a discussion with @abellgithub. In time, EigenUtils will be renamed to something else.


hobu:
> It also moves calculateBounds from PointView to EigenUtils

👍

pdal/PointView.hpp
```diff
@@ -182,19 +182,6 @@ class PDAL_DLL PointView : public PointContainer
         getFieldInternal(dim, idx, buf);
     }
 
-    /*! @return a cumulated bounds of all points in the PointView.
```
abellgithub: This is API change.

hobu: What is the case to remove PointView::calculateBounds()? Can we keep the method and simply forward down to eigenutils or wherever?

chambbj: The case is only, when I wanted to add demean and transform to the PointView, it was suggested that calculateBounds could probably move out. I don't care either way.

abellgithub:
I think we should redirect and give people a couple of releases' notice.

pdal/PointView.cpp
```diff
@@ -100,26 +101,13 @@ void PointView::setFieldInternal(Dimension::Id dim, PointId idx,
...
 void PointView::calculateBounds(BOX2D& output) const
 {
-    for (PointId idx = 0; idx < size(); idx++)
-    {
-        double x = getFieldAs<double>(Dimension::Id::X, idx);
-        double y = getFieldAs<double>(Dimension::Id::Y, idx);
-
-        output.grow(x, y);
-    }
+    pdal::calculateBounds(const_cast<PointView&>(*this), output);
```
chambbj: @abellgithub Is this what you'd envisioned in terms of a redirect?

abellgithub: Exactly.

abellgithub:
Right now I think we have almost everything either in a namespace or a class.  This removes that for the functions that are in EigenUtils.  The only reason I like the namespacing is that it gives me a cue as where to look for the code.  Thoughts anyone?

chambbj:
I'm happy to keep this namespaced. As much as anything this was a step away from thinking this group of utils is explicitly Eigen-related, and I couldn't think of a more appropriate name.

abellgithub:
I figured as much.  A good name would be nice.

doc/stages/filters.icp.rst
```diff
@@ -78,6 +76,17 @@ The metadata output might start something like:
...
 
-None.
+max_iter
+  Maximum number of iterations. [Default: **100**]
+
+max_similar
+  Max number of similar transforms to consider converged. [Default: **0**]
+
+mse_abs
+  Absolute threshold for MSE. [Default: **1e-12**]
```
gadomski: In both the documentation and in the usage text in the code, might be helpful to use the full phrase instead of the acronym.

PR #2577, opened 2019-06-18
TileDB: Add append option to writer

normanb:
This change adds the option to append to an existing TileDB array. This change is useful when merging datasets.

### DIFF
--- a/doc/stages/writers.tiledb.rst
+++ b/doc/stages/writers.tiledb.rst
@@ -56,6 +56,9 @@ compression
 compression_level
   TileDB compression level for chosen compression [Optional]
 
+append
+  Append to an existing TileDB array with the same schema [Optional]
+
 stats
   Dump query stats to stdout [Optional]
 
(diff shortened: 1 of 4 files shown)


abellgithub:
Do you need/want some check to make sure that the existing attributes are the same as the new ones when appending?

normanb:
@abellgithub it is going to throw an exception https://github.com/TileDB-Inc/TileDB/blob/88b047879ab4696ecc84042ea6803d17526fc8be/tiledb/sm/query/writer.cc#L473 in the case the attributes are not matching. Do you want me to add this in the PDAL driver as well?

abellgithub:
No.  If you want, it would be nice to catch TileDB exceptions and rethrow wrapped in a PDAL exception so that the built-in handling we have for those gets to run, but not a big deal.

normanb:
@abellgithub  I will look to catch the error and rethrow as a PDAL exception, I won't get to this until tomorrow.

PR #2581, opened 2019-06-21
Replace WIN32 macro use with _WIN32

### LINKED ISSUE #2576: Use _WIN32 instead of WIN32
WIN32 isn't always defined

abellgithub:
Close #2576


hobu:
This looks fine except that _WIN64 is defined and no _WIN32 is defined by MSVC2017 in the 64 bit case. Not sure what to do about that. We were able to get by with "WIN32" for all MSVC compiler stuff. There used to be a specific #ifdef for that, but it seems to be gone now.

abellgithub:
Strange.  That's not what the docs say:

```
_WIN32 Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, or x64. Otherwise, undefined.

_WIN64 Defined as 1 when the compilation target is 64-bit ARM or x64. Otherwise, undefined.
```

PR #2590, opened 2019-06-26
Modified arbiter to get valid GCS access token on windows.

pravinshinde825:
- `Libexecinfo` and `Libunwind` are not for windows, Modified to ignore them on windows.
- `ShLwApi` is already in path.
- `utcOffsetSeconds` throws an null pointer exception on windows. Which causes the failure in retriving the access token from GCS. Added new method for windows named `utcToLocalOffsetSeconds` which will return the UTC to Local time difference. After this arbiter can retrieve a valid access token from 
GCS.

### DIFF
--- a/cmake/execinfo.cmake
+++ b/cmake/execinfo.cmake
@@ -1,4 +1,6 @@
 #
 # Unwind
 #
-find_package(Libexecinfo QUIET)
+if(NOT MSVC)
+    find_package(Libexecinfo )
+endif()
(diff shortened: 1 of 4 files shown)


hobu:
Upstream for this code is at https://github.com/connormanning/arbiter. It would be best to make a PR there and then we can integrate this into both Entwine and PDAL.

pravinshinde825:
@hobu , Yes make sense. Thanks

PR #2594, opened 2019-06-27
remove greyhound driver

hobu:


hobu:
Still need to remove docs

abellgithub:
StageFactory, StageExtensions have references.  cmake directory does too.

hobu:
Workshop materials need to be updated to use USGS EPT resources instead of greyhound

PR #2598, opened 2019-07-02
Fix global symbol export for python

abellgithub:
A change to improve flexibility (#2399) broke the loading of python symbols for Ubuntu.

Note: On ubuntu extension modules aren't liked with libpython3 (or whatever).  This means that when loaded from a dll, the python symbols aren't available.  In our case the extension is numpy, and it can't find python symbols because it hasn't been linked with python.  So we globally export the python symbols when our shared library is loaded so that they're available to satisfy the symbols in numpy.

This uses the base name from the compile-time python library as the source of the symbols.  It allows for override by setting the env variable PDAL_PYTHON_LIBRARY.

### DIFF
--- a/pdal/util/Utils.hpp
+++ b/pdal/util/Utils.hpp
@@ -64,15 +64,15 @@ namespace Utils
 {
 
 #if defined(__APPLE__) && defined(__MACH__)
-    const std::string dynamicLibExtension = ".dylib";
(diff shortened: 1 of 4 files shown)


hobu:
Windows CI fails to find/configure Python

abellgithub:
That seems something for another ticket.

PR #2603, opened 2019-07-10
Streamable EPT Reader

pravinshinde825:
Updated EPT reader to support streamable mode.

PR #2612, opened 2019-07-15
Use vector to avoid leaking streaming EPT buffers in case of exception.

connormanning:
Related to #2603.

### DIFF
--- a/io/EptReader.cpp
+++ b/io/EptReader.cpp
@@ -818,11 +818,10 @@ void EptReader::loadNextOverlap()
 void EptReader::fillPoint(PointRef& point)
 {
     DimTypeList dims = m_bufferPointView->dimTypes();
-    char* buffer = new char[m_bufferPointView->pointSize()];
-    m_bufferPointView->getPackedPoint(dims, m_currentIndex, buffer);
-    point.setPackedData(dims, buffer);
+    std::vector<char> buffer(m_bufferPointView->pointSize());
+    m_bufferPointView->getPackedPoint(dims, m_currentIndex, buffer.data());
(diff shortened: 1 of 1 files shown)


io/EptReader.cpp
```diff
@@ -818,11 +818,10 @@ void EptReader::loadNextOverlap()
 void EptReader::fillPoint(PointRef& point)
 {
     DimTypeList dims = m_bufferPointView->dimTypes();
-    char* buffer = new char[m_bufferPointView->pointSize()];
-    m_bufferPointView->getPackedPoint(dims, m_currentIndex, buffer);
-    point.setPackedData(dims, buffer);
+    std::vector<char> buffer(m_bufferPointView->pointSize());
```
abellgithub: This is fine, but it still seems like it should be done once -- the point size is fixed.

pravinshinde825: @connormanning , We can make it as a member of EptReader class like: `std::vector<char> m_temp_buffer` and reserve memory for it in `if(pipelineStreamable)` at line 441 in `addDimensions()` like : `m_temp_buffer.reserve(layout->pointSize())`. So here we only need to use `m_temp_buffer.data()`.
@abellgithub , In this way buffer will be allocated only once.

pravinshinde825: @connormanning , Shall I do that??

connormanning: Go ahead if you'd like @pravinshinde825, my only concern for this PR was the `new`/`delete`.

pravinshinde825: @connormanning , Please check https://github.com/PDAL/PDAL/pull/2617

PR #2617, opened 2019-07-16
Ept reader streaming exception safety

pravinshinde825:


abellgithub:
I think we can dispense with the whole temp buffer thing by using:
```
point.setPackedData(dims, m_bufferPointView->getPoint(m_currentIndex));
```

pravinshinde825:
@abellgithub , I tried the way you suggested while implementing the streamable EPT reader in https://github.com/PDAL/PDAL/pull/2603, It was expected to work fine but something was wrong with colors (Please refer the image, here the reader was EPT and Writer was LAS). Thats the reason I choosed to use intermediate buffer to retrive point from PointView and set PointRef.
[
Capture_both
](url)

pravinshinde825:
Added one more exception safety here!!
- Fixed the crash if there are 0 points in the overlap. Image shows execution log after fix.

Capture

abellgithub:
Is there a test for the case you're fixing?

pravinshinde825:
> Is there a test for the case you're fixing?

Hi @abellgithub , I need to prepare some input data to reproduce the crash, since I cannot share the data where I noticed this. Shall I do that in my upcomming PR?? where I'm optimizing EPT reader's performance in Stream mode.

PR #2619, opened 2019-07-17
E57 check

abellgithub:
@nicolas-chaulet Would you please look this over -- especially the comments labeled "ABELL".

Most of the changes are formatting.

Thanks.


plugins/e57/io/E57Writer.cpp
```diff
@@ -167,12 +174,12 @@ bool E57Writer::processOne(PointRef& point)
          point.getFieldAs<double>(Dimension::Id::Y),
          point.getFieldAs<double>(Dimension::Id::Z));
 
-    // Write point
-    if (m_chunkWriter)
-        m_chunkWriter->write(point);
-    else
-        throw pdal_error("Something went wrong, e57 writer has not been initialised correctly.");
+    //ABELL - Do we need to check this for every point?
+    if (!m_chunkWriter)
```
nicolas-chaulet: It's probably not needed.

plugins/e57/io/E57Writer.cpp
```diff
@@ -184,26 +191,31 @@ void E57Writer::done(PointTableRef table)
...
+    if (Utils::contains(m_dimensionsToWrite, "colorRed"))
     {
         // found color info
         e57::StructureNode colorbox = e57::StructureNode(*m_imageFile);
         for (const std::string& color: {"Red", "Green","Blue"})
         {
-            auto minmax = pdal::e57plugin::getPdalBounds(pdal::Dimension::Id::Red);
-            colorbox.set("color"+color+"Minimum",e57::IntegerNode(*m_imageFile,minmax.first));
-            colorbox.set("color"+color+"Maximum",e57::IntegerNode(*m_imageFile, minmax.second));
+//ABELL - this seems wrong.  Why would we always use red?
```
nicolas-chaulet: We alsways use Red because Pdal treates all colors the same. But I guess we could change the code and use the actual color.

plugins/e57/io/E57Writer.cpp
```diff
@@ -242,66 +259,79 @@ void E57Writer::setupFileHeader()
...
 
     // Create creationDateTime structure TODO add actual time
     e57::StructureNode creationDateTime = e57::StructureNode(*m_imageFile);
     creationDateTime.set("dateTimeValue", e57::FloatNode(*m_imageFile, 0.0));
-    creationDateTime.set("isAtomicClockReferenced", e57::IntegerNode(*m_imageFile, 0));
+    creationDateTime.set("isAtomicClockReferenced",
+        e57::IntegerNode(*m_imageFile, 0));
     m_rootNode->set("creationDateTime", creationDateTime);
 
+//ABELL - This description seems too specific.
```
nicolas-chaulet: Yes indeed :)
Happy to change to "E57 file generated by PDAL version ..."

plugins/e57/io/E57Writer.cpp
```diff
@@ -32,49 +32,59 @@
...
 
 void E57Writer::ChunkWriter::write(pdal::PointRef& pt)
 {
-    // If buffer full, write to disk and reinitialise buffer
-    if (m_currentIndex == m_defaultChunkSize -1)
+// If buffer full, write to disk and reinitialise buffer
+
+//ABELL - This seems like it's off by one.
+// We increment after setting the entry in the map, so I would think we'd
+// want to use up to m_defaultChunkSize entries.
```
nicolas-chaulet: You are correct, no idea how it could have worked and passed unit tests...

plugins/e57/io/E57Writer.cpp
abellgithub: But you're fetching the min/max values of the dimension, right?  Wouldn't those be different (at least potentially) for red, green and blue?  If this isn't what you're doing, maybe a comment is in order.

nicolas-chaulet:
Yes they would potentially.

Le mer. 17 juil. 2019 à 18:22, Andrew Bell <notifications@github.com> a
écrit :

abellgithub:
@nicolas-chaulet : Want to look this over one more time before I merge?

nicolas-chaulet:
All good! Thanks for taking a stab

PR #2624, opened 2019-07-19
Add native PCD I/O (removing reader/writer from PCL plugin)

chambbj:
- Supports reading and writing ASCII
- Supports reading and writing binary
- Readers are streamable, while writers are not
- Does NOT yet support binary compressed
- Removes need for PcdCommon header
- Adds rudimentary tests for PCD reader/writer

A note about the PCD writer. Similar to other writers, the PCD writer can
output a subset of the dimensions if keep_unspecified is set to false. In that
case, only the dimensions specified in the order parameter are written in the
output. The dimension type ("F", "I", or "U") and size (in bytes) can also be
specified, along with precision in the case of ASCII output.

This native driver for PCD will remove a long-standing issue with the previous
PCL-backed drivers, which required a conversion to a fixed, single-precision
point schema, that resulted in loss of extra dimensions (beyond XYZRGBA) and
confusion surrounding subtraction of offsets to avoid loss of precision for
certain coordinate systems.


io/PcdReader.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+        std::string buf;
+
+        std::getline(*m_istreamPtr, buf);
+        m_line++;
+        if (buf.empty())
+            continue;
+
+        Utils::trim(buf);
+        m_fields = Utils::split(buf, pdalboost::algorithm::is_any_of("\t\r "));
```
abellgithub: I'd vote for no boost and just making a lambda.

chambbj:
> I'm going to vote for `Dimension=PDAL_Type, Dimension=PDAL_Type` for binary and `Dimension[:precision], Dimension[:precision]` for ASCII output. This is substantially consistent with other stages.

But ASCII will also need to know type. Maybe `Dimension=PDAL_Type[:precision]`?
 
> Don't know how much you want to keep messing with this. I can see optimization things, but maybe we don't care right now.

First objective was to get it working, but understand it may not be optimal. Feel free to make improvements over time! And if there are some you'd like to see right away (you've noted a couple possibilities), I can address these prior to merge.

abellgithub:
Well, precision only applies for floating point, right?  I don't know.  Guess if it were me I'd just like to write the types as PDAL has them (we already know if something is signed/unsigned/float), but I guess people will want to specify, so maybe your suggestion is fine.  I don't understand the notion of specifying type for ASCII output, but I didn't create the PCD format.

abellgithub:
You *could* take the type from the precision (0 precision is integer representation).  You'd still need to differentiate signed/unsigned, but that's problematic no matter -- a conversion error will be thrown if you try to move a negative value into an unsigned.

abellgithub:
@chambbj : Want to merge master into this and see if tests pass?

PR #2626, opened 2019-07-20
Implement coordoperation filter

vilaa:
Related to #2616


abellgithub:
This looks nice. I'll look it over again on Monday. Thanks!

filters/CoordOperationFilter.cpp
```diff
@@ -0,0 +1,143 @@
...
+* AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+* OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+* OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY
+* OF SUCH DAMAGE.
+****************************************************************************/
+
+#include "CoordOperationFilter.hpp"
+
+#include <gdal_version.h>
+#if GDAL_VERSION_MAJOR >= 3
```
hobu: I think the entire driver should be protected at the cmake level by a GDAL 3 guard. If GDAL 3 is not available, the driver is simply not activated.

filters/CoordOperationFilter.cpp
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    viewSet.insert(outView);
+    return viewSet;
+}
+
+
+bool CoordOperationFilter::processOne(PointRef& point)
+{
+    double x(point.getFieldAs<double>(Dimension::Id::X));
```
hobu: Should we allow the user to map other dimensions to XYZ? I can see scenarios where data is modified in a pipeline and copied to other dimensions.

filters/CoordOperationFilter.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+//#include <memory>
+
+namespace pdal
+{
+
+static StaticPluginInfo const s_info
+{
+    "filters.coordoperation",
+    "Transform coordinates using Proj pipeline string, WKT2 coordinate operations or URN definition",
```
hobu: @vilaa Thanks for the contribution. It brings up some design and organization discussion the project needs to have in regards to this capability.

@abellgithub this filter really does exactly the same thing as `filters.projection` if we add the necessary capability to it. What value is there in having two filters with somewhat similar names doing slightly different things?

filters/CoordOperationFilter.cpp
vilaa: resolved in https://github.com/PDAL/PDAL/pull/2626/commits/3c3c1d019c058301e8261152fbcdbcbac3346ccb

filters/CoordOperationFilter.cpp
vilaa: I'm not familiar with the mechanism that allows data to be copied to other dimensions. In the issue discussion, you told me that the transformation filter has this capability, but watching to the code I don't see clearly where in the code this is enabled. 
I'm new to PDAL development so maybe I've missed something.

hobu: What I would propose is adding options that allow users to map "X=>OtherX", and upon `initialize`, the filter caches a reference to the OtherX dimension and uses it as X (same for Y and Z). That way if there were other dimensions in the PointView that were different kinds of XYZ, the user wouldn't have to `filters.ferry` data into XYZ and destroy their original values (important when constructing transformations).

I also wonder if we shouldn't allow output of the new XYZ coordinates into **new** XYZ dimensions to preserve the old coordinates too, again for the same reasons as above.

abellgithub: I prefer ferry.

abellgithub:
Aurelien,

We had a long talk about this yesterday, trying to figure out where this should go.  In the end, we decided that it should stand alone as you've implemented it with a name change.

There are a few changes that I'd like to see.  You can make them or I can after the merge.  Just let me know:

1) To make it clear what this filter is doing, rename to `proj_pipeline`.
2) Rename `a_srs` to `in_srs`.
3) Add `out_srs` as an option to allow an optional assignment of an SRS at pipeline completion.
4) Remove `ogr_spatialref.h` from the filter header file.  We're trying to remove GDAL references from header files.  There are a few stragglers where this isn't the case, but we're not wanting to introduce new instances.  In this case this means moving the definition of CoordOperationFilter from the header to the source file, I think.

Thanks,

vilaa:
@abellgithub 

Thanks for your review. I will do the change quickly.
But I don't understand the need for the `in_srs` option since the filters
does not take into account SRS information in input. 
For me, It is the `a_srs` option that must be renamed to `out_srs`.
Could you confirm me that before i do the change ?

Thanks

abellgithub:
Sorry, yes.  `a_srs` -> `out_srs`.  You can use `addSynonym("out_srs", "a_srs")` if you want to have both.

vilaa:
@abellgithub 

Everything resolved in 1896559a0a102054d4fd91149dfde20ac04147cc

PR #2630, opened 2019-07-23
Remove Python2 support

hobu:


cmake/python.cmake
```diff
@@ -3,62 +3,31 @@
...
-    endif()
+    set(PYTHON_LIBRARY ${Python3_LIBRARIES}
+        CACHE FILEPATH "Python library")
+    set(PYTHON_INCLUDE_DIR ${Python3_INCLUDE_DIRS}
+        CACHE PATH "Location of Python include files.")
+     set(PYTHON_EXECUTABLE ${Python3_EXECUTABLE})
+     find_package(NumPy 1.5 REQUIRED)
     set(PDAL_HAVE_PYTHON 1)
 else()
     find_package(PythonInterp 3 QUIET)
```
abellgithub: QUIET should be removed on these.

PR #2641, opened 2019-07-30
default WITH_ZSTD to TRUE if its headers are found

hobu:


abellgithub:
Are you sure that works if you don't have ZSTD?  I would think it would error.

hobu:
I followed the same pattern as LASzip's config. Checking...

hobu:
Tweaked in 8bccb090362d610d89c7ed4a4ded9a209565bc2b

PR #2644, opened 2019-07-30
remove linking of GDAL library from 'pdal' application #2350

hobu:


hobu:
Let's merge this for now and kick dealing with windows on this issue ahead to a different ticket.

abellgithub:
Note that the issue here is, I believe, that GDAL functions are exposed in GDALUtils.hpp.

Issue #2642, opened 2019-07-30
KNN queries should add a stride option, defaulting to 1


chambbj:
Several recently published papers have suggested increasing the local receptive field of a point by increasing the stride of the k-nearest neighbors so that the number of computations is not also increased. Maybe nanoflann already supports something like this, but it would be an easy add to KDIndex.

chambbj:
- http://arxiv.org/abs/1907.12046
- http://arxiv.org/abs/1907.09798

chambbj:
Closed via #2650

PR #2650, opened 2019-07-31
Add stride option (default=1) to KDIndex 3D neighbor search

chambbj:


hobu:
What does the stride parameter do?

chambbj:
Give you every nth neighbor.

chambbj:
A little more detail provided in #2642. It's a way to keep the computational complexity of a neighborhood fixed while increasing the area.

pdal/KDIndex.hpp
```diff
@@ -260,25 +262,29 @@ class PDAL_DLL KD3Index : public KDIndex<3>
         pt.push_back(y);
         pt.push_back(z);
         m_index->findNeighbors(resultSet, &pt[0], nanoflann::SearchParams());
+        std::valarray<PointId> temp(output.data(), output.size());
```
abellgithub: This seems like a lot of copying for very little benefit.  And it made me look up what a valarray is.

if (stride > 1)
{
   for (i = 1; i < k; i++)
      output[i] = output[i * stride]
   output.resize(k);
}

Also, you should make it clear in a comment that you're still getting 'k' neighbors, but you're really extracting k * stride neighbors and then taking every stride'th item.

chambbj: Right, so that'll work. Probably went too far out of my way to get slice to work. Wish I could simply use it directly on the vector.

PR #2651, opened 2019-08-01
First whack at removing PCL plugin

chambbj:
Ground filtering tutorial switch from PCL to PDAL is being addressed in https://github.com/PDAL/PDAL/tree/ground-tutorial-refresh.


abellgithub:
@hobu : were we wanting to do this for the release?

chambbj:
I believe so. The two requirements were that we replace ICP (done) and PCD (PR pending).

hobu:
> were we wanting to do this for the release?

*I* sure want to be rid of the PCL burden as soon as we can do it. We have been signaling this move for multiple releases. I think part of the process should be moving the old plugin's codebase to https://github.com/PDAL/unsuppported-plugins which will give someone who was depending on it a way forward (if hand rolled and messy).

chambbj:
1. Do you want to retain any of the old docs in said repo?
2. Do you want to include prior versions of stages that are now included in PDAL (ICP, PCD)? Or just the ones that were left behind?

hobu:
1. Yes, maybe useful. Make a `./docs` directory in the top-level and stuff it there?
2. I say no on this because we didn't change the stage names. @abellgithub ?

chambbj:
For now, I've started with https://github.com/PDAL/unsuppported-plugins/pull/1, which pulls over the entire PCL plugin directory with full commit history. I'll continue to evolve that PR to remove the ported stages and pull over the relevant docs as well.

PR #2653, opened 2019-08-01
Don't build EPT if we don't have arbiter

### LINKED ISSUE #2357: Don't build EPT if no arbiter
EPT needs arbiter to work properly.

abellgithub:
Close #2357


abellgithub:
Now I'm wondering why I did this.  Arbiter is vendor'd.  This was obviously a problem or I wouldn't have messed with it.  Has something changed?  I don't remember.

connormanning:
Initially (pre-EPT) Arbiter wasn't #ifdef'd properly for the absence of things like libcurl and libcrypto since it presumed HTTP as a primary target (it was only used for the GreyhoundReader which was necessarily HTTP, and optionally enabled for remote IO if libcurl was found).

At this point it properly falls back to allow local access to work normally if those libs are not found, but I'd guess that this issue is a hold-over from the original behavior.

abellgithub:
Should I just kill this?

connormanning:
I'd think so.

Senfourtou:
I think so

PR #2654, opened 2019-08-01
New options for writers.gdal

### LINKED ISSUE #2415: change writers.gdal bounds option
writers.gdal has a `bounds` option.  This currently does two things in one: it sets the raster origin and also sets the number of cells in the raster.  It would be more clear if there were separate options to describe these attributes.

abellgithub:
Close #2415


abellgithub:
We don't need to deprecate.  You don't need bounds for streaming anymore.  I'll fix that in the docs (thought I had). 

We've had a history of bounds confusing people at the edges -- they don't get the exact number of cells they expect.  I don't know why they would care, but it has come up more than once.

PR #2655, opened 2019-08-01
Add 2-pass check for streamable pipelines

### LINKED ISSUE #2376: filters.colorinterp broken in stream mode
filters.colorinterp doesn't deal properly with m_min and m_max in stream mode.

abellgithub:
This fixes the issue with the colorinterp filter and min/max.
Close #2376


abellgithub:
@hobu: I made a bunch of changes to this to make it better (IMO).  Would you look it over once again before I merge?

There's support here to force both streaming/standard mode and preference for stream mode.  We can change the argument handling for kernels to support this.  Right now --nostream will always run in standard mode, but --stream is the same as providing no option -- it runs in streaming if possible, but standard mode otherwise.  We could change things easily now so that --stream means to ONLY run in stream mode, which as you pointed out, has value.

hobu:
LGTM. It would be nice to add `pdal --stream` as forced-stream mode now with an error message stating that the user might remove the `--stream` argument from the invocation to let PDAL revert to bulk mode if it needs.

abellgithub:
What are your thoughts on doing that for this release?

hobu:
> What are your thoughts on doing that for this release?

I think it is worthy of putting in. We introduced `--stream` a long time ago, and we aren't changing its behavior in such a disruptive way IMO if we do this. If I asked for `--stream` and you can't because of some filter, ok, I need to adjust my pipeline or remove the `--stream` request.

abellgithub:
I don't think there's much risk.  I'll add it to this PR.

PR #2659, opened 2019-08-05
Updates for writers.fbx

hobu:


plugins/fbx/io/FbxWriter.cpp
```diff
@@ -130,11 +131,11 @@ void FbxWriter::done(PointTableRef table)
     // ...
     const char *format = m_ascii ? "FBX ascii (*.fbx)" : "FBX binary (*.fbx)";
     int writer = registry->FindWriterIDByDescription(format);
-    /**
+
```
abellgithub: This is just debug code.  I don't think it should be uncommented.

PR #2662, opened 2019-08-08
Make simple GDAL version handling.

abellgithub:


cmake/gdal.cmake
```diff
@@ -1,19 +1,22 @@
...
-    # if we have GDAL_VERSION (as a version type, not a string)
-    # don't run gdal-config
-    if ( ${GDAL_VERSION})
-        # cmake is dumb. if we test NOT ${GDAL_CONFIG} it is true when GDAL_CONFIG
-        # is set to a value. I don't know why.
-    else()
-        execute_process(COMMAND ${GDAL_CONFIG} "--version" OUTPUT_VARIABLE GDAL_VERSION)
-    endif()
-    string(COMPARE GREATER ${GDAL_VERSION} "3.0.0" GDAL_3)
+    gdal_find_version(GDAL_VERSION)
```
hobu: The issue I kept having is FindGDAL.cmake from CMake is setting a `GDAL_VERSION` too. We're now overriding that, and if the user set a different GDAL_ROOT or something, I think we're going to get the wrong one.

abellgithub: I don't know what GDAL_ROOT is.  If the user is setting some other copy of GDAL with that env variable but it doesn't impact the FindGDAL script, we're screwed.

The version info here is coming from the found copy of the header.  I don't know how we can make it any more right.  I could name the variable something else, but I'm not sure why we'd do that.  Overriding seems fine to me since the problem is that it's not always available/right.

hobu: re: `-DGDAL_ROOT=/path/to/gdal` is how you hint FindGDAL.cmake in CMake-speak to give it a user-defined path/prefix to override and look for your own instance of the library. Packagers use this all the time.

My fix attempts were trying to see if `GDAL_VERSON` had already been set appropriately by FindGDAL.cmake, and if not, resort to monkey patching our own methods for extracting it. If CMake's syntax was such 🗑, I might have succeeded.

abellgithub: If this is the case, then we're getting the version from the proper place according to GDAL_ROOT, since we're using the found include directory to locate the version info.  I don't think it's worth bothering with the module's version information if it's not always right.

I did consider that someone might change the format of the gdal_version.h file, but that seems unlikely and I'm not sure it's guaranteed to be any more stable than anything else.

PR #2667, opened 2019-08-12
support 'polygon' query limiter, which smartly limits fetches based on box intersections of the polygon(s) given

hobu:


msmitherdc:
Is the clip a actual POLYGON or does it use the POLYGON to create a bbox? And are MULTIPOLYGONs supported?

hobu:
The idea is that it works the same as `filters.crop` but the polygon is used to pre-filter the EPT requests instead of testing the points. Whatever is supported in the `polygon` option of `filters.crop` would be supported as far as POLYGON/MULTIPOLYGON/GeoJSON, etc.

msmitherdc:
Do you need to put something in to limit it to POLY/MULTIPOLYGON?

abellgithub:
I don't understand the advantage of this over just adding a crop filter.  We're not eliminating any point reading with this.

hobu:
Connor and I discussing that when you put a `polygon` on the `readers.ept`, you always want the equivalent of `filters.crop` output – ie, the points cropped too. That means baking in a `filters.crop` into the `readers.ept` or something.

msmitherdc:
does it prevent pulling that data over the wire is the important question?

hobu:
> does it prevent pulling that data over the wire is the important question?

Yes it does, but then you also want the points cropped to that same polygon. Why did you bother otherwise?

msmitherdc:
Does the spatial reference parameter apply to this?

io/EptReader.cpp
```diff
@@ -215,12 +216,19 @@ void EptReader::initialize()
     // Set geometry from polygons.
     if (m_args->m_polys.size())
     {
-        m_polys.clear();
+        m_queryPolys.clear();
         for (Polygon& poly : m_args->m_polys)
         {
             // Throws if invalid.
             poly.valid();
```
abellgithub: valid just returns a bool.  It doesn't throw.

https://github.com/PDAL/PDAL/blob/master/pdal/Geometry.cpp#L176

abellgithub:
@hobu: This could still be better as far as eliminating processing checks, but I think it's OK for a first cut.  Please have a look when you're back to work.

PR #2677, opened 2019-08-20
Don't write metadata with more precision than exists.

### LINKED ISSUE #2663: Metadata precision causes LAS forwarding issues
We seem to be writing double-precision values with so much precision that we're bleeding bad values into metadata.  This can be seen when forwarding scale values.

abellgithub:
Close #2663


hobu:
Is it intended that this fix will make the 2.0 release?

abellgithub:
Yes.  It's tagged as such.

abellgithub:
Now to figure out why SQLite is still broken here...

PR #2683, opened 2019-08-21
Workshop updates

adamsteer:
Updates to PDAL workshop materials ahead of FOSS4G 2019. Predominantly replacing windows-ified paths and using entwine instead of greyhound.


abellgithub:
Thanks for the careful doc work.

PR #2690, opened 2019-08-22
EPT reader headers/query parameters forwarding for HTTP-based endpoints

connormanning:
Adds `headers` and `query` options to the EPT reader, which will be forwarded to remote endpoints.  The typical use-case here is for authenticated endpoints.

Also adds capability for `filename` values ending with `"ept.json"` to be routed to the EPT reader, which can be more convenient than the alternative `ept://` protocol method.

### DIFF
--- a/io/EptReader.cpp
+++ b/io/EptReader.cpp
@@ -312,7 +312,7 @@ void EptReader::initializeHttpForwards()
         for (const auto& entry : obj.items())
         {
             if (!entry.value().is_string())
-                throwError("Invalid " + type + "parameters: "
+                throwError("Invalid " + type + " parameters: "
                     "expected string->string mapping");
             map[entry.key()] = entry.value().get<std::string>();
         }
--- a/io/EptReader.cpp
+++ b/io/EptReader.cpp
@@ -137,8 +137,10 @@ public:
     double m_resolution = 0;
     std::vector<Polygon> m_polys;
     NL::json m_addons;
-};
 
+    NL::json m_query;
+    NL::json m_headers;
+};
 
 EptReader::EptReader() : m_args(new EptReader::Args), m_currentIndex(0)
 {}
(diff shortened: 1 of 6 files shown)


io/EptReader.cpp
```diff
@@ -159,6 +161,38 @@ void EptReader::addArgs(ProgramArgs& args)
     args.add("polygon", "Bounding polygon(s) to crop requests",
         m_args->m_polys).setErrorText("Invalid polygon specification. "
             "Must be valid GeoJSON/WKT");
+    args.add("header", "Header fields to forward with HTTP requests",
```
abellgithub: Ugh to JSON here.  But I know I lose.

connormanning:
The parameters are validated as either `null` or object mappings whose values are all strings [here](https://github.com/PDAL/PDAL/pull/2690/files#diff-6c6aa6fc314618e64694ec420896da89R304), and an error occurs otherwise:

```
pdal info http://na.entwine.io/autzen/ept.json --readers.ept.query="{\"asdf\":42}"
PDAL: readers.ept: Invalid query parameters: expected string->string mapping
```

connormanning:
Or if not an object:

```
pdal info http://na.entwine.io/autzen/ept.json --readers.ept.query=42
PDAL: readers.ept: Invalid query parameters: expected object
```

abellgithub:
Sorry.  I saw that check.  I mean when those query/header parameters are sent downstream.

connormanning:
If the service endpoint errors due to the *contents* of the parameters, for example if a `401` status code were returned, then the error handling is the same as any other endpoint error:
```
PDAL: readers.ept: Could not read from http://my-server.com/whatever/ept.json
```

This is potentially a spot for improvement, but presumably for users opting in to supplying headers/query options for authenticated endpoints, the contents of those options would be the obvious place to look.

Issue #2692, opened 2019-08-23
writers.pgpointcloud selects wrong pcid


sensor-freak:
I believe that the pgpointcloud plugin contains an anomaly when creating a new table in the target database.

Steps to reproduce the issue:
- (create new database and enable reqquired extensions postgis, pointcloud and pointcloud_postgis)
- create a pipeline using a writer for `writers.pgpointcloud` (and arbitrary input data with no assigned SRS)
- define a valid SRS in the writer stage (e.g. `"srsid":"25832"`)
- execute the pipeline
- the database should now contain a new table with a column of type `PcPatch`
- create a seconds pipeline using a different table name and different valid SRS (e.g. `"srsid":"31467"`)
- execute the pipeline
- check the SRS assigned to the geometries in the two tables (e.g. `select st_srid(geometry(pa)) from <tablename>`). The results should have (IMHO) the respective srsid values as specified in the pipeline. Actually (as of version 1.9.1) both tables indicate the same srsid, specifically the one from the first table created.

The fix for this seems to be somewhat trivial: just add the srid argument when searching for a matching entry in the pointcloud_formats table. This might be done by adding a WHERE clause in line PgWriter.cpp:234 `WHERE srid = ?`, or by extending the SELECT statement to return column `srid` and compare the value against the variable `srid` in PgWriter.cpp:240.

(Currently i have no setup for building PDAL myself, therefore I cannot validate my suggestion.)

Any comments are welcome!

hobu:
If you can turn these steps into SQL and provide some data that demonstrates the issue, we can make a test case and a bug fix that attacks it.

sensor-freak:
I have not checked if the fix is compatible with master, since I currently have to stay with 1.9.1. But please feel free to pick the change also for the master branch.

hobu:
Closing as this appears merged.

PR #2696, opened 2019-08-26
Fix selecting wrong schema pcid on pgpointcloud writing (#2692)

sensor-freak:
Also added a unit test to ensure the fix works as expected.

### DIFF
--- a/plugins/pgpointcloud/io/PgWriter.cpp
+++ b/plugins/pgpointcloud/io/PgWriter.cpp
@@ -230,8 +230,10 @@ uint32_t PgWriter::SetupSchema(uint32_t srid)
     // Do any of the existing schemas match the one we want to use?
     if (schema_count > 0)
     {
-        PGresult *result = pg_query_result(m_session,
-            "SELECT pcid, schema FROM pointcloud_formats");
+        oss.str("");
+        oss.clear();
+        oss << "SELECT pcid, schema FROM pointcloud_formats WHERE srid = " << srid;
+        PGresult *result = pg_query_result(m_session, oss.str());
         for (int i = 0; i < PQntuples(result); ++i)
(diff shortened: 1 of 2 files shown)


sensor-freak:
I have pushed wrong commit. Will fix it soon.

plugins/pgpointcloud/io/PgWriter.cpp
```diff
@@ -230,8 +230,10 @@ uint32_t PgWriter::SetupSchema(uint32_t srid)
     // Do any of the existing schemas match the one we want to use?
     if (schema_count > 0)
     {
-        PGresult *result = pg_query_result(m_session,
-            "SELECT pcid, schema FROM pointcloud_formats");
+        oss.str("");
+        oss.clear();
+        oss << "SELECT pcid, schema FROM pointcloud_formats WHERE srid = " << srid;
+        PGresult *result = pg_query_result(m_session, oss.str());
```
sensor-freak: Here I'm using oss to construct the query. oss.str("") and oss.clear() ensure that the stream is really empty.
The modified SQL query only selects those schema entries, that have the wanted SRS column. This modification will allow a new row to be generated, when the same schema is used with a different SRS. (Generation of the new entry is performed below with the INSERT statement (line 288ff).

sensor-freak:
I have reverted some changes that shouldn't have gone to the repo at all.

sensor-freak:
> 
> 
> > I don't understand the changes in this function. Can you explain?
> 
> The reader is not making joins properly in some situations.

I cannot comment on the reader, since I'm using only the writer.

abellgithub:
Integrated via 59ad8229fffbb70a4f6e3b51e61d2c89bae2e654

Issue #2629, opened 2019-07-23
Streamable version of voxel grid


nicolas-chaulet:
Hi, I was wondering if there was a way to introduce some sort of streamble version of the voxel grid filter. The use case would be that the input point cloud is 1mm precise say and we apply a voxel grid of 10mm. The output of this would be quite light but with the current setup we have to load all the data in memory before moving forward which seems unecessary to me.

Any idea about how we could go about that?  Is there a way to run part of the steps in a pipeline in streaming mode (the reader in that case) and part of the steps in normal mode? Such a strategy would reduce drastically the memory footprint of those voxel filters because we only need to store in memory one point per voxel unit as opposed to the whole point cloud.

chambbj:
It would be a bit different than the two native voxel-type filters, and more like the "original" PCL filter. I believe it would be possible using online algorithms for computing the mean coordinate per occupied cell. You'd have to scrap (or otherwise interpolate) additional dimension data though. You'd probably also need to either specify your voxel grid offsets or ensure you have trusted bounds in the file header such that you are not depending on computing any of that directly from the data.

nicolas-chaulet:
Regarding the functionality, I think we could ahve the same as the current functionality, just lighter. 
Regarding the size of the grid, you have a point. An option would be to have vaid headers, another way would be to have a dynamic (sparse) grid that would scale with new data points.

chambbj:
Current functionality is to store the single point closest to either the centroid or the voxel center. So, I guess an online update to voxel center is easy enough, as long as the voxel size and position is static. Centroid seems problematic to me, as you won't know the centroid until you've see all of the points, but you don't want to store the points, so how do you get back to the one closest to it? Maybe this is the hybrid step you proposed, but I don't think that will be possible in the current architecture. I don't see how you do any of this with a dynamic grid. Maybe I don't follow what you are proposing though.

abellgithub:
I'm not even sure what you'd be doing with the center algorithm.  How would you mark points as they pass through?  How would you know what points to skip?

chambbj:
> I'm not even sure what you'd be doing with the center algorithm. How would you mark points as they pass through? How would you know what points to skip?

I had imagined each point that is processed first is checked to see if it's corresponding voxel was occupied. If so, and the current point was closer than the existing point, swap them out. If it's not occupied, simply add it.

nicolas-chaulet:
The dynamic grid would be stored in a map with index being the integer coodinate of the voxel. 
[...]

PR #2699, opened 2019-08-27
Feature/first-in-voxel filter

pravinshinde825:
- Added `first-in-voxel filter`.
      - This filter will do a pointcloud downsizing based on voxel-grid. 
      - One of the limitation to the `voxelcenternearestneibour` filter is it cannot run in streamable mode. To overcome the limitation, I wrote first-in-voxel filter. This filter will only accept the first point entering in the voxel and mark the voxel as populated, After that any point entering in the same voxel will be ignored/skipped.


nicolas-chaulet:
> I guess I'm not convinced the center approach couldn't be made streamable (see comments in #2629), but if this works for you then I have no objections. I can take a closer look at the code today.

The voxel centre is streamable and could be an option to this filter I guess. The one that is not streamable is the voxel centroid :)

filters/FirstInVoxelFilter.cpp
```diff
@@ -73,24 +73,53 @@ PointViewSet FirstInVoxelFilter::run(PointViewPtr view)
...
-    size_t const h3(std::hash<int>{}(static_cast<uint32_t>(POSITIVE_VALUE((gz - m_Bounds.minz)))));
-    size_t key = std::hash<size_t>{}(HASH_COMBINE((HASH_COMBINE(h1, h2)), h3));
+	/*
+	* Calculate the local voxel coordinates for incoming point, Using the Pivot voxel.
+	* Generate a combined hash key from local coordinates.
+	*/
+    size_t const hx(std::hash<int>{}(std::floor(gx - m_Bounds.minx)));
+    size_t const hy(std::hash<int>{}(std::floor(gy - m_Bounds.miny)));
+    size_t const hz(std::hash<int>{}(std::floor(gz - m_Bounds.minz)));
+    size_t key = std::hash<size_t>{}(HASH_COMBINE((HASH_COMBINE(hx, hy)), hz));
```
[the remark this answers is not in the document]
pravinshinde825: Updated to use `std::tuple`.

filters/FirstInVoxelFilter.cpp
```diff
@@ -0,0 +1,132 @@
...
+ * OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY
+ * OF SUCH DAMAGE.
+ ****************************************************************************/
+
+#include "FirstInVoxelFilter.hpp"
+
+namespace pdal
+{
+
+static StaticPluginInfo const s_info{"filters.first-in-voxel",
```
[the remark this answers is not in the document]
pravinshinde825: Done!!

filters/FirstInVoxelFilter.cpp
```diff
@@ -0,0 +1,132 @@
...
+CREATE_STATIC_STAGE(FirstInVoxelFilter, s_info)
+
+std::string FirstInVoxelFilter::getName() const
+{
+    return s_info.name;
+}
+
+void FirstInVoxelFilter::addArgs(ProgramArgs& args)
+{
+    args.add("cell", "Cell size", m_cell).setPositional();
```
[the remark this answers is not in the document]
pravinshinde825: done!!

filters/FirstInVoxelFilter.cpp
```diff
@@ -73,24 +73,53 @@ PointViewSet FirstInVoxelFilter::run(PointViewPtr view)
...
 
-    m_Bounds.grow(gx, gy, gz);
+	static bool initialized=false;
+    if (!initialized)
+    {
+        /*
+		* Save global coordinates of first incoming point. 
+		* This will act as a Pivot for calculation of local coordinates of the voxels.
+		*/
+		m_Bounds.grow(gx, gy, gz);
```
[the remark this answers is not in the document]
pravinshinde825: Changed to array.

filters/FirstInVoxelFilter.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+bool FirstInVoxelFilter::voxelize(const PointRef point)
+{
+    double gx = point.getFieldAs<double>(Dimension::Id::X) / m_cell;
+    double gy = point.getFieldAs<double>(Dimension::Id::Y) / m_cell;
+    double gz = point.getFieldAs<double>(Dimension::Id::Z) / m_cell;
+
+    m_Bounds.grow(gx, gy, gz);
+
+    size_t const h1(std::hash<int>{}(static_cast<uint32_t>(POSITIVE_VALUE((gx - m_Bounds.minx)))));
```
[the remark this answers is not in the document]
pravinshinde825: Not needed.

pravinshinde825:
@chambbj , Sorry for the delayed response.
I addressed comments and updated code accordingly. Below is a snapshot for output according to latest changes.
Capture

filters/FirstInVoxelFilter.hpp
```diff
@@ -0,0 +1,68 @@
...
+    std::string getName() const;
+
+private:
+    double m_cell;
+
+    virtual void addArgs(ProgramArgs& args) override;
+    virtual PointViewSet run(PointViewPtr view) override;
+    bool processOne(PointRef& point) override;
+    bool voxelize(const PointRef point);
+    std::set<std::tuple<int,int,int>> m_populatedVoxels;
```
nicolas-chaulet: @pravinshinde825 I think we would get some nice speed boosts by making that one an `unordered_set`, did you look into that?

pravinshinde825: @nicolas-chaulet , Yes I checked with unordered_set as well. 
- In performance point of view `unordered_set` is a bit faster than `std::set` but not much. 
- In memory point of view I noticed approx 60% more memory usage in `unordered_set` than `set`.
- Also we may need to inject a hash funtion for `std::tuple<int,int,int>`.
So, I think `set` is fine.

filters/FirstInVoxelFilter.cpp
```diff
@@ -0,0 +1,130 @@
...
+ * OF SUCH DAMAGE.
+ ****************************************************************************/
+
+#include "FirstInVoxelFilter.hpp"
+
+namespace pdal
+{
+
+static StaticPluginInfo const s_info{"filters.firstInVoxel",
+                                     "First Entry Voxel Filter", ""};
```
chambbj: Third (currently empty field) should contain link to the docs, which will be http://pdal.io/stages/filters.first-in-voxel.html as things stand currently, though I think the doc page should be updated to match the name of the filter (having dropped the hyphens now).

doc/stages/filters.first-in-voxel.rst
```diff
@@ -0,0 +1,37 @@
...
+    :ref:`filters.voxelcenternearestneighbor` offers a similar solution,
+    using
+    the coordinates of the voxel center as the query point in a 3D nearest neighbor search. 
+    The nearest neighbor is then added to the output point cloud, along with any existing dimensions.
+
+Options
+-------------------------------------------------------------------------------
+
+cell
+  Cell size in the ``X``, ``Y``, and ``Z`` dimension.
```
chambbj: Please indicate the default value for the cell size by adding `[Default: 0.001]` to the end of the description.

test/unit/filters/FirstInVoxelFilterTest.cpp
```diff
@@ -0,0 +1,165 @@
...
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+ * OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY
+ * OF SUCH DAMAGE.
+ ****************************************************************************/
+
+#include <pdal/pdal_test_main.hpp>
+
+#include <pdal/StageFactory.hpp>
+#include<pdal/Reader.hpp>
```
chambbj: These three lines should have spaces after `#include` for consistency.

chambbj:
@pravinshinde825 Things are looking pretty good I think. I get believable results now. Just a few minor edits requested and then you should be good to go.

abellgithub:
I just wonder what, if any, advantage this has over the existing filter.  Could this logic, along with a check for distance, be added to VoxelCenterNearestNeighbor?  Is there some functional advantage to choosing the first point seen, rather than the one nearest to the center of the voxel?

nicolas-chaulet:
@abellgithub I can't thing of a way to get VoxelCenterNearestNeighbor to work in streaming mode. You need to go through the whole file before you can decide which point to keep. The idea here was to have a voxel downsizing that works in streaming mode. Same filter could be used to pick the centre just like PCL does it (as opposed to the first point)

chambbj:
I think it's as simple as, when you encounter an already initialized voxel, check to see if the distance for the current point is smaller than that of the one that first initialized the voxel, and if it is, update the dimensions/swap the point.

nicolas-chaulet:
If you are streaming, how can you swap? The first point has already been given to the next stage no?

chambbj:
This is true. Sorry.

abellgithub:
Sorry.  For some reason I was thinking of this filter as a writer -- that the voxel set was the output.

filters/FirstInVoxelFilter.cpp
pravinshinde825: Added link to the doc.

doc/stages/filters.first-in-voxel.rst
pravinshinde825: Added link to the doc.

test/unit/filters/FirstInVoxelFilterTest.cpp
pravinshinde825: Added spaces after `#include`

filters/FirstInVoxelFilter.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+    /*
+     * Calculate the voxel coordinates for the incoming point.
+     * gx, gy, gz will be the global coordinates from (0, 0, 0).
+     */
+    int gx = point.getFieldAs<double>(Dimension::Id::X) / m_cell;
+    int gy = point.getFieldAs<double>(Dimension::Id::Y) / m_cell;
+    int gz = point.getFieldAs<double>(Dimension::Id::Z) / m_cell;
+
+    static bool initialized = false;
```
abellgithub: This needs to be moved to be a member variable or multiple instances of this filter will be confused.

filters/FirstInVoxelFilter.cpp
```diff
@@ -0,0 +1,131 @@
...
+        m_pivotVoxel[1] = gy; // Y Coordinate of an Pivot voxel
+        m_pivotVoxel[2] = gz; // Z Coordinate of an Pivot voxel
+        initialized = true;
+    }
+
+    /*
+     * Calculate the local voxel coordinates for incoming point, Using the Pivot
+     * voxel.
+     */
+    auto t=std::make_tuple(gx - m_pivotVoxel[0], gy - m_pivotVoxel[1],
```
abellgithub: Spacing

filters/FirstInVoxelFilter.cpp
pravinshinde825: Yeah, make sense.
Changed to member variable

filters/FirstInVoxelFilter.cpp
pravinshinde825: Done!!

pravinshinde825:
Thanks @nicolas-chaulet for the accurate description.
@chambbj , @abellgithub , I addressed all the comments.

PR #2703, opened 2019-08-29
Limit VoxelCentroidNearestNeighbor results to points within voxel

### LINKED ISSUE #2664: VoxelCentroidNearestNeighbour doesn't check NN point is in the current voxel
The VoxelCentroidNearestNeighbour downsampling filter does not check that the nearest neighbour point to the calculated centroid of the voxel cell is actually within that current voxel. As a result a voxel may _choose_ an output point that is not inside it.

I've attached a small point cloud and a pipeline which exhibits this behaviour. Point number 28 (X=6126.592, Y=5145.142, Z=50.000) appears twice in the output file. 

image

chambbj:
Previously, VoxelCentroidNearestNeighbor could return points outside the
current voxel. This happened because we search a global KDIndex for the nearest
neighbor of the voxel centroid, which could in fact come from a neighboring
voxel.

The refactored implementation ensures that we are only considering neighbors
within the current voxel when determining which neighbor is nearest the
centroid.

Furthermore, this refactor also addresses some special cases, namely voxels
containing only one point (it simply passes through), and those containing two
points which by definition are equidistant to the centroid. In this case, we
revert to the VoxelCenterNearestNeighbor approach and select the point that is
closest to the voxel center.

Fix #2664

### DIFF
--- a/filters/VoxelCentroidNearestNeighborFilter.cpp
+++ b/filters/VoxelCentroidNearestNeighborFilter.cpp
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright (c) 2017, Bradley J Chambers (brad.chambers@gmail.com)
+ * Copyright (c) 2019, Bradley J Chambers (brad.chambers@gmail.com)
  *
  * All rights reserved.
  *
@@ -35,7 +35,6 @@
 #include "VoxelCentroidNearestNeighborFilter.hpp"
 
 #include <pdal/EigenUtils.hpp>
-#include <pdal/KDIndex.hpp>
 
 #include <map>
 #include <string>
@@ -71,8 +70,6 @@ PointViewSet VoxelCentroidNearestNeighborFilter::run(PointViewPtr view)
     BOX3D bounds;
     calculateBounds(*view, bounds);
 
-    KD3Index& kdi = view->build3dIndex();
-
     // Make an initial pass through the input PointView to index PointIds by
     // row, column, and depth.
     std::map<std::tuple<size_t, size_t, size_t>, std::vector<PointId>>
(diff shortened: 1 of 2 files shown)


filters/VoxelCentroidNearestNeighborFilter.cpp
```diff
@@ -71,8 +70,6 @@ PointViewSet VoxelCentroidNearestNeighborFilter::run(PointViewPtr view)
     BOX3D bounds;
     calculateBounds(*view, bounds);
```
abellgithub: calculateBounds makes a full pass through the data.  I think this can be avoided since it seems like we're just using the min x/y/z as the origin for the voxel space.  I think using the value of the first point would do just as well if the r/c/d calculation were modified a bit.

filters/VoxelCentroidNearestNeighborFilter.cpp
```diff
@@ -93,10 +90,67 @@ PointViewSet VoxelCentroidNearestNeighborFilter::run(PointViewPtr view)
...
+        else
+        {
+            // Else there are more than two neighbors, so choose the one
+            // closest to the centroid.
+
+            // Compute the centroid.
+            Eigen::Vector3d centroid = computeCentroid(*view, t.second);
+
+            // Compute distance from each point in the voxel to the centroid,
+            // retaining only th closest.
```
abellgithub: Missing 'e'

dbrookes96:
This may not be the correct place for this, but its something I thought worth mentioning. This downsampler and the alternative VoxelCentreNearestNeighbour make use of std::map to index the cells. Running a profiler over this show that a lot of time is spent in std::map find. There is likely a measurable performance improvement available by moving to std::unordered_map instead. Hashing should generally be faster than a tree search here.

Note this may require a hash operator to be defined (I don't remember if std::hash supports something like std::tuple<size_t,size_t,size_t> out of the box) i.e.:
```
struct CellIndex
  {
  size_t x;
  size_t y;
  size_t z;
  };

struct CellIndexHash
  {
  size_t operator()(const CellIndex& cellIndex) const
    {
    size_t out = std::hash<size_t>()(cellIndex.x)
      ^ (std::hash<size_t>()(cellIndex.y) << 1)
      ^ (std::hash<size_t>()(cellIndex.z) << 2);
    return out;
    }
  };

struct CellIndexEqual
  {
  bool operator()(const CellIndex& lhs, const CellIndex& rhs) const
    {
    return lhs.x == rhs.x && lhs.y == rhx.y && lhs.z == rhs.z;
    }
  };

using CellIndexMap = std::unordered_map<CellIndex, std::vector<PointId>, CellIndexHash, CellIndexEqual>;
```
This shouldn't require many other change to the code which is bonus. I would expect a >1/3 performance improvement (for large point clouds). Ofcourse the possibility of multi-threading this by dividing up the input may also be a future thing to look at for larger point clouds.

abellgithub:
You're welcome to submit a PR.  I think that for this application, the normal case is that the X/Y/Z values are relatively small, so I would think you could get near-perfect hashing by preserving the low bits of each.

PR #2704, opened 2019-08-29
Add power parameter to GDAL writer

### LINKED ISSUE #2550: Request for user-specified power in IDW with writers.gdal
Spatial interpolation using inverse distance weighting (IDW) requires a power term (AKA weighting coefficient). The power term currently employed in writers.gdal is not reported in the the documentation, which limits reproducibility (_I would have expected to see it [here](https://pdal.io/stages/writers.gdal.html) under IDW, apologies if I missed it elsewhere_).

I think it would be helpful to:
 
1.  State the default power term in the documentation (_2 is a common default, e.g. [ESRI's ArcMap](https://pro.arcgis.com/en/pro-app/tool-reference/spatial-analyst/idw.htm)_)

2. Implement an optional 'power' argument to writers.gdal, to allow users to customise this parameter for different applications.

Less important, but it might also be helpful to be explicit in the documentation about how coincident (identical) points are handled, as this can differ between implementations of IDW (some will take all points, others the first/last only; for further discussion of this nuance see [here](https://pro.arcgis.com/en/pro-app/tool-reference/spatial-analyst/idw.htm)).

chambbj:
IDW can be more heavily influenced by near points if desired. Default value of 2.0 represents a change from the previous behavior, when no power was considered.

Close #2550

### DIFF
--- a/io/private/GDALGrid.hpp
+++ b/io/private/GDALGrid.hpp
@@ -62,7 +62,7 @@ public:
 
     // Exported for testing.
     PDAL_DLL GDALGrid(size_t width, size_t height,
-        double edgeLength, double radius, int outputTypes, size_t windowSize);
+        double edgeLength, double radius, int outputTypes, size_t windowSize, double power=2.0);
 
     void expand(size_t width, size_t height, size_t xshift, size_t yshift);
 
@@ -90,6 +90,7 @@ private:
     size_t m_windowSize;
     double m_edgeLength;
     double m_radius;
+    double m_power;
 
     typedef std::vector<double> DataVec;
     typedef std::unique_ptr<DataVec> DataPtr;
--- a/io/private/GDALGrid.hpp
+++ b/io/private/GDALGrid.hpp
@@ -62,7 +62,7 @@ public:
 
     // Exported for testing.
     PDAL_DLL GDALGrid(size_t width, size_t height,
-        double edgeLength, double radius, int outputTypes, size_t windowSize, double power=2.0);
+        double edgeLength, double radius, int outputTypes, size_t windowSize, double power);
 
     void expand(size_t width, size_t height, size_t xshift, size_t yshift);
 
--- a/doc/stages/writers.gdal.rst
+++ b/doc/stages/writers.gdal.rst
@@ -112,6 +112,15 @@ radius
     Radius about cell center bounding points to use to calculate a cell value.
     [Default: resolution_ * sqrt(2)]
 
+power
+    Exponent of the distance when computing IDW. Close points have higher
+    significance than far points. [Default: 2.0]
+
+    .. note::
+	Power is a newly added, optional argument, only used when computing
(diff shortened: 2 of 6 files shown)


abellgithub:
Is there a good reason to change the previous behavior?

io/private/GDALGrid.hpp
```diff
@@ -62,7 +62,7 @@ class GDALGrid
 
     // Exported for testing.
     PDAL_DLL GDALGrid(size_t width, size_t height,
-        double edgeLength, double radius, int outputTypes, size_t windowSize);
+        double edgeLength, double radius, int outputTypes, size_t windowSize, double power=2.0);
```
abellgithub: Don't see any reason to set this as a default here.

chambbj:
I was on the fence about it. OP seemed to indicate that 2.0 may be the norm for users who are looking for it, but I don't love that unsuspecting users may get stuck with it either.

abellgithub:
I vote to leave it at one, but put a note in that says that the ESRI tools use a default of 2.

hobu:
> I vote to leave it at one, but put a note in that says that the ESRI tools use a default of 2.

I concur.

AndrewCunliffe:
Thanks folks, sorry I saw this thread late.

A default of 1 that can be changed would be fine for my purpose and most I can imagine.

I think the default power for IDW is normally 2 across a range of software ([ESRI's ArcPro](https://pro.arcgis.com/en/pro-app/tool-reference/spatial-analyst/idw.htm), [QGIS](https://docs.qgis.org/2.8/en/docs/user_manual/processing_algs/gdalogr/gdal_analysis/gridinvdist.html), and also in most research articles ([example 1](https://www.sciencedirect.com/science/article/pii/S027843431100197X), [example 2](http://spatial-accuracy.org/system/files/ChaokuiLi2008accuracy.pdf)[, example 3](https://besjournals.onlinelibrary.wiley.com/doi/full/10.1111/2041-210X.13178)), [ Example 4](https://www.tandfonline.com/doi/pdf/10.1080/21580103.2010.9671977). The influence of this parameter (1 vs. 2) is illustrated [here ](https://gisgeography.com/inverse-distance-weighting-idw-interpolation/).

I'm not saying that a value of 2 is always appropriate, but you should be aware that it does seem to be the most common default.

Thanks again for all of your efforts to keep making PDAL awesome!

hobu:
> The default power for IDW is normally 2 across a range of software (ESRI's ArcPro, QGIS and the R package 'lidR', and GITTA).

ok, alignment with these tools is compelling for me. I see no reason to be different. The value of 1 originated in Points2grid, I believe.

abellgithub:
Not convinced.  Someone running EXISTING code with this released is going to get different behavior that matters.  It think we need some incredibly compelling reason to change this and I don't think the choice of other software packages is sufficient.  People have lived without this feature and it's easy to set the value as you wish.

PR #2716, opened 2019-09-03
Draft bug_report.md

chambbj:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,55 @@
...
+Please note that GitHub issues are meant to report bugs or feature requests, and are not intended for "How to" questions or general support. GitHub issues requesting such support will be closed. Luckily, we have an active community on the GIS StackExchange [1] and the PDAL Mailing List [2] that are more than happy to chime in with advice in these situations!
+
+[1] https://gis.stackexchange.com/questions/tagged/pdal
+[2] https://pdal.io/community.html#mailing-list
+-->
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
```
abellgithub: Think this should be merged with the above.  What I want 99% of the time is the exact command someone has run along with a pipeline file if that's relevant.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,55 @@
...
+Steps to reproduce the behavior.
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**System/installation information:**
+Please provide information on your PDAL version and system.
+
+If installed via Conda, please paste the output of `conda list` and `conda info`:
```
abellgithub: This seems like way too much junk to wade through in most cases.  I really don't need this except with specific install problems and to me it just clutters up the report otherwise.

hobu: I need it sometimes. We end up asking for it and it gives them template to give us what we need to figure out what versions of dependencies might be used and causing problems.

.github/ISSUE_TEMPLATE/bug_report.md
chambbj: Agreed, and I had actually intended to ask for the pipeline too, just forgot.

.github/ISSUE_TEMPLATE/bug_report.md
chambbj: I know we've debated this before. I mean, at least it is folded up in the details block, and so not necessarily cluttering the report when viewed via GitHub. At the same time, maybe not too hard to ask for it when needed. I guess I don't have a super strong opinion either way, though I personally lean a little towards collecting as much information up front as possible to reduce the number of roundtrips with the reporter, but then I'm not fielding most bug reports either.

abellgithub: I think you can ask for it if you need it, but if we're going to do it, put the conda junk at the end.  Put the PDAL version/system items up top.

chambbj:
Tweaked wording/order a bit @abellgithub @hobu. Let me know if you think it addresses the concerns.

.github/ISSUE_TEMPLATE/bug_report.md
hobu: > put the conda junk at the end. Put the PDAL version/system items up top

Agree.

abellgithub:
Do we have to do something to get it to actually work?

chambbj:
Once merged, submitters should be prompted with this content to edit as necessary. Is that what you are asking?

abellgithub:
Just checking.  Seems fine to me.

Issue #2717, opened 2019-09-04
Type decorated dimensions for performance and simpler compile time type checking


wrenoud:
Adding structs for each Dimension decorated with the C type in Dimension.hpp (with the DimBuilder) would allow for skipping some switches and casting in getting and setting dimensions that should improve performance and push type checking out to the code trying to access the dimension at compile time.

This does not remove the need for existing dimension architecture which is needed for dimensions registered at runtime, this would just accelerate access to compile-time known dimensions.

Below is an outline of the idea in code.

`Dimension.hpp`
```cpp
namespace Dimension{
namespace Trait{
struct X
{
    typedef double_t value_type;
    static constexpr Id id() { return Id::X; }
    static constexpr Dimension::Type type() { return Dimension::Type::Double; }
};
...
}
}
```

New templated overloads on `PointRef` and potential child classes of `PointContainer` can then make use of the Trait struct to skip switches on the type enum and type casting.

`PointRef.hpp`
```cpp
class PointRef
{
public:
    ...
    template<typename T>
    void setField(typename T::value_type val)
    {
        m_container.setFieldInternal(T::id(), m_idx, &val);
    }

    template<typename T>
    typename T::value_type getFieldAs() const
    {
        typename T::value_type val;
        m_container.getFieldInternal(T::id(), m_idx, &val);
        return val;
    }
    ...
};
```

Finally, passing the trait struct as the template parameter provides the quicker access.
`LasReader.hpp`
```cpp
void LasReader::loadPointV10(PointRef& point, laszip_point& p)
{
    const LasHeader& h = m_header;

    double x = p.X * h.scaleX() + h.offsetX();
    ...
    point.setField<Dimension::Trait::X>(x);
    ...
}
```

I have a PoC that I will add to a pull request. Initial testing on Windows with MSVC 15 (2017) showed a 30% speedup for LasReader::loadPointV10

## Reading 12 laz with the existing interface (~9sec)
![image](https://user-images.githubusercontent.com/2761756/64243069-ffceb100-cf06-11e9-8df0-5b09c4ec11ab.png)

## Reading 12 laz with the proposed interface (~6sec)
![image](https://user-images.githubusercontent.com/2761756/64243176-33114000-cf07-11e9-806c-5a4172a77da2.png)

Both approaches spent approximately 4 seconds in `void SimplePointTable::setFieldInternal(Dimension::Id id, PointId idx, const void *value)`

abellgithub:
[...]

PR #2718, opened 2019-09-04
WIP Dimension traits

wrenoud:
PoC for #2717

### DIFF
--- a/dimbuilder/DimBuilder.cpp
+++ b/dimbuilder/DimBuilder.cpp
@@ -228,7 +228,8 @@ void DimBuilder::extractDim(NL::json& dim)
             "string.";
         throw dimbuilder_error(oss.str());
     }
-    d.m_type = Dimension::type(dimType.get<std::string>());
+    d.m_typeName = dimType.get<std::string>();
+    d.m_type = Dimension::type(d.m_typeName);
     if (d.m_type == Dimension::Type::None)
     {
         std::ostringstream oss;
@@ -330,6 +331,8 @@ void DimBuilder::writeOutput(std::ostream& out)
     out << "\n";
     writeIds(out);
     out << "\n";
+    writeTraitStructs(out);
+    out << "\n";
     writeDescriptions(out);
     out << "\n";
     writeNameToId(out);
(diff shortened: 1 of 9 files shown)


abellgithub:
This won't work because we don't know the dimension sizes at compile time.  Sorry.

wrenoud:
> This won't work because we don't know the dimension sizes at compile time. Sorry.

Ah! Bummer. So the size may change despite the intended c type?

abellgithub:
The size that's listed in the dimension file are default sizes that are used if a stage doesn't register a dimension with a specific size.  File formats have varying ways of storing data.  For instance, BPF uses single-precision floating point for all dimensions other than X/Y/Z.  If you're using BPF, you may well want 4 bytes for classification.  On the other hand, some people are really concerned with memory burn and want fields as small as possible that still accommodate their data.  There is a definite run-time cost to this as you've seen, but it provides a flexibility and ease-of-use that is part of the tradeoff.

For LAS, which is the dominant format, only loading the fields you really need seems like the low-hanging fruit for a throughput boost (most people don't care about all the fields, especially in 1.4 formats).  There might also be some other optimizations to be made WRT avoiding the switch statements, but I haven't spent tons of time there.  I did make an attempt to change some ordering to things in the LAS reader to try to get speculative execution to make a dent, but it didn't make enough difference to matter.

We _assume_ that X,Y and Z are double-precision floating, and we could probably codify that with no downside and get a little boost, but the cast is already eliminated in the set/get template in those cases.  You still have to switch on type, though.  You could do other things as a programmer, but we're also trying to make this easy for API users-- there's often a tradeoff between performance and ease of use.

We do little threading (on purpose, generally) and no GPU-accelerated processing.  Provided it can be done well, I'd have no problem using these tools.

I'm all for improving performance, but I also have tried to 1) maintain the existing interfaces 2) maintain flexibility 3) make the API relatively easy to use and difficult to misuse.

wrenoud:
@abellgithub again, thanks for taking the time to respond. I appreciate the design decisions you've made and understand the tradeoffs. I thought I had a quick win I could carry over from my own internal implementation of point abstraction. But yeah, I have a little more flexibility around the tradeoffs.

wrenoud:
Ok, I just made the accessors fall back to normal pipeline if types don't match. So this would accelerate sometimes, at the cost of I think a trivial comparison.

wrenoud:
Although, just tested again and I think my recent change burns up most of the gain... which means it will slow down normal access.

abellgithub:
I did this as a test:

https://github.com/PDAL/PDAL/tree/quicktest

On OSX with release mode I get essentially zero improvement reading/writing LAS.  This surprises me if you're seeing 30% improvement with your change.  Wondering what might be going on.

wrenoud:
@abellgithub looking at your code and knowing you don't see an improvement suggests the switch isn't what was costly that I skipped. I'm looking at the dimension detail lookup from the layout. Looks like it gets done a couple of times. I'm wondering if it could be passed on instead of accessed multiples times.

abellgithub:
I tried avoiding that as well by placing it in a vector that was built at the time that the layout was finalized, also to little effect.

wrenoud:
Tried minimizing detail lookup here: https://github.com/wrenoud/PDAL/tree/try-reducing-details-lookup

Nothing impressive, as you noted, but marginally better _on windows_, which may be the factor here... At my work we've definitely seen better performance with some task in our software on non-windows platforms.

abellgithub:
I'll run stuff on Windows and see what I find.

hobu:
Closing this for now. Please reopen if there's to be more activity on it.

PR #2719, opened 2019-09-05
Visual Studio CMake tweaks

wrenoud:
These are a couple of tweaks I've made to make the Visual Studio environment a little friendlier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,16 +190,16 @@ if (APPLE)
 endif()
 
 file(GLOB BASE_SRCS
-    ${PDAL_FILTERS_DIR}/*.cpp
+    ${PDAL_FILTERS_DIR}/*.?pp
     ${PDAL_IO_DIR}/*.cpp
-    ${PDAL_KERNELS_DIR}/*.cpp
-    ${PDAL_SRC_DIR}/*.cpp
-    ${PDAL_SRC_DIR}/compression/*.cpp)
+    ${PDAL_KERNELS_DIR}/*.?pp
+    ${PDAL_SRC_DIR}/*.?pp
+    ${PDAL_SRC_DIR}/compression/*.?pp)
 file(GLOB_RECURSE PRIVATE_SRCS
-    ${PDAL_FILTERS_DIR}/private/*.cpp
-    ${PDAL_IO_DIR}/private/*.cpp
-    ${PDAL_KERNELS_DIR}/private/*.cpp
-    ${PDAL_SRC_DIR}/private/*.cpp)
+    ${PDAL_FILTERS_DIR}/private/*.?pp
+    ${PDAL_IO_DIR}/private/*.?pp
+    ${PDAL_KERNELS_DIR}/private/*.?pp
+    ${PDAL_SRC_DIR}/private/*.?pp)
 list(APPEND SRCS ${BASE_SRCS} ${PRIVATE_SRCS})
 
 #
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -106,6 +106,25 @@ include(${PDAL_CMAKE_DIR}/options.cmake)
 
 set(PDAL_BUILD_TYPE ${CMAKE_BUILD_TYPE})
 
+# detect if we're setting up a visual studio solution
```
abellgithub: This shouldn't go here.  If you want vsconfig.cmake in the cmake subdir, that's fine.

I don't know what this is about, but it looks pretty specific and subject to churn.

What does this file do?

wrenoud: Unless you start Visual Studio from the command line in a conda environment, trying to run executables doesn't work because it can't find dependencies. This is just trying to preserve the path to the conda env so Visual Studio will set it in the environment when it runs the executable.

An alternative would be to set the working directory, but then pdal wouldn't see plugins unless they're installed to the conda env

abellgithub: I get that the conda path isn't set when you start VS by clicking, but how does this help?  Just trying to understand VS, which I use almost never.

wrenoud: The `user.props` file I'm making can be referenced/linked by a visual studio project, and will be the new defaults for the project settings unless specifically overridden by the project.

Specify the `LocalDebuggerEnvironment` will set environment variables, in this case I'm overriding the `PATH` variable to add the path to the dependencies (aka `%CONDA_PREFIX%/Libraries/bin`) to make them discoverable, kind of manual getting the conda environment setup.

I'm assuming the plugin install path `PDAL_PLUGIN_INSTALL_PATH` is the appropriate path here. But it may be worth adding a more specific cmake variable.

abellgithub:
Yeah, I just don't understand why in a dev environment the PDAL install path would be at all related to conda. I never set the install path when doing development.

hobu:
Is this ready for merge?

abellgithub:
I don't see how this works in a general case.  See comment above.

CMakeLists.txt
```diff
@@ -190,55 +194,55 @@ if (APPLE)
...
 file(GLOB BASE_SRCS
-    ${PDAL_FILTERS_DIR}/*.cpp
+    ${PDAL_FILTERS_DIR}/*.?pp
     ${PDAL_IO_DIR}/*.cpp
-    ${PDAL_KERNELS_DIR}/*.cpp
-    ${PDAL_SRC_DIR}/*.cpp
-    ${PDAL_SRC_DIR}/compression/*.cpp)
+    ${PDAL_KERNELS_DIR}/*.?pp
+    ${PDAL_SRC_DIR}/*.?pp
+    ${PDAL_SRC_DIR}/compression/*.?pp)
```
chambbj: If globbing the headers too, why skip those in the `PDAL_IO_DIR`?

cmake/vsconfig.cmake
```diff
@@ -0,0 +1,23 @@
...
+    set(VCPROJ ON)
+endif()
+
+# add user props to build directory to add to project settings
+# for running targets in the IDE
+if(VCPROJ)
+    file(GENERATE
+        OUTPUT "${CMAKE_BINARY_DIR}/user.props"
+        CONTENT "<?xml version='1.0' encoding='utf-8'?>
+<Project ToolsVersion='15.0' xmlns='http://schemas.microsoft.com/developer/msbuild/2003'>
```
chambbj: This looks fair specific to a single version of MSVC. Is that not the case?

CMakeLists.txt
abellgithub: I kind of hate this in general.  This seems like something for VS to deal with, not a project like PDAL.  Maybe it wouldn't bug me if it were put off in a corner somewhere for VS where it didn't pollute the rest of the build system.

cmake/vsconfig.cmake
abellgithub: Right.  And what I don't get is why we'd be using PDAL_PLUGIN_INSTALL_PATH.  The issue was finding DLLs from the debugger env.  There was talk of using the conda PATH.  This makes sense, but I'm not sure what the PDAL_PLUGIN_INSTALL_PATH has to do with a PATH you would use to find libgdal.dll.  The assumption here is that they're the same, I think.  That hasn't been the case for me when I've used VS.  I haven't even set the install path.  I looked for some ENV VAR that conda sets as its part of the PATH env, but didn't find one.  And even in that case, it's env. specific.  You might not be using conda.  I guess this just doesn't seem like a PDAL build issue.  It seems like a general environment issue.  Maybe someone should make a program that creates this `user.props` given a set of paths that you run for your environment.

All that said, I could be misunderstanding.

CMakeLists.txt
naruephon2541: > ฉันเกลียดสิ่งนี้โดยทั่วไป ดูเหมือนว่าบางสิ่งบางอย่างสำหรับ VS ที่จะจัดการไม่ใช่โครงการอย่าง PDAL บางทีมันอาจจะไม่ใช่ข้อผิดพลาดของฉันถ้ามันถูกนำไปวางไว้ที่มุมหนึ่งสำหรับ VS ซึ่งมันไม่ได้ทำให้ระบบการสร้างที่เหลืออยู่เสียไป

mloskot:
Generally, polluting 'generic' CMake configuration with IDE specific features smells and I'm not sure why CMake even allows that. My experience with CMake taught me that CMake scripts should be simple and minimal, written as much declarative as possible, without too much of clever engineering, with as little custom modules and macros as possible. Unless, maintainers are truly have an affection to CMake scripting.

Specifically, for VS 2017/2019, there are intermediary file sto handle VS specifics while building a project with CMake.
Namely, `CMakeSettings.json` and `launch.vs.json` which can be used for setting up build, debugger and run-time environment.

What projects like PDAL can do is offer such files with reasonable defaults as examples. This, is what did for example here https://github.com/boostorg/gil/tree/develop/example/cmake and https://github.com/nanodbc/nanodbc/tree/master/utility/vs2017

### References:

- https://docs.microsoft.com/en-us/cpp/build/cmakesettings-reference
- https://docs.microsoft.com/en-us/cpp/build/launch-vs-schema-reference-cpp
- https://docs.microsoft.com/en-us/cpp/build/configure-cmake-debugging-sessions
- https://docs.microsoft.com/en-us/visualstudio/ide/customize-build-and-debug-tasks-in-visual-studio

hobu:
Closing for now.

PR #2720, opened 2019-09-05
Small voxel filter fixes

abellgithub:
Make filter name case consistent with others.
Add filter to stage index doc.
Compute voxel using floor() to eliminate errors around origin.
Make Streamable public base to allow direct execution.
Use return value of insert() to eliminate the need to find() when inserting into map.


doc/stages/filters.rst
```diff
@@ -227,6 +228,9 @@ the input. These filters will invalidate an existing KD-tree.
     Remove points that are in a raster cell but have a value far from the
     value of the raster.
 
+:ref:`filters.firstinvoxel`
+    Filter out points that aren't the first encountered that lie in a voxel.
```
chambbj: Only comment on the PR is that this wording is bit awkward still. Maybe "Retain only the first detected point per voxel."

Issue #2721, opened 2019-09-06
E57 reader Windows performance issue.


dbrookes96:
Hi Guys.

We include the PDAL C++ API in our software to convert point cloud files between formats.
We've upgraded to PDAL 2.0-maintenance to try out the new E57 plugin.
The E57 reader plugin suffers from a large performance issue (maybe only on Windows) caused by its use of exceptions inside its processOne loop. The code in question is in `E57Reader::ChunkReader::setPoint`.
```
            double value;
            // Rescale the value if needed so that if fits Pdal expectations
            try
            {
                auto minmax = m_scan->getLimits(pdalDimension);
                value = pdal::e57plugin::rescaleE57ToPdalValue(keyValue.first,keyValue.second[index],minmax);
            }
            catch (std::out_of_range &e)
            {
                value = keyValue.second[index];
            }
            point.setField(pdalDimension,value);
```
The code relies on `std::map` throwing `std::out_of_range` to determine if it contains limits for the dimension. The is being thrown/caught multiple times for each point (for each dimension). I think using `std::map` `find` would be a better option here.
I've patched this locally and for a point cloud of ~4 million points it took the processing time down from ~1 minutes to just ~6 seconds. It is even worse when running with a debugger taking around ~100ms per point.
I'll try and submit a PR for this soon.

I appreciate that this reader is very new and performance is not the priority over stability. The reader seems to be working well for the files I've tested so far. There may also be some other places we can squeeze out more performance but this seemed the most immediate.

Regards.

**System/installation information:**
PDAL 2.0-maintenance @ d1eaa6d5c70724c3a4637301fc864c293ad43bee
Built on VS2019 v142

nicolas-chaulet:
Awesome!!! I knew something was wrong with performance but could not pin point what... Thanks! There might be something up with the writer as well...

nicolas-chaulet:
I think a PR would be welcome!

abellgithub:
Fixed with https://github.com/PDAL/PDAL/commit/a27c1beadbe14ef6487f7a82ab2b507f08288eb2

PR #2722, opened 2019-09-06
Use map find instead of at (#2721)

dbrookes96:
Fixes performance issues on Windows caused by exceptions.

### DIFF
--- a/plugins/e57/io/E57Reader.cpp
+++ b/plugins/e57/io/E57Reader.cpp
@@ -103,13 +103,13 @@ void E57Reader::ChunkReader::setPoint(point_count_t pointIndex, PointRef point,
         {
             double value;
             // Rescale the value if needed so that if fits Pdal expectations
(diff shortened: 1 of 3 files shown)


abellgithub:
I've got no problem with this change, but I'll note that I think there's a lot that can be improved in the general logic to eliminate a whole bunch of this code.

PDAL doesn't limit the size of any dimension.  That size is limited at the time of registration by readers/filters.  If you need signed 32 bits to represent values that come from your source, register the dimension as Signed32.  PDAL takes care of fixing up any discrepancies between various stages using the same dimension and casting values as necessary to harmonize values coming from different sources.  I thought about making a pass at cleaning this up, but haven't had time.  But if someone else wants to take a whack at it, that would be great.  I'd be happy to provide feedback.

hobu:
I think this should be merged back to `2.0-maintenance`.

PR #2724, opened 2019-09-08
Separate scan lines filter

gui2dev:
This is a basic implementation for filter that separates scan lines, using the ``EdgeOfFlightLine`` dimension.


filters/SeparateScanLineFilter.cpp
```diff
@@ -0,0 +1,98 @@
...
+{
+    PointLayoutPtr layout(table.layout());
+    if (!layout->hasDim(Dimension::Id::EdgeOfFlightLine))
+        throwError("Layout does not contains EdgeOfFlightLine dimension.");
+}
+
+PointViewSet SeparateScanLineFilter::run(PointViewPtr inView)
+{
+    PointViewSet result;
+    std::vector<PointViewPtr> views;
```
abellgithub: Think this is unused.

filters/SeparateScanLineFilter.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+class PDAL_DLL SeparateScanLineFilter : public Filter
+{
+public:
+    SeparateScanLineFilter();
+
+    std::string getName() const;
+
+private:
+    std::map<uint64_t, PointViewPtr> m_viewMap;
```
abellgithub: Unused

filters/SeparateScanLineFilter.hpp
```diff
@@ -0,0 +1,67 @@
...
+    uint64_t m_groupBy;
```
abellgithub: The type of this variable should be the same as the type of `i` in the processing loop.

filters/SeparateScanLineFilter.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+    virtual void addArgs(ProgramArgs& args);
+    virtual void prepared(PointTableRef table);
+    virtual PointViewSet run(PointViewPtr view);
+
+    SeparateScanLineFilter& operator=(const SeparateScanLineFilter&); // not implemented
```
abellgithub: If you could move these to be tagged as `= deleted` that would be great.

gui2dev: @abellgithub, not sure what you mean

abellgithub: NP  I'll take care of it.  Thanks again.

PR #2725, opened 2019-09-08
[readers.rxp] Expose edge of flight line

gui2dev:
Get scan line start info from riegl rxp format.

### DIFF
--- a/plugins/rxp/io/RxpPointcloud.cpp
+++ b/plugins/rxp/io/RxpPointcloud.cpp
@@ -52,10 +52,11 @@ Dimension::Id getTimeDimensionId(bool syncToPps)
 }
 
 
-Point::Point(scanlib::target target, unsigned int returnNumber, unsigned int numberOfReturns)
+Point::Point(scanlib::target target, unsigned int returnNumber, unsigned int numberOfReturns, bool edgeOfFlightLine)
     : target(target)
     , returnNumber(returnNumber)
     , numberOfReturns(numberOfReturns)
+    , edgeOfFlightLine(edgeOfFlightLine)
 {}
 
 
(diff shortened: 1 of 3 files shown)


plugins/rxp/io/RxpPointcloud.cpp
```diff
@@ -130,6 +132,7 @@ void RxpPointcloud::copyPoint(const Point& from, PointRef& to) const {
     to.setField(Id::Deviation, from.target.deviation);
     to.setField(Id::BackgroundRadiation, from.target.background_radiation);
     to.setField(Id::IsPpsLocked, from.target.is_pps_locked);
+    to.setField(Id::EdgeOfFlightLine, from.edgeOfFlightLine?1:0);
```
abellgithub: Some spacing around the `? :` would be nice.

plugins/rxp/io/RxpPointcloud.cpp
```diff
@@ -163,9 +166,22 @@ void RxpPointcloud::on_echo_transformed(echo_type echo)
     unsigned int returnNumber = 1;
     for (scanlib::pointcloud::target_count_type i = 0; i < target_count; ++i, ++returnNumber)
     {
-        m_points.emplace_back(targets[i], returnNumber, target_count);
+        m_points.emplace_back(targets[i], returnNumber, target_count,m_edge);
+        if (m_edge)
```
abellgithub: A comment explaining that all points after the first aren't marked as EOF might be nice.

A space before `m_edge` above as well.

gui2dev: @abellgithub, actually, first point is marked as EOF where it's actually BOF

PR #2727, opened 2019-09-09
Check if inView is empty before setting m_size

### LINKED ISSUE #2726: [filters.divider] Bad behaviour when inView is empty
**Describe the bug**
When trying to divide a empty cloud, PDAL starts allocating memory in what seems a never ending loop.

<details>
```
{
 "pipeline":
 [
  {
   "type":"readers.faux",
   "bounds":"([0,1000000],[0,1000000],[0,100])",
   "count":"10000",
   "mode":"random"
  },
  {
   "type":"filters.range",
   "limits":"Z[200:300]"
  },
  {
   "type":"filters.divider",
   "capacity":100
  }
 ]
}
```
</details>

**Expected behavior**
PDAL should exit smoothly

```
$ uname -a
Linux DESKTOP-N8C3JIA 4.4.0-17763-Microsoft #379-Microsoft Wed Mar 06 19:16:00 PST 2019 x86_64 x86_64 x86_64 GNU/Linux
```

```
$ pdal --version
pdal 2.0.0 (git-version: 5c2dab)
```
Also on windows 10 with version 1.9

gui2dev:
Check if inView is empty before setting m_size to avoid integer overflow

Fixes #2726

### DIFF
--- a/filters/DividerFilter.cpp
+++ b/filters/DividerFilter.cpp
@@ -109,11 +109,18 @@ void DividerFilter::initialize()
 
 PointViewSet DividerFilter::run(PointViewPtr inView)
 {
-    if (m_sizeMode == SizeMode::Capacity)
-	m_size = inView->empty() ? 1 : (((inView->size() - 1) / m_size) + 1);
-
-
     PointViewSet result;
+
+    if (inView->empty())
+    {
+	result.insert(inView);
+	return result;
+    }
+
+    if (m_sizeMode == SizeMode::Capacity)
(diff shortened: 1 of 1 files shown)


abellgithub:
Thanks for finding this. I think the better solution is just to return in this case and avoid having to figure out what happens when you set the value to one.

gui2dev:
So  returning an empty ```PointViewSet``` would be okay?

abellgithub:
It doesn't matter. Either is fine. I would just return the input PointView in the PointViewSet.

gui2dev:
That should do the trick then.

PR #2737, opened 2019-09-11
Allow user to seed random number generator when shuffling points

### LINKED ISSUE #2736: filters.randomize should be seeded, switch from random_shuffle to shuffle
`filters.randomize` is not currently seeded and can return not-really-all-that-random results.

As an aside, it looks like std::random_shuffle is eventually [going away](https://meetingcpp.com/blog/items/stdrandom_shuffle-is-deprecated.html) and given that we are using C++11, we may as well go ahead and switch to std::shuffle.

In addition to creating a uniform random number generator and properly seeding it, we could allow the user to specify a seed for reproducible results (especially useful in the limited unit tests).

chambbj:
This change also defaults to using random_device to seed the mt19937 random
number generator.

Close #2736

### DIFF
--- a/filters/RandomizeFilter.cpp
+++ b/filters/RandomizeFilter.cpp
@@ -1,5 +1,6 @@
 /******************************************************************************
  * Copyright (c) 2015, Hobu Inc. (hobu@hobu.co)
+ * Copyright (c) 2019, Bradley J Chambers (brad.chambers@gmail.com)
  *
  * All rights reserved.
  *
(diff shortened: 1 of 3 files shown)


chambbj:
Yeah, of course the shuffle is going to depend on the machine/compiler. Will have to think of a better way to test this.

chambbj:
Tests could be simplified and commented, otherwise ready for review.

PR #2741, opened 2019-09-11
Add Bartels & Wei skewness balancing approach to ground segmentation

### LINKED ISSUE #2740: Add Skewness Balancing approach to DTM generation
Bartels and Wei published a couple of papers on DTM generation using a method called Skewness Balancing [1, 2]. While these do not necessarily compete with more recent/mature approaches like SMRF, they are straightforward and conveniently parameter-free. Plus, I coded the basic approach [1] several years ago, just never committed. Part of the work in [2] goes to revise the DTM by recursively apply Skewness Balancing on the detected object points, which is supposed to help in sloped terrain.

[1] Bartels, Marc, Hong Wei, and David C. Mason. "DTM generation from LIDAR data using skewness balancing." 18th International Conference on Pattern Recognition (ICPR'06). Vol. 1. IEEE, 2006.

[2] Bartels, Marc, and Hong Wei. "Threshold-free object and ground point separation in LIDAR data." Pattern recognition letters 31.10 (2010): 1089-1099.

chambbj:
The basic approach only; no bells and whistles.

Future work _could_ include the Revised Skewness Balancing approach and addition of commonly used convenience parameters such as limiting points to certain return groups or ignoring specified DimRanges.

Close #2740

### DIFF
--- a/filters/SkewnessBalancingFilter.cpp
+++ b/filters/SkewnessBalancingFilter.cpp
@@ -1,0 +1,137 @@
+/******************************************************************************
+ * Copyright (c) 2019, Bradley J Chambers (brad.chambers@gmail.com)
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following
+ * conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided
+ *       with the distribution.
+ *     * Neither the name of Hobu, Inc. or Flaxen Geo Consulting nor the
+ *       names of its contributors may be used to endorse or promote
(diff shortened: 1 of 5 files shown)


filters/SkewnessBalancingFilter.cpp
```diff
@@ -0,0 +1,137 @@
...
+    auto cmp = [](const PointIdxRef& p1, const PointIdxRef& p2) {
+        return p1.compare(Dimension::Id::Z, p2);
+    };
+    std::sort(view->begin(), view->end(), cmp);
+
+    point_count_t n(0);
+    point_count_t n1(0);
+    double delta, delta_n, term1, M1, M2, M3;
+    M1 = M2 = M3 = 0.0;
+    std::map<PointId, double> skewness;
```
abellgithub: I don't understand why this is a map.  Unless I'm missing something, skewness is always accessed by pointId, so PointIdList would work.

chambbj: I probably changed course along the way. That said, it's not a PointIdList, as we are actually interested in the skewness, which is a double. I think vector<double> will do the trick.

PR #2747, opened 2019-09-17
[WIP] Add little description on transformation filter

vvmruder:
I puzzled around a little with the transformation filter and thought some details in description might be welcome.


abellgithub:
This renders OK for me.

Actually, this confused me a bit.  I was trying to figure out what the pictures were representing, and
I think that you're just showing how the matrix is linearized, right?  There's no reason that the
JSON can't be written as a matrix.  Do you think that would be more clear?  I do like having the examples with the various transforms shown.

vvmruder:
Yes, I tried to make clear how the linearisation should be done. Maybe I should add some words about that. I needed some time to recognize how it should be done. The reason was, that I was used to another order inside the 4x4 transformation matrix. Maybe there are different ways to write it? Trying to solve my examples I first thought row-major is simply wrong in the docs because the way I was used to see 4x4 matrix would lead to column-major order. Then I recognized in some US paper that obviously the matrix can be written in different way. That was why I chose to make some simple examples to connect usual 4x4 matrix style with representation expected by this filter.

What do you mean with examples? Should I add some more details about coordinates systems and how it works?

My idea was to add some detailed steps in the tutorial section. Especially to show the handling with georeferenced point clouds.

vvmruder:
To show you how its different have a look at these two links:
http://www-lehre.informatik.uni-osnabrueck.de/~mm/skript/7_3_3D_Transformation.html

http://www.math.tau.ac.il/~dcor/Graphics/cg-slides/geom3d.pdf

They have the translation components of written in different places in the matrix. The second link uses ordering like you probably expect :)

chambbj:
I'm not opposed to a full-blown tutorial treatment of the transformation matrices, but I also think that for brevity in the core filter documentation we could just as easily state that we expect the matrix in row major order and then given some examples. Our docs can render LaTeX math via mathjax, so that would be an option too, e.g., for transformations:

```latex
\begin{matrix}
  1 & 0 & 0 & t_x \\
  0 & 1 & 0 & t_y \\
  0 & 0 & 1 & t_z \\
  0 & 0 & 0 & 1
\end{matrix}
```

vvmruder:
@chambbj Ok. So I will remove graphics and add LaTeX math. This would look much nicer. Meanwhile I figured out how to build the docs to see how it actually looks.

I will put PR back to WIP and work on that.

abellgithub:
I'm going to close this if there's no follow-up shortly.

PR #2749, opened 2019-09-17
Initial commit of DBSCAN clustering

### LINKED ISSUE #2748: DBSCAN clustering would be a nice addition to the filters toolbox
Wikipedia version of the algorithm is at https://en.wikipedia.org/wiki/DBSCAN

> Ester, Martin, et al. "A density-based algorithm for discovering clusters in large spatial databases with noise." Kdd. Vol. 96. No. 34. 1996.

chambbj:
Close #2748


doc/stages/filters.dbscan.rst
```diff
@@ -0,0 +1,45 @@
...
+          "type":"writers.bpf",
+          "filename":"output.bpf",
+          "output_dims":"X,Y,Z,ClusterID"
+      }
+  ]
+
+Options
+-------
+
+min_points
```
abellgithub: Number of dimensions?  Do you mean x, y and z?

filters/DBSCANFilter.cpp
```diff
@@ -0,0 +1,148 @@
...
+{
+    args.add("min_points", "Min points per cluster", m_minPoints,
+             static_cast<uint64_t>(6));
+    args.add("eps", "Epsilon", m_eps, 1.0);
+}
+
+void DBSCANFilter::addDimensions(PointLayoutPtr layout)
+{
+    m_cluster =
+        layout->registerOrAssignDim("ClusterID", Dimension::Type::Signed64);
```
abellgithub: This type seems huge for this and just adds to memory consumption.

filters/DBSCANFilter.cpp
```diff
@@ -0,0 +1,148 @@
...
+            // Select first neighbor and move it to the visited set.
+            PointId p = *neighbors_next.begin();
+            neighbors_next.erase(neighbors_next.begin());
+            neighbors_visited.insert(p);
+
+            // Reassign cluster label to neighbor previously marked as noise.
+            if (view.getFieldAs<int64_t>(m_cluster, p) == -1)
+                view.setField(m_cluster, p, cluster_label);
+
+            // Neighbor has already been labeled, so move on to next.
```
abellgithub: Do you mean this as an else case?  If not, perhaps better just to fetch the value once from the point view and set when we're done.

filters/DBSCANFilter.hpp
```diff
@@ -0,0 +1,68 @@
...
+private:
+    uint64_t m_minPoints;
+    double m_eps;
+    Dimension::Id m_cluster;
+
+    virtual void addArgs(ProgramArgs& args);
+    virtual void addDimensions(PointLayoutPtr layout);
+    virtual void filter(PointView& view);
+
+    DBSCANFilter& operator=(const DBSCANFilter&); // not implemented
```
abellgithub: We've been marking these as deleted.

filters/DBSCANFilter.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+namespace pdal
+{
+
+class ProgramArgs;
+
+class PDAL_DLL DBSCANFilter : public Filter
+{
+public:
+    DBSCANFilter() : Filter()
```
abellgithub: This should go in CPP file to avoid any issues on windows with construction/destruction of objects in different DLLs.

doc/stages/filters.dbscan.rst
chambbj: Yeah, so a little confusing if you are thinking about number of "PDAL Dimensions" = number of attributes in a point record. I do mean XYZ in this context, though in #2748 it has been suggested that clustering could in theory be performed on 1) more than just the three XYZ dimensions, or 2) a custom set of dimensions such as NxNyNz.

filters/DBSCANFilter.cpp
chambbj: In practice, this is huge, and would hardly be useful. That said, poor parameter choice could lead you down the path of landing with one cluster per point. `point_count_t` is uint64, so even larger than what we are providing here, but maybe there is a better way to guard against blowing up, even if the utility of the result is questionable.

filters/DBSCANFilter.cpp
chambbj: Maybe this can be rewritten as

```c++
int64_t cluster_val = view.getFieldAs<int64_t>(m_cluster, p) == -1);
if (cluster_val ==-1)
{
    view.setField(m_cluster, p, cluster_label);
    continue;
}
else if (cluster_val != -2)
{
    continue;
}
```

Not sure if that's any clearer, and I'd need to double check the logic. Point being that as written it correctly sets a cluster label if it was noise before, and regardless of whether or not it was originally labelled, or was relabeled from noise, as long as it now has a label, we hit the continue at line 126.

abellgithub:
Is this ready @chambbj ?

chambbj:
No, in addition to your suggested changes, I was working on some features that were suggested in the associated issue. We can close and re-open later, or mark as WIP if you'd like.

pdal/KDIndex.hpp
```diff
@@ -373,6 +373,251 @@ class PDAL_DLL KD3Index : public KDIndex<3>
         return radius(x, y, z, r);
     }
 
+};
+
+class PDAL_DLL KDNormalIndex : public KDIndex<3>
```
chambbj: I think I'll take this back out. It is not used in this PR and is a result of me learning how to work with nanoflann to create additional index adapters.

doc/stages/filters.dbscan.rst
chambbj: Updated description slightly. The addition of the new `dimensions` argument will help to highlight the treatment of dimensions and number used.

chambbj:
@abellgithub Assuming that everything checks out, I'm ready to put this one to bed.

hobu:
- [ ] Add versionadded entry to docs.

PR #2754, opened 2019-09-19
Add PointIdList typedef for std::vector<PointId>

chambbj:


abellgithub:
Wow.  There are a lot of them.  Does make me wonder about using a deque.  Trying to think if there are any downsides.  In any of these cases do we really care about contiguous memory?

chambbj:
I did start a separate, follow-on PR to begin looking into the deque as you suggested. We could combine. It's a considerably smaller subset of instances from what I can tell. I did not do any profiling of the two.

Issue #2755, opened 2019-09-20
Ability to specify data types of custom dimensions when created by the Python filter stage


pramukta:
Thanks for all the work on PDAL, it's super awesome!  Was playing around prototyping a stage in Python and found that, currently, it looks like the `add_dimension` option always creates double precision floats:

https://github.com/PDAL/PDAL/blob/b65725eb7ecae77562f284f8f1e53f031389f370/plugins/python/filters/PythonFilter.cpp#L97

Also looks like a similar behavior happens in `filters.ferry`:

https://github.com/PDAL/PDAL/blob/08fb529e63db71491e530f3f1b737bf2b2ffe612/filters/FerryFilter.cpp#L97

The result is that I couldn't find a way to prototype a stage in Python that needs to return a uint64.  For example, this pipeline/script that tries to add a global 64-bit z-ordering: https://gist.github.com/pramukta/ff37596e180001cefbd49aa7030c3e99

Could/should a `datatype` option be added to the stage config to support this kind of use case?  I'm not very familiar with the codebase but will give it a shot as well.  If there's another way to approach this that I'm not seeing, that would be good to know too.

abellgithub:
This can be done, but your use case is unusual. Needing 64 bits of integer
precision seems strange and isn't supported by the ply format anyway.

pramukta:
Oh sorry, the ply writer is just a placeholder for testing.

abellgithub:
I think I would just add an optional type to the current `add_dimensions` option:

```
"add_dimensions" : "foo, bar=uint64_t"
```

pramukta:
AH, missed the ply writer example before this PR.  It looks like it does a type lookup and uses some utils I didn't realize existed.  Can update it on Monday if that is the preferred behavior.

pramukta:
Moving discussion over to the PR

pramukta:
Closing this issue following merge of #2756

PR #2756, opened 2019-09-21
Ability to add alternate dimension types other than float64 via the Python Filter stage

pramukta:
Related to the github issue #2755, this implements @abellgithub's suggestion to use the syntax `newDimension=dataType` to specify types of dimensions added via the Python filter stage.  If a type isn't specified, the current behavior is maintained (ie creating a float64 dimension).

### DIFF
--- a/plugins/python/filters/PythonFilter.cpp
+++ b/plugins/python/filters/PythonFilter.cpp
@@ -37,6 +37,7 @@
 #include <nlohmann/json.hpp>
 
 #include <pdal/PointView.hpp>
+#include <pdal/DimUtil.hpp>
 #include <pdal/util/ProgramArgs.hpp>
 #include <pdal/util/FileUtils.hpp>
 
@@ -93,8 +94,17 @@ void PythonFilter::addArgs(ProgramArgs& args)
 
 void PythonFilter::addDimensions(PointLayoutPtr layout)
 {
-    for (const std::string& s : m_args->m_addDimensions)
-        layout->registerOrAssignDim(s, pdal::Dimension::Type::Double);
+    for (const std::string& s : m_args->m_addDimensions) {
+        std::size_t sep = s.find('=');
+        if (sep != std::string::npos)
+        {
+            std::string tval = s.substr(sep+1, s.length() - sep - 1),
+                        dval = s.substr(0, sep);
+            layout->registerOrAssignDim(dval, pdal::Dimension::type(tval));
+        } else {
+            layout->registerOrAssignDim(s, pdal::Dimension::Type::Double);
+        }
+    }
 }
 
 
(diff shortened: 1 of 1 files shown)


abellgithub:
If you wanted to modify this to make it more like what's in PLY as you noted elsewhere, that would be great.

pramukta:
I made the switch to using the utility function to split.  On second glance, the layout lookups and instantiations happening in PlyWriter didn't seem to be applicable here.  I did, however, notice a potential issue with current behavior.  

With existing (pre-PR) functionality, specifying an existing dimension name will change it's datatype to double, since `PointLayout::registerOrAssignDim` passes the data type through without a lookup: https://github.com/PDAL/PDAL/blob/404e3c86be8004b532c9f28f5746cfdce0437fb9/pdal/PointLayout.cpp#L123

Similarly, passing a known but not registered dimension name results in it being created as a double by default, which would create a mismatch between the layout and the known dimension specs.  Post PR, current behavior is retained, but wondering if it should be adjusted, either here or in a separate PR?  It looks like this is easily fixable either centrally in `PointLayout` or specially in `PythonFilter` but either adjustment would create slightly different behavior than what currently exists.  Any thoughts?

abellgithub:
Each stage registers dimensions with the type it needs.  PDAL will create dimensions to satisfy the total range requested.  For example, if one stage requests the dimension Foo to be uint8_t and another requests it as int8_t, PDAL will actually use int16_t, since it allows for all the requested values.

Your case is a special one, in that you're trying to create what amounts to a bit pattern rather than a value.  Normally, if one stage were to request Zorder as double and another were to request it as uint64_t, we'd use double, figuring that it would be OK to reduce the precision of the uint64_t in order to get the range of the double.  This isn't what you want, but if you use a proprietary dimension, you shouldn't have any trouble.

pramukta:
Ah got it cool, that makes sense.  So seems like should be good then, let me know if there's any other mods needed.

plugins/python/filters/PythonFilter.cpp
```diff
@@ -93,8 +94,15 @@ void PythonFilter::addArgs(ProgramArgs& args)
 
 void PythonFilter::addDimensions(PointLayoutPtr layout)
 {
-    for (const std::string& s : m_args->m_addDimensions)
-        layout->registerOrAssignDim(s, pdal::Dimension::Type::Double);
+    for (const std::string& s : m_args->m_addDimensions) {
```
abellgithub: The code should provide feedback to the user if the syntax of the argument is bad.  Here, spec size should only be 1 or 0 and a check should be made to make sure spec[1] is a valid type.

abellgithub:
Close #2755

PR #2757, opened 2019-09-23
Adding enhancements in E57 plugins

pravinshinde825:
In this PR I tried to add some enhancements and features in E57 plugins, Below are the highlights:
1. E57 reader is more accurately scale the scalable dimensions.
2. E57 reader can now read multiple scans with different dimensions.
3. E57 writer can write accurate Limits for dimension in E57 header.
4. Both reader and writer supports `extra_dims` arguments.
5. Improved Scan class.


plugins/e57/io/E57Reader.cpp
```diff
@@ -1,228 +1,127 @@
...
-             keyValue.second.data(), m_defaultChunkSize, true,
-             (prototype.get(keyValue.first).type() == e57::E57_SCALED_INTEGER));
-
-    m_pointOffset = pointOffset;
-
-    // Setup the reader
-    m_dataReader = std::unique_ptr<e57::CompressedVectorReader>(
-        new e57::CompressedVectorReader(vectorNode.reader(m_e57buffers)));
-}
+static PluginInfo const s_info{"readers.e57", "Reader for E57 files", ""};
```
abellgithub: This should contain a link to the stage documentation.

plugins/e57/io/E57Reader.cpp
```diff
@@ -235,143 +134,120 @@ QuickInfo E57Reader::inspect()
     Dimension::IdList dims = layout->dims();
     for (auto di = dims.begin(); di != dims.end(); ++di)
         qi.m_dimNames.push_back(layout->dimName(*di));
-    qi.m_pointCount = m_pointCount;
+    qi.m_pointCount = e57plugin::numPoints(*m_data3D);
```
abellgithub: Have you checked that this works?  I don't think m_data3d would be populated.  You need to call initialize() if you want that behavior in inspect().

plugins/e57/io/E57Reader.cpp
```diff
@@ -235,143 +134,120 @@ QuickInfo E57Reader::inspect()
...
+        // Either we are at very begining or finished processing all points in
+        // current batch. Its time to read new points batch.
+        m_pointsInCurrentBatch = readNextBatch();
     }
-    catch(const e57::E57Exception& e)
+
+    if (!m_pointsInCurrentBatch)
     {
-        throwError(std::to_string(e.errorCode()) + " : " + e.context());
+        // We're done with reding
```
abellgithub: Spelling

plugins/e57/io/Scan.cpp
```diff
@@ -59,17 +59,6 @@ e57::CompressedVectorNode Scan::getPoints() const
...
-    if (itMinMax != m_valueBounds.end())
-    {
-        minMax = itMinMax->second;
-        return true;
-    }
-
-    return false;
-}
 
 bool Scan::hasPose() const
```
abellgithub: Don't you just want to call this one time?  Or one time per chunk?  It seems like it's getting called for every point.  Can you just check m_hasPose at the beginning and exit if you've already checked or something like that?

plugins/e57/io/E57Reader.cpp
pravinshinde825: initialize() is already called before accessing `m_data3D`.

abellgithub: Sorry .  That code got rolled up in github.

pravinshinde825:
@abellgithub , Please give me some time. I'll resolve all comments once I get some bandwidth. Thanks.

plugins/e57/io/Scan.cpp
```diff
@@ -130,70 +119,118 @@ void Scan::decodeHeader()
...
-        //ABELL - This seems weird.  Could you not have one value that's
-        //  NaN and another not NaN?
-        if (minmax == minmax) // not nan
-            m_valueBounds[pdal::e57plugin::e57ToPdal(field)] = minmax;
+        auto minmax = std::make_pair(0.0, 0.0);
+        if (pdal::e57plugin::getLimits(*m_rawData, field, minmax))
+        {
+            m_rescaleFactors[pdal::e57plugin::e57ToPdal(field)] =
+                (std::numeric_limits<uint16_t>::max)() /
+                (minmax.second - minmax.first);
```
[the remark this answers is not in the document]
pravinshinde825: - For intensities the scaling would be 0->1 to 0->65535. But cannot not be same for all, Colors may need to be scaled from 0->255 to 0->65535.

- `supportedFields` is a list of E57 dimensions which are supported in PDAl.

plugins/e57/io/Scan.cpp
pravinshinde825: This needs to be called once per e57 scan.
We can do the following : 

- Call this once a new scan is setup and capture status in `E57Reader::m_needsTransform` (new member) and instead making call to `hasPose()`, Can just check `m_needsTransform` for every point.

-  Just make hasPose inline.

What do you think?

plugins/e57/io/Scan.cpp
```diff
@@ -130,70 +119,118 @@ void Scan::decodeHeader()
...
+            m_translation[0] = e57::FloatNode(transNode.get("x")).value();
+            m_translation[1] = e57::FloatNode(transNode.get("y")).value();
+            m_translation[2] = e57::FloatNode(transNode.get("z")).value();
+        }
+    }
+    else
+        m_hasPose = false;
+}
+
+double Scan::rescale(pdal::Dimension::Id dim, double value)
```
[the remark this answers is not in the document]
pravinshinde825: Changed to array.

plugins/e57/io/Scan.cpp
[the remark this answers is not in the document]
abellgithub: But this isn't what you're doing.  You're scaling the current range -- whatever that is -- to 0-65535.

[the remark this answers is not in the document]
pravinshinde825: Hmm got your point.. 
I think we need to get rid of this scaling stuff. We should forward what we get from E57.
Only intensities are required to scale if they are between 0-1. Since we cannot forward them as a double value.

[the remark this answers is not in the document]
pravinshinde825: Done.

plugins/e57/io/Scan.cpp
```diff
@@ -141,13 +141,16 @@ void Scan::decodeHeader()
...
     std::fill_n(m_rescaleFactors, pdal::Dimension::COUNT, 1.0); //Re-initialize all rescale factors to 1.0
-    for (auto& field : supportedFields)
+    auto rescaleableFields = pdal::e57plugin::rescalableE57Types();
+    for (auto& field : rescaleableFields)
     {
         auto minmax = std::make_pair(0.0, 0.0);
         if (pdal::e57plugin::getLimits(*m_rawData, field, minmax))
+        {
             m_rescaleFactors[(int)pdal::e57plugin::e57ToPdal(field)] =
                 (std::numeric_limits<uint16_t>::max)() /
```
nicolas-chaulet: This assumes that the target of the rescaling is 65534 which will not always be the case. Maybe you can change `rescalableE57Types` into a map and store the target scale there? Alternatively you can ask PDAL for the type of that dimension and use that instead of uint16_t

pravinshinde825:
@abellgithub , @nicolas-chaulet , 
https://github.com/PDAL/PDAL/pull/2757/commits/15ca4e90d353f2b9289c12c8df697badf79cbf36 : This commit adds more enhancements and features to E57 plugins. Please check PR description for more details.

plugins/e57/io/E57Reader.cpp
```diff
@@ -51,30 +50,122 @@ std::string E57Reader::getName() const
...
 }
 
 void E57Reader::addDimensions(PointLayoutPtr layout)
 {
-    for (auto& keyVal : m_doubleBuffers)
+    auto supportedFields = e57plugin::supportedE57Types();
+    auto numScans = m_data3D->childCount();
+    std::unique_ptr<Scan> tempScan;
+
+    for (auto& dimension : supportedFields)
```
abellgithub: This is confusing to me.  Can you explain what the gist of it is?  The `extra_dims` refers to writing, but this is a reader.

pravinshinde825: `extra_dims` can be used with both reader and writer. 

Consider a dimension `abcd` written by a writer.
Since the dimension is not in `supportedE57Types`, We also need to have support in reader to read that dimension. In that case the dimension need to be provided in `extra_dims`.
Here I am registering supported types which are present in E57 scan. And then same for extra dims.

nicolas-chaulet: I believe the las reader has the same logic at the moment

plugins/e57/io/E57Reader.cpp
pravinshinde825: corrected !!

plugins/e57/io/E57Reader.cpp
pravinshinde825: done!!

plugins/e57/io/E57Reader.cpp
```diff
@@ -1,228 +1,127 @@
...
+std::string E57Reader::getName() const
 {
-    point_count_t actualIndex = index - m_pointOffset;
-    return (actualIndex < m_maxPointRead);
+    return s_info.name;
 }
 
-bool E57Reader::ChunkReader::isInChunk(point_count_t index) const
+E57Reader::E57Reader()
+    : Reader(), Streamable(), m_currentIndex(0), m_pointsInCurrentBatch(0), m_defaultChunkSize(1000000),
```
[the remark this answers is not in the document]
pravinshinde825: moved to ready.

abellgithub:
This has been through enough that I think it needs a fresh look.  How about closing this PR and opening it again and we can take another turn at it.

pravinshinde825:
@abellgithub , As  suggested, I reopened this PR in #2847

PR #2762, opened 2019-09-25
Add farthest point sampling (FPS) filter

### LINKED ISSUE #2761: Add farthest point sampling filter
It would be nice to have a farthest point sampling (FPS) filter. The output would be visually very similar to that of `filters.outlier` but with FPS we can specify the number of points in the output exactly.

chambbj:
- Similar characteristics to Poisson filtering

- Allows user to specify an exact number of points for the output

- This is a greedy, brute force approach to FPS, and will not scale well to
  large point clouds

Close #2761

### DIFF
--- a/filters/FarthestPointSamplingFilter.cpp
+++ b/filters/FarthestPointSamplingFilter.cpp
@@ -77,8 +77,7 @@ PointViewSet FarthestPointSamplingFilter::run(PointViewPtr inView)
     PointViewPtr outView = inView->makeNew();
 
     // Construct a KD-tree of the input view.
-    KD3Index kdi(*inView);
-    kdi.build();
+    KD3Index& kdi = inView->build3dIndex();
 
     // Seed the output view with the first point in the current sorting.
     PointId seedId(0);
--- a/filters/FarthestPointSamplingFilter.cpp
+++ b/filters/FarthestPointSamplingFilter.cpp
@@ -1,0 +1,128 @@
+/******************************************************************************
+ * Copyright (c) 2019, Bradley J Chambers (brad.chambers@gmail.com)
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following
+ * conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided
+ *       with the distribution.
+ *     * Neither the name of Hobu, Inc. or Flaxen Geo Consulting nor the
+ *       names of its contributors may be used to endorse or promote
+ *       products derived from this software without specific prior
+ *       written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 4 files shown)


filters/FarthestPointSamplingFilter.cpp
```diff
@@ -0,0 +1,128 @@
...
+    std::vector<double> sqr_dists(inView->size());
+    kdi.knnSearch(seedId, inView->size(), &indices, &sqr_dists);
+
+    // Sort distances by PointId.
+    std::vector<double> min_dists(inView->size());
+    for (PointId i = 0; i < inView->size(); ++i)
+        min_dists[indices[i]] = sqr_dists[i];
+
+    // Proceed until we have m_count points in the output PointView.
+    for (PointId i = 1; i < m_count; ++i)
```
abellgithub: I guess I don't understand what's being done with min_dists.

filters/FarthestPointSamplingFilter.cpp
```diff
@@ -0,0 +1,128 @@
...
+    // Return empty PointViewSet if the input PointView has no points.
+    PointViewSet viewSet;
+    if (!inView->size() || (inView->size() < m_count))
+        return viewSet;
+
+    // Otherwise, make a new output PointView.
+    PointViewPtr outView = inView->makeNew();
+
+    // Construct a KD-tree of the input view.
+    KD3Index kdi(*inView);
```
abellgithub: You can use the input view's KDIndex here: `inView->build3dIndex()`

chambbj: I still tend to forget about this one. Thanks for the reminder.

filters/FarthestPointSamplingFilter.cpp
```diff
@@ -0,0 +1,128 @@
...
+    // Seed the output view with the first point in the current sorting.
+    PointId seedId(0);
+    outView->appendPoint(*inView, seedId);
+
+    // Compute distances from seedId to all other points.
+    PointIdList indices(inView->size());
+    std::vector<double> sqr_dists(inView->size());
+    kdi.knnSearch(seedId, inView->size(), &indices, &sqr_dists);
+
+    // Sort distances by PointId.
```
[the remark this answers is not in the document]
chambbj: It does represent the furthest point (from seedId in this case), which would be sufficient if we were to end here. Unfortunately, we need to continue finding the next furthest point, which is not simply the furthest point from the next point we add, but the furthest from the collection of points that are being added to the output view. This is why we order the distances by PointId and store as min_dists, which I'll describe in your other comment.

filters/FarthestPointSamplingFilter.cpp
chambbj: So the key thing happening with `min_dists` is

https://github.com/PDAL/PDAL/blob/c678bff41306818d99e1d7adc8975c7ad065600b/filters/FarthestPointSamplingFilter.cpp#L119-L120

Keep in mind that from the initialization of `min_dists` above, we had a distance of 0.0 for the `seedId`. Within this loop, we will also have a distance of 0.0 at `idx` (i.e., the most recently added furthest point). By keeping the minimum distance, we now have two (in the first iteration) 0.0 distance entries, one for each point in our output view, meaning they cannot be repeatedly detected as the furthest point, and are removed from further consideration. It also means that of the remaining neighbors, we are now storing the current closest distance to any point that has been added to the output view, which is critical when picking the next furthest point.

filters/FarthestPointSamplingFilter.cpp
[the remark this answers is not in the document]
abellgithub: Sorry.  I figured stuff out and thought I deleted this comment.

filters/FarthestPointSamplingFilter.cpp
abellgithub: Same with this.  Sorry.

chambbj:
All comments resolved at this time.

PR #2766, opened 2019-09-26
Add Cloth Simulation Filter (CSF)

### LINKED ISSUE #2765: Add Cloth Simulation Filter (CSF) for ground segmentation
Cloth Simulation Filtering (CSF) aims to segregate returns into ground/nonground classes (much like SMRF, PMF, etc.).

> Zhang, Wuming, et al. "An easy-to-use airborne LiDAR data filtering method based on cloth simulation." Remote Sensing 8.6 (2016): 501.

The original code is Apache 2.0 licensed [1]. It has also been integrated as a CloudCompare plugin [2]. It would be quite trivial to also wrap the original implementation as a PDAL filter.

[1] https://github.com/jianboqi/CSF

[2] http://www.cloudcompare.org/doc/wiki/index.php?title=CSF_(plugin)

chambbj:
Close #2765

### DIFF
--- a/filters/private/csf/CSF.cpp
+++ b/filters/private/csf/CSF.cpp
@@ -1,0 +1,211 @@
+// ======================================================================================
+// Copyright 2017 State Key Laboratory of Remote Sensing Science, 
+// Institute of Remote Sensing Science and Engineering, Beijing Normal University
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+//     http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ======================================================================================
+
+#define DLL_IMPLEMENT
+
+#include "CSF.h"
+#include "XYZReader.h"
+#include "Vec3.h"
+#include "Cloth.h"
+#include "Rasterization.h"
+#include "c2cdist.h"
+#include <fstream>
+#include <pdal/Log.hpp>
+
+
+CSF::CSF(int index) {
+    params.bSloopSmooth     = true;
+    params.time_step        = 0.65;
+    params.class_threshold  = 0.5;
+    params.cloth_resolution = 1;
+    params.rigidness        = 3;
+    params.interations      = 500;
+
+    this->index = index;
+}
+
+CSF::CSF() {
+    params.bSloopSmooth     = true;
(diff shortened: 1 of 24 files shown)


CMakeLists.txt
```diff
@@ -125,6 +125,13 @@ include(${PDAL_CMAKE_DIR}/arbiter.cmake)
 include(${PDAL_CMAKE_DIR}/nlohmann.cmake)
 include(${PDAL_CMAKE_DIR}/openssl.cmake) # Optional
 
+find_package(OpenMP)
+if (OPENMP_FOUND)
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
+    set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
+endif()
+
```
hobu: Put this in cmake/openmp.cmake

filters/private/csf/CSF.cpp
```diff
@@ -0,0 +1,206 @@
+// ======================================================================================
```
hobu: The couts in this file should be aliased and set via method so we can wire up PDAL's logging.

dbrookes96:
Cool to see this available in PDAL.

I cobbled together a pretty basic CSF filter myself last year thats more or less the same as this. Mine lacks the ignored values and include last returns options though, which are nice.

There is two things I would like to mention however...

I would recommend adding a try catch around `c.do_filtering` as depending on the bounds of the input points and the cloth resolution CSF will attempt to allocate a **huge** amount of cloth nodes. It doesn't check this internally (maybe it should). Not sure how PDAL traditionally handles these sorts of things though?

In our build of CSF I've removed all uses of `#pragma omp parallel for` as somewhere inside this introduces non-deterministic behavior. An alternative is to set `OMP_NUM_THREADS` to `1`. This was causing our regression tests to fail. CloudCompare has patched one of these away, but it still persists (https://github.com/jianboqi/CSF/issues/21 and https://github.com/CloudCompare/CloudCompare/issues/953). I don't have exact numbers but this didn't seem to completely destroy the performance of CSF. Again, not sure if you guys consider non-deterministic behavior a problem.

Hope this helps.

chambbj:
Thanks for the comments @dbrookes96!

Good news I guess is that we don't actually build with OpenMP support, so I think this non-deterministic behavior concern is somewhat mitigated for the time being. We should consider addressing it more definitively though should someone ever turn OMP back on at some point in the future.

As for the try/catch for the filtering, I can look more closely at how we could guard against potential issues.

chambbj:
@abellgithub @hobu I'm done fiddling with this one for now.

hobu:
:shipit:

PR #2769, opened 2019-10-01
TileDB CLI argument synonyms for r/w

### LINKED ISSUE #2768: PDAL CLI - TileDB
The PDAL CLI cannot parse a TileDB array as it doesn't have a file extension hence it is necessary to override the driver. This causes the following scenarios

```
pdal info --driver readers.tiledb sample_array 
PDAL: readers.tiledb: Missing value for positional argument 'array_name'

pdal info --driver readers.tiledb --readers.tiledb.array_name=sample_array
PDAL: No input file specified.

usage: pdal info [options] [input]
:
:
```

I would expect to be able to issue

```pdal info --driver readers.tiledb sample_array```

or

```pdal info --driver readers.tiledb --readers.tiledb.array_name=sample_array```

I can take a look at this in the near future. In the TileDB case (is this a more general problem with the CLI?) the `array_name` parameter is positional and shouldn't need to be named.

normanb:
Fixes #2768

### DIFF
--- a/plugins/tiledb/io/TileDBReader.cpp
+++ b/plugins/tiledb/io/TileDBReader.cpp
@@ -89,6 +89,7 @@ Dimension::Type getPdalType(tiledb_datatype_t t)
 void TileDBReader::addArgs(ProgramArgs& args)
 {
     args.add("array_name", "TileDB array name", m_arrayName).setPositional();
+    args.addSynonym("array_name", "filename");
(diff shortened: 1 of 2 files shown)


abellgithub:
I made these changes as an alternative: https://github.com/PDAL/PDAL/pull/2853

normanb:
:+1:

PR #2770, opened 2019-10-01
Documentation fixes

jrobichaud:
I found a few issues while I was reading the documentation.

The first three of them are minor (see separate commits for details).

The fourth is an issue of duplicated label

```bash
/Users/user/git/PDAL/doc/stages/filters.stats.rst:38: WARNING: duplicate label dimensions, other instance in /Users/user/git/PDAL/doc/dimensions.rst
```

It breaks several links to the [dimensions](https://pdal.io/dimensions.html) section.

image

I used a workaround to rename the anchor to `#stats-dimensions`. I would have liked to preserve the `#dimensions` anchors but it seems to be a Restructured Text limitation.


abellgithub:
Thanks for this!

PR #2772, opened 2019-10-01
Make EPT reader asynchronous in streaming mode

connormanning:
The current streaming implementation for the EPT reader (#2603) is synchronous, so it is many times slower than the non-streaming version which loads many buffers at once.  This PR adds asynchronous look-ahead buffering to the streaming mode of the EPT reader (the number of which respects the existing `threads` option) which is significantly faster.

The diff is very noisy since it replaces the existing implementation - the relevant bits are `load`, `next`, and `processOne` in the EPT reader.

### DIFF
--- a/io/EptReader.cpp
+++ b/io/EptReader.cpp
@@ -62,7 +62,6 @@ namespace
         { "ept" }
     };
 
-    const std::size_t streamNodeCount(8);
     const std::string addonFilename { "ept-addon.json" };
 
     Dimension::Type getRemoteType(const NL::json& dim)
@@ -906,7 +905,7 @@ void EptReader::load()
     // Asynchronously trigger the fetching and point-view execution of
     // a lookahead buffer of nodes.
     while (
-        m_upcomingNodeBuffers.size() < streamNodeCount &&
+        m_upcomingNodeBuffers.size() < m_pool->size() &&
         m_overlapIt != m_overlaps.end())
     {
         const auto nodeId(m_nodeId++);
--- a/io/EptReader.cpp
+++ b/io/EptReader.cpp
@@ -996,7 +996,7 @@ bool EptReader::processOne(PointRef& point)
     auto& sourceView(m_currentNodeBuffer->view);
     const auto& layout(*m_currentNodeBuffer->table.layout());
 
-    for (const auto& id : m_currentNodeBuffer->table.layout()->dims())
+    for (const auto& id : layout.dims())
     {
         point.setField(
             id,
@@ -1004,10 +1004,7 @@ bool EptReader::processOne(PointRef& point)
             sourceView.getPoint(m_pointId) + layout.dimOffset(id));
     }
 
-    if (++m_pointId == m_currentNodeBuffer->view.size())
-    {
-        m_currentNodeBuffer.reset();
-    }
+    if (++m_pointId == sourceView.size()) m_currentNodeBuffer.reset();
 
     return true;
 }
--- a/io/EptReader.cpp
+++ b/io/EptReader.cpp
@@ -918,7 +918,8 @@ void EptReader::load()
         // Insert an empty placeholder node to keep track of the outstanding
         // nodes that are currently being fetched/executed.
         std::unique_lock<std::mutex> lock(m_mutex);
-        auto& loadingBuffer = m_upcomingNodeBuffers[nodeId];
+        m_upcomingNodeBuffers.emplace_front();
+        auto& loadingBuffer = m_upcomingNodeBuffers.front();
         lock.unlock();
 
         m_pool->add([this, &loadingBuffer, nodeId, key]()
@@ -949,8 +950,8 @@ EptReader::NodeBufferIt EptReader::findBuffer()
     return std::find_if(
         m_upcomingNodeBuffers.begin(),
         m_upcomingNodeBuffers.end(),
-        [this](const NodeBufferPair& p)
-            { return static_cast<bool>(p.second); });
+        [](const std::unique_ptr<NodeBuffer>& n)
+            { return static_cast<bool>(n); });
 }
 
 bool EptReader::next()
@@ -974,7 +975,7 @@ bool EptReader::next()
     if (it == m_upcomingNodeBuffers.end()) return false;
 
     m_pointId = 0;
-    m_currentNodeBuffer = std::move(it->second);
+    m_currentNodeBuffer = std::move(*it);
     m_upcomingNodeBuffers.erase(it);
 
     return true;
--- a/io/EptReader.cpp
+++ b/io/EptReader.cpp
@@ -944,6 +944,14 @@ void EptReader::load()
     }
 }
 
+EptReader::NodeBufferIt EptReader::findBuffer()
+{
+    return std::find_if(
+        m_upcomingNodeBuffers.begin(),
+        m_upcomingNodeBuffers.end(),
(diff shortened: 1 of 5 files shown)


nicolas-chaulet:
Awesome!!

io/EptReader.cpp
```diff
@@ -902,86 +893,120 @@ Dimension::Type EptReader::getCoercedTypeTest(const NL::json& j)
...
+{
+    // Asynchronously trigger the loading of some nodes.
+    load();
+
+    // Now wait for a node to be populated, or for there to be no outstanding
+    // nodes left, in which case we're all done.
+    std::unique_lock<std::mutex> lock(m_mutex);
+    m_cv.wait(lock, [this]()
+    {
+        return m_upcomingNodeBuffers.empty() ||
```
abellgithub: Think this empty check isn't needed.

io/EptReader.cpp
```diff
@@ -902,86 +893,120 @@ Dimension::Type EptReader::getCoercedTypeTest(const NL::json& j)
...
+            findBuffer() != m_upcomingNodeBuffers.end();
+    });
+
+    // Processing is complete.
+    if (m_upcomingNodeBuffers.empty()) return false;
```
abellgithub: Or here.

io/EptReader.cpp
```diff
@@ -902,86 +893,120 @@ Dimension::Type EptReader::getCoercedTypeTest(const NL::json& j)
...
+
+    // We know at least one node is populated from our `wait` predicate, so find
+    // the first one and transfer it out of our `upcoming` pool to make it the
+    // current active node.
+    const auto it = findBuffer();
```
abellgithub: The result of this iterator should be sufficient.

io/EptReader.cpp
connormanning: How could it be omitted?  This structure being empty is our exit condition, and otherwise we will wait here forever once processing is complete and the `upcoming` list is empty.

io/EptReader.cpp
connormanning: Resolved in 807b0be20.

io/EptReader.cpp
connormanning: Resolved in 807b0be20.

io/EptReader.cpp
```diff
@@ -902,86 +893,120 @@ Dimension::Type EptReader::getCoercedTypeTest(const NL::json& j)
...
-    // In that case proceed to load next overlap.
-    if (m_bufferPointView->size())
-        fillPoint(point);
-    else
-        return processOne(point);
+    for (const auto& id : m_currentNodeBuffer->table.layout()->dims())
+    {
+        point.setField(
+            id,
+            m_currentNodeBuffer->view.getFieldAs<double>(id, m_pointId));
```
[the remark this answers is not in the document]
connormanning: Resolved in e4f280d.

io/EptReader.cpp
```diff
@@ -902,86 +893,120 @@ Dimension::Type EptReader::getCoercedTypeTest(const NL::json& j)
...
-        m_temp_buffer.data());
-    point.setPackedData(dims, m_temp_buffer.data());
-    m_currentIndex++;
+    return std::find_if(
+        m_upcomingNodeBuffers.begin(),
+        m_upcomingNodeBuffers.end(),
+        [this](const NodeBufferPair& p) { return !!p.second; });
 }
 
+bool EptReader::next()
```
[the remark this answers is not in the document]
connormanning: I'm not sure what to change exactly - to me the current code matches how I understand the workflow exactly: if we've finished processing a node (or are just beginning operation), first make sure we have `n` lookahead nodes loading, and then simply `wait` for one to be loaded and make it the active one for processing.  If you have a different idea, I'd need some more specific direction on exactly what extra complexity I could remove.

io/EptReader.hpp
```diff
@@ -101,16 +103,22 @@ class PDAL_DLL EptReader : public Reader, public Streamable
...
+        PointId startId = 0) const;
 
     // To allow testing of hidden getRemoteType() and getCoercedType().
     static Dimension::Type getRemoteTypeTest(const NL::json& dimInfo);
     static Dimension::Type getCoercedTypeTest(const NL::json& dimInfo);
 
-    // For streamable pipeline.
+    // For streaming operation.
+    struct NodeBuffer;
+    using NodeBufferMap = std::map<uint64_t, std::unique_ptr<NodeBuffer>>;
```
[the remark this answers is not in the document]
connormanning: It's mostly something to serve as a unique identifier for the `key` since we insert empty `unique_ptr` instances as values into the map to indicate a currently loading node until it gets populated.

io/EptReader.cpp
abellgithub: Isn't this wait the same as:

```
while (!upcomingNodeBuffers.empty() && findBuffer() == something)
   wait();
```
If the buffers are empty, you'll never get in the loop.  If they are, it's equivalent to saying that you found something.  In other words, you can't find something if the buffers are empty.

io/EptReader.cpp
```diff
@@ -902,86 +893,119 @@ Dimension::Type EptReader::getCoercedTypeTest(const NL::json& j)
...
+    // Now wait for a node to be populated, or for there to be no outstanding
+    // nodes left, in which case we're all done.
+    std::unique_lock<std::mutex> lock(m_mutex);
+    m_cv.wait(lock, [this]()
+    {
+        return m_upcomingNodeBuffers.empty() ||
+            std::any_of(
+                m_upcomingNodeBuffers.begin(),
+                m_upcomingNodeBuffers.end(),
+                [this](const NodeBufferPair& p) { return !!p.second; });
```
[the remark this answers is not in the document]
connormanning: Replaced with `static_cast<bool>` in d6ca66b.

io/EptReader.hpp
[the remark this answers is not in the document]
abellgithub: I didn't understand why it wouldn't be a vector<unique_ptr<NodeBuffer>>

io/EptReader.cpp
abellgithub: I'm reading this: https://en.cppreference.com/w/cpp/thread/condition_variable/wait

connormanning: Maybe I'm just confused.  Replacing this check with simply `return findBuffer() != m_upcomingNodeBuffers.end();` ends up in an infinite `wait` at the end of processing like I'd expect - as you say, you can't find something if the buffers are empty.  And if the buffers are empty we want to exit for good, but without the `empty` check we hang because we're waiting for non-existent buffers to get populated.

What code exactly are you suggesting to be in this `wait` predicate?  I think I'm missing something.

io/EptReader.hpp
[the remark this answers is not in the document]
connormanning: Since we're erasing from random locations within the structure it doesn't fit the typical vector model.

io/EptReader.cpp
abellgithub: The confusion is mine.  I'm used to seeing it == container.end(), so that's what I was thinking.  Sorry.

io/EptReader.hpp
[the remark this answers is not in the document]
abellgithub: I guess then I'd use a list?  Using a key that you never use is confusing.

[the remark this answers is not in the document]
connormanning: Good call, switched to a `list` in 2f3dc3f.

PR #2774, opened 2019-10-03
JSON fixes in documentation codeblocks

NicolaiMogensen:
Several JSON code snippets in the documentation were not valid, i fixed them. Trailing commas on the last entry was the most common problem, which the JSON specification does not allow.

### DIFF
--- a/doc/python.rst
+++ b/doc/python.rst
@@ -73,15 +73,14 @@ results as `Numpy`_ arrays:
 
 
     json = """
-    {
-      "pipeline": [
+    [
         "1.2-with-color.las",
         {
             "type": "filters.sort",
             "dimension": "X"
         }
-      ]
-    }"""
+    ]
+    """
 
     import pdal
     pipeline = pdal.Pipeline(json)
--- a/doc/python.rst
+++ b/doc/python.rst
@@ -74,7 +74,7 @@ results as `Numpy`_ arrays:
 
     json = """
     {
-      [
+      "pipeline": [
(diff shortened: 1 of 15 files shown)


doc/python.rst
```diff
@@ -74,7 +74,7 @@ results as `Numpy`_ arrays:
 
     json = """
     {
-      [
```
abellgithub: I think the intent of the original change was to remove the base object and "pipeline" key.

doc/stages/readers.tindex.rst
```diff
@@ -43,7 +43,7 @@ merge the data.
           "type":"readers.tindex",
           "filter_srs":"+proj=lcc +lat_1=43 +lat_2=45.5 +lat_0=41.75 +lon_0=-120.5 +x_0=399999.9999999999 +y_0=0 +ellps=GRS80 +units=ft +no_defs",
           "filename":"index.sqlite",
-          "where":"location LIKE \'%nteresting.las%\'",
+          "where":"location LIKE \"%nteresting.las%\"",
```
abellgithub: I think the original was fine here.

NicolaiMogensen: I think so too, might have been my linter 5ec3f92

doc/python.rst
NicolaiMogensen: Right, removing the opening and closing {} then as {[]} is not valid JSON 5ec3f92

abellgithub:
Thanks!

PR #2776, opened 2019-10-07
Support simplification without topology

hobu:


filters/HexBinFilter.hpp
```diff
@@ -71,6 +71,7 @@ class PDAL_DLL HexBin : public Filter, public Streamable
     bool m_outputTesselation;
     bool m_doSmooth;
     point_count_t m_count;
+    bool m_preserve_topology;
```
abellgithub: Doesn't appear to be used.

PR #2778, opened 2019-10-08
Avoid potential descending for-loop wrapping

connormanning:
Fixes warning:
```
[258/396] Building CXX object CMakeFiles/pdalcpp.dir/filters/SkewnessBalancingFilter.cpp.o
../filters/SkewnessBalancingFilter.cpp:86:42: warning: comparison of unsigned expression >= 0 is always true [-Wtautological-compare]
    for (PointId i = view->size() - 1; i >= 0; --i)
                                       ~ ^  ~
1 warning generated.
```

### DIFF
--- a/filters/SkewnessBalancingFilter.cpp
+++ b/filters/SkewnessBalancingFilter.cpp
@@ -83,7 +83,8 @@ std::set<PointId> SkewnessBalancingFilter::processGround(PointViewPtr view)
     }
 
     PointId j(0);
-    for (PointId i = view->size() - 1; i >= 0; --i)
+    const auto size(view->size());
+    for (PointId i = size - 1; i < size; --i)
     {
         if (skewness[i] <= 0)
         {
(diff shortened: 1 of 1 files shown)


chambbj:
Thanks for catching this! Looks good.

filters/SkewnessBalancingFilter.cpp
```diff
@@ -83,7 +83,8 @@ std::set<PointId> SkewnessBalancingFilter::processGround(PointViewPtr view)
     }
 
     PointId j(0);
-    for (PointId i = view->size() - 1; i >= 0; --i)
+    const auto size(view->size());
```
abellgithub: Umm, I'm not a fan.  It makes you think about wrapping.  How about just writing the loop differently?

```
while
{
} while (--i != 0)
```

abellgithub: If the view is empty we can just early exit at the top.

chambbj:
@connormanning @abellgithub Minor tweak to this PR to address Andrew's comment. Maybe we can merge and move on.

PR #2780, opened 2019-10-10
support pdal tindex of non-filename data such as ept://https://...

hobu:


abellgithub:
This change eliminates the current feedback when files are missing and doesn't provide any for the remotely accessed files.

Issue #2707, opened 2019-08-29
E57 library bleeds symbols out of headers


abellgithub:
The E57 reader/writer include files that force direct linking of the E57 support library on some platforms.  This shouldn't be necessary.

norbertwenzel:
@abellgithub Are you talking about linking E57Format to the respective reader/writer tests? Or some different issue I do not fully understand yet?
I was trying to look into this but the more I did, the more I was unsure I was attacking the issue you were referring to. So I've done some basic cleanup (for a header that's never included except in the associated unit tests) in a [branch](https://github.com/PDAL/PDAL/compare/master...norbertwenzel:issue-2707) but could you clarify whether this is what you had in mind with this issue?

abellgithub:
No users of PDAL should have to explicitly link the E57 library.  I think some of the tests do now.  I'd have to look back to give you more info.

norbertwenzel:
The only uses of the E57 library I managed to `grep` outside the e57 reader/writer plugins where in their respective tests. This seems to be caused by the architecture of the tests, which are testing the PDAL reader/writer interfaces against the direct e57 implementation available in these plugins. So they have to link against the e57 library because the library is actively used in the test, not just because the reader/write include files force linking the library (which they previously did, but that dependency is removed in my branch).

It also seems that a lot of functions are explicitly exported from these plugins using the `PDAL_DLL` annotation just to use them in the tests. The number of such exported functions could be reduced (up to the bare PDAL `Reader/Writer/Streamable` interface functions) but this breaks compatibility with current interfaces (of which I'm not sure anybody uses these outside of the mentioned tests).

I removed most includes of E57 specifics from the headers, except from [`Scan.hpp`](https://github.com/PDAL/PDAL/blob/master/plugins/e57/io/Scan.hpp#L55) which directly returns a copy of an E57 library type (`e57::CompressedVectorNode`). Again I assume this is for testing only and no user including only `E57(Reader|Writer).hpp` would need to include `Scan.hpp` or link to the E57 library directly I believe.

So I will clean up my mess in the following days and make a PR for you to review. This PR will focus on removing the e57 internals from all headers where possible, without changing functions which I believe were intentionally exported from the shared library. This will remove the direct linking of the e57 library for the `pdal_io_e57_write_test`, but not the other test (which actively uses the library in test code).

PR #2782, opened 2019-10-13
Remove E57 dependency from PDAL plugin headers

norbertwenzel:
As discussed with @abellgithub in #2707 the E57 reader/writer plugins included the E57Format library in the public headers and therefore required clients and tests to link against this library. In this PR I removed the depedencies on the E57Format library as much as possible (hopefully) without any major changes to functions exported by the plugins.

The only change I've (intentionally) made is that the (already previously) `private` internal classes `ChunkReader` & `ChunkWriter` are now only forward declared in header files and marked with `PDAL_LOCAL`. These classes depend on E57 library types and were therefore removed from the headers.

This change allows to remove the dependency on the E57Format library for the writer test. The dependency remains for the reader test, since this test uses the publicly exported `e57::Scan` class to directly access E57 types in the tests.
For users of the plugins (`E57Reader` & `E57Writer`) a dependency on `e57::Scan` and therefore the E57Format library is not required as long as scans stored in E57 are not accessed individually.

I ran the corresponding tests successfully on my computer (Ubuntu Bionic with Clang 8). For the tests I've built all e57 code (both plugins and the E57Format library) with default and inline visibility set to `hidden`. After removing all tests that required `e57::Scan` both tests could be built, linked and ran successfully *without* linking to E57Format.

### DIFF
--- a/plugins/e57/io/E57Reader.cpp
+++ b/plugins/e57/io/E57Reader.cpp
@@ -33,13 +33,47 @@
 ****************************************************************************/
 
 #include <iostream>
+#include <E57Format.h>
 #include <E57Exception.h>
 
 #include "E57Reader.hpp"
+#include "Scan.hpp"
 #include "Utils.hpp"
 
 namespace pdal
 {
+
+class E57Reader::ChunkReader
+{
+public:
+    ChunkReader(const point_count_t &pointOffset,
+                const point_count_t &maxPointRead,
+                const std::shared_ptr<e57::Scan> &scan,
+                const std::set<std::string> &e57Dimensions);
+
+    ~ChunkReader();
+
+    // returns false if the index falls out of the
+    // [pointOffset,pointOffset + m_maxPointRead] interval
+    bool isInScope(point_count_t index) const;
+
+    bool isInChunk(point_count_t index) const;
+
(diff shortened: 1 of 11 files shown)


plugins/e57/io/E57Reader.cpp
```diff
@@ -33,14 +33,48 @@
...
 #include <E57Exception.h>
 
 #include "E57Reader.hpp"
+#include "Scan.hpp"
 #include "Utils.hpp"
 
 namespace pdal
 {
 
+class E57Reader::ChunkReader
```
pravinshinde825: Do we really need this class?
I believe its always better to have specific class to handle specific class.
But I think we can handle things better without this class, Please check #2757 .

norbertwenzel: Sorry, I didn't see your PR linked in #2707 where I started from.
My goal was to change as little as possible in order to remove the E57 stuff from header files only. So these classes have only been moved to the implementation and are unchanged otherwise. But if I were to completely reimplement these plugins I would probably remove them, yes.

abellgithub:
Dealt with in #2868 after integration of e57 changes.

PR #2783, opened 2019-10-14
EPT Zstandard support

### LINKED ISSUE #2750: ZStandard not included in ept.reader
The zstandandard compression format seems not to be included in ept.reader

Repeat:

- Configure and build PDAL with
```
$ cmake .. -DWITH_ZSTD=YES
```
- Build some ept with entwine, zstandard will be the default
- Try to read the ept repository with:
```
$ pdal info --driver readers.ept /my/path/to/ept.json
PDAL: readers.ept: Unrecognized EPT dataType: zstandard
```

```
$ pdal --version
pdal 2.0.1 (git-version: Release)
```
There was a merge request some time ago at:
https://github.com/PDAL/PDAL/pull/2561
that apparantly had no follow-up. Likely because this branch is waiting to be merged?
https://github.com/PDAL/PDAL/tree/ept-zstandard-updates

What is the current advise on using zstandard?

connormanning:
Closes #2750.  Initial code is from #2561 by @nicolas-chaulet, and with the addition of some test data and adding Zstandard streaming support since EPT streaming has landed on master since the initial code was written.

The commit history has some extraneous commits, since it includes some pre-squash streaming commits, but should be fine once it's squashed.


io/EptReader.cpp
```diff
@@ -710,8 +712,10 @@ PointViewSet EptReader::run(PointViewPtr view)
 
             if (m_info->dataType() == EptInfo::DataType::Laszip)
                 startId = readLaszip(*view, key, nodeId);
-            else
+            else if (m_info->dataType() == EptInfo::DataType::Binary)
                 startId = readBinary(*view, key, nodeId);
+            else
```
abellgithub: This is a bad default condition.

io/EptReader.cpp
```diff
@@ -710,8 +712,10 @@ PointViewSet EptReader::run(PointViewPtr view)
...
+                startId = readZstandard(*view, key, nodeId);
```
abellgithub: Seems like an #ifdef for ZSTD here would be cleaner.

io/EptReader.cpp
```diff
@@ -793,6 +797,45 @@ PointId EptReader::readBinary(PointView& dst, const Key& key,
...
+    dec.decompress(compressed.data(), compressed.size());
+
+    ShallowPointTable table(
+        *m_remoteLayout,
+        uncompressed.data(),
+        uncompressed.size());
+
+    PointRef pr(table);
+
+    std::lock_guard<std::mutex> lock(m_mutex);
```
abellgithub: This seems like a confusing and unnecessarily large mutex scope.  I would think it could go around the `setField()` calls in process().

io/EptReader.cpp
```diff
@@ -940,6 +983,8 @@ void EptReader::load()
 
             if (m_info->dataType() == EptInfo::DataType::Laszip)
                 readLaszip(nodeBuffer->view, key, nodeId);
+            else if (m_info->dataType() == EptInfo::DataType::Zstandard)
+                readZstandard(nodeBuffer->view, key, nodeId);
```
abellgithub: Default here is different from above.  Seems better to be explicit for all cases.

io/EptReader.cpp
connormanning: ef58e49a6596edcbfe80b30169142b7b1f734a96

io/EptReader.cpp
connormanning: 96a672f33f4b550e13ffb3bc1acf3b4c7f1c4f8c

io/EptReader.cpp
connormanning: 77890cf36cfe98559fc6975cc4fc48d37d673470

io/EptReader.cpp
connormanning: I'm not sure about this - we are appending points to a single `PointView` during each call to `process` and I'd be hesitant to introduce parallelism there.  I personally find this scope much easier to reason about - the multi-threading is used only to acquire chunks (and potentially decompress them into something we can `process`) from potentially remote data sources, and once a chunk is ready for processing we lock and then mutate the output `PointView` in a single-threaded manner.

I made a change that is loosely related to this comment [here](a415808813fd507c95534496ac7a652de429f563), although I did not change the mutex scope.  I think I'd prefer a separate issue be made if you'd like to change how we are threading things here, since the current PR maintains existing behavior of the binary reading logic.

io/EptReader.cpp
```diff
@@ -772,11 +784,10 @@ PointId EptReader::readLaszip(PointView& dst, const Key& key,
...
-        const uint64_t nodeId) const
+PointId EptReader::processPackedData(PointView& dst, const uint64_t nodeId,
+    char* data, const uint64_t size) const
 {
-    auto data(getBinary("ept-data/" + key.toString() + ".bin"));
-    ShallowPointTable table(*m_remoteLayout, data.data(), data.size());
+    ShallowPointTable table(*m_remoteLayout, data, size);
     PointRef pr(table);
 
     std::lock_guard<std::mutex> lock(m_mutex);
```
abellgithub: There's an extra `pointId` declaration below.  Can't figure out how to mark that line.

connormanning: Good catch.  d3dc6344d5bf4bf3a90616f5ac13d107fab68c97

PR #2784, opened 2019-10-14
Upper-bound point count for EPT info summary

connormanning:
If a spatial query is passed to `pdal info --summary` for an EPT resource, determine an upper bound for the point count.


io/EptReader.cpp
```diff
@@ -436,12 +436,28 @@ QuickInfo EptReader::inspect()
     {
         initialize();
 
-        qi.m_bounds = toBox3d(m_info->json()["boundsConforming"]);
+        const BOX3D fullBounds(toBox3d(m_info->json()["boundsConforming"]));
```
abellgithub: Why has this changed?

connormanning: We use fullBounds below as well.

PR #2802, opened 2019-11-08
Enable Extended Local Maximum in filters.elm, add 'cls_dimension' and 'z_dimension' options to allow control of outputs

hobu:


filters/ELMFilter.hpp
```diff
@@ -58,9 +58,23 @@ class PDAL_DLL ELMFilter : public Filter
     double m_cell;
     double m_threshold;
     uint8_t m_class;
+    bool m_maximum;
+
+    Dimension::Id m_Zdim;
```
abellgithub: These names don't follow our normal convention.  It should probably be "m_zDim", "m_classDim", etc.

filters/ELMFilter.cpp
```diff
@@ -122,16 +116,61 @@ void ELMFilter::filter(PointView& view)
...
+    for (PointId id = 0; id < view.size(); ++id)
+    {
+        double x = view.getFieldAs<double>(Dimension::Id::X, id);
+        double y = view.getFieldAs<double>(Dimension::Id::Y, id);
+        double z = view.getFieldAs<double>(m_Zdim, id);
+
+        size_t c = static_cast<size_t>(std::floor(x - bounds.minx) / m_cell);
+        size_t r = static_cast<size_t>(std::floor(y - bounds.miny) / m_cell);
+
+        plus_hash[c * rows + r].emplace(z, id);
```
abellgithub: This is a relatively expensive way to do this.  We're just gridding points, so we're using twice as much memory to have a minimum and maximum grid.  Changing the cull function to look through the points from the other direction could eliminate the extra grid.

filters/ELMFilter.cpp
```diff
@@ -122,16 +116,61 @@ void ELMFilter::filter(PointView& view)
...
+
+}
+
+
+void ELMFilter::filter(PointView& view)
+{
+    BOX2D bounds;
+    calculateBounds(view, bounds);
+
+    size_t cols =
```
abellgithub: This is all pretty expensive as it takes three passes through the points.  One could certainly eliminate the bounds calculation -- you don't need a full grid, you just need cells.

filters/ELMFilter.cpp
```diff
@@ -66,43 +66,37 @@ void ELMFilter::addArgs(ProgramArgs& args)
     args.add("cell", "Cell size", m_cell, 10.0);
     args.add("class", "Class to use for noise points", m_class, uint8_t(7));
     args.add("threshold", "Threshold value", m_threshold, 1.0);
-}
+    args.add("maximum", "Compute Extended Local Maximum too", m_maximum, false);
```
abellgithub: I think this a confusing name.  Perhaps some enumeration would be more clear?  Noise::Low, Noise::High, Noise::Both or some such?  Default would be Noise::High.

filters/ELMFilter.cpp
abellgithub: Actually, this whole thing can be made much more lean -- you don't need to store all the point IDs to do this.  Most of the time you can determine at insertion time that a point isn't noise because its distance from some other point is less than the threshold.  Perhaps that could be improved in another PR.

filters/ELMFilter.hpp
chambbj: I also wonder how meaningful it is to offer an alternative to Z. Did you have another dimension in mind @hobu? Intensity? If we need to retain it, we can call it `m_fltDim` or similar.

filters/ELMFilter.cpp
chambbj: You can indeed do this without computing bounds and constructing the grid that way. I've created a new branch that offers a slightly modified approach. It doesn't really get us away from storing PointId's though, and I'm not yet convinced that we can. The problem is that when we consider whether or not a point is noise (either min or max), we need to compare it to the next highest/lowest point, which in turn is compared to the next highest/lowest point. We only ever stop iterating once we've found a point that is sufficiently close to the next in order of elevation. We don't really have enough information for a given cell until we've seen all points contained within that cell.

hobu:
The thought of multiple Z dimensions (ferry or ept_addon) was why I added that.

hobu:
We can do this later if there's a need.

PR #2807, opened 2019-11-13
Allow setting of spatial reference in filters.transform

### LINKED ISSUE #2796: assign spatialreference of filters.transformation
`filters.transformation` should have a `spatialreference` option to allow the user to override or assign the spatialreference from that point forward.

abellgithub:
Close #2796


abellgithub:
It's what we use in the readers.  We were migrating away from `spatialreference` for clarity.  `override_srs` seemed most consistent and most like the use elsewhere.  If we don't like that, we could use `out_srs`, which is what's in `filters.reprojection`, but that kind of implies that there's an `in_srs` and there isn't.

hobu:
`override_srs` is fine with me. Just making sure things are consistent.

PR #2820, opened 2019-11-22
cache TileDB attributes and remove extraneous call to schema

normanb:
Reviewing the `processOne` code there is a call to retrieve attributes that isn't required, this PR removes that and also caches the fetch of the attributes during the schema check in the `ready` function.


abellgithub:
Please see the error log.  The call to `schema` is missing `()`

Issue #2821, opened 2019-11-23
Inspect remote las without downloading the whole file


nicolas-chaulet:
The `pdal info --summary` is a very useful command to figure out what's in a file and is very efficient for local files. However for a file on a remote server it requires to download the whole file first which can be very time consuming.

abellgithub:
I think this should be dinner outside of the file handling. The reader should be agnostic to the data source.

@connormanning : I think we've been through this and you found a reasonable solution, correct?

connormanning:
Yes, Entwine handles LAS/LAZ specially during a shallow scan by downloading only the metadata portions and creating a valid LAS file with all the metadata pieces but without points.  Then we can simply run normal PDAL calls on this file since it's valid LAS but empty of points.

The code is [here](https://github.com/connormanning/entwine/blob/master/entwine/builder/scan.cpp#L178).  The steps are basically:

- Download 375 bytes (max header size of LAS versions 0-4)
- Deep-inspect the header for its true size (which may be 375 or 227 bytes - truncate it if it's only 227), and the VLR/EVLR offsets/sizes
- Download the VLRs if they exist, append them to the header
- If there are EVLRs, concatenate them to the buffer and set the EVLR offset in this buffer to point to the end of the VLRs

Now we can run PDAL's `preview` on this file.  You might also need to fudge the point count and legacy point count header fields depending on exactly what you're doing with the file.

nicolas-chaulet:
@abellgithub it seems complicated to take this out of the reader since every format would be handled differently (for the oens that have headers). Let me update the PR by using the whole entwine solution before calling initialize.

abellgithub:
I think this needs some more thought as to the best way to handle.

nicolas-chaulet:
Yes agree... I'll use `entwine scan` for now!

PR #2822, opened 2019-11-23
Lazy inspection os las files on remote servers

nicolas-chaulet:
Potential solution (inspired by entwine codebase) for #2821


nicolas-chaulet:
Looks like I broke quite a few things related to `preview()` and the management of options for readers. Any suggestion about how to solve those?

PR #2823, opened 2019-11-25
Adding EVLR offset to las files

nicolas-chaulet:
It seems that the EVlr offset is never set. This PR fixes that (with a test)

### DIFF
--- a/test/unit/io/LasWriterTest.cpp
+++ b/test/unit/io/LasWriterTest.cpp
@@ -1183,6 +1183,39 @@ TEST(LasWriterTest, pdal_add_vlr)
     EXPECT_EQ(nodes.size(), 2UL);
 }
 
+TEST(LasWriterTest, evlroffset)
+{
+    std::string outfile(Support::temppath("evlr.las"));
+
+    FileUtils::deleteFile(outfile);
+    {
+        LasReader r;
+        Options ro;
+        ro.add("filename", Support::datapath("las/wontcompress3.las"));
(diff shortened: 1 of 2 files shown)


test/unit/io/LasWriterTest.cpp
```diff
@@ -1183,6 +1183,39 @@ TEST(LasWriterTest, pdal_add_vlr)
...
+        wo.add("filename", outfile);
+        w.setOptions(wo);
+        w.setInput(r);
+
+        PointTable t;
+        w.prepare(t);
+        w.execute(t);
+        LasTester tester;
+        LasHeader *h = tester.header(w);
+        EXPECT_GT(h->eVlrOffset(),0);
```
abellgithub: There's really no reason not to test the actual value.  There are a lot of wrong values that are greater than 0.

nicolas-chaulet: Sure... Will add a better test.

test/unit/io/LasWriterTest.cpp
```diff
@@ -1189,16 +1189,19 @@ TEST(LasWriterTest, evlroffset)
 
     FileUtils::deleteFile(outfile);
     {
-        LasReader r;
+        StageFactory f;
+        Stage& r = *(f.createStage("readers.faux"));
```
abellgithub: No need to read 0 points.  Just pitch this and execute the writer.

abellgithub:
Close #2824

PR #2831, opened 2019-12-02
make sure to initialize outArrays

hobu:


abellgithub:
I saw this a couple of weeks ago but forgot to change.  I'm not sure, though, that nullptr initialization is sufficient.  I think there are some places that need to be checked/guarded from a nullptr.

hobu:
IIRC we are ok with a nullptr in this particular instance.

PR #2833, opened 2019-12-09
Add a_srs, t_srs, and default_srs hidden options to 'pdal info'

hobu:
It is common to want to see coordinates in a different coordinate system if you are normalizing a bunch of different data to a common coordinate system, and having `pdal info` be able to apply a `filters.reprojection` eliminates a full i/o operation that writes a new file that was reprojected to get coordinate statistics and information. 

Having done this, I can see I'd want `filters.transformation` also, but no other filters are reasonable to inline with `pdal info`. There isn't a need to add a `--json` switch like `pdal translate` has that takes in a generic pipeline.

### DIFF
--- a/kernels/InfoKernel.cpp
+++ b/kernels/InfoKernel.cpp
@@ -40,6 +40,7 @@
 #include <pdal/pdal_features.hpp>
 
 #include <filters/InfoFilter.hpp>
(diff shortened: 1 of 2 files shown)


abellgithub:
Why are we doing this?  Something for entwine?

PR #2838, opened 2019-12-10
tiledb: use internal array metadata for pipeline serialization

normanb:
In TileDB 1.7+ we support internal array metadata which removes the need for a sidecar JSON file. This PR adds support for internal metadata and has backward support for sidecar files for older versions.

I have added a conditional test for this change.

### DIFF
--- a/plugins/tiledb/io/TileDBReader.cpp
+++ b/plugins/tiledb/io/TileDBReader.cpp
@@ -308,9 +308,9 @@ void TileDBReader::localReady()
     {
         tiledb::VFS vfs(*m_ctx, m_ctx->config());
         tiledb::VFS::filebuf fbuf(vfs);
-        std::string metaFName = m_arrayName + pathSeparator + "pdal.json";
+        std::string metaFName = m_filename + pathSeparator + "pdal.json";
 
-        if (vfs.is_dir(m_arrayName))
+        if (vfs.is_dir(m_filename))
         {
             auto nBytes = vfs.file_size(metaFName);
             tiledb::VFS::filebuf fbuf(vfs);
--- a/plugins/tiledb/io/TileDBReader.cpp
+++ b/plugins/tiledb/io/TileDBReader.cpp
@@ -322,7 +322,9 @@ void TileDBReader::localReady()
         }
     }
 
-    if ((meta != nullptr) && (meta["writers.tiledb"].count("spatialreference") > 0))
+    if ((meta != nullptr) &&
+        (meta.count("writers.tiledb") > 0) &&
+        (meta["writers.tiledb"].count("spatialreference") > 0))
     {
         SpatialReference ref(meta["writers.tiledb"]["spatialreference"]);
         setSpatialReference(ref);
--- a/plugins/tiledb/io/TileDBReader.cpp
+++ b/plugins/tiledb/io/TileDBReader.cpp
@@ -34,8 +34,17 @@
 
 #include <algorithm>
 
+#include <nlohmann/json.hpp>
+
 #include "TileDBReader.hpp"
 
+
+const char pathSeparator =
+#ifdef _WIN32
+        '\\';
+#else
+        '/';
+#endif
 namespace pdal {
 
 static PluginInfo const s_info
@@ -281,6 +290,42 @@ void TileDBReader::localReady()
             subarray.push_back(kv.second.second);
         }
         m_query->set_subarray(subarray);
+    }
+
+    // read spatial reference
+    NL::json meta = nullptr;
+
+#if TILEDB_VERSION_MAJOR >= 1 && TILEDB_VERSION_MINOR >= 7
+    tiledb_datatype_t v_type = TILEDB_UINT8; // CPLSerializeXMLTree returns char*
+    const void* v_r;
+    uint32_t v_num;
+    m_array->get_metadata("_pdal", &v_type, &v_num, &v_r);
+    if (v_r != NULL)
+        meta = NL::json::parse(static_cast<const char*>(v_r));
+#endif
+
+    if (meta == nullptr)
+    {
+        tiledb::VFS vfs(*m_ctx, m_ctx->config());
+        tiledb::VFS::filebuf fbuf(vfs);
+        std::string metaFName = m_arrayName + pathSeparator + "pdal.json";
+
+        if (vfs.is_dir(m_arrayName))
+        {
+            auto nBytes = vfs.file_size(metaFName);
+            tiledb::VFS::filebuf fbuf(vfs);
+            fbuf.open(metaFName, std::ios::in);
+            std::istream is(&fbuf);
+            std::string s { std::istreambuf_iterator<char>(is), std::istreambuf_iterator<char>() };
+            fbuf.close();
+            meta = NL::json::parse(s);
+        }
+    }
+
+    if ((meta != nullptr) && (meta.count("spatialreference") > 0))
+    {
+        SpatialReference ref(meta["spatialreference"]);
+        setSpatialReference(ref);
     }
 
     // initialize read buffer variables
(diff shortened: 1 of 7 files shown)


normanb:
I am going to read the spatial reference written from the TileDBReader. Can we hold off merging until I have added this?

normanb:
I have added spatial reference support in the reader. This is ready for review.

plugins/tiledb/io/TileDBWriter.cpp
```diff
@@ -347,7 +347,7 @@ void TileDBWriter::ready(pdal::BasePointTable &table)
             else
             {
                 // check attribute exists in original tiledb array
-                auto attrs = m_schema->attributes();
+                auto attrs = m_array->schema().attributes();
```
normanb: I have reverted this back to `m_array->schema()` which was how it was in the original PR https://github.com/PDAL/PDAL/pull/2828 and this fixes a segfault in the test.

plugins/tiledb/io/TileDBReader.cpp
```diff
@@ -255,6 +264,42 @@ void TileDBReader::ready(PointTableRef)
...
+    m_array->get_metadata("_pdal", &v_type, &v_num, &v_r);
+    if (v_r != NULL)
+        meta = NL::json::parse(static_cast<const char*>(v_r));
+#endif
+
+    if (meta == nullptr)
+    {
+        tiledb::VFS vfs(*m_ctx, m_ctx->config());
+        tiledb::VFS::filebuf fbuf(vfs);
+        std::string metaFName = m_arrayName + pathSeparator + "pdal.json";
```
abellgithub: Maybe I'm missing something, but if what you're trying to do is read the contents of `pdal.json` into a string, you can just use `pdal::FileUtils::readFileIntoString(filename)`

plugins/tiledb/test/TileDBReaderTest.cpp
```diff
@@ -154,6 +158,49 @@ class TileDBReaderTest : public ::testing::Test
...
+
+#if TILEDB_VERSION_MAJOR >= 1 && TILEDB_VERSION_MINOR >= 7
+    TEST_F(TileDBReaderTest, spatial_reference)
+    {
+        tiledb::Context ctx;
+        tiledb::VFS vfs(ctx);
+        std::string pth = Support::temppath("tiledb_test_srs");
+
+        Options options;
+        std::string sidecar = pth + "/pdal.json";
```
abellgithub: Unused

plugins/tiledb/io/TileDBReader.cpp
normanb: @abellgithub I use VFS for the cases where this file in on S3, this PR moves away from that to use internal array metadata, this code is there to keep compatibility with older library versions.

plugins/tiledb/test/TileDBReaderTest.cpp
normanb: yes, thanks, will remove this

normanb:
I will resolve the conflicts.

normanb:
@abellgithub I have included the stage metadata in the saved metadata with the array. Is this sufficient, or can I also save the entire pipeline?

abellgithub:
@normanb That seems sufficient to me unless you have some special purpose for saving metadata to be re-read.

abellgithub:
@normanb  Are you ready for this to be merged?

normanb:
@abellgithub thank you, this is ready to be merged.

normanb:
@abellgithub actually, I have found a bug with the CLI, I will work through it and update this PR.

normanb:
I have debugged this issue @abellgithub if I validate the pipeline (in Python) for executing then the stage metadata is `[]`, if I don't validate then the stage metadata (for this dataset) is `{
  "writers.tiledb":
  {
  }
}`. How do I capture the whole pipeline metadata? e.g in python this is the dictionary returned in `pipeline.metadata`.

abellgithub:
Sorry not to have gotten back to you.  I don't understand exactly what the issue is or what you're trying to do.  You're sticking metadata from the PDAL pipeline in the tiledb file.  I get that much.  But what is the issue with the metadata that you're currently writing?

normanb:
There is an inconsistency in the stage metadata between validating a pipeline and not validating a pipeline in that once the pipeline has been validated then the stage metadata is empty.

It may be better to have the complete pipeline metadata in the output array. My question is whether that is a good thing to do.

abellgithub:
I don't understand.  You're writing some metadata to the output TileDB file.  But...so what?  What is happening with the data after that?  What are you doing WRT validating the data you've written?

normanb:
If I run the following python code

``` 
pipeline = pdal.Pipeline(req)
pipeline.validate()
pipeline.execute()
```
then the metadata that is written to the TileDB array (the stage metadata) is empty i.e. `[]` whereas without the call to `validate` then the metadata is `tiledb.writer ...`, the complete stage metadata.

However am I better off writing the complete pipeline metadata as this contains the spatial reference system information and other processing steps used to create this array?

```
pipeline.metadata
```

The metadata written to the array is read back on reading the array to parse the spatial reference system. https://github.com/PDAL/PDAL/pull/2838/files#diff-97af259ddfac6e5f2e7574faac1dd54eR327 and should be written consistently.

abellgithub:
Hmm.  Still not sure I understand exactly.  Maybe we can just talk in person in a week or so?

I will say that if what you want is to write CRS to your TileDB metadata store, just do it.  You don't need to rely on the PDAL metadata mechanism for that.

normanb:
@abellgithub yes, talking in person next week works for me :+1:

normanb:
@abellgithub as we discussed, I have created https://github.com/PDAL/PDAL/issues/2891 to capture the bug and this PR is ready to be merged.

The CI failures seem to be unrelated to this PR.

normanb:
This is ready to be merged.

hobu:
bouncing for CI churn

plugins/tiledb/io/TileDBReader.cpp
```diff
@@ -34,8 +34,17 @@
 
 #include <algorithm>
 
+#include <nlohmann/json.hpp>
+
 #include "TileDBReader.hpp"
 
+
+const char pathSeparator =
```
hobu: FIXME: When we go to C++17, we should use std::filesystem:: preferred_separator

plugins/tiledb/io/TileDBReader.cpp
```diff
@@ -283,6 +292,44 @@ void TileDBReader::localReady()
         m_query->set_subarray(subarray);
     }
 
+    // read spatial reference
+    NL::json meta = nullptr;
+
+#if TILEDB_VERSION_MAJOR >= 1 && TILEDB_VERSION_MINOR >= 7
+    tiledb_datatype_t v_type = TILEDB_UINT8; // CPLSerializeXMLTree returns char*
```
hobu: Dead comment about CPLSerializeXMLTree?

normanb: yes, sorry, will fix that.

plugins/tiledb/io/TileDBReader.cpp
normanb: :+1:

normanb:
Can this be added to the 2.1 milestone? It doesn't add a user feature so I understand if not, but does enable versioned metadata going forward.

hobu:
Added. How close to done is it?

normanb:
@hobu it is done, @abellgithub and I discussed it at the sprint and we created #2891 to capture the bug

hobu:
bouncing CI

hobu:
I wonder why CI doesn't finish here. can you rebase it so we can try with whatever our latest CI churn has been?

normanb:
rebased and enabled TileDB tests in cmake.

PR #2841, opened 2019-12-12
pass extra options through to the reader/writer in tile application

jtfell:
Adds the `override_srs`, `use_eb_vlr`, `in_compression` and `out_compression` options to `pdal tile` application by passing the options through to the `reader.las` and `writer.las` stages.

I"m happy to change the option naming if someone has better ideas.

### DIFF
--- a/doc/apps/tile.rst
+++ b/doc/apps/tile.rst
@@ -22,15 +22,31 @@ differs in several ways.  The ``tile`` command:
 
 ::
 
-    --input, -i     Input filename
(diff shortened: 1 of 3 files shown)


abellgithub:
Stage-specific settings should get passed using the standard syntax `--readers.stage.option=`.  If `pdal tile` doesn't process these options, that's what needs to be modified.  If you want to pursue this, you might look at `Kernel.cpp`.

PR #2845, opened 2019-12-13
tiledb: Fix name clash with filters

normanb:
In the `TileDBWriter` we had an option called `filters` which was fine in code usage (as exercised by the tests), but clashed with the CLI. This PR changes the filter option name and updates the tests.

### DIFF
--- a/doc/stages/writers.tiledb.rst
+++ b/doc/stages/writers.tiledb.rst
@@ -62,7 +62,7 @@ append
 stats
   Dump query stats to stdout [Optional]
 
-filters
+attr_filters
   JSON array or object of compression filters for either `coords` or `attributes` of the form {coords/attributename : {"compression": name, compression_options: value, ...}} [Optional]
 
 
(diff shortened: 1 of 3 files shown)


abellgithub:
What was the use case that exposed this as a problem?

normanb:
The use case was using the `pdal pipeline` cli with a pipeline that contained multiple attribute compression filters. "Filters" clashed as a name.

```
{
	"pipeline": [{
			"type": "readers.las",
			"filename": "ot_C320000_4172000.laz"
		},
		{
			"type": "writers.tiledb",
			"config_file": "tiledb.config",
			"array_name": "sample_comp_array",
			"attr_filters": {
				"coords": {
					"compression": "zstd",
					"compression_level": 75
				},
				"Intensity": {
					"compression": "bzip2",
					"compression_level": 5
				},
				"ReturnNumber": {
					"compression": "zstd",
					"compression_level": 95
				}
			}
		}
	]
}
```

abellgithub:
When I run this pipeline with `attr_filters` replaced with `filters`, it seems to run with the current code base, so I'm not understanding the issue.  (I did have to remove the `config_file` entry.)

normanb:
I can't recreate the issue with a clean environment, so we can close this.

PR #2846, opened 2019-12-13
Improve filters.hag for delaunay case

abellgithub:


abellgithub:
Maybe we should talk about this when you have time?  I still have a question or two.

doc/stages/filters.hag.rst
```diff
@@ -122,23 +127,44 @@ or :ref:`filters.smrf` to label ground returns, as shown
...
+    the `delaunay`_ option. [Default: 1]
+
+_`delaunay`
+    If true, create a delaunay triangulation of neighboring ground points
+    to aid in interpolation of a ground height to use as a difference to
+    a point's non-ground height. [Default: false]
+
+max_distance
+    Eliminate non-ground points from consideration when their distance
+    from a ground point exceeds this amount.  Not used with the `delaunay`_
```
chambbj: I'm not sure this is actually what we are doing, whether or not it's what we intend to do. The way I'm reading the code, the loop over the ground neighbors breaks whenever the threshold is exceeded, but any neighbors encountered up until that point still impact the result.

filters/HAGFilter.cpp
```diff
@@ -136,103 +213,88 @@ void HAGFilter::filter(PointView& view)
...
         PointRef point = ngView->point(i);
-        double x0 = point.getFieldAs<double>(Dimension::Id::X);
-        double y0 = point.getFieldAs<double>(Dimension::Id::Y);
-        double z0 = point.getFieldAs<double>(Dimension::Id::Z);
-        auto ids = kdi.neighbors(point, m_count);
-        double z1 = std::numeric_limits<double>::quiet_NaN();
+        double x0 = point.getFieldAs<double>(Id::X);
+        double y0 = point.getFieldAs<double>(Id::Y);
+        double z0 = point.getFieldAs<double>(Id::Z);
+        PointIdList ids = kdi.neighbors(point, m_count);
```
chambbj: If we use `knnSearch`, then we get the squared distances for free and do not need to recalculate within the loop below.

doc/stages/filters.hag.rst
abellgithub: But the neighbors come in order.  Closest first...

chambbj: Right, but if you look at that section of code, as I'm reading it, the first neighbors may very well be under the max distance, and will begin to update `z_accumulator` and `weights`. When a neighbor exceeds the distance, it bails and does the final update to `z1` based on the values accumulated up to that point. Maybe the wording is just backwards and should be "Eliminate ground points from consideration when their distance from a non-ground point exceeds..."? The way it is written now could lead one to believe that non-ground points will not have HAG computed if the ground neighbors are too far away (whether that's the nearest ground point, or just any of the neighbors is unclear).

abellgithub: Yes, backwards.  Do you have better wording?  I hate to get too verbose.

doc/stages/filters.hag.rst
```diff
@@ -164,7 +164,11 @@ max_distance
...
-    Does no extrapolation.  For ``HeightAboveGround`` to be calculated,
-    a ground point must be at the same location as the non-ground point
-    being evaluated.  Not used with the `delaunay`_ option.
+    If false and a non-ground point lies outside of the bounding box of
+    all ground points, its ``HeightAboveGround`` is set to 0.  If true
+    and ``delaunay`` is set, the ``HeightAboveGround`` is set to the
+    difference between the heights of the non-ground point and nearest
+    ground point.  If true and using neighbor interpolation (``delaunay``
+    is false), extrapolation is used to assign the ``HeightAboveGround``
+    value.
```
chambbj: > extrapolation is used to assign the ``HeightAboveGround`` value.

Still a little confusing, as it is all a form of extrapolation, regardless of if you are using Delaunay or neighbor interpolation. As I read it, the neighbor interpolation just continues to do what it does inside the bounding box, which is effectively IDW. I need to reread, but I'm not sure we ever even say this. IIRC, we call it a "weighted" approach rather than using the common name.

filters/HAGFilter.cpp
```diff
@@ -215,85 +254,53 @@ void HAGFilter::filter(PointView& view)
...
         std::vector<double> sqr_dists(m_count);
         kdi.knnSearch(x0, y0, m_count, &ids, &sqr_dists);
 
-        double z1 = gView->getFieldAs<double>(Id::Z, ids[0]);
+        // Closest ground point.
+        double x = gView->getFieldAs<double>(Id::X, ids[0]);
+        double y = gView->getFieldAs<double>(Id::Y, ids[0]);
+        double z = gView->getFieldAs<double>(Id::Z, ids[0]);
+
         assert(ids.size() > 0);
```
chambbj: Originally thought this assert should be moved before the preceding three lines of code (275-277) where it is used, but is it even worth checking? We already check that the ground view is not empty at line 250. I can't think of a way that `ids` could be empty with a non-empty ground view.

filters/HAGFilter.cpp
```diff
@@ -45,6 +45,133 @@
...
+{
+    double weights = 0;
+    double z_accumulator = 0;
+
+    for (size_t j = 0; j < ids.size(); ++j)
+    {
+        auto z = gView->getFieldAs<double>(Dimension::Id::Z, ids[j]);
+        double sqr_dist = sqr_dists[j];
+        if (maxDistance2 > 0 && sqr_dist > maxDistance2)
+            break;
```
chambbj: If we are going to keep the current behavior, which is to not give a result at all (well, give the default Z that is) if any of the neighbors exceeds the max distance, then we can just compare the last element of `sqr_dist` before we even enter the for loop and save a few iterations here and there.

doc/stages/filters.hag.rst
chambbj: Perhaps something like

> Use default Z value whenever any neighbors in the ground set exceeds the maximum distance to the non-ground point.

doc/stages/filters.hag.rst
abellgithub: Not exactly.  For delaunay interpolation there is no extrapolation, really.  If the switch is set to true and the test point is outside of the triangulation, we just use the height of the closest ground point as the base.

I get that this is confusing, which is why I wondered if we needed it at all.  I'm unsure what to do.

filters/HAGFilter.cpp
abellgithub: This isn't what is happening.  We use all the neighbors whose distance from the test point is less than the max distance that was specified.

chambbj: Right, I got it confused again. Sorry. I'll take another look at the documentation as I'm still not convinced it reflects what's being done here.

doc/stages/filters.hag.rst
chambbj: Based off our other conversation on the topic

> Use only ground points within `max_distance` of non-ground point when performing neighbor interpolation.

filters/HAGFilter.cpp
```diff
@@ -59,180 +186,122 @@ std::string HAGFilter::getName() const
...
-            m_count, point_count_t(1));
-    args.add("max_distance", "The maximum distance to the farthest nearest neighbor before the height above ground is not calculated [default: 0 (disabled)]", m_max_distance);
-    args.add("allow_extrapolation", "If true and count > 1, allow extrapolation [default: true].",
-            m_allow_extrapolation, true);
-    args.add("delaunay", "Construct local Delaunay fans and infer heights from them [default: false].",
-            m_delaunay, false);
+    args.add("count", "The number of points to fetch to determine the "
+        "ground point [default: 1].", m_count, point_count_t(1));
+    args.add("max_distance", "The maximum distance to the farthest nearest "
+        "neighbor before the height above ground is not calculated "
```
chambbj: Perhaps

> Ground points beyond the maximum distance will not influence nearest neighbor interpolation of height above ground.

abellgithub:
@chambbj : I changed the `max_distance` text.  If you don't have any issues with the functionality, I'll go ahead and merge when tests pass.  If you want to improve the doc text, I encourage you to do so.

BC14w:
I'm getting errors using pdal.hag_delaunay. I have an area I scanned that was pretty flat but not perfectly flat and I'm trying to create my hag, but I get an error: All point collinear. I wouldn't get this error in the past when I was just using pdal.hag. Any clues why this is happening?

abellgithub:
The nearest neighbors to one of the points in your set are all collinear with the point in question. You can change "count" to make a larger set, or you can use one of the other hag filters.

PR #2864, opened 2020-01-07
Hope this addresses the issue Even brought up.  We need a test.

abellgithub:
See https://github.com/PDAL/PDAL/commit/ec08617b363e5f0772ff7289d8777326868ee2cd#commitcomment-36659616


chambbj:
@abellgithub I see it's already merged, but this looks like a good solution. I'd only add that we could probably break the for loop once the skewness becomes positive.

abellgithub:
I found an issue, but I don't understand this so maybe we should talk about what we're trying to do.  After looking again, I don't get why skewness is being stored in a vector at all.

chambbj:
We can talk any time. Agreed. The way it is now setup, the vector is no longer needed.

PR #2865, opened 2020-01-07
Prevent filters.hexbin from creating invalid geometry

abellgithub:


abellgithub:
Sorry.  We were creating invalid geometry.  What I *thought* at first was happening is that we were removing negative space containing positive space (islands).  We certainly could do that with the baseline code.  So, I wrote things so this wouldn't happen.  It turns out that this wasn't the problem.  There was some issue with removing incorrect rings altogether.  I don't know what it was and would have to go back to investigate.

I also thought there might have been a problem in the hexbin code that invoked the simplification and in order to figure out what was going on, I reorganized things so that the bits that were associated were together.  In the end, I don't believe that was a problem, but I also don't think that the code is worse for the reorganization.  I can revert that if you think it would be helpful.

abellgithub:
If you want to walk through the changes on a call, that would be fine.

hobu:
> If you want to walk through the changes on a call, that would be fine.

No need. Your comment is enough background. The PR title isn't descriptive enough. Maybe it should be "Prevent `preserve_topology` option of `filters.hexbin` from creating invalid geometry"

PR #2866, opened 2020-01-08
More cleanup of skewness filter

abellgithub:
I'd still like to move the skewness calculation to another function, but I'll do that separately.  Think it's easier to review in pieces.


filters/SkewnessBalancingFilter.cpp
```diff
@@ -67,7 +67,17 @@ std::set<PointId> SkewnessBalancingFilter::processGround(PointViewPtr view)
     point_count_t n1(0);
     double delta, delta_n, term1, M1, M2, M3;
     M1 = M2 = M3 = 0.0;
-    std::vector<double> skewness(view->size());
+
+    auto setGround = [&view](PointId start, PointId end)
```
connormanning: To me it's unexpected that `end` would be inclusive here, when I see `begin`/`end` type ranges to a function I'd expect STL-like behavior which wouldn't include the `end`.  However the scope here is very localized so I could live with it.

abellgithub: Is it the name you don't like?  The point is not to overflow.

filters/SkewnessBalancingFilter.cpp
```diff
@@ -67,7 +67,17 @@ std::set<PointId> SkewnessBalancingFilter::processGround(PointViewPtr view)
...
+    {
+        // 2 is the ASPRS LAS classification for ground.
+        for (PointId idx = start; idx <= end; ++idx)
+            view->setField(Dimension::Id::Classification, idx, 2);
```
chambbj: This isn't so much a comment on what you've done here, but it dawns on me that we should probably change this filter to behave more like the other ground filters, setting those points that are not ground to a Classification value of 1. Would clearly happen elsewhere in the code though.

filters/SkewnessBalancingFilter.cpp
```diff
@@ -79,61 +89,32 @@ std::set<PointId> SkewnessBalancingFilter::processGround(PointViewPtr view)
...
-    std::set<PointId> groundIdx;
-    for (PointId i = 0; i <= j; ++i)
-        groundIdx.insert(i);
-
-    log()->get(LogLevel::Debug)
-        << "Stopped with " << groundIdx.size()
-        << " ground returns and skewness of " << skewness[j] << std::endl;
-
-    return groundIdx;
+        lastSkewness = skewness;
```
chambbj: Still trying to convince myself that this actually does what we want. It seems entirely possible that as you iterate through the points the skewness could at times be positive and others negative. You allow this no doubt, but it seems as though only those points between a negative skewness and the next positive skewness will be classed ground. The intent of the algorithm is to only lop off those positive skewness values once the whole distribution is known.

filters/SkewnessBalancingFilter.cpp
abellgithub: I can do that here.  Do you want me to do that?

filters/SkewnessBalancingFilter.cpp
abellgithub: I believe that this is slightly different implementation of the same algorithm.  I've run the tests on the old and new code and the results match.

What you say is true -- we classify all points between a negative skewness and a positive one as ground.  But we do this repeatedly,  Ultimately this means that we classify all points from the beginning to the last point with non-positive skewness as ground, even if the skewness flips back and forth between non-positive and positive.  I don't think that this is necessary.  I believe that because the points are sorted, skewness will necessarily be monotonically increasing so there will only be one transition to positive, but I wasn't sure when I started so I did this.

filters/SkewnessBalancingFilter.cpp
chambbj: I think we should. Thanks.

filters/SkewnessBalancingFilter.cpp
chambbj: I'm convinced now. To the question of monotonically increasing skewness, it is not guaranteed, but the logic holds regardless. It was just not immediately obvious to me (overlooked the bookkeeping on the last transition to positive as the start of the range to set as ground).

abellgithub:
@chambbj  - Would you look over the small changes I just pushed?

chambbj:
Yes, these still look fine @abellgithub.

PR #2867, opened 2020-01-10
Support passing a reference to a .py script to readers.numpy along with function/module/args options that reference a function that returns a Numpy array

hobu:


plugins/python/io/NumpyReader.cpp
```diff
@@ -147,6 +166,34 @@ PyArrayObject* load_npy(std::string const& filename)
...
 }
 
+PyArrayObject* load_npy_script(std::string const& source,
+                               std::string const& module,
+                               std::string const& function,
+                               std::string const& fargs)
+{
+
+    MetadataNode m;
+    std::unique_ptr<plang::Script> script(new plang::Script(source, module, function));
```
abellgithub: Is there some reason for these not just to be on the stack?

plugins/python/io/NumpyReader.cpp
```diff
@@ -147,6 +166,34 @@ PyArrayObject* load_npy(std::string const& filename)
...
+    {
+        PyObject* arg = PyUnicode_FromString(args[i].c_str());
+            if (!arg)
+                throw pdal_error(plang::getTraceback());
+        PyTuple_SetItem(scriptArgs, i, arg);
+
+    }
+
+    PyObject *array = PyObject_CallObject(method->m_function, scriptArgs);
+    if (!array)
```
abellgithub: scriptArgs needs a decref here.

plugins/python/io/NumpyReader.cpp
```diff
@@ -163,14 +210,25 @@ void NumpyReader::initialize()
...
     {
-        m_array = load_npy(m_filename);
+        m_array = load_npy_file(m_filename);
     }
-    if (m_array)
-        if (!PyArray_Check(m_array))
-            throw pdal::pdal_error("Object in file  '" + m_filename +
-                "' is not a numpy array");
+
+    if (!PyArray_Check(m_array))
```
abellgithub: This error is misleading when loading from a script.  Perhaps it should be moved into the second case and a different error provided for the first.

plugins/python/io/NumpyReader.cpp
```diff
@@ -71,6 +75,14 @@ static PluginInfo const s_info
     ""
 };
 
+struct NumpyReader::Args
+{
+    std::string module;
+    std::string function;
+    std::string source;
```
abellgithub: Don't thing `source` belongs in this struct.

plugins/python/io/NumpyReader.cpp
```diff
@@ -239,6 +297,13 @@ void NumpyReader::addArgs(ProgramArgs& args)
     m_orderArg = &args.add("order", "Order of dimension interpretation "
         "of the array.  Either 'row'-major (C) or 'column'-major (Fortran)",
         m_order);
+
+    args.add("module", "Python module containing the function to run",
+        m_args->module, "");
```
abellgithub: You don't need the empty string defaults.

test/data/plang/threedim.py
```diff
@@ -0,0 +1,8 @@
+import numpy as np
+
+def load(filename):
+    array = np.load(filename)
+    return array
+
+#a = load('threedim.npy')
```
abellgithub: Dead code.

plugins/python/test/NumpyReaderTest.cpp
```diff
@@ -236,3 +236,60 @@ TEST(NumpyReaderTest, threedim_col)
...
+TEST(NumpyReaderTest, threedim_row_script)
+{
+    StageFactory f;
+
+    Options opts;
+    opts.add("filename", Support::datapath("plang/threedim.py"));
+    opts.add("order", "row");
+    opts.add("function", "load");
+    opts.add("module", "threedim");
+    opts.add("fargs",  Support::datapath("plang/threedim.npy"));
```
abellgithub: This seems wrong.  The function takes no args, right?

plugins/python/io/NumpyReader.cpp
```diff
@@ -147,6 +166,34 @@ PyArrayObject* load_npy(std::string const& filename)
...
 
+PyArrayObject* load_npy_script(std::string const& source,
+                               std::string const& module,
+                               std::string const& function,
+                               std::string const& fargs)
+{
+
+    MetadataNode m;
+    std::unique_ptr<plang::Script> script(new plang::Script(source, module, function));
+    std::unique_ptr<plang::Invocation> method(new plang::Invocation(*script, m, fargs));
```
abellgithub: I don't understand this.  Can't you just call the method with the args without using Script and Invocation?  It seems like that's all you're doing in the end anyway.

plugins/python/test/NumpyReaderTest.cpp
hobu: The user function can take in the user's args. For example, in the test, we take in a reference to a npy name to load with our python script and pass back to the `readers.numpy`

plugins/python/io/NumpyReader.cpp
hobu: Invocation is what does the compilation of the user's script. We need the code they gave us (through `m_filename` or as the `source` option, compiled to Python objects before we can call any functions.

test/data/plang/threedim.py
hobu: addressed by fddec900a5c59978842bd541e53784e578215084

plugins/python/io/NumpyReader.cpp
hobu: addressed by fddec900a5c59978842bd541e53784e578215084

plugins/python/io/NumpyReader.cpp
hobu: addressed by fddec900a5c59978842bd541e53784e578215084

plugins/python/io/NumpyReader.cpp
hobu: addressed by fddec900a5c59978842bd541e53784e578215084

plugins/python/io/NumpyReader.cpp
hobu: addressed by fddec900a5c59978842bd541e53784e578215084

plugins/python/io/NumpyReader.cpp
hobu: We need something to carry the source of the `m_filename` we were given and pass it into plang::Script to be compiled. I'm not sure what you are suggesting with its removal.

abellgithub: It's not an option that's handled in `addArgs`.  It can just be a variable on the stack at the point of assignment if I'm not mistaken.

hobu: Ah. I kept the interface of `filters.numpy` because I suppose we might also want to directly pass in python source via arg rather than via m_filename. I was on the fence on that and left it off for now.

PR #2874, opened 2020-01-15
Fixes memory leaks found when runing ctest

rouault:
Found with a -fsanitize=address build

### DIFF
--- a/io/BpfCompressor.cpp
+++ b/io/BpfCompressor.cpp
@@ -71,7 +71,7 @@ void BpfCompressor::compress()
     uint32_t rawWritten = (uint32_t)m_out.position();
 
     // Pop our temp stream so that we can write the real output file.
-    m_out.popStream();
+    delete m_out.popStream();
 
     m_rawSize += rawWritten;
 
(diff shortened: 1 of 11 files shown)


rouault:
Even if the fixes don't seem important for production use, they are a precondition if one day you want to run ctest on a ASAN enabled build (which would be a good engineering practice IMHO).
I'm not sure which exception should be caught exactly in the stream reading part ?

abellgithub:
I've run ASAN on things at times.  I guess I don't get too shook if it reports errors that I understand.

I'll merge this and go back and try to handle the wildcard exceptions so that it's clear in the code what the failure might be.

rouault:
>  I guess I don't get too shook if it reports errors that I understand.

my point was to enable it on CI, so that pull requests get tested for correctness. In such setup, you want everything reported to have been fixed.

abellgithub:
Yes, I understand.  I'm not sure that I agree with the principle.  I've spent many weeks of my life satisfying lint-type systems for no real benefit and more obtuse code.  (Casts that change no behavior but quiet the compiler are, to me, particularly annoying, and I think a net harm.)

But sometimes such tools find important issues.  I'm not sure what the best solution is.

Thanks for taking the time.  I'm not sure that I'm pleased or disappointed that you didn't find something more interesting.

PR #2875, opened 2020-01-15
Create enum for classification labels

chambbj:


pdal/PointView.hpp
```diff
@@ -559,7 +559,7 @@ void PointView::setField(Dimension::Id dim, PointId idx, T val)
         ok = convertAndSet<T, uint64_t>(dim, idx, val);
         break;
     case Dimension::Type::None:
-        val = (T)0;
+        val = 0;
```
abellgithub: I'm thinking that the cast was there for a reason.  Is there a reason to remove it?

chambbj: I had introduced it in 6eae9061404f1bb18b131d1bb7eb1e5bb5da80ae to try to work through the casting in calls to setField.

abellgithub: 👍

PR #2879, opened 2020-01-22
Python filter doc

jrobichaud:
While I was trying to test the python filter I found a few issues in this documentation.

I updated it.

Each changes are separated in different commits.

It was not specified explicitly in the documentation but Numpy is obviously required otherwise the process will just crash without warning even if you just want to put a print in the script.

image


hobu:
Thanks for the doc update. 

>  Numpy is obviously required otherwise the process will just crash without warning even if you just want to put a print in the script.

We should fix THAT bug too.

Issue #2638, opened 2019-07-29
Add "where" DimRange option to PDAL filters


hobu:
Add a `where` option to filters that are mixed-in (via subclass?) that applies filtering operations to only points that meet the DimRange constraint.

#1957 is similar but opposite.

PR #2881, opened 2020-01-23
Add raster option to HAGFilter

jtfell:
I had a look at adding a test for this mode, but didn't understand the geometry used in the existing test suite (https://github.com/PDAL/PDAL/blob/master/test/data/filters/hagtest.jpg).

If this is likely to get accepted I'll go ahead and add documentation and try to figure out how to create a DEM raster to use in the test suite.

### DIFF
--- a/doc/stages/filters.hag.rst
+++ b/doc/stages/filters.hag.rst
@@ -21,7 +21,7 @@ also be referred to as *height above ground* (HAG) or *above ground level* (AGL)
 heights. In the end, it is simply a measure of a point's relative height as
 opposed to its raw elevation value.
 
-There are two modes of operation:
+There are three modes of operation:
 
 Weighted Nearest Point
 ----------------------
@@ -51,6 +51,19 @@ If, for example, all the ground points near a non-ground point lay on
 one side of that non-ground point, finding a containing triangle will fail.
 
 .. embed::
+
+DEM Raster Comparison
+---------------------
+Used when the `raster`_ option is set. The filter loads a GDAL readable raster
+specifying the DEM to calculate ``HeightAboveGround`` from. The ``Z`` value
+of each point is compared against the value at the corresponding X,Y point in
+the DEM raster.
+
+If `respect_ground_classification`_ is set, any points classified as ground will
+have their ``HeightAboveGround`` value set to 0 regardless of the raster.
+
+Any points that do not have a corresponding value in the raster DEM will not have
+a ``HeightAboveGround`` value set.
 
 Example #1
 ----------
@@ -158,6 +171,9 @@ _`delaunay`
     to aid in interpolation of a ground height to use as a difference to
     a point's non-ground height. [Default: false]
 
+_`raster`
+    GDAL-readable raster to use for DEM (uses band 1, starting from 1)
+
 max_distance
     Use only ground points within `max_distance` of non-ground point when
     performing neighbor interpolation.  Not used with the `delaunay`_
@@ -172,3 +188,8 @@ allow_extrapolation
     is false), extrapolation is used to assign the ``HeightAboveGround``
     value.
     [Default: false]
+
+respect_ground_classification
+    If true, set HAG of ground-classified points to 0 rather than comparing
+    ``Z`` value to raster DEM
+    [Default: true]
--- a/filters/HAGFilter.cpp
+++ b/filters/HAGFilter.cpp
@@ -196,12 +196,16 @@ void HAGFilter::addArgs(ProgramArgs& args)
 {
     args.add("count", "The number of points to fetch to determine the "
         "ground point [Default: 1].", m_count, point_count_t(1));
-    args.add("raster", "GDAL-readable raster to use for DEM (uses band 1)", m_rasterName, "NONE");
+    args.add("raster", "GDAL-readable raster to use for DEM (uses band 1, "
+        "starting from 1)", m_rasterName, "");
     args.add("max_distance", "Ground points beyond this distance will not "
         "influence nearest neighbor interpolation of height above ground."
         "[Default: None]", m_maxDistance);
     args.add("allow_extrapolation", "If true and count > 1, allow "
         "extrapolation [Default: true].", m_allowExtrapolation, true);
+    args.add("respect_ground_classification", "If true, set HAG of ground"
+        "-classified points to 0 rather than comparing Z value to raster DEM",
+        m_respectGroundClassification, true);
     args.add("delaunay", "Construct local Delaunay fans and infer heights "
         "from them. [Default: false].", m_delaunay, false);
 }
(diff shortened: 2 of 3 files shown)


filters/HAGFilter.cpp
```diff
@@ -195,6 +196,7 @@ void HAGFilter::addArgs(ProgramArgs& args)
 {
     args.add("count", "The number of points to fetch to determine the "
         "ground point [Default: 1].", m_count, point_count_t(1));
+    args.add("raster", "GDAL-readable raster to use for DEM (uses band 1)", m_rasterName, "NONE");
```
abellgithub: I don't see any reason to use a sentinel value here ("NONE").  An empty string seems better.

filters/HAGFilter.cpp
```diff
@@ -229,6 +242,27 @@ void HAGFilter::filter(PointView& view)
 {
     using namespace pdal::Dimension;
 
+    // If using a raster as the DTM, check each point against it and write
```
abellgithub: This should go in a function.

This behavior is somewhat different than the other methods: using other methods we always set HAG for points classified as ground to 0.  This doesn't guarantee that.  @chambbj , do you have thoughts on this?

filters/HAGFilter.cpp
```diff
@@ -229,6 +242,27 @@ void HAGFilter::filter(PointView& view)
...
+    // the HAG dimension
+    if (m_rasterName != "NONE") {
+      static std::vector<double> data;
+      for (PointId i = 0; i < view.size(); ++i)
+      {
+        double x = view.getFieldAs<double>(Id::X, i);
+        double y = view.getFieldAs<double>(Id::Y, i);
+        double z = view.getFieldAs<double>(Id::Z, i);
+
+        if (m_raster->read(x, y, data) == gdal::GDALError::None)
```
abellgithub: I wonder what should be done in the "else" case.

filters/HAGFilter.cpp
chambbj: It is true that the original method, which relied on ground classifications, would set HAG equal to 0 for these ground classified points. For the rest, it was valid to have both positive and negative HAG values.

In this case, I suppose you could apply some logic that if a point was classified as zero and was sufficiently close to the DEM, then you could force the HAG to zero. That said, with the external DEM, if you are treating it as truth, maybe we don't even require ground classifications as a prerequisite anymore, and just let the HAG calculation carry through as written.

abellgithub: I don't disagree, but I have two concerns:

1) The behavior is inconsistent.
2) You will probably create points classified as ground that have a non-zero HeightAboveGround.

Not sure how important those concerns are.

chambbj: We can think about how to make it more consistent.

To your second point, I guess I'm suggesting that points don't have to be classified as anything first, but you are right, if they are, they could be non-zero.

I could envision a number of possibilities to address the second point. One, flip things around and use HAG directly to classify ground points (if abs(HAG) < threshold, set Classification to ground). Two, use Classification to modify HAG (if Classification = ground, set HAG to 0). Both of these are reasonable post-processing steps that we don't really have a filter to do (outside of Python). They could be baked in here, or we could wait on #2638.

abellgithub: I get a little concerned about complication that might be involved in the above suggestion.  Options that change significant behavior for a filter don't get me excited.

I'd rather just see things operate the same way they do currently -- if the point is classified as ground, set HAG to 0.  If not, use the raster value.  There is also a question of what to do if there is no raster cell where a point falls.

Classifying points based on a raster seems more like something for `filters.dem` or a new filter.

chambbj: Right, but `filters.dem` culls points. Even more confusing to stick it there. New filter, maybe.

Perhaps the simplest thing to do right now would be to add an option to force HAG to 0 if Classifications exist and it is ground - default would be ON. I do think that we should allow the filter to proceed without Classifications at all if the DEM option is exercised.

filters/HAGFilter.cpp
jtfell: Yeah good point, fixing now.

filters/HAGFilter.cpp
jtfell: I expect it should do nothing in the default case. If there is no ground defined at that point then there is no HAG, right? Maybe we can add an option to allow:

- default to a configurable value
- the Z of the point

filters/HAGFilter.cpp
jtfell: I've implemented what you suggested @chambbj - I've used `respect_ground_classification` as the option but don't know if its the best choice. Open to suggestions here.

jtfell:
Thanks for the feedback @chambbj @abellgithub - I've made the suggested changes so I think its ready for another review

filters/HAGFilter.cpp
chambbj: I think this is the right way to approach this. A more succinct option name would be good. I'm at a bit of a loss, maybe `zero_ground`?

filters/HAGFilter.cpp
```diff
@@ -224,43 +228,57 @@ void HAGFilter::ready(PointTableRef table)
...
         throwError("Option 'count' must be at least 3 when using the "
             "'delaunay' option.");
 
     const PointLayoutPtr layout(table.layout());
-    if (m_rasterName == "NONE" && !layout->hasDim(Dimension::Id::Classification))
+    if (m_rasterName == "" && !layout->hasDim(Dimension::Id::Classification))
         throwError("Missing Classification dimension in input PointView.");
 }
 
+void HAGFilter::set_HAG_from_raster(PointView& view)
```
abellgithub: We don't do underscores.

doc/stages/filters.hag.rst
```diff
@@ -172,3 +188,8 @@ allow_extrapolation
     is false), extrapolation is used to assign the ``HeightAboveGround``
     value.
     [Default: false]
+
+respect_ground_classification
```
abellgithub: This name is way too long, IMO.  Its behavior needs to be applied to all processing modes.

chambbj: Yes, but ground points are already 0 by definition in the other two modes because HAG is only calculated for non-ground points.

abellgithub: We zero HAG for ground in those modes.  If we're going to do this, I think it should be consistent.  Either you ignore ground points altogether, or you set to 0.

jtfell: I'm not sure I understand @abellgithub - By default this mode is zeroing HAG for ground nodes, the option just allows you to opt out.

Any ideas for a better/shorter name? I'm happy with `zero_ground` or maybe `ignore_classification` (but default to false and the logic is backwards)?

abellgithub:
@jtfell : After some discussion, we would like to break the modes of HAG calculation into separate filters.  Besides eliminating some confusion WRT handling of ground points, the raster mode that you've provided would support streaming mode, which the other methods don't support.

If you would like to take a crack at this and call it filters.hag_raster or filters.hag_dem (your choice), that would be great.  If you don't want to do this, I can make it happen.

jtfell:
Ok that sounds good to me. I will have a crack on Monday and update the PR, thanks for the feedback so far 👍

abellgithub:
Please create a new PR for this work.

jtfell:
Superseded by #2894

PR #2883, opened 2020-01-24
Added support for not offsetting the nominal xyz values while writing las.

floriankramer:
This pr adds an additional parameter to the las writer which disables the offset when computing the output xyz coordinates. The offset is still written into the las header.
Using this new parameter las files with high offsets can be created without having to transform all points before writing the las file. For very high offsets (such as those creating when georeferencing pointclouds) this avoid floating point precision problems (which can reduce the accuracy of point positions down to half a meter or less).

### DIFF
--- a/doc/stages/writers.las.rst
+++ b/doc/stages/writers.las.rst
@@ -166,6 +166,7 @@ scale_x, scale_y, scale_z
   dimensions to range from [0, 2147483647].  [Default: .01]
 
   Note: written value = (nominal value - offset) / scale.
+  ``transform_points_by_offset`` may change this.
 
 offset_x, offset_y, offset_z
    Offset to be subtracted from the X, Y and Z nominal values, respectively,
@@ -174,6 +175,14 @@ offset_x, offset_y, offset_z
    dimension.  [Default: 0]
 
    Note: written value = (nominal value - offset) / scale.
+   ``transform_points_by_offset`` may change this.
+
+transform_points_by_offset
+  If true the offset is only written to the header but not applied to the
+  X, Y and Z nominal values. The scaling is still applied.  [Default: false]
+
+  Note: If this is true, the written values is nominal value / scale.
+
 
 filesource_id
   The file source id number to use for this file (a value between
--- a/io/LasWriter.cpp
+++ b/io/LasWriter.cpp
@@ -148,6 +148,7 @@ void LasWriter::addArgs(ProgramArgs& args)
     args.add("offset_x", "X offset", m_offsetX);
     args.add("offset_y", "Y offset", m_offsetY);
     args.add("offset_z", "Z offset", m_offsetZ);
+    args.add("transform_points_by_offset", "Only scale the nominal values and write the offset to the header.", m_transformPointsByOffset, false);
     args.add("vlrs", "List of VLRs to set", *m_userVLRs);
 }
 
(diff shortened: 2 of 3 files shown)


abellgithub:
I don't understand this at all.

floriankramer:
We have pointclouds in various formats with their points in a local coordinate system (the origin is somewhere in the pointcloud). We also have a position of that pointcloud in the [utm](https://en.wikipedia.org/wiki/Universal_Transverse_Mercator_coordinate_system) coordinate system. We want to store these pointclouds using their utm coordinates as .las files.
With the way the las writer currently works we have to first transform the points from their local coordinate system, in which their positions are relatively small floats, to the utm coordinate system. Due to the scale of the utm system we loose a lot of precision during that transformation. When writing the pointcloud we can then specify the offset to be the utm position of the origin of the pointcloud. The las.write then undoes the transformation we applied manually and writes the offset into the las header. This results in the correct las file (for our requirements) but looses a lot of precision in the process.
To avoid the loss of precision we incur when transforming the points into the utm coordinate system we'd like to skip that transformation, and instead only write the offset into the las file's header.

abellgithub:
I still don't understand.  An example with some numbers would be very helpful.

If you're losing precision transforming to UTM, I'm not sure how you're getting it back at the time you're writing the data, regardless of the offset.  Normally, applying an offset (which you can specify) will INCREASE output precision, as LAS provides for fewer places of precision than the double-precision floating-point that PDAL uses internally.  Lopping off the high order bits means that you're going to write more of the low order bits into the output.

floriankramer:
@abellgithub We did some more experimentation. If we simply read and then write a las file with a high offset we do not lose any precision. When using the pipeline posted below we do loose precision though.
```
[
    {
        "type" : "readers.pcd",
        "filename" : "kp_ascii.pcd"
    },
    {
        "type":"filters.transformation",
        "matrix":"1 0 0 32414748 0 1 0 5316177 0 0 1 401.778 0 0 0 1"
    },
    {
        "type" : "writers.las",
        "a_srs": "EPSG:25832",
        "scale_x": "0.001",
        "scale_y": "0.001",
        "scale_z": "0.001",
        "offset_x": "32414000",
        "offset_y": "5316000",
        "offset_z": "0",
        "filename" : "kp_transform.laz"
    }
]

```
With the new parameter in the pr we can do
```
[
    {
        "type" : "readers.pcd",
        "filename" : "kp_ascii.pcd"
    },
    {
        "type":"filters.transformation",
        "matrix":"1 0 0 748 0 1 0 177 0 0 1 401.778 0 0 0 1"
    },
    {
        "type" : "writers.las",
        "a_srs": "EPSG:25832",
        "scale_x": "0.001",
        "scale_y": "0.001",
        "scale_z": "0.001",
        "offset_x": "32414000",
        "offset_y": "5316000",
        "offset_z": "0",
        "transform_points_by_offset": "true",
        "filename" : "kp_directoffset.laz"
    }
]

```
which doesn't have the precision problem. The accuracy loss we incur would indicate that the values are stored as floats at some point, but I haven't been able to find any usage of floats by looking at the code. We ran the two pipelines on the code in this pr.
These screenshots show the output of the two pipelines. In the first one you can see the precision problems created by the first pipeline. The second one shows the result of the second pipeline.
Screenshot from 2020-01-28 11-53-55
Screenshot from 2020-01-28 11-53-45

abellgithub:
Can you add the input file to the ticket?

Tones29:
Hi Andrew,
you can find the compressed input file here: https://we.tl/t-dYkHSMfBeX. The data is attached to EPSG:25832 (UTM zone 32N).
Good find, the x-value is very high. This is due to our customer's convention to put the `32` in front of the x-coordinate. So an offset of 414000 becomes 32414000.
Best, Andreas

abellgithub:
The issue you're seeing is that the PCD reader is creating the X, Y and Z dimensions as single-precision floating point.  I've opened the ticket #2892 to address this.

Tones29:
Thank you for investigating. I can confirm that the problem seems to be the input. We could not reproduce the problem with a .las file as input file.

PR #2894, opened 2020-01-29
Add filters.hag_dem

jtfell:
Created a new filter for calculating HAG values from a raster-based DEM based on feedback from  #2881

### DIFF
--- a/filters/HAGDEMFilter.cpp
+++ b/filters/HAGDEMFilter.cpp
@@ -1,5 +1,5 @@
 /******************************************************************************
-* Copyright (c) 2016, Bradley J Chambers (brad.chambers@gmail.com)
+* Copyright (c) 2020, Julian Fell (hi@jtfell.com)
 *
 * All rights reserved.
 *
--- a/filters/HAGDEMFilter.cpp
+++ b/filters/HAGDEMFilter.cpp
@@ -1,0 +1,128 @@
+/******************************************************************************
+* Copyright (c) 2016, Bradley J Chambers (brad.chambers@gmail.com)
+*
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following
+* conditions are met:
+*
+*     * Redistributions of source code must retain the above copyright
+*       notice, this list of conditions and the following disclaimer.
+*     * Redistributions in binary form must reproduce the above copyright
+*       notice, this list of conditions and the following disclaimer in
+*       the documentation and/or other materials provided
+*       with the distribution.
+*     * Neither the name of Hobu, Inc. or Flaxen Geo Consulting nor the
+*       names of its contributors may be used to endorse or promote
+*       products derived from this software without specific prior
+*       written permission.
+*
+* THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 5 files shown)


chambbj:
This looks very nice @jtfell. I personally would only suggest updating the copyright for current year and your own name/contact info as this is now unrelated to the original HAG filter in that regard.

jtfell:
Cheers @chambbj - I've updated the copyright info

filters/HAGDEMFilter.cpp
```diff
@@ -0,0 +1,130 @@
...
+}
+
+
+HAGDEMFilter::HAGDEMFilter()
+{}
+
+
+void HAGDEMFilter::addArgs(ProgramArgs& args)
+{
+    args.add("raster", "GDAL-readable raster to use for DEM (uses band 1, "
```
abellgithub: If you use `args.add(...).setPositional()`, you will this option will be required and you won't need to check that it was set in `prepared()`.

filters/HAGDEMFilter.cpp
```diff
@@ -0,0 +1,130 @@
...
+    m_raster.reset(new gdal::Raster(m_rasterName));
+    m_raster->open();
+}
+
+void HAGDEMFilter::prepared(PointTableRef table)
+{
+  if (m_rasterName == "")
+    throwError("Must specify a 'raster' DEM");
+  if (m_band <= 0)
+    throwError("'band' must be greater than 1");
```
abellgithub: greater than 0

filters/HAGDEMFilter.cpp
```diff
@@ -0,0 +1,130 @@
...
+    // If "zero_ground" option is set, all ground points get HAG of 0
+    if (m_zeroGround && point.getFieldAs<uint8_t>(Id::Classification) == ClassLabel::Ground)
+    {
+        point.setField(Id::HeightAboveGround, 0);
+    }
+    else
+    {
+        double x = point.getFieldAs<double>(Id::X);
+        double y = point.getFieldAs<double>(Id::Y);
+        double z = point.getFieldAs<double>(Id::Z);
```
abellgithub: The fetch of Z could be deferred until you know you need it.

filters/HAGDEMFilter.cpp
jtfell: I was following the convention from https://github.com/PDAL/PDAL/blob/master/filters/DEMFilter.cpp#L104 where the band counts from 1. Happy to change it, but I thought it best to be consistent.

hobu: Bands in GDAL count from one. A band of 0 is an error. The conditional is bad (in both places!) and should be changed to `m_band < 0`.

abellgithub: The text needs to say "must be greater than 0"

jtfell: I agree that the error message is wrong but it seems to me that the conditional is fine.

jtfell: I have pushed a fix for the error message (in both places) @abellgithub @hobu - Does this look right to you?

abellgithub:
Merged into https://github.com/PDAL/PDAL/tree/issue-2886
Removed static type on vector (bad in case people want to thread) and made formatting changes.

abellgithub:
@jtfell : Do you have a test for this that I can integrate?

PR #2895, opened 2020-01-29
move Python to https://github.com/PDAL/python

hobu:
The Python embedded plugins – `readers.numpy` and `filters.python` have been moved to the Python tree at https://github.com/PDAL/python and are now managed and released with that software instead of with PDAL.


hobu:
I don't understand this comment

abellgithub:
I was making note of remaining references to PDAL.

Issue #2890, opened 2020-01-28
Require MSVC 2017 


hobu:
Bump our base requirement for PDAL compilation on Windows to MSVC 2017 with -std=c++17.

PR #2896, opened 2020-01-29
require C++17 #2890

hobu:


hobu:
We have since reverted this after working through some issues and found support for C++17 to be too uneven for all of the platforms we support. We will revisit in another year.

PR #2900, opened 2020-01-30
Add reader, now, and file_size keys to pdal info output

hobu:
`reader`, `file_size`, and `now` entries have been added to `pdal info` output


hobu:
> My only comment is that having an output field called `now` seemed a little odd, 

I was trying to keep with programmer convention. The other possible names seems long-winded.

connormanning:
`analyzedAt` (or `ranAt`?) would be preferable to me, but this is fine.

PR #2901, opened 2020-01-30
update to next OSX azp image

hobu:


abellgithub:
Y?  I'm still on 13

hobu:
Apple and Azure of EoL'd 10.13, and Azure is removing the 10.13 images on 20 MAR 2020.

PR #2903, opened 2020-01-31
tiledb: add defaults for attribute filters

normanb:
This is a WIP to add compression defaults for data attributes.

### DIFF
--- a/plugins/tiledb/io/TileDBWriter.cpp
+++ b/plugins/tiledb/io/TileDBWriter.cpp
@@ -72,83 +72,33 @@ struct TileDBWriter::Args
     std::string m_compressor;
     int m_compressionLevel;
     NL::json m_filters;
+    NL::json m_defaults;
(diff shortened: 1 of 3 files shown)


normanb:
@abellgithub there are failing tests with the writer that are not being picked up by the CI. I will update this PR when the tests are working.

plugins/tiledb/io/TileDBWriter.cpp
```diff
@@ -75,6 +75,80 @@ struct TileDBWriter::Args
     bool m_append;
 };
 
+NL::json attributeDefaults =
```
hobu: It would be helpful to reflect these defaults or the behavior of the defaults into the ``writers.tiledb`` docs so that users will be able to expect how these behave. That can be done in another PR though.

normanb:
This is ready for merge. Tests all pass and defaults filters are documented.

PR #2907, opened 2020-02-03
PMF classification fix

### LINKED ISSUE #2898: PMF can output Classification of 0
I would expect that PMF (like SMRF) would classify **all points** as either 1 (unclassified) or 2 (ground), but I've seen several examples lately, where some number of points retain the classification value 0 (created, never classified). This should only be possible if some points are ignored, either through the ignore option or through return filtering.

chambbj:
- Do not touch classification of points exluded by return information

- Initialize all points passed to processGround() by setting
  classification to ClassLabel::Unclassified

- Apply clang-format

Fix #2898

### DIFF
--- a/test/unit/filters/PMFFilterTest.cpp
+++ b/test/unit/filters/PMFFilterTest.cpp
@@ -99,5 +99,6 @@ TEST(PMFFilterTEst, classOneAndTwoOnly)
         if (cl == ClassLabel::CreatedNeverClassified)
             classZero++;
     }
+    EXPECT_EQ(v->size(), 110000u);
     EXPECT_EQ(classZero, 0u);
 }
--- a/test/unit/filters/PMFFilterTest.cpp
+++ b/test/unit/filters/PMFFilterTest.cpp
@@ -1,39 +1,42 @@
 /******************************************************************************
-* Copyright (c) 2018, Bradley J Chambers (brad.chambers@gmail.com)
-*
-* All rights reserved.
-*
-* Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


abellgithub:
Shouldn't all the points in the second group be marked as Unclassified, too?  Should we just classify everything as Unclassified?

chambbj:
@abellgithub I debated this, and could still be convinced, but my rationale is this. Any points excluded from processing, either by `ignore` or return filtering, should go untouched, preserving their incoming classification tags. I'm certain that's the right thing to do for `ignore`, because those were very likely tagged as noise. For return filtering, I could probably go either way honestly.

abellgithub:
I'm OK with whatever, but I think it needs clear documentation in the code.

chambbj:
@abellgithub I added some comments and better names for the various PointViews. Also realized the same thing could've happened in SMRF, just had never been observed.

test/unit/filters/PMFFilterTest.cpp
```diff
@@ -60,3 +63,42 @@ TEST(PMFFilterTest, validReturns)
     PointTable table;
     EXPECT_NO_THROW(filter.prepare(table));
 }
+
+TEST(PMFFilterTEst, classOneAndTwoOnly)
```
abellgithub: Capitalization typo in name.

chambbj:
@abellgithub If you are talking whitespace and newlines, it's just because I ran `clang-format` which is my typical go-to. If you'd like, I could back them out. If you are talking variable renaming, I'd argue to keep those - the original names were bad IMO.

Issue #2804, opened 2019-11-11
Allow OAMS_TRADITIONAL_GIS_ORDER for GDAL 3+


hobu:
Because of the inadequacies in how SRSs can be defined with WKT1 and their interaction with GDAL/PROJ, it is sometimes necessary for users to override the axis ordering of the definition. GDAL provides 

`srs->SetAxisMappingStrategy(OAMS_TRADITIONAL_GIS_ORDER)`

to achieve this task. Background for reasons why can be found in the [GDAL Barn RFC](https://gdal.org/development/rfc/rfc73_proj6_wkt2_srsbarn.html).

PDAL needs a mechanism for the user to control the behavior of `OAMS_AUTHORITY_COMPLIANT` or `OAMS_TRADITIONAL_GIS_ORDER` when it uses GDAL 3.0+.

hobu:
I think this should be provided as part of the ``spatialreference`` option.

For example, this would use the SRS of the LAS file but switch the axis ordering:
```
{
    "type": "readers.las",
    "spatialreference": "OAMS_TRADITIONAL_GIS_ORDER"
}
```
Overriding the SRS would follow the same adhoc SRS definition syntax we're using for assigning it on geometries: 
```
{
    "type": "readers.las",
    "spatialreference": "EPSG:4326 / OAMS_TRADITIONAL_GIS_ORDER"
}
```

I think that we should NOT default to assigning OAMS_TRADITIONAL_GIS_ORDER, however. It is too easy to get the wrong behavior, and we should let GDAL/PROJ figure it out.

msmitherdc:
we will need this as not having this will negatively affect existing data. But shouldn't that perhaps be an option on the spatial reference. I don't really like the " srs / option" syntax

hobu:
> an option on the spatial reference.

`spatialrefernce` isn't compound (and neither is `bounds` or `geometry`). I like the compound syntax better myself, but that's not what PDAL does.

abellgithub:
What do you mean that it isn't compound?  It's an opaque string to us.

hobu:
I think @msmitherdc is asking for this style of definition:

```
{ 
    "spatialreference": 
    {
        "definition": "PROJ_CS ....",
        "axis_ordering": "OAMS_TRADITIONAL_GIS_ORDER"
    }
}
```

abellgithub:
I don't care, but I don't think using a GDAL environment variable name is the way to go.

Does this only matter when you're doing a transform?

[...]

PR #2916, opened 2020-02-06
Implement in_axis_ordering and out_axis_ordering for filters.reprojection

hobu:


hobu:
Implementation of #2804 for ``filters.reprojection`` at least. A full implementation of axis awareness will have to wait.

hobu:
My build script changes are now removed.

filters/ReprojectionFilter.cpp
```diff
@@ -91,7 +93,39 @@ void ReprojectionFilter::createTransform(const SpatialReference& srsSRS)
             throwError("source data has no spatial reference and "
                 "none is specified with the 'in_srs' option.");
     }
-    m_transform.reset(new SrsTransform(m_inSRS, m_outSRS));
+
+    if (m_inAxisOrdering.size() || m_outAxisOrdering.size())
```
abellgithub: This can be eliminated by using a decent default.

abellgithub: This should be done in `prepared()` so that processing can be stopped during argument parsing.  Also, it would be good to check that the vector and values are reasonable and provide some feedback.  You should only ever have vector sizes of 2 or 3 and they should always be the values 1,2 or 1,2,3 when sorted.

test/unit/SpatialReferenceTest.cpp
```diff
@@ -513,6 +513,51 @@ TEST(SpatialReferenceTest, set_srs)
...
         std::string::npos);
 }
+
 #endif // PDAL_HAVE_LIBXML2
 
+#if GDAL_VERSION_MAJOR >= 3
+// Test setting EPSG:4326 from User string
+TEST(SpatialReferenceTest, axis_ordering)
+{
+    SpatialReference wgs84("EPSG:4326");
```
abellgithub: unused

filters/ReprojectionFilter.cpp
hobu: > a decent default.

I don't know what this means. The default is managed by GDAL, and any setting of it potentially triggers something to happen internally with the OSR class. 

I'll implement this in `prepared()`.

CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ mark_as_advanced(CMAKE_VERBOSE_MAKEFILE)
 set(PDAL_API_VERSION "9")
 set(PDAL_BUILD_VERSION "10")
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 11)
```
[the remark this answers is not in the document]
hobu: It is 11 in master now. See a8f82c2cf15307b775132039b0c9af729624d188

filters/ReprojectionFilter.cpp
hobu: See latest. I added a member to SpatialReference to getAxisOrdering, which forwards the ordering from OSR to provide a default if needed, but it is not used.

abellgithub: 👍

PR #2925, opened 2020-02-11
Add some initial support for handling synthetic points

### LINKED ISSUE #2876: Improve handling of synthetic points
I've started a branch https://github.com/PDAL/PDAL/tree/synthetic-handling that aims to improve handling of synthetic points within PDAL.

The first baby step it takes is to test the parsing of the synthetic bit when reading/writing LAS. The tests are far from exhaustive, and should probably also give similar attention to keypoint and withheld bits as well.

Second, it implements a method to segment an incoming PointView into Kept/Ignored PointViews (perhaps these should instead be named Real/Synthetic for clarity and to not be too presumptuous). This segmentation is exercised in SMRFilter to optionally passthrough synthetic points without them having any bearing on the ground extraction.

The multiple SMRF options for ignoring DimRanges, returns by ReturnNumber/NumberOfReturns, and now synthetic bit makes the code a bit of a hot mess, and it still feels like a general purpose, filter-level solution would make this cleaner, especially if we want to implement similar for other filters. Perhaps priority can be given to #2638, with the additional requirement that the WHERE syntax be able to support the synthetic (and other) bits.

Thoughts?

chambbj:
* Add tests for LasReader and LasWriter

* Add method to Segmentation.cpp to separate PointViews into synthetic
  and real points

* Add arguments to SMRF to allow/ignore synthetic points during
  processing

* Add Synthetic to the list of ClassLabels (though this doesn't exactly
  hold for LAS1.4)

Close #2876


filters/SMRFilter.cpp
```diff
@@ -119,6 +120,8 @@ void SMRFilter::addArgs(ProgramArgs& args)
     args.add("ignore", "Ignore values", m_args->m_ignored);
     args.add("returns", "Include last returns?", m_args->m_returns,
              {"last", "only"});
+    args.add("synthetic", "Allow synthetic returns?", m_args->m_allowSynthetic,
+             true);
```
hobu: What about `withheld`? The user can't specify ignoring those currently given how our dimRange stuff works. Maybe put this in as a future TODO.

pdal/pdal_types.hpp
```diff
@@ -190,6 +190,7 @@ namespace ClassLabel
     const uint8_t IgnoredGround = 20;
     const uint8_t Snow = 21;
     const uint8_t TemporalExclusion = 22;
+    const uint8_t Synthetic = 32;
```
hobu: Add Withheld and Key-Point too?

chambbj: Oh, right. I'd thought about this before and then promptly forgot. I think I got disillusioned at the prospect of having to add separate arguments to the filters to support each. Perhaps we could change it from a boolean to a comma-separated string of bits to ignore, e.g., `ignore_bits="synthetic,withheld"`.

chambbj: @hobu Look at latest commit and let me know what you think.

filters/SMRFilter.cpp
rphillips91: This is something important for NZ national QC I'm trying to calculate if extreme scan angles are found in withheld flagged points vs non-withheld. If they are in non-withheld it could potentially impact our products e.g. CL2 point cloud files, DSM and DEM. Thank you again for an epic piece of open source software.

abellgithub: I'm not sure what's up with this comment. The current code supports all the classification filter bits.

rphillips91: pipelines I have tried using pdal 2.3.0: to filter out (1) unclassified points flagged as withheld;

1.
```
{
    "pipeline" : [
    {
        "type" : "readers.las",
        "filename" : "input.las"
    },
	{
		"type":"filters.range",
		"limits":"Classification![129:129]"
	},	
    {
        "type" : "writers.las",
        "a_srs" : "EPSG:2193+7839",
        "filesource_id" : "0",
        "filename" : "output.las"
    }
    ]
  }

```
2.
```
{
    "pipeline" : [
    {
        "type" : "readers.las",
        "filename" : "input.las"
    },
	{
		"type":"filters.range",
		"limits":"Classification![129:129]",
		"where":"Classification!=129"
	},	
    {
        "type" : "writers.las",
        "a_srs" : "EPSG:2193+7839",
        "filesource_id" : "0",
        "filename" : "output.las"
    }
    ]
  }
```
  
3. 
` pdal info --stats C:\Temp\scan_angle\test_file\CL2_BW24_2020_1000_4526.las --filters.range.limits=Classification![129:129] --filters.stats.dimensions=Classification --filters.stats.count=Classification`

 All my outputs have these points still intact in their files and the pdal info doesn't appear to support this filter.

abellgithub: Filtering out withheld points is not related to this PR unless you're attempting to do this when using the SMRF filter.  If you're having a general issue with withheld points, you should post to the mailing list or open another PR if appropriate.

rphillips91: Thank you for the heads up - I will post in pdal mailing list this query.

PR #2927, opened 2020-02-11
Split HAG filter into three

### LINKED ISSUE #2886: Break apart HAG filter

abellgithub:
Close #2886.


filters/HagDelaunayFilter.hpp
```diff
@@ -0,0 +1,69 @@
+/******************************************************************************
+* Copyright (c) 2016, Bradley J Chambers (brad.chambers@gmail.com)
```
chambbj: I know it's a nit, but I had very little to do with the Delaunay method (if anything at all). Maybe it would be better to include James McClain (2019) for these files.

doc/stages/filters.hag_delaunay.rst
```diff
@@ -0,0 +1,97 @@
...
+
+.. image:: ./images/autzen-height.png
+   :height: 400px
+
+::
+
+    $ pdal translate autzen.laz autzen-height-as-Z-smrf.bpf \
+        smrf hag_delanay ferry \
+        --filters.ferry.dimensions="HeightAboveGround=Z" \
+        --filters.hag_delanay.count=15
```
chambbj: A couple of instances here that should be `delaunay` and not `delanay`.

chambbj:
I believe you are already on top of it, but the existing HAG tests fail, referencing the former filter name. Related to that, are we going to keep the old `filters.hag` around for a release cycle, or just rip off the bandaid?

filters/HagDelaunayFilter.hpp
abellgithub: I'm just going to remove it.  I don't know why we do this or how any individual can really claim copyright on this.  But I'm not an IP attorney, thank goodness.

abellgithub:
@chambbj : I don't know WRT keeping old HAG.  Perhaps we should.

PR #2929, opened 2020-02-11
fix #2904 - support cropping with 3D bounds

### LINKED ISSUE #2904: filters.crop does not work with 3D bounds
``filters.crop`` currently [downcasts bounds](https://github.com/PDAL/PDAL/blob/master/filters/CropFilter.cpp#L124) to 2d for cropping. We should be able to support 3D box clipping with ``filters.crop``

NOTE: Spherical cropping, where the user inputs a point and a radius, does work in 3D.

hobu:


abellgithub:
This is a behavior change.  I'm not really a fan.  If we're going to do this I think you should have to ask for it explicitly.  This was done as is intentionally and people may well be using the current behavior.

hobu:
> This is a behavior change.

It was a behavior change to REMOVE 3D box cropping previously.

abellgithub:
If so, it's like in version 10.  Too old to be relevant.

pdal/GDALUtils.cpp
```diff
@@ -195,6 +195,18 @@ bool reprojectBounds(BOX3D& box, const std::string& srcSrs,
 }
 
 
+bool reprojectBounds(Bounds& box, const std::string& srcSrs,
```
abellgithub: I don't understand this.  You shouldn't treat Bounds as BOX3D.  It also doesn't seem to be used.

filters/CropFilter.cpp
```diff
@@ -257,6 +275,29 @@ bool CropFilter::crop(const PointRef& point, const BOX2D& box)
     return (m_args->m_cropOutside != box.contains(x, y));
 }
 
+void CropFilter::crop(const Bounds& box, PointView& input, PointView& output)
```
abellgithub: This doesn't seem to be used.

hobu: It is used https://github.com/PDAL/PDAL/blob/ed3c8e5035574cc723e88a697756f5e18757315e/filters/CropFilter.cpp#L245

abellgithub: OK.  Then I think the BOX2D version isn't used.

I think this logic would be better in the caller.  is3D is invariant.  No need to check for every point.  The compiler might optimize this, but IMO would still be cleaner to eliminate the function altogether and do:

```
if (box.is3d())
  crop(box.to3d(), *view, *outView);
else
  crop(box.to2d(), *view, *outView);
```

Or eliminate the BOX2D version.  Do what you want.

PR #2934, opened 2020-02-13
HDF Reader

rjpals:
Creates readers.hdf which allows reading of simple data types from an HDF file, given a JSON object that maps from PDAL Dimension Name to HDF Dataset Path, e.g.

```
{ "PdalDimensionName1" : "HdfDatasetPath1",  "PdalDimensionName2" : "HdfDatasetPath2" }
```


doc/stages/readers.hdf.rst
```diff
@@ -0,0 +1,96 @@
...
+
+
+`ICESAT2 Data products Documentation <https://icesat-2.gsfc.nasa.gov/science/data-products>`_
+~                                                                                              
+~             
+Options
+-------
+
+filename
+  File to read from [Required]
```
rjpals: Remove

doc/stages/readers.hdf.rst
```diff
@@ -0,0 +1,96 @@
...
+
+.. include:: reader_opts.rst
+
+map
```
rjpals: Removes dimensions

plugins/hdf/io/HdfReader.cpp
```diff
@@ -0,0 +1,175 @@
...
+    log()->get(LogLevel::Info) << m_pathDimMap << std::endl;
+
+    if(m_pathDimMap.is_null()) {
+        throw pdal_error("Required option 'map' was not set");
+    } else if(!m_pathDimMap.is_object()) {
+        throw pdal_error("Option 'map' must be a JSON object, not a " +
+            std::string(m_pathDimMap.type_name()));
+    }
+    
+    for(auto& [dimName, datasetName] : m_pathDimMap.items()) {
```
rjpals: No deconstructing

plugins/hdf/io/HdfReader.cpp
```diff
@@ -0,0 +1,175 @@
...
+    for(const auto& [key, value] : m_pathDimMap.items()) {
+        log()->get(LogLevel::Info) << "Key: " << key << ", Val: " << value <<std::endl;
+    }
+    validateMap();
+
+    // ICESat-2 datasets are EPSG:7912
+    setSpatialReference(SpatialReference("EPSG:7912"));
+}
+
+void HdfReader::validateMap()
```
rjpals: Move to anonymous namespace

rjpals: Or make a lambda

rjpals: parse json and return map<string><string>

plugins/hdf/io/HdfReader.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+void HdfReader::initialize()
+{
+    for(const auto& [key, value] : m_pathDimMap.items()) {
+        log()->get(LogLevel::Info) << "Key: " << key << ", Val: " << value <<std::endl;
+    }
+    validateMap();
+
+    // ICESat-2 datasets are EPSG:7912
+    setSpatialReference(SpatialReference("EPSG:7912"));
```
rjpals: delete this and add information to a note in the `readers.hdf` docs

rjpals: Add warning about HDF reader not setting coordinate system

plugins/hdf/io/HdfReader.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+    virtual void addDimensions(PointLayoutPtr layout);
+    virtual void prepared(PointTableRef table);
+    virtual void addArgs(ProgramArgs& args);
+    virtual void initialize();
+    virtual void ready(PointTableRef table);
+    virtual point_count_t read(PointViewPtr view, point_count_t count);
+    virtual bool processOne(PointRef& point);
+    virtual void done(PointTableRef table);
+    virtual bool eof();
```
rjpals: mark with override

plugins/hdf/io/HdfReader.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include <pdal/pdal_features.hpp>
+
+#include <pdal/Reader.hpp>
+#include <pdal/Streamable.hpp>
+#include <pdal/Options.hpp>
+#include <pdal/StageFactory.hpp>
+#include <nlohmann/json.hpp>
+
+#include "Hdf5Handler.hpp"
```
rjpals: remove

doc/stages/readers.hdf.rst
```diff
@@ -0,0 +1,101 @@
+.. _readers.hdf:
+
+readers.hdf
+===============
+
+The **HDF reader** reads from files in the HDF5 format. See the
```
hobu: The **HDF reader** reads data from files in the (link) HDF5 format. You must explicitly specify a mapping of HDF datasets to PDAL dimensions using the ``dimensions`` parameter. ALL dimensions must be scalars and be of the same length. Compound types are not supported at this time.

doc/stages/readers.hdf.rst
```diff
@@ -0,0 +1,101 @@
...
+            "map": 
+            { 
+                "X" : "gt1l/heights/lon_ph", 
+                "Y" : "gt1l/heights/lat_ph", 
+                "Z" : "gt1l/heights/h_ph", 
+            } 
+        }, 
+        { 
+            "type" : "writers.las", 
+            "filename": "output.las"
```
hobu: You should set ``scale_x``, ``scale_y``, and ``scale_z`` here along with auto on the offsets. A user following this example is going to get trimmed data due to the XYs being Lat/lon.

plugins/CMakeLists.txt
```diff
@@ -23,6 +23,13 @@ if(BUILD_PLUGIN_ICEBRIDGE)
     add_subdirectory(icebridge)
 endif()
 
+if(BUILD_PLUGIN_HDF)
+    if (NOT PDAL_HAVE_LIBXML2)
+        message(FATAL_ERROR "Can't build hdf plugin without libxml2")
```
hobu: HDF driver should not require libxml2. That was true of the Icebridge driver for its metadata thing. I think this can be removed.

plugins/hdf/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+#
+# HDF plugin CMake configuration
+#
+
+include (${PDAL_CMAKE_DIR}/hdf5.cmake)
+
+
+if (NOT PDAL_HAVE_HDF5)
+    message(FATAL "HDF5 not found but is required for HDF.")
```
hobu: HDF5 library is required to build HDF support

plugins/hdf/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+else()
```
hobu: you can remove this else() conditional entirely. You've already kicked them out above this.

plugins/hdf/io/HdfReader.hpp
```diff
@@ -0,0 +1,91 @@
...
+    NL::json m_pathDimJson;
+    std::map<std::string,std::string> m_pathDimMap;
+    Dimension::IdList m_idlist;
+    struct BufferInfo {
+        BufferInfo(const hdf5::DimInfo& info);
+        std::unique_ptr<hdf5::DimInfo> info;
+        uint8_t * buffer = nullptr;
+    };
+    std::vector<BufferInfo> m_info;
+    // std::vector<std::unique_ptr<hdf5::DimInfo>> m_infos;
```
hobu: lint

plugins/hdf/test/HdfReadertest.cpp
```diff
@@ -0,0 +1,119 @@
...
+    EXPECT_EQ(view->size(), 1065u);
+    //point equality
+    Support::check_p0_p1_p2(*view);    
+    PointViewPtr view2 = view->makeNew();
+    view2->appendPoint(*view, 100);
+    view2->appendPoint(*view, 101);
+    view2->appendPoint(*view, 102);
+    Support::check_p100_p101_p102(*view2);
+    //color equality
+    //should work, but the GPS time is 0.00004 units off
```
hobu: hmm, I wonder if there's some truncation happening in the python script I used to convert the Autzen file to HDF.

hobu:
OSX compiler warnings:
```
[69/138] Building CXX object plugins/hdf/CMakeFiles/pdal_plugin_reader_hdf.dir/io/Hdf5Handler.cpp.o
../plugins/hdf/io/Hdf5Handler.cpp:139:46: warning: adding 'H5T_class_t' to a string does not append to the string [-Wstring-plus-int]
            throw pdal_error("Unkown type: " + vague_type);
                             ~~~~~~~~~~~~~~~~^~~~~~~~~~~~
../plugins/hdf/io/Hdf5Handler.cpp:139:46: note: use array indexing to silence this warning
            throw pdal_error("Unkown type: " + vague_type);
                                             ^
                             &               [           ]
../plugins/hdf/io/Hdf5Handler.cpp:108:28: warning: comparison of integers of different signs: 'uint64_t' (aka 'unsigned long long') and 'hssize_t' (aka 'long long') [-Wsign-compare]
            if(m_numPoints != dspace.getSelectNpoints()) {
               ~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~
```

```
[135/138] Building CXX object plugins/hdf/CMakeFiles/pdal_plugin_reader_hdf.dir/io/HdfReader.cpp.o
In file included from ../plugins/hdf/io/HdfReader.cpp:35:
../plugins/hdf/io/HdfReader.hpp:60:17: warning: 'getName' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    std::string getName() const;
                ^
../pdal/Stage.hpp:261:25: note: overridden virtual function is here
    virtual std::string getName() const = 0;
                        ^
1 warning generated.
```

hobu:
I added a random-chunk version of the autzen test file and I *think* I fixed the GpsTime fp comparison. Grab it from https://github.com/PDAL/PDAL/tree/howard-hdf

plugins/hdf/io/HdfReader.cpp
```diff
@@ -124,17 +128,36 @@ void HdfReader::addArgs(ProgramArgs& args)
...
+void HdfReader::validateMap()
+{
+    log()->get(LogLevel::Info) << "**JSON map**" << std::endl;
+    log()->get(LogLevel::Info) << m_pathDimMap << std::endl;
+
+    if(!m_pathDimMap.is_object()) {
+        throw pdal_error("Option 'map' must be a JSON object, not a " +
+            std::string(m_pathDimMap.type_name()));
+    }
+    for(auto& [dimName, datasetName] : m_pathDimMap.items()) {
```
abellgithub: This syntax isn't valid before C++17.  You'll have to get back the actual object and split it apart later.  If it builds on your system, you're getting lucky.

plugins/hdf/io/HdfReader.cpp
```diff
@@ -102,10 +105,11 @@ point_count_t HdfReader::read(PointViewPtr view, point_count_t count)
 }
 
 
-bool HdfReader::processOne(PointRef& point) {
-    for(int index = 0; index < m_infos.size(); ++index) {
+bool HdfReader::processOne(PointRef& point)
```
abellgithub: There are a few things here that can be improved, but I think the big thing to me is that it brings all the HDF logic into this function instead of the handler.  That means that you have to store a vector<void *> in this class and know how things are laid out in the HDF5Handler, which seems unfortunate to me.  I think I'd just want to fetch a pointer to the data for each field in the point from the handler.  Something like:
```
for (DimInfo& di : dimensions)
{
  void *fieldData = m_hdf5Handler.value(di, m_index);
  point.setField(di.id, di.pdal_type, fieldData);
}
```
All the rest can be done in the handler.  That means that this code has to know nothing about chunks or store information about the handler's data.

The other thing I'd be careful of is naming.  Calling lots of thing "index" is confusing, especially when they're different.  `m_index` is a point ID.  `index` is a loop variable.  You can eliminate the loop variable by using what's called a "range-based for loop."  It's a terrible name, but it just lets you get back the actual object in the collection you're looping.  In your case:

```
for (auto& dimInfo : m_infos)
...
```
I don't love `auto` if the type is simple.  If you do:

```
for (DimInfo& dimInfo : m_infos)
  ...
```

I think it's more clear for the reader - you have to do less searching if you're not using an IDE.  And again, naming matters.  Calling something `info` really doesn't convey much information.  That said, sometimes coming up with a good name is hard.  Also, if you explicitly use the type when defining the name and the scope of the name is small, then I don't much mind index-type/shortcut names.  But if you use something like `auto` and an ambiguous name like `info`, then I'm left searching around...  I have to go figure out the type of `m_infos` to know what's going on.

plugins/hdf/io/Hdf5Handler.cpp
```diff
@@ -0,0 +1,199 @@
...
+        std::string const& dimName = entry.first;
+        std::string const& datasetName = entry.second;
+        m_dimInfos.emplace_back(DimInfo(dimName, datasetName, m_h5File.get()));
+    }
+
+    // Check that all dimensions have equal lengths
+    m_numPoints = m_dimInfos.at(0).getNumPoints();
+    for( DimInfo& info : m_dimInfos) {
+        if(m_numPoints != info.getNumPoints()) {
+            throw pdal_error("All given datasets must have the same length");
```
hobu: It would be helpful to tell the user which dimension they gave that doesn't have the same count as the others.

rjpals:
Something I just realized: If you simply do something like `pdal info file.h5`, PDAL will default to readers.icebridge. Should that be changed to readers.hdf?

abellgithub:
I would think "no" unless you've changed things so that the new HDF reader can work without the input map.

hobu:
Ready to merge

PR #2938, opened 2020-02-20
Fix #2937 to support writers.las assignment of VLRs

### LINKED ISSUE #2937: writers.las can't accept a list of VLRs from a pipeline
Because m_userVLRs is a JSON blob, the pipeline reader can't set a list of VLRs in the pipeline.

hobu:


abellgithub:
Closing as discussed.

PR #2941, opened 2020-02-21
Fix tests for voxel filter.

### LINKED ISSUE #2940: filters.volxeldownsize tests don't work
Some tests don't really check anything and pass by not processing any data.

abellgithub:
Close #2940

### DIFF
--- a/doc/stages/filters.voxeldownsize.rst
+++ b/doc/stages/filters.voxeldownsize.rst
@@ -41,7 +41,7 @@ cell
 mode
   Mode for voxel based filtering. [Default: voxelcenter]
   **voxelcenter**: Coordinates of the first point found in each voxel will
-  be modified to those at the center of the voxel.
+  be modified to be the center of the voxel.
   **firstinvoxel**: Only the first point found in each voxel is retained.
 
 .. warning::
--- a/doc/stages/filters.voxeldownsize.rst
+++ b/doc/stages/filters.voxeldownsize.rst
@@ -20,7 +20,7 @@ Example
       {
           "type":"filters.voxeldownsize",
           "cell":1.0,
-          "mode":"voxelcenter"
+          "mode":"center"
       },
       "output.las"
   ]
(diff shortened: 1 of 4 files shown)


doc/stages/filters.voxeldownsize.rst
```diff
@@ -39,9 +40,11 @@ cell
...
+  **voxelcenter**: Coordinates of the first point found in each voxel will
+  be modified to be the center of the voxel.
+  **firstinvoxel**: Only the first point found in each voxel is retained.
 
 .. warning::
-    If you choose **voxelcenter** mode, you are loosing your dimensional data/point coordinates.
-  
\ No newline at end of file
+    If you choose **voxelcenter** mode, you are overwriting the X, Y and Z
+    values of retained points.
```
chambbj: Originally he says "losing your dimensional data", which I suppose isn't true, the dimension still exists, but it is invalidated in the sense that it is simply inheriting the dimension data of the first point, which may or may not be a good thing. Maybe we should still try to capture this fact in some manner.

abellgithub: Updated.  Let me know if you have better wording.

doc/stages/filters.voxeldownsize.rst
```diff
@@ -39,9 +40,11 @@ cell
 
 mode
   Mode for voxel based filtering. [Default: voxelcenter]
-  **voxelcenter**: Retain center of a populated voxel. Coordinates of a first point detected will be modified and set to the center of a populated voxel.
-  **firstinvoxel**: Retain first point detected in each voxel.
+  **voxelcenter**: Coordinates of the first point found in each voxel will
+  be modified to be the center of the voxel.
+  **firstinvoxel**: Only the first point found in each voxel is retained.
```
chambbj: Docs show these as `voxelcenter` or `firstinvoxel` with the former being the default. The updated code indicates the choices would be `center` or `first` with `center` still being the default. A call to PDAL options however displays the following message, which is misleading.

```
$ pdal --options=filters.voxeldownsize
filters.voxeldownsize -- http://pdal.io/stages/filters.voxeldownsize.html
-----------------------------------------------------------------------------------------------------
  user_data
      User JSON

  log
      Debug output filename

  option_file
      File from which to read additional options

  cell [0.001]
      Cell size

  mode [centerfirst]
      Method for downsizing : center / first
```

abellgithub: Oops

PR #2942, opened 2020-02-21
Handle VLRs as pipeline arguments in writers.las

### LINKED ISSUE #2937: writers.las can't accept a list of VLRs from a pipeline

abellgithub:
Close #2937

### DIFF
--- a/io/LasVLR.cpp
+++ b/io/LasVLR.cpp
@@ -67,7 +67,16 @@ std::istream& operator>>(std::istream& in, LasVLR& v)
     NL::json j;
     in >> j;
 
-    std::cout << j << "!\n";
+    // Make sure there isn't stuff in the input stream after the JSON
+    // object.
+    char c;
+    do
+    {
+        in >> c;
+    } while (std::isspace(c));
+    if (!in.eof())
+        throw pdal_error("Invalid characters following LAS VLR JSON object.");
+    in.clear();
 
     if (!j.is_object())
         throw pdal_error("LAS VLR must be specified as a JSON object.");
--- a/io/LasVLR.cpp
+++ b/io/LasVLR.cpp
@@ -34,6 +34,11 @@
 
 #include "LasVLR.hpp"
 
+#include <limits>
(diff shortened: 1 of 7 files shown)


io/LasVLR.cpp
```diff
@@ -57,6 +62,115 @@ ILeStream& operator>>(ILeStream& in, LasVLR& v)
...
+        throw pdal_error("LAS VLR must contain 'data' member.");
+    if (userId.empty())
+        throw pdal_error("LAS VLR must contain 'user_id' member.");
+    if (std::isnan(recordId))
+        recordId = 1;
+
+    v.m_userId = userId;
+    v.m_recordId = recordId;
+    v.m_description = description;
+    v.m_data = Utils::base64_decode(b64data);
```
hobu: Do we need to protect against too large of a size?

abellgithub: The data?

hobu: Sorry, yeah. The data's size needs to be under a VLR.

abellgithub: It can be as large as any 64 bit integer value, right?  So, no, if that's what you mean.

hobu: EVLRs can be as large as any 64 bit integer value. VLRs can only be as large as a 16 bit value.

abellgithub: True.

We treat everything as an EVLR if it's over the VLR limit.  If we can't do that (pre 1.4), then we throw an error at the time we try to add the VLR to the list of VLRs.  But I think things are fine here.

hobu: Ok. Anyone pushing 4gb of base64 through pipeline JSON deserves their fate anyway.

abellgithub: https://github.com/PDAL/PDAL/blob/227855a35532b7e978245a6e1a1072aa00bfd36a/io/LasWriter.cpp#L573-L593

PR #2944, opened 2020-02-21
Support range-based for loops on PointView with PointRef as reference type

### LINKED ISSUE #2560: Add begin/end for point view
Would be nice if PointView had begin/end to support range-based loops.

abellgithub:
Close #2560


test/unit/filters/SortFilterTest.cpp
```diff
@@ -48,11 +48,12 @@ using namespace pdal;
 namespace
 {
 
-void doSort(point_count_t count, const std::string & dim="X", const std::string & order="")
+void doSort(point_count_t count, Dimension::Id dim,
```
hobu: I think there was a reason to start sorting with the X dimension. This change seems extraneous to the PointRef iteration. Should these changes be moved to something else or are they part of something else?

abellgithub: It wouldn't work as it was written with any dimension other than X.  I fixed that.  If you want those changes in another PR, that's fine and seems reasonable.

hobu: No need. Merging.

Issue #2943, opened 2020-02-21
LasWriter needs to write WKT1_GDAL when PROJ 6 or 7 is used


hobu:
The LAS 1.4 specification states that GDAL (or OGC) style WKT1 flavor will be written into the CRS VLRs. If the user is using PROJ 7, they will get WKTv2 by default in some transformation situations. We should prevent writing this into LAS files, but it is perfectly fine to use it as the WKT moving through the PDAL system.

hobu:
#2945

PR #2945, opened 2020-02-21
ensure WKTv1 writing for LAS 1.4 #2943

hobu:


pdal/SpatialReference.cpp
```diff
@@ -438,6 +438,28 @@ std::string SpatialReference::prettyWkt(const std::string& wkt)
...
+    OGRScopedSpatialReference srs = ogrCreateSrs(wkt);
+    if (!srs)
+        return outWkt;
+
+    char *buf = nullptr;
+
+    char **papszOptions = nullptr;
+    papszOptions = CSLAddNameValue(papszOptions, "FORMAT", "WKT1_GDAL");
+    srs->exportToWkt(&buf, papszOptions);
+
```
rouault: missing CSLDestroy(papszOptions).

Alternatively you could use:
```
const char* apszOptions[] = { "FORMAT=WKT1_GDAL", nullptr };
srs->exportToWkt(&buf, apszOptions);
```

pdal/SpatialReference.cpp
```diff
@@ -438,6 +438,28 @@ std::string SpatialReference::prettyWkt(const std::string& wkt)
...
+    outWkt = buf;
```
rouault: You should likely test buf for nullptr before assigning it to outWkt (in case the CRS wouldn't be compatilble of WKT1, like a Geographic 3D CRS)

PR #2953, opened 2020-03-01
Howard windows updates

hobu:
Catching up some windows build stuff to silence a few warnings and get it working on my 2019 build

### DIFF
--- a/filters/StatsFilter.hpp
+++ b/filters/StatsFilter.hpp
@@ -89,7 +89,7 @@ public:
     {
         if (M2 == 0 || m_cnt <= 2 || !m_advanced)
             return 0.0;
-        double c(m_cnt);
+        double c((double)m_cnt);
         return populationSkewness() * std::sqrt(c) * std::sqrt(c - 1) / (c - 2);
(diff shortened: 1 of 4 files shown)


filters/StatsFilter.hpp
```diff
@@ -89,7 +89,7 @@ typedef std::vector<double> DataVector;
     {
         if (M2 == 0 || m_cnt <= 2 || !m_advanced)
             return 0.0;
-        double c(m_cnt);
```
abellgithub: We shouldn't have to do this.  What terrible warning is MS giving?

hobu: an implicit conversion warning.

abellgithub: Wonder if we should suppress.  That seems anal.

abellgithub:
This is all fine, but I really hate such casts.  They confuse me -- why would I be doing this?  I'd rather suppress the compiler warning, but it's up to you.  Merge as is or just suppress the warning.  Up to you.

Issue #2939, opened 2020-02-21
Inconsistent transformation results when applying matrix from `filters.icp`


jomey:
I am getting different results, when applying the matrix reported back by the `filters.icp` and using the transform with `filters.transformation` for the same point could.

Steps I used:
1. Run below pipeline with the `--metadata` options as described in the docs.
```JSON
[
   "fixed.laz",
   "moving.laz",
    {
        "type": "filters.icp"
    },
    "aligned_icp.laz"
]
```
2. Take the `transform` information from the metadata output and put it into a new pipeline to apply the same transformation to the initial moving cloud.
```JSON
[
    "moving.laz",
    {
        "type": "filters.transformation",
        "matrix": "           1 -7.96375e-06 -2.36573e-06  -0.00623153\n  7.9637e-06            1 -2.50978e-05     0.048159\n 2.36593e-06  2.50977e-05            1    -0.418688\n           0            0            0            1"
    },
    {
        "type": "writers.las",
        "compression": "laszip",
        "forward": "all",
        "filename": "aligned_transformation.laz"
    }
]
```
3. I checked the `aligned_icp.laz` summary versus the `alinged_transformation.laz` summary and they are not the same.

* Summary: `moving.laz`
```JSON
...
  "summary":
  {
    "bounds":
    {
      "maxx": 336552,
      "maxy": 4322445,
      "maxz": 4107.3,
      "minx": 322518,
      "miny": 4305588,
      "minz": 2657.95
    },
...
```

* Summary: `alingned_icp.laz`
```JSON
...
  "summary":
  {
    "bounds":
    {
      "maxx": 336552.064,
      "maxy": 4322445.055,
      "maxz": 4106.927776,
      "minx": 322517.9357,
      "miny": 4305587.991,
      "minz": 2657.314217
    },
...
```

* Summary: `aligned_transformation.laz`
```JSON
...
  "summary":
  {
    "bounds":
    {
      "maxx": 336517.6937,
      "maxy": 4322447.596,
      "maxz": 4216.000738,
      "minx": 322483.5654,
      "miny": 4305590.533,
      "minz": 2766.387178
    },
...
```
**Notes**
When I only apply the translation and not the rotation, the results are closer. Applying both, translation and rotation, produces the results that are a lot different.

**Expected behavior**
Identical transformation for the moving cloud using the output from `filters.icp` and applying it with `filters.transformation`

**System/installation information:**

```
$ uname -a
[...]

PR #2962, opened 2020-03-06
filters.icp: output centroid to metadata

klassenjs:
This is needed to apply the same transformation to a different set
of points using filters.transform. (#2939)

### DIFF
--- a/doc/stages/filters.icp.rst
+++ b/doc/stages/filters.icp.rst
@@ -62,10 +62,43 @@ The metadata output might start something like:
         {
             "filters.icp":
             {
+                "centroid": "    583394  5.2831e+06   498.152",
                 "converged": true,
                 "fitness": 0.01953125097,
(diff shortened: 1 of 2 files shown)


jomey:
I could live with this as a documented first pass solution. As discussed in the mailing list, this is not the most user friendly version and likewise I see this as the nature of applying an ICP transformation manually and outside of the `icp.filter`.

chambbj:
I honestly think that we can address this by removing the centering step altogether. All of our operations use double precision, so I'm not terribly concerned with loss of precision. I can push up a branch for you both to test if you like.

jomey:
> I honestly think that we can address this by removing the centering step altogether. All of our operations use double precision, so I'm not terribly concerned with loss of precision. I can push up a branch for you both to test if you like.

Happy to test your solution

chambbj:
Give this a shot. https://github.com/PDAL/PDAL/tree/issue-2939/icp-without-centroid

abellgithub:
@chambbj : Are you OK with this as submitted?

chambbj:
This works for this release @abellgithub.

PR #2963, opened 2020-03-06
Return correct ICP transformation

### LINKED ISSUE #2939: Inconsistent transformation results when applying matrix from `filters.icp`

chambbj:
* Remove the centering step, which is unnecessary and can cause problems when applying the same transformation after the fact
* Add test to recover rotation (5 degrees about Y axis)

Supersedes #2962 

Fixes #2939

### DIFF
--- a/filters/IterativeClosestPoint.cpp
+++ b/filters/IterativeClosestPoint.cpp
@@ -109,8 +109,8 @@ PointViewPtr IterativeClosestPoint::icp(PointViewPtr fixed,
     // be reasonable to alternately accept an initial guess.
     Eigen::Matrix4d final_transformation = Eigen::Matrix4d::Identity();
 
-    // Construct 3D KD-tree of the centered, fixed PointView to facilitate
-    // nearest neighbor searches in each iteration.
+    // Construct 3D KD-tree of the fixed PointView to facilitate nearest
+    // neighbor searches in each iteration.
     KD3Index& kd_fixed = fixed->build3dIndex();
 
     // Iterate to the max number of iterations or until converged.
@@ -119,8 +119,8 @@ PointViewPtr IterativeClosestPoint::icp(PointViewPtr fixed,
     int num_similar(0);
     for (int iter = 0; iter < m_max_iters; ++iter)
     {
-        // At the beginning of each iteration, transform our centered, moving
-        // PointView by the current final_transformation.
+	// At the beginning of each iteration, transform our moving PointView
+	// by the current final_transformation.
         PointViewPtr tempMovingTransformed =
             transform(*moving, final_transformation.data());
 
@@ -131,9 +131,9 @@ PointViewPtr IterativeClosestPoint::icp(PointViewPtr fixed,
         moving_idx.reserve(tempMovingTransformed->size());
         double mse(0.0);
 
-        // For every point in the centered, moving PointView, find the nearest
-        // neighbor in the centered fixed PointView. Record the indices of each
-        // and update the MSE.
+	// For every point in the moving PointView, find the nearest neighbor
+	// in the fixed PointView. Record the indices of each and update the
+	// MSE.
         for (PointRef&& point : *tempMovingTransformed)
         {
             // Find the index of the nearest neighbor, and the square distance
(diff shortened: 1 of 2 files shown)


klassenjs:
FYI: On one of my datasets with points centered around 581500, 5284200, with this PR, I am seeing about a 2.6m difference between the points output from filters.icp and the points output from filters.transformation using the matrix from filters.icp.  Using #2962 which does the centering the difference between the output of filters.icp and the center/transform/uncenter is about 0.005m.  I presume the difference is due to serializing the 4x4 matrix to text and then converting back to a double.

The difference between point clouds output produced directly by ICP before and after this PR differ by 30 to 60 cm (changes as you move across the cloud).  I don't know if before or after this PR is a better overall result.

So, for best results when applying the ICP matrix to another transform, the user should consider manually centering their point cloud as was mentioned on the mailing list.  I believe this PR and having the user explicitly recenter their cloud is better than #2962 from the standpoint of reproducibility and from principle of least surprise.  This is probably worth a note in the docs and maybe a runtime warning if the points are grouped too far from (0,0).

jomey:
I gave this branch a test with my use case and the performance is much worse not using a centroid. See the attached logs for more details on the iterations. Having a centroid runs for my two point clouds (~32 Mio points each) in around 45 mins. Using this branch without the centroid did not finish within 2 hours, but got close to getting to the threshold.

I think having the option to give an initial guess for the difference between the two could help reduce the iterations and also boost performance. Not sure if this should be part of this PR though.

[no_centroid.log.zip](https://github.com/PDAL/PDAL/files/4303436/no_centroid.log.zip)
[with_centroid.log.zip](https://github.com/PDAL/PDAL/files/4303438/with_centroid.log.zip)

marcorios40:
no:assignee

chambbj:
We can close this. I do have another branch that I'll push to output the full concatenated transform. The rotation test is also probably worth keeping.

PR #2964, opened 2020-03-09
Update docker/CI to support tiledb and use conda base

abellgithub:


hobu:
Grab master again and the CI should succeed

PR #2967, opened 2020-03-09
Allow zero results from TileDB query

normanb:
If making a bounding box query to TileDB within a pipeline it is possible to get an result of zero length if the bounding box specified doesn't contain any points. This PR adds a test for this and initializes the point table result size to zero.


normanb:
This is a bug fix that would be good to get in 2.1 if possible.

Issue #2262, opened 2018-10-31
Installed targets lack an include interface for pdal headers


mathstuf:
The installed targets from the CMake build lack `INTERFACE_INCLUDE_DIRECTORIES` which means that `target_include_directories` is required in addition to `target_link_libraries`. Something like [this](https://github.com/libLAS/libLAS/pull/153/files#diff-95e351a3805a1dafa85bf20b81d086e6R236) should be done. I'd do it, but PDAL has a much more complex CMake setup that I'm not sure exactly where best to put it.

abellgithub:
It would be best if you could report a specific problem that you're having so that we can fix it.  Knowing how your suggested change might impact others is difficult to know with what you've submitted.

mathstuf:
PDAL offers imported targets when `find_package(PDAL)` is done (via `install(EXPORT)`). These targets should be usable as-is without having to also use `PDAL_INCLUDE_DIRS` by doing, somewhere:

```cmake
target_include_directories(pdal_target
    INTERFACE
        $<INSTALL_INTERFACE:include>)
```

abellgithub:
Fixed with #2981

PR #2981, opened 2020-03-13
Add interface includes for PDAL library target.

abellgithub:


abellgithub:
Close #2262

CMakeLists.txt
```diff
@@ -378,6 +378,9 @@ install(
     DESTINATION
         "${PDAL_LIB_INSTALL_DIR}/cmake/PDAL")
 include(${PDAL_CMAKE_DIR}/config.cmake)
+target_include_directories(${PDAL_BASE_LIB_NAME}
+    INTERFACE
+        $<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/include>)
```
hobu: Do we need the same for pdal_util as well?

abellgithub: Not need.  The only reason is if you're using the utilities without the main PDAL lib.  But I can add it.

hobu: we should do so in case someone follows a pattern from a PDAL filter/reader in their own plugin or something.

PR #2999, opened 2020-03-23
set INSTALL_INTERFACE to a relative ./include path rather than a hard path of CMAKE_INSTALL_PREFIX

hobu:


CMakeLists.txt
```diff
@@ -257,8 +258,8 @@ target_include_directories(${PDAL_BASE_LIB_NAME}
         ${ZSTD_INCLUDE_DIRS}
         ${NLOHMANN_INCLUDE_DIR}
     PUBLIC
-        ${GDAL_INCLUDE_DIR}
-        ${LASZIP_INCLUDE_DIR}
+        $<BUILD_INTERFACE:${GDAL_INCLUDE_DIR}>
```
abellgithub: Why?

hobu: https://stackoverflow.com/a/25681179/498396

abellgithub: But then it seems pointless without an INSTALL_INTERFACE entry.  Also, I don't think these should be public anymore, but I didn't want to change it right before release.

hobu: > But then it seems pointless without an INSTALL_INTERFACE entry.

Those libraries don't expose _INTERFACE at all, so we would have to create those for laszip and GDAL too.

Issue #3006, opened 2020-03-30
Declaration hides previous local declaration - warning


cgodkin:
**Describe the bug**

Reusing the same variable name on an inner scope causes a compiler warning from (at least) VS2017 with warnings turned on.  Code would do the right thing but the warning is annoying.

```
D:\pdal-2.1.0\include\pdal/util/ProgramArgs.hpp(1264): warning C4456: declaration of 'i' hides previous local declaration
D:\pdal-2.1.0\include\pdal/util/ProgramArgs.hpp(1249): note: see declaration of 'i'
```

**Expected behavior**

Use a different variable name in the inner scope.

**System/installation information:**

Git master (noticed with PDAL 2.1.0).

PR #3007, opened 2020-03-30
Fixed hiding of local variable (principally the warning)

cgodkin:
Addresses issue #3006.


abellgithub:
What compiler/version emits an error on this?

cgodkin:
Visual Studio 2017 with -W4 (which we always use).  I can #pragma disable C4250 this if you prefer to overload "i" in this manner for some reason I can't see.

PR #3015, opened 2020-04-07
Add seed support to faux reader

abellgithub:
Also remove broken normal/uniform number generation.

### DIFF
--- a/pdal/util/Utils.cpp
+++ b/pdal/util/Utils.cpp
@@ -81,25 +81,6 @@ double Utils::random(double minimum, double maximum)
 }
 
 
-double Utils::uniform(const double& minimum, const double& maximum,
-    uint32_t seed)
-{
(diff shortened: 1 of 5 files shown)


pdal/util/Utils.cpp
```diff
@@ -81,25 +81,6 @@ double Utils::random(double minimum, double maximum)
 }
 
 
-double Utils::uniform(const double& minimum, const double& maximum,
```
hobu, line 84: Do any filters still use this? I suppose this is pre-c++11 stuff.

 It seems out of scope to remove these capabilities in this PR, or the description should be updated to reflect we're removing them.

abellgithub: They were only used by filters.faux and they were broken, because you had to set a seed for each call, which was dumb.  I probably wrote them.

PR #3017, opened 2020-04-08
Return proper position for seek on stream.

abellgithub:
Properly position stream when opened.


hobu:
Please cherry-pick this to `2.1-maintenance` too.

PR #3022, opened 2020-04-10
Create UTM SRS from zone number in one place

### LINKED ISSUE #3019: Centralize handling of UTM SRS creation

abellgithub:
Close #3019


pdal/SpatialReference.cpp
```diff
@@ -398,6 +398,31 @@ int SpatialReference::calculateZone(double lon, double lat)
 }
 
 
+/**
```
connormanning, line 401: It might cause confusion that this will always return the WGS84 code and not NAD83 which is more common for users in the US (approx. zones 10-20).  E.g. if I passed in `13` from Colorado I might expect `EPSG:26913` instead of `EPSG:32713`.  Alternatively there could be an optional second parameter `bool nad83 = false` that returned codes in the NAD83 range, and make it clear from the function prototype that it's WGS84 by default.

PR #3023, opened 2020-04-12
Modified not to assume std namespace.

cgodkin:
We applied our proposed fix to https://github.com/LASzip/LASzip/issues/69 locally.  

Concerned about downstream implications of removing "using namespace std;" from a header, I rebuilt PDAL to see if it would have any problem and found that PDAL-2.1.0-src/tools/lasdump/Dumper.cpp (probably inadvertently) depends on this std namespace injection.

This fix simply adds the std namespace in three places.

An alternative would be to add "using namespace std;" at the top of Dumper.cpp of course.

Thanks for your consideration.

### DIFF
--- a/doc/python.rst
+++ b/doc/python.rst
@@ -27,9 +27,8 @@ their own applications.
 Versions
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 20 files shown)


cgodkin:
Thanks!

PR #3030, opened 2020-04-16
Changes to support TileDB 2.0 (in `dev`) in PDAL 2.1

normanb:
This is a minor refactor that is backwards compatible with the PDAL driver and TileDB 1.4+. In TileDB 2.0 it is recommended to set the coordinate buffers directly. We can also no longer reuse the query object in the PDAL TileDB driver so this is now created and used in `FlushCache`.

I changed the `utm16` reader test to be more tolerant of different proj versions.

### DIFF
--- a/plugins/tiledb/io/TileDBReader.cpp
+++ b/plugins/tiledb/io/TileDBReader.cpp
@@ -295,7 +295,7 @@ void TileDBReader::localReady()
     // read spatial reference
     NL::json meta = nullptr;
 
-#if TILEDB_VERSION_MAJOR >= 1 && TILEDB_VERSION_MINOR >= 7
+#if TILEDB_VERSION_MAJOR > 1 || TILEDB_VERSION_MINOR >= 7
     tiledb_datatype_t v_type = TILEDB_UINT8;
     const void* v_r;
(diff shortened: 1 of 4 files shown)


normanb:
I will rerun the tests once TileDB 2.0 is out.

plugins/tiledb/io/TileDBReader.cpp
```diff
@@ -295,7 +295,7 @@ void TileDBReader::localReady()
     // read spatial reference
     NL::json meta = nullptr;
 
-#if TILEDB_VERSION_MAJOR >= 1 && TILEDB_VERSION_MINOR >= 7
+#if TILEDB_VERSION_MAJOR > 1 || TILEDB_VERSION_MINOR >= 7
```
abellgithub, line 298: This doesn't seem robust going forward.

normanb: In what way? If the version of TileDB is greater than 1.7 then use internal TIleDB metadata.

abellgithub: It's fine.  Sorry.

PR #3032, opened 2020-04-17
Improve hole removal to avoid confusion on referenced polygons

abellgithub:


abellgithub:
@rouault : I think things were OK before given how we only had one layer of polys in a multi-poly, but I think this is more clear.  If you want to take a look, that would be great.

PR #3038, opened 2020-04-22
Update readers.ept.rst (headers -> header)

chelm:
I noticed that the code for this (which is a very very helpful feature) has `header` and not `headers`. Tripped me up today so i thought that maybe the docs could be updated.


abellgithub:
Thanks.

PR #3043, opened 2020-04-23
Write more glTF attributes, including color and normals

KeyboardSounds:
**Summary:**
- Adds a `color_vertices` flag, which defaults to false, which tells the writer to include color information for each vertex in the output glTF. 
- Writes out normals for each vertex if they are present.
- New unit test file for GltfWriter, with some very basic tests.

**Background:**
I was using `filters.poisson` on a point cloud to generate a glTF mesh, and noticed that the normals were missing in the resulting .glb file. @nf-s asked me to add the color attribute while I was at it, since he was doing something similar for a colored point cloud.

If anyone is ever attempting something similar, you have to run `filters.normal` on the output of `filters.poisson` for the resulting mesh to have normal information.

### DIFF
--- a/io/GltfWriter.cpp
+++ b/io/GltfWriter.cpp
@@ -36,7 +36,6 @@
 
 #include <iostream>
 #include <nlohmann/json.hpp>
-
 #include <pdal/PointView.hpp>
 #include <pdal/util/FileUtils.hpp>
 #include <pdal/util/OStream.hpp>
@@ -76,6 +75,10 @@ void GltfWriter::addArgs(ProgramArgs& args)
     args.add("alpha", "Alpha factor [0-1]", m_alpha, 1.0);
     args.add("double_sided", "Whether the material should be applied to "
              "both sides of the faces.", m_doubleSided);
+    args.add("color_vertices", "If color data is present for each point, write "
+             "it to the output vertices. Note that most renderers will "
+             "interpolate the color of each vertex across a face, so this may "
+             "look odd.", m_colorVertices, true);
 }
 
 
@@ -122,6 +125,17 @@ void GltfWriter::write(const PointViewPtr v)
         vd.m_vertexByteLength += v->size() * sizeof(float) * 3; // 3 for X,Y,Z
     }
 
+    // Only color vertices if color dimensions are present and the option is
+    // enabled
+    m_colorVertices = v->hasDim(Dimension::Id::Red)
+                      && v->hasDim(Dimension::Id::Green)
+                      && v->hasDim(Dimension::Id::Blue)
+                      && m_colorVertices;
+    if (m_colorVertices) {
+        // Add the length of 3 colors to the vertex byte length
+        vd.m_vertexByteLength += v->size() * sizeof(float) * 3; // 3 for R,G,B
+    }
+
     m_binSize += vd.m_indexByteLength + vd.m_vertexByteLength;
     m_totalSize = static_cast<size_t>(out.position()) + m_binSize;
     if (m_totalSize > (std::numeric_limits<uint32_t>::max)())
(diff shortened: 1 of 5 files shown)


io/GltfWriter.cpp
```diff
@@ -173,97 +204,165 @@ void GltfWriter::writeJsonChunk()
...
+    u_short nextAccessorIndex = 0;
+    u_short normalAccessorIndex = 0;
+    u_short colorAccessorIndex = 0;
+    u_short positionAccessorIndex = 0;
+    u_short faceAccessorIndex = 0;
     for (const ViewData& vd : m_viewData)
     {
+        // Buffer views
+        // Vertex indices (faces)
+        const u_short faceBufferViewIndex = 0;
```
abellgithub, line 230: I don't understand this change.  If there are more than one data view, all the views get combined.

abellgithub:
This should be nice.  What's your experience with the mesh creation in PDAL?  Are you using `filters.poisson`?

io/GltfWriter.hpp
```diff
@@ -79,6 +80,20 @@ class PDAL_DLL GltfWriter : public Writer
     double m_blue;
     double m_alpha;
     bool m_doubleSided;
+    bool m_colorVertices;
 };
 
+
+struct GltfWriter::ViewData
```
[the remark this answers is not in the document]
KeyboardSounds: I had to move it into the header so that GltfWriterTest.cpp would compile. I'm possibly showing how rusty I am at C++ here, there's probably a better way to fix this that I've forgotten.

abellgithub:
I fixed the issue you were having with needing ViewData in the header. 
 You should be able to merge this into your branch: https://github.com/PDAL/PDAL/tree/gltf-tests

io/GltfWriter.cpp
KeyboardSounds: As far as I understand, the original version of this code before my changes created a separate BufferView for each ViewData, and a separate Accessor for each BufferView.

But then in in the actual primitive added to the gltf, it did this:
```
    NL::json mesh;
    mesh["primitives"].push_back(
        {
            { "attributes", {{"POSITION", 1}} },
            { "indices", 0 },
            { "material", 0 }
        }
    );
```
Each integer there is an index of an Accessor. Regardless of how many Accessors were actually created, it only ever references two: Accessor 0, which has the indices for the faces, and accessor 1, which describes the positions of the first BufferView. As far as I can see, all other accessors are ignored entirely, and therefore their geometry isn't actually rendered by an application when it loads the glTF. Indeed, a gltf validator will give warnings that accessors with indices > 1 aren't used. This seems incorrect to me, but I presumed it was like that for a reason, so I tried to make my code replicate this behaviour-- although I think I've done this wrong, and it will use the _last_ accessors created rather than the first ones.

I think a better way to do this would be to create a separate primitive for each ViewData, referencing the appropriate accessors for that ViewData. I'm happy to implement this, it just seemed beyond the scope of what I was trying to achieve with this PR.

Which would you prefer? I can update my code to keep working in a similar way, or I can change to code to make a primitive for each ViewData.

KeyboardSounds:
> This should be nice. What's your experience with the mesh creation in PDAL?
The current project I'm working on is the first time I've ever used PDAL

I've been trying to make a mesh out of a fairly large point cloud (~77 million points) and PDAL has done a really good job. The results have been highly satisfactory, if slightly too smooth looking-- but my source data doesn't have any normals, so that's to be expected. It uses about 50GB of RAM and takes roughly 20 hours, which is inconvenient, but I presume that's unavoidable for a point cloud this size. I haven't had any crashes using `pdal pipeline` and `filter.poisson` except when I ran out of RAM :)

> I fixed the issue you were having with needing ViewData in the header.
> You should be able to merge this into your branch: https://github.com/PDAL/PDAL/tree/gltf-tests
Much appreciated, thank you! I'll update this PR soon.

io/GltfWriter.cpp
abellgithub: I think you're probably correct.  My recollection is foggy.  I doubt that it was well tested and I found the terminology confusing.  Making separate primitives for each ViewData seems the way to go.  Thanks.

abellgithub:
Is this ready?

KeyboardSounds:
It is now! Sorry about the delay, I got pulled into doing other urgent work.

PR #3046, opened 2020-04-24
Enable clamping in colorinterp filter

### LINKED ISSUE #2729: Colorinterp can make points invisible
The ColorinterpFilter consciously hides points that fall outside the given range (assuming that RGB dimensions are empty), which could lead users to believe that points have been culled.

https://github.com/PDAL/PDAL/blob/84d7e666e331140c308a40132f8fc5b013e68213/filters/ColorinterpFilter.cpp#L272-L274

We should at least provide an option to saturate the coloration at either the high or low end, rather than falling back to existing (possibly zero) values.

MattCsencsits:
Fixes #2729.

Previously values outside a determined range would be left with
default values, potentially providing unexpected results. This
commit adds a `clamp` parameter to a user can explicitly control
whether values outside the given range will be colored.

There apparently was an explicit design decision to not include
values equal to the maximum in the coloring. When clamping, this
design then resulted in values that were greater than the max to
still be left uncolored. Since nothing in the available user
documentation explicitly states that the valid range is exclusive
of the maximum value it has been changed to be inclusive.

This required modifying two unit tests which were testing for
maximum values to be excluded from the coloring.

Another approach could have been to clamp to a value slightly
less than m_max (e.g. m_max - DBL_EPSILON). But this would still
leave a discrepancy between the documentation and the implementation.

### DIFF
--- a/filters/ColorinterpFilter.cpp
+++ b/filters/ColorinterpFilter.cpp
@@ -37,6 +37,7 @@
 #include <pdal/GDALUtils.hpp>
 #include <pdal/PointView.hpp>
 #include <pdal/util/ProgramArgs.hpp>
+#include <pdal/util/Utils.hpp>
 
 #include <array>
 #include <algorithm>
@@ -111,6 +112,9 @@ void ColorinterpFilter::addArgs(ProgramArgs& args)
         std::numeric_limits<double>::quiet_NaN());
     args.add("maximum", "Maximum value to use for scaling", m_max,
         std::numeric_limits<double>::quiet_NaN());
+    args.add("clamp",
+        "Clamp and color values outside the range [minimum, maximum]",
+        m_clamp, false);
     args.add("ramp", "GDAL-readable color ramp image to use", m_colorramp,
         "pestel_shades");
     args.add("invert", "Invert the ramp direction", m_invertRamp, false);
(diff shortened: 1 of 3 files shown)


filters/ColorinterpFilter.cpp
```diff
@@ -269,14 +273,24 @@ bool ColorinterpFilter::processOne(PointRef& point)
...
+    if (v < m_min || v > m_max)
         return true;
 
     double factor = (v - m_min) / (m_max - m_min);
     size_t img_width = m_redBand.size();
     size_t position = size_t(std::floor(factor * img_width));
 
+    // Handle the case that v == m_max (position == img_width) by clamping
+    // position to img_width - 1
+    position = pdal::Utils::clamp(position, 0UL, img_width - 1);
```
abellgithub, line 292: Maybe just std::min(position, image_width - 1); ?

filters/ColorinterpFilter.cpp
```diff
@@ -269,14 +273,24 @@ bool ColorinterpFilter::processOne(PointRef& point)
 {
     double v = point.getFieldAs<double>(m_interpDim);
 
-    // Don't color points that aren't in the min/max range.
-    if (v < m_min || v >= m_max)
+    if (m_clamp) v = pdal::Utils::clamp(v, m_min, m_max);
```
abellgithub, line 276: No one-line if's, please.

filters/ColorinterpFilter.cpp
```diff
@@ -269,16 +285,33 @@ bool ColorinterpFilter::processOne(PointRef& point)
...
         return true;
+    }
 
     double factor = (v - m_min) / (m_max - m_min);
     size_t img_width = m_redBand.size();
     size_t position = size_t(std::floor(factor * img_width));
 
+    // Handle the case that v == m_max (position == img_width) by clamping
+    // position to img_width - 1
+    position = std::min(position, img_width - 1);
```
abellgithub, line 309: This needs to be guarded:

`(std::min)(position, img_width - 1);`

Sorry I didn't mention this before.

ntr-808:
thanks @MattCsencsits and @abellgithub. this is a great feature for PDAL

PR #3049, opened 2020-04-28
Nanoflann update

abellgithub:


abellgithub:
Just as a note, there's not much new here.  I think I saw like one real bug fix.  Also, I purged a ton of junk from this that we don't use in hope of reducing compilation time and making it easier to follow.  It does modernize a bit of code.

PR #3070, opened 2020-05-14
Fix row/column calculation.

### LINKED ISSUE #3069: SMRF can put data in the wrong cells
The calculation of the row and column position in `filters.smrf` is incorrect, which may lead to point data being placed into incorrect cells.

abellgithub:
Close #3069


abellgithub:
@chambbj : I didn't know what to do WRT the size of r/c.  In one place it was `int`, in the other `size_t`.  I didn't know if we should check for any overflow.

chambbj:
> @chambbj : I didn't know what to do WRT the size of r/c. In one place it was `int`, in the other `size_t`. I didn't know if we should check for any overflow.

No idea how we ended up with both `size_t` and `int`. I imagine you could always shoot yourself in the foot by picking cell size small enough. Do we protect against it elsewhere? Maybe GDAL writer or similar?

PR #3075, opened 2020-05-15
Initial commit of Lloyd's algorithm for k-means clustering

chambbj:
Putting this out there to see if there is any interest in a k-means clustering approach. This is perhaps a not very optimal, but easy to understand implementation of [Lloyd's algorithm](https://en.wikipedia.org/wiki/Lloyd%27s_algorithm).

Default cluster dimensions are XYZ, but user can provide a comma separated list of alternate dimensions to use.

Initial cluster centers are seeded using farthest point sampling, so this sampling method has been moved from filters/FarthestPointSampling.cpp to private/Segmentation.cpp.

Any conditioning or normalization of the dimensions will need to take place prior to clustering.

KDFlexIndex has been extended slightly to support nearest neighbor searches on a flexible list of dimensions.

### DIFF
--- a/filters/LloydKMeansFilter.cpp
+++ b/filters/LloydKMeansFilter.cpp
@@ -49,6 +49,8 @@ static StaticPluginInfo const s_info{
 
 CREATE_STATIC_STAGE(LloydKMeansFilter, s_info)
 
+LloydKMeansFilter::LloydKMeansFilter() {}
+
 std::string LloydKMeansFilter::getName() const
 {
     return s_info.name;
@@ -100,10 +102,7 @@ void LloydKMeansFilter::filter(PointView& view)
     // additional dimensions as specified via filter options
     PointTable table;
     table.layout()->registerDims({Id::X, Id::Y, Id::Z});
-    for (auto const& dimid : m_dimIdList)
-    {
-        table.layout()->registerDim(dimid);
-    }
+    table.layout()->registerDims(m_dimIdList);
 
     // create view of initial cluster centers
     PointViewPtr centers(new PointView(table));
@@ -157,9 +156,9 @@ void LloydKMeansFilter::filter(PointView& view)
                 centers->setField(m_dimIdList[i], id, M1[i][id]);
             }
         }
+        centers->invalidateProducts();
+        centers->build3dIndex();
     }
-    centers->invalidateProducts();
-    centers->build3dIndex();
 }
 
 } // namespace pdal
--- a/filters/LloydKMeansFilter.cpp
+++ b/filters/LloydKMeansFilter.cpp
@@ -1,0 +1,165 @@
+/******************************************************************************
+ * Copyright (c) 2020, Bradley J Chambers (brad.chambers@gmail.com)
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following
+ * conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided
+ *       with the distribution.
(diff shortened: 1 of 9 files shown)


chambbj:
`libssh2` Windows issue now documented in https://github.com/conda-forge/libssh2-feedstock/pull/37. It's listed as "broken" in Anaconda too, so I'm guessing if we do another pull without pinning, it should revert to v1.8.2. Will try shortly.

abellgithub:
Is this ready for review?

chambbj:
Yes, ready for review. Thanks!

filters/LloydKMeansFilter.cpp
```diff
@@ -0,0 +1,165 @@
...
+        // adjust clusters based on newly added points
+        for (PointId id = 0; id < m_k; ++id)
+        {
+            for (size_t i = 0; i < m_dimIdList.size(); ++i)
+            {
+                centers->setField(m_dimIdList[i], id, M1[i][id]);
+            }
+        }
+    }
+    centers->invalidateProducts();
```
abellgithub, line 161: I don't understand this invalidate/build.  The `centers` is about to be destroyed.

filters/LloydKMeansFilter.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+namespace pdal
+{
+
+class ProgramArgs;
+
+class PDAL_DLL LloydKMeansFilter : public Filter
+{
+public:
+    LloydKMeansFilter() : Filter() {}
```
abellgithub, line 49: This should be moved to the C++ file to help avoid future issues with using STL with unknown types.

filters/private/Segmentation.cpp
```diff
@@ -266,5 +266,51 @@ void segmentReturns(PointViewPtr input, PointViewPtr first,
...
+    {
+        // Find the max distance in min_dists, this is the farthest point from
+        // any point currently in the output PointView.
+        auto it = std::max_element(min_dists.begin(), min_dists.end());
+
+        // Record the PointId of the farthest point and add it to the output
+        // PointView.
+        PointId idx(it - min_dists.begin());
+	ids[i] = idx;
+
```
abellgithub, line 300: There may be a spacing there here, but it may just be the diff tool.

abellgithub: I think you can lose the last loop iteration and just assign ids[count - 1] to avoid another KDtree search.

filters/LloydKMeansFilter.cpp
```diff
@@ -0,0 +1,165 @@
...
+        std::vector<point_count_t> cnt(m_k, 0);
+
+        // for every point, find closest cluster center and assign ClusterID
+        for (PointRef p : view)
+        {
+            PointId q = kdi.neighbor(p);
+            p.setField(Id::ClusterID, q);
+
+            // update cluster size and mean
+            cnt[q]++;
```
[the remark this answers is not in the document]
chambbj: Because it is drawn from `kdi` which is the tree built from `centers` which itself is constructed from only the `m_k` ids sampled at line 97. Is there something we can do to make this clearer? Maybe better variable names?

filters/LloydKMeansFilter.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+    // come up with k random samples for initial cluster centers (based on
+    // spatial farthest point sampling)
+    PointIdList ids = Segmentation::farthestPointSampling(view, m_k);
+
+    // setup table with at least XYZ as required by KDIndex, plus any
+    // additional dimensions as specified via filter options
+    PointTable table;
+    table.layout()->registerDims({Id::X, Id::Y, Id::Z});
+    for (auto const& dimid : m_dimIdList)
```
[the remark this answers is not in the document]
chambbj: Fixing in next commit

filters/LloydKMeansFilter.cpp
chambbj: I too am surprised not to see this within the iteration loop. Maybe accidentally moved it outside the loop at some point!

PR #3085, opened 2020-05-21
Use signed int for clustering/grouping

### LINKED ISSUE #3079: Groupby not accepting signed integers
Not sure if this is an issue or not, but I have difficulty working around the following situation:
```
{
        "type":"filters.dbscan",
        "min_points":10,
        "eps":1.2,
        "dimensions":"X,Y,Z"
    },{
        "type":"filters.groupby",
        "dimension":"ClusterID"
    }
```
The output of dbscan is signed because of the unclassified points receiving a -1
Groupby does not accept this and errs out with
`PDAL: Unable to fetch data and convert as requested: ClusterID:int64_t(-1) -> unsigned long`

The use case is that I do another iteration in the pipeline where I want the unclustered points to be considered as well. Currently I can circumvent this by assigning a very high integer to ClusterID=-1 but this doesn't feel like the right way.
Would it make sense to have groupby accept signed longints?

chambbj:
Modify DBSCAN to use the standard ClusterID dimension, which is now
defined as int64. The Groupby filter is similarly instructed to evaluate
int64. This allows clustering algorithms like DBSCAN to use -1 to
indicate noise.

Close #3079

### DIFF
--- a/filters/DBSCANFilter.cpp
+++ b/filters/DBSCANFilter.cpp
@@ -45,6 +45,8 @@
 namespace pdal
 {
 
+using namespace Dimension;
+
 static StaticPluginInfo const s_info{
     "filters.dbscan", "DBSCAN Clustering.",
     "http://pdal.io/stages/filters.dbscan.html"};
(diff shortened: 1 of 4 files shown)


abellgithub:
Seems good to me!

Issue #3084, opened 2020-05-21
Extend Euclidean cluster extraction to work in 2D as well


chambbj:
There can be instances where it may make sense to run cluster extraction in a 2D mode. As a motivating example, consider only non-ground points after running a ground segmentation step such as SMRF. If I want to cluster sparsely sampled points on a vertical pole using 3D cluster extraction, I can end up with many clusters even though they are all more or less coincident in X and Y. Converting the clustering to only consider X and Y would place all "overlapping" points in the same cluster.

chambbj:
Closed via #3086

PR #3086, opened 2020-05-21
Add option to perform Euclidean cluster extraction in 2D (#3084)

chambbj:
Changed to return clusters as deque of PointIdLists (rather than
vector). Default behavior will continue to be to cluster in 3D.


chambbj:
Yes, it is. And you are right, it could be simplified.

chambbj:
It occurs to me that this needs a slight doc tweak (and a test or two for 2D clusters).

filters/private/Segmentation.hpp
```diff
@@ -91,10 +91,68 @@ std::ostream& operator<<(std::ostream& out, const PointClasses& classes);
   \param[in] tolerance the tolerance for adding points to a cluster.
   \returns a vector of clusters (themselves vectors of PointIds).
```
abellgithub, line 92: No longer vectors.

In this case, I can't imagine it much matters if you use deque or vector.

filters/private/Segmentation.hpp
```diff
@@ -91,10 +91,68 @@ std::ostream& operator<<(std::ostream& out, const PointClasses& classes);
...
 */
-PDAL_DLL std::vector<PointIdList> extractClusters(PointView& view,
-                                                  uint64_t min_points,
-                                                  uint64_t max_points,
-                                                  double tolerance);
+template <class T>
```
abellgithub, line 94: T might be named better in this case, like KDINDEX or something.

PR #3087, opened 2020-05-21
Allow covariance features filter to obtain neighbors by radius

### LINKED ISSUE #3083: Add radius option for covariance features filter
Several authors have suggested that radius is more robust than k nearest neighbors, especially for non-uniform point spacing, when computing point features. The covariance features filter (and possibly others?) should provide a radius argument as an alternative to k nearest neighbors.

chambbj:
Close #3083


chambbj:
@nicolas-chaulet Are you still using this filter? What are your thoughts on providing radius search as an alternative to k nearest neighbors? Happy to hear your thoughts on the proposed change...

PR #3042, opened 2020-04-22
Follow SOVERSION naming rules

hobu:
Address #3039, PDAL_BUILD_VERSION naming to again

PR #3096, opened 2020-05-26
Implement SOVERSION naming rules and document behavior 

hobu:
Fixes #3042

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,9 +26,25 @@ mark_as_advanced(CMAKE_VERBOSE_MAKEFILE)
 # PDAL general settings
 #------------------------------------------------------------------------------
 
-set(PDAL_API_VERSION "10")
-set(PDAL_BUILD_VERSION "11")
+#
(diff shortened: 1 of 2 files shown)


hobu:
This needs to be merged back to `2.1-maintenance` with the SOVERSION properly set to increment what was the last `2.0-maintenance` branch.

hobu:
@sebastic we are planning to release a 2.1.1 maintenance release before the end of June. What would you suggest we do with the SONAME given the broken value that was pushed out for 2.1.0?

sebastic:
The current state of the 2.1-maintenance branch looks fine, it keeps the `SOVERSION` at `10` but also fixes the real file to have the `.10.0.0` suffix instead of `.11`.

If there is an ABI break caused by the changes since 2.1.0 then bump the `SOVERSION` (via `PDAL_SOLIB_MAJOR`) accordingly. Although ABI breakage is not really appropriate for a patch release.

hobu:
> If there is an ABI break caused by the changes since 2.1.0

This is not expected at all.

PR #3097, opened 2020-05-26
Handle errors properly from Raster::statistics

abellgithub:
Close #3097


pdal/GDALUtils.hpp
```diff
@@ -848,7 +867,7 @@ class PDAL_DLL Raster
 
     GDALError wake();
 
-    std::string m_errorMsg;
+    mutable std::string m_errorMsg;
```
hobu, line 870: Why did this change to mutable too?

abellgithub: So you can set error state in const functions.

PR #3099, opened 2020-05-26
Use SrsTransform to make sure that traditional axis order is used for the tindex kernel

hobu:


kernels/TIndexKernel.cpp
```diff
@@ -712,7 +703,7 @@ gdal::Geometry TIndexKernel::prepareGeometry(const FileInfo& fileInfo)
     }
     try
     {
-       g = prepareGeometry(fileInfo.m_boundary, srcSrs, tgtSrs);
+       g = prepareGeometry(fileInfo.m_boundary, fileInfo.m_srs, &transform);
```
abellgithub, line 706: This isn't how this should be done.  I'll offer:
```
Geometry g(fileInfo.m_boundary, fileInfo.m_srs);
g.transform(m_tgtSrsString);
```

That's it.

PR #3103, opened 2020-05-26
Format ICP metadata using IOFormat in full precision

chambbj:


klassenjs:
Tested both transformation with `"combined"` and transformation with `-"centroid"*"transform"+"centroid"` workflows now both identically match the direct point cloud output of `filters.icp`.

PR #3106, opened 2020-05-27
Updates based on review.

abellgithub:
There are still private headers in EptReader.


io/EptReader.cpp
```diff
@@ -159,6 +141,28 @@ void EptReader::addArgs(ProgramArgs& args)
         m_args->m_ogr);
 }
 
+namespace
+{
+
+// Extract a string map from JSON.  Used in setForwards().
+StringMap toStringMap(const NL::json& fwd)
```
connormanning, line 148: It still seems strange to be reimplementing error handling that already exists in our JSON library: https://gist.github.com/connormanning/08462b92f1d9e29fe6454e07708c0d88

The only places where this is used are already wrapped in a try/catch which rethrows as a `pdal_error`.  Are we just doing this to avoid the `[json.exception.type_error.302]` part of the error message?

abellgithub: Oh, I tried `>>` and it didn't compile, so I just left it, figuring it wasn't worth more effort, but I can change this.  Didn't try assignment.

PR #3111, opened 2020-06-01
Add LiTreeFilter to segment individual trees

chambbj:
I tried to stick pretty close to the paper (available at [ResearchGate](https://www.researchgate.net/profile/Maggi_Kelly/publication/236325443_A_New_Method_for_Segmenting_Individual_Trees_from_the_Lidar_Point_Cloud/links/02e7e5179b5b619d96000000/A-New-Method-for-Segmenting-Individual-Trees-from-the-Lidar-Point-Cloud.pdf)).

> Li, W., Guo, Q., Jakubowski, M. K., & Kelly, M. (2012). A new method for segmenting individual trees from the lidar point cloud. Photogrammetric Engineering & Remote Sensing, 78(1), 75-84.

The method requires HeightAboveGround, which itself requires ground Classifications. In my test pipelines, I've also range filtered on HeightAboveGround to remove ground and near ground points that shouldn't be part of the trees. As such, as reasonable pipeline may be something like

```
[
    {
        "type":"filters.assign",
        "assignment":"Classification[:]=0"
    },
    {
        "type":"filters.smrf"
    },
    {
        "type":"filters.hag"
    },
    {
        "type":"filters.range",
        "limits":"HeightAboveGround[3:]"
    },
    {
        "type":"filters.litree"
    }
]
]
```

This can be very slow, especially if the near ground points are included. I'm open to any ideas for possible optimization, as I may have overlooked something.

I think the basic approach can work on any object that extends above the ground plane sufficiently, e.g., buildings and towers. Be mindful though that there needs to be sufficient separation laterally such that closely spaced objects do not get merged together.

### DIFF
--- a/doc/stages/filters.litree.rst
+++ b/doc/stages/filters.litree.rst
@@ -1,0 +1,52 @@
+.. _filters.litree:
+
+===============================================================================
+filters.litree
+===============================================================================
+
+The purpose of the Li tree filter is to segment individual trees from an input
+``PointView``. In the output ``PointView`` points that are deemed to be part of
+a tree are labeled with a ``TreeID``. Tree IDs start at 1, with non-tree points
+given a ``TreeID`` of 0.
+
+.. note::
+
+  The filter differs only slightly from the paper in the addition of a few
+  conditions on size of tree, minimm height above ground for tree seeding, and
+  flexible radius for non-tree seed insertion.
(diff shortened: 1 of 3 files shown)


hobu:
🌴 🎄 🌳 🌲 🎋

PR #3131, opened 2020-06-16
Report final number of ground/non-ground points

chambbj:
Our original logs only report ground/non-ground cells, which may be misleading.


filters/SMRFilter.cpp
```diff
@@ -380,10 +382,21 @@ void SMRFilter::classifyGround(PointViewPtr view, std::vector<double>& ZIpro)
...
         else
+        {
+            g++;
             view->setField(Id::Classification, i, ClassLabel::Ground);
+        }
     }
+    double p(100.0 * double(ng) / double(view->size()));
+    log()->floatPrecision(2);
+    log()->get(LogLevel::Debug) << "\t" << g << " ground points"
+                                << "\t" << ng << " non-ground points"
```
abellgithub, line 398: Is it clear here that `p` refers to the percent of ground points out of the total?

PR #3132, opened 2020-06-16
Enhance eigenvalues filter with radius and stride options for defining neighborhood

chambbj:
Default behavior does not change at all, but adds radius and strided neighborhood options which can be better in irregularly sampled data.


filters/EigenvaluesFilter.cpp
```diff
@@ -65,6 +65,9 @@ void EigenvaluesFilter::addArgs(ProgramArgs& args)
 {
     args.add("knn", "k-Nearest neighbors", m_knn, 8);
     args.add("normalize", "Normalize eigenvalues?", m_normalize, false);
+    args.add("stride", "Compute features on strided neighbors", m_stride, size_t(1));
```
abellgithub, line 68: There are some checks/feedback that could be provided WRT various options being incompatible.

chambbj: let me know if latest commit bring some more clarity

PR #3149, opened 2020-06-24
Support 'filename' to specify VLR content when adding VLRs using writers.las

hobu:


abellgithub:
The test doesn't actually check the value of the VLR. Did you do this by hand?

hobu:
> The test doesn't actually check the value of the VLR. 

The tests never tested the values of the data of the VLRs.

> Did you do this by hand?

I'm not sure what you mean. I decoded the data's base64 from the second case and stored that as a file.

PR #3151, opened 2020-06-25
Provide option to apply inverse transformation matrix

### LINKED ISSUE #3126: Inverse of Transformation support in filters.transformation
In certain cases, it would be convenient to apply the inverse of the affine transformation applied using filters.transformation.  We sometimes need to do this when projecting back to the original (raw) data collected in the scanner coordinate system when examining instrument calibration for example.

For forward transformation, 4 by 4 matrix M consists of P a 3 by 3 rotation, and v the 3 by 1 translation:

M = | P    v  |
       |  0   1  |

Inverse is given as:

M^-1 = | P^-1.   (-P^-1)*v |
            |   0                 1     |

chambbj:
Close #3126

### DIFF
--- a/doc/stages/filters.transformation.rst
+++ b/doc/stages/filters.transformation.rst
@@ -41,6 +41,10 @@ This example rotates the points around the z-axis while translating them.
 Options
 -------
 
+invert
+  If set to true, applies the inverse of the provided transformation matrix.
+  [Default: true]
+
 _`matrix`
   A whitespace-delimited transformation matrix.
   The matrix is assumed to be presented in row-major order.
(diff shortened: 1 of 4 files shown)


hobu:
Docs need update

chambbj:
> Docs need update

Always forget this...

chambbj:
Done messing with this for now. We can completely replace with Eigen down the line if we deem it necessary.

abellgithub:
If we do more on this, perhaps it makes sense to do quaternions w/translation.

chambbj:
Luckily the Eigen support _should_ make that easier.

PR #3161, opened 2020-07-08
Create raster based on face interpolation.

abellgithub:


filters/FaceRasterFilter.cpp
```diff
@@ -0,0 +1,142 @@
...
+        double x3 = v.getFieldAs<double>(Dimension::Id::X, t.m_c) - m_limits.xOrigin;
+        double y3 = v.getFieldAs<double>(Dimension::Id::Y, t.m_c) - m_limits.yOrigin;
+        double z3 = v.getFieldAs<double>(Dimension::Id::Z, t.m_c);
+
+        double xmax = (std::max)((std::max)(x1, x2), x3);
+        double xmin = (std::min)((std::min)(x1, x2), x3);
+        double ymax = (std::max)((std::max)(y1, y2), y3);
+        double ymin = (std::min)((std::min)(y1, y2), y3);
+
+        double halfEdge = m_limits.edgeLength / 2;
```
chambbj, line 116: This _could_ be moved to the top of `filter()` and `halfEdge` used when setting `xOrigin` and `yOrigin`. Also, no need to recompute for each triangle. Doubt it makes any noticeable difference to do so though.

pdal/private/MathUtils.cpp
```diff
@@ -384,5 +384,55 @@ void writeMatrix(Eigen::MatrixXd data, const std::string& filename,
...
+//
+/// Return the interpolated Z value at location X/Y. If the X/Y location is not
+/// in the triangle, return infinity.  If the determinant is 0, the input points
+/// aren't a triangle (they're collinear).
+/// \param x1, y1, z1  Coordinates of point 1.
+/// \param x2, y2, z2  Coordinates of point 2.
+/// \param x3, y3, z3  Coordinates of point 3.
+/// \param x, y  X and Y coordinates of location to find an interpolated Z
+/// \return  Interpolated Z value or infinity.
+double barycentricInterpolation(double x1, double y1, double z1,
```
chambbj, line 399: I'm guessing you didn't change anything when moving to MathUtils (good call doing so BTW). We don't have any tests for this though. Would probably be a good thing to add going forward.

pdal/private/Raster.hpp
```diff
@@ -0,0 +1,172 @@
...
+    const RasterLimits& limits() const
+        { return m_limits; }
+
+private:
+    RasterLimits m_limits;
+    std::vector<T> m_data;
+    T m_initializer;
+};
+
+using Rasterd = Raster<double>;
```
chambbj, line 169: Totally bikeshedding, but I don't like `Rasterd`. Maybe `RasterD` is better? Doesn't really matter in the end.

Why `using` and not a `typedef`?

pdal/private/Raster.hpp
```diff
@@ -0,0 +1,172 @@
...
+    double edgeLength;
+
+private:
+    Arg *m_xOriginArg;
+    Arg *m_yOriginArg;
+    Arg *m_widthArg;
+    Arg *m_heightArg;
+};
+
+// This is a raw raster suitable for sticking into GDAL. X goes right to left, Y goes from
```
chambbj, line 84: Right to left? This always trips me up, but I thought Y was the axis that typically gets flipped when not careful.

pdal/private/MathUtils.hpp
```diff
@@ -63,10 +63,13 @@ typedef std::shared_ptr<PointView> PointViewPtr;
 namespace math
 {
 
-PDAL_DLL PointViewPtr demeanPointView(const PointView& view);
-PDAL_DLL PointViewPtr demeanPointView(const PointView& ,double* centroid);
-PDAL_DLL PointViewPtr transform(const PointView&, double* matrix);
-PDAL_DLL void transformInPlace(PointView&, double* matrix);
+PointViewPtr demeanPointView(const PointView& view);
```
chambbj, line 66: Deliberately not exporting these now private functions? I wonder if any plugins have depended on these?

pdal/PointView.hpp
```diff
@@ -301,6 +319,7 @@ class PDAL_DLL PointView : public PointContainer
     std::queue<PointId> m_temps;
     SpatialReference m_spatialReference;
     std::map<std::string, std::unique_ptr<TriangularMesh>> m_meshes;
+    std::map<std::string, std::unique_ptr<Rasterd>> m_rasters;
```
chambbj, line 322: Would the existing GDAL raster bands also get created and added in such a manner? Maybe not, as those are bands for a single raster out.

pdal/private/MathUtils.cpp
abellgithub: There are tests in hag that exercise this.   I could extract, I think.

chambbj: Right, I suppose by proxy via https://github.com/PDAL/PDAL/blob/master/test/unit/filters/HAGFilterTest.cpp#L47.

pdal/private/Raster.hpp
abellgithub: I was following the Eigen way on the name. I don't care.

I like `using` better. It works in template cases where `typedef` doesn't. And if you declare function pointers as types, the `typedef` way is yuck. That's all I got.

pdal/private/Raster.hpp
abellgithub: Oops.

pdal/private/MathUtils.hpp
abellgithub: This perhaps needs some discussion. Here I'm making the assumption here that someone writing a plugin has the code and can build in private things if they need them. This came up with `rotate`, which I also stuck in here but was used by the I3S/SLPK plugin.

I was just working on making more stuff private, which I think it should be, but it's also not a fight that I want to get into.

pdal/PointView.hpp
abellgithub: I've reworked the GDALWriter to use the Raster class, but right now it's all still self-contained.  That bit is a work in progress. I have mixed feelings about all of this, as I did with the mesh code. I'm just not sure where it ends.

filters/FaceRasterFilter.cpp
abellgithub: I expect the compiler would do this for me, but I just missed this. Thx.

pdal/private/Raster.hpp
chambbj: I guess you are right, it's also lower case in Eigen, just separated by the dimensionality. It just looked at first like a typo when I first encountered it in the FaceRasterFilter code.

:+1: to `using`. It just caught me off guard and we use plenty of `typedefs` elsewhere hence the curiosity. No need to change.

pdal/private/MathUtils.hpp
chambbj: > it's also not a fight that I want to get into

Nor I. We've been pretty consistent with that approach to private code as of late, so we can continue it here.

PR #3172, opened 2020-07-14
Ept summary bounds

### LINKED ISSUE #3168: pdal info --summary for ept origin returns bounds for whole dataset
when running pdal info --summary on an ept dataset and passing an reader.ept.origin param, the points value returned is just for the specific file but the bounds are for the whole dataset. It should be for just the particular origin specified to be consistent

connormanning:
Closes #3168.

### DIFF
--- a/doc/stages/readers.ept.rst
+++ b/doc/stages/readers.ept.rst
@@ -17,7 +17,7 @@ with `Potree`_ or `Plasio`_.
 Example
 --------------------------------------------------------------------------------
 
-This example downloads a small area around the the Statue of Liberty from the New York City data set (4.7 billion points) which can be viewed in its entirety in `Potree`_ or `Plasio`_.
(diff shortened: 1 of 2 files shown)


doc/stages/readers.ept.rst
```diff
@@ -53,7 +53,7 @@ Options
 --------------------------------------------------------------------------------
 
 filename
-    EPT resource from which to read.  Because EPT resources do not have a file extension, to specify an EPT resource as a string, it must be prefixed with ``ept://``.  For example, ``pdal translate ept://http://na.entwine.io/autzen autzen.laz``. [Required]
+    Path to the EPT resource from which to read, ending with ``ept.json``.  For example, ``/Users/connor/entwine/autzen/ept.json`` or ``http://na.entwine.io/autzen/ept.json``.  For remote paths which are not on the local filesystem, PDAL must be built with Curl support. [Required]
```
abellgithub, line 56: Looking at the code, it seems that curl is now required.

connormanning: Removed Curl comment. 5abea26.

Issue #111, opened 2012-11-09
Add switch to apps to tell us if we're in release or debug mode


hobu:
It would be helpful to be able to determine via a switch if PDAL was compiled with debug defined or not.

PR #3173, opened 2020-07-14
Add note on PDAL_DRIVER_PATH usage.

jedfrechette:
Add note on usage of PDAL_DRIVER_PATH to installation instructions.

### DIFF
--- a/doc/quickstart.rst
+++ b/doc/quickstart.rst
@@ -78,8 +78,12 @@ the PDAL maintenance branch.
 
     It is a good idea to install PDAL in it's own environment (or
     add it to an existing one). You will **NOT** want to add it to your default
-    environment named ``base``. Managing environments is beyond the scope of
-    the quickstart, but you can read more about it `here <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html#managing-envs>`_.
+    environment named ``base``. If you do add it to the ``base`` environment you
+    will also want to set the environment variable ``PDAL_DRIVER_PATH`` so that
+    all available plugins can be detected. On *nix the path would typically be
+    ``$CONDA_PREFIX/lib`` and on Windows ``%CONDA_PREFIX%\Library\bin``.
+    Managing environments is beyond the scope of the quickstart, but you can
+    read more about it `here <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html#managing-envs>`_.
 
 To install the PDAL package so that we can use it to run PDAL commands, we run
 the following command to create an environment named ``myenv``, installing PDAL


hobu:
I'm confused by this recommendation. Things should Just Work ™️ if you use a normal environment and activate it. I don't understand the need for this note.

jedfrechette:
Perhaps it would be better somewhere else, but I think it would be good to document what run time configuration options affect PDAL's execution environment a little better. Are there any others besides `PDAL_DRIVER_PATH`? 

For example both [GDAL](https://gdal.org/user/configoptions.html) and [PROJ_LIB](https://proj.org/usage/environmentvars.html) have dedicated pages in their docs. That seems a little excessive though if PDAL only has one option. The only documentation I could find for PDAL_DRIVER_PATH was in the FAQ and and refers to a specific error message, so it's not a very obvious place to look for general configuration info. This is compounded by the fact that most of PDAL's drivers are built in and not affected by PDAL_DRIVER_PATH. It also isn't easy for a user to tell which drivers are plugins that will be effected. All of this combines to make the effects of having PDAL_DRIVER_PATH set incorrectly pretty subtle for a user who doesn't know to look for them.

What about moving the FAQ entry, more or less intact, to an Environment Variables section of the Quickstart guide? That way it will be obvious to users when they're first setting up their environment what is affecting that environment. Most can ignore it, but it will be easier to discover for those who do need to manage in manually.

abellgithub:
I mean, if you can't load a stage, it literally tells you the web page to read that explains the issue.

If you're dealing with a standard installation and need to set this environment variable, we should fix something.

jedfrechette:
> I mean, if you can't load a stage, it literally tells you the web page to read that explains the issue

Sure, but I can't assume that all stages are going to be available by default so if I run `pdal --drivers` to find out which ones are available and PDAL_DRIVER_PATH is set incorrectly then I won't get an any errors which might lead me to the FAQ entry. If I don't know about PDAL_DRIVER_PATH then it isn't obvious at all why some of the drivers in the docs are available and some aren't.

As an aside, what's the difference between a driver and a stage? Is a driver simply a stage that does IO and if so is the reason `--drivers` lists filters as well just historical?

abellgithub:
"driver" and "stage" are synonyms from a PDAL perspective. I don't know how that came about.

I don't have any problem with better documentation, but I'd rather fix issues that cause people to need PDAL_DRIVER_PATH.

hobu:
> As an aside, what's the difference between a driver and a stage?

Driver is GDAL's word for it, which is where it came from very early on. We should add a `--stage` synonym for the `--driver` argument to reduce this ambiguity.

jedfrechette:
I've encountered two scenarios where I needed to worry about the environment variable, both in the context of using the conda packages on Windows. See [pdal-feedstock issues #111](https://github.com/conda-forge/pdal-feedstock/issues/111).

1. If pdal is installed to the base environment PDAL_DRIVER_PATH doesn't get set because activate isn't run.
2. If you use Git bash instead of the native Windows shell the variable gets set incorrectly because the wrong activate is run.

Those are both probably edge cases but I'm sure there are others. From my perspective, while it would be ideal if the environment variable wasn't necessary, as long as it exists and is affecting the runtime environment I think just having it documented a little more prominently is fine. Once I knew it existed it was obvious what needed to happen. Unfortunately, the only reason I learned about it was by filing a ticket after spending a significant amount of time debugging behavior that seemed pretty mysterious at the time.

abellgithub:
My claim is that you shouldn't need the environment variable for a good installation. I think the answer is that we should make the lookup DLL-relative as it is now in PROJ.

abellgithub:
I believe this has been made obsolete by ba4b9ac4b85ab31a76eebf416e431fbfbd3a6716 and #3181

Please close this if you agree.

jedfrechette:
Looks good to me. Thanks!

PR #3176, opened 2020-07-15
Remove filters.hag.

abellgithub:


hobu:
I think we should keep `filters.hag.rst` but remove it from the index. Anyone with older software is going to be getting links to that and will be confused when it 404s.

abellgithub:
You mean remove the filter, but just keep the doc for it around?

hobu:
yeah, keep the doc's location but update it to point at the new filters and state it was removed in 2.2.

PR #3177, opened 2020-07-15
Add "where" support for skipping points

### LINKED ISSUE #3118: Add generic "where" argument for filters
There have been numerous requests to have general handling of specifying a subset of points to be handled by a filter.

abellgithub:
Close #3118


abellgithub:
Note that there's no doc for this. I'm going to stick that in another PR since it's a significant change by itself.

PR #3179, opened 2020-07-16
Add "where" option support to documentation.

abellgithub:


doc/stages/expression.rst
```diff
@@ -0,0 +1,104 @@
...
+
+Selects points with a classification of 7 or 8 and number of returns equal to 1.  Note
+that in this case the parentheses are necessary.
+
+::
+
+    X > 2500 && X < 4700 && Y > 0
+
+Selects points with an X between 2500 and 4700 and a positive Y value.
+
```
hobu, line 104: Some more detail needed

* what happens for dimension names that are not known by PDAL? Are they addressable?
* a statement about case (in)sensitivity
* a few more examples of common ``filters.range`` scenarios such as ReturnNumber or NumberOfReturns
* An example of statement negation ``(!(Classification == 2 || ReturnNumber == 1)``

doc/stages/filters.icp.rst
```diff
@@ -128,3 +128,6 @@ rt
 
 tt
   Translation threshold. [Default: **9e-8**]
+
```
hobu, line 131: The file's encoding or something seems wonky here.

abellgithub: I see that pretty regularly. No clue.

doc/stages/expression.rst
abellgithub: > what happens for dimension names that are not known by PDAL? Are they addressable?

? When a pipeline is run, we know all the dimension names.

> a statement about case (in)sensitivity

This doesn't seem appropriate here. It's the same as it is for every other option that uses dimension names.

PR #3184, opened 2020-07-22
Make readers.head streamable

hobu:


filters/HeadFilter.cpp
```diff
@@ -50,4 +50,47 @@ std::string HeadFilter::getName() const
...
+        "to skip from the beginning.", m_invert, false);
+}
+
+bool HeadFilter::processOne(PointRef& point)
+{
+
+    bool keep = false;
+    if (m_index < m_count)
+        keep = true;
+    m_index++;
```
abellgithub, line 68: m_index needs to be set to 0 in `ready()`

hobu: it was set to 0 in the ctor. Is that not what we should be doing?

abellgithub: This handles the case where someone may execute a stage more than once.

PR #3188, opened 2020-07-24
PGPointcloud: Add streaming support to reader

foggel:
Thia patch adds streaming support for the pgpointcloud reader.


foggel:
Thank you, I will modify code and come back!

PR #3191, opened 2020-07-24
PointViewSets returned by ground segmentation filters can simply contain the input PointView

### LINKED ISSUE #3185: SMRF can be converted from run() to filter(), avoiding creation of new PointView
The current structure of the filter dates back to when we were only returning (at least as an option) the ground returns. Now that we are just attributing the Classification dimension, this no longer needs to be the case. It will play much more nicely with the forthcoming "where" syntax, especially "where_merge" if it operates in place. (Note: this probably applies to other filters as well, e.g., PMF, in which case the ticket should be expanded as such.)

chambbj:
Close #3185

### DIFF
--- a/test/unit/filters/CSFilterTest.cpp
+++ b/test/unit/filters/CSFilterTest.cpp
@@ -62,5 +62,7 @@ TEST(CSFilterTest, emptyView)
     filter->prepare(table);
 
     PointViewSet s = filter->execute(table);
-    EXPECT_EQ(s.size(), 0u);
+    EXPECT_EQ(s.size(), 1u);
+    PointViewPtr v = *s.begin();
+    EXPECT_EQ(v->size(), 0u);
 }
--- a/filters/CSFilter.cpp
+++ b/filters/CSFilter.cpp
@@ -115,7 +115,7 @@ void CSFilter::prepared(PointTableRef table)
     for (auto& r : m_args->m_ignored)
     {
         r.m_id = layout->findDim(r.m_name);
-        if (r.m_id == Dimension::Id::Unknown)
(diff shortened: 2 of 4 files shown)


test/unit/filters/CSFilterTest.cpp
```diff
@@ -62,5 +62,7 @@ TEST(CSFilterTest, emptyView)
     filter->prepare(table);
 
     PointViewSet s = filter->execute(table);
-    EXPECT_EQ(s.size(), 0u);
+    EXPECT_EQ(s.size(), 1u);
```
abellgithub, line 65: Wondering what changed here.

chambbj:
It's just simpler IMO. Thanks for pointing out that we could do this in the first place. Longer term, I could see an argument for switching from `run()` to `filter()` as this now behaves much more like the latter, but there is no burning need. This will play nicely with #3177, which is what I was after.

abellgithub:
Agreed WRT `filter()`. Not terribly important.

test/unit/filters/CSFilterTest.cpp
chambbj: Simply that I add the input PointView to the output PointViewSet via list initialization in each of the three filters now, prior to checking whether or not the PointView has any points. I had previously returned with an empty PointViewSet in that case, but now return the PointViewSet, which just happens to contain an empty PointView. It could be reverted if it makes more sense the original way.

abellgithub: Nope. That's fine. Thanks for the explanation. I wouldn't have questioned if not for the change.

PR #3192, opened 2020-07-24
Migrate from Azure Pipelines to Github Actions

hobu:
Migrate from Azure Pipelines to Github Actions

* Build Conda forge PDAL feedstock for all three platforms
* Upload Conda packages as artifacts to the build
* Build a "fixed" configuration that includes a pinned GDAL
* Build a "floating" configuration that grabs whatever dependencies are current for PDAL from Conda
* Docs are built and pushed as artifacts. Commits to the `*-maintenance` will cause the docs to be built and pushed to pdal.io as before
* Remove Azure Pipelines, Travis, and AppVeyor configuration

Some more configuration tweaks that could be made include:

- [ ] pinning the compilers to known versions. The MSVC compiler is whatever is activated by the environment, for example
- [ ] adding a static analysis builder
- [ ] compiler matrix. It would be nice to have coverage of various flavors of gcc and clang on osx and linux.


hobu:
I would like to merge this back to the ``2.2-maintenance`` branch as well after it is squashed and merged into master.

scripts/ci/linux/setup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+echo "Configuring build type '$BUILD_TYPE'"
+mkdir build
+
+conda update -n base -c defaults conda
+conda install cmake ninja conda-build compilers -y
```
chambbj, line 7: Is `conda-build` actually required?

scripts/ci/linux/setup.sh
```diff
@@ -0,0 +1,22 @@
...
+
+if [ "$BUILD_TYPE" == "fixed" ]; then
```
chambbj, line 9: Only problematic on Linux? Or just overkill to matrix this to other OSs also?

hobu: I was replicating what we had on AzP. It would be a good idea to spread it to the other builds too.

scripts/ci/linux/setup.sh
hobu: conda-build gets you compilers, cmake, and ninja. Since they are explicitly stated, we don't need it.

.github/workflows/osx.yml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Test
+      shell: bash -l {0}
+      run: |
+          source ../scripts/ci/osx/test.sh
+      working-directory: ./build
+
+    - name: Examples
+      shell: bash -l {0}
+      run: |
+          source ./scripts/ci/osx/examples.sh
```
chambbj, line 48: Just hoping to catch it if we break one? Why not on Windows too?

hobu: yes, this should be extended to the windows builds. The examples were never tested on windows AFAIK.

scripts/ci/linux/setup.sh
hobu: The idea of 'fixed' is a fixed GDAL dependency that doesn't change, which should pin most of the other PDAL dependencies. This needs to specified individually for each setup.sh of all builders.

chambbj:
No real gripes with anything. Just the minor comments.

It would be beneficial to eventually have as much parity as possible between the three builds. I feel like we rely on defaults in some places, but are more explicit in others (e.g., only Windows turned off CPD explicitly). I didn't take the time to check every little switch, but it looks at a glance as though they differ, at least slightly. At a minimum, all should mirror what the conda-forge feedstocks are expected to build I should think.

scripts/ci/linux/setup.sh
hobu: Added floating and fixed to osx and windows

.github/workflows/osx.yml
hobu: done.

scripts/ci/linux/setup.sh
hobu: conda-build removed.

chambbj:
:+1:

PR #3202, opened 2020-08-05
Improve expression support in filters.assign

### LINKED ISSUE #1468: Enhance filters.assign
filters.assign could do more than a single blanket value assignment quite easily.

abellgithub:
Close #1468


abellgithub:
This is unrelated to `where` or `where_merge`. Those only deal with points that skip a stage. This option applies to all points passed to a stage, like all other options.

ahuarte47:
This is a fantastic new feature. Thanks for your work. What about supporting multiple assignments separated by ", "? We could assign dimensions in one unique step

abellgithub:
Any number of assignment value options are supported by placing then in a JSON array in PDAL pipeline syntax.

ahuarte47:
Ok. Thanks. You are right. Using multiple assignments would be more efficient and it would have a more compact style. It was only an idea. Your work here is great

abellgithub:
It's all one thing. See the doc page for more info.

ahuarte47:
> Any number of assignment value options are supported by placing then in a JSON array in PDAL pipeline syntax.

Sorry @abellgithub, I did not explain myself properly, I was asking for an expression with this pattern:
  Dimension = ValueExpression[, Dimension1 = ValueExpression1...] [WHERE ConditionalExpression)]
  Example:
  X=2, Y=3, Z=4 WHERE Class=45

Regards

abellgithub:
This can now be achieved as follows:

```
...
{
  "type" : "filters.assign",
  "value": [
    "X = 2", "Y = 3", "Z = 4"
  ],
  "where" : "Class == 45"
}
...
```

The functionality isn't identical to what you suggest, but I think it's close enough in practical situations to suffice.

ahuarte47:
Ok thanks

PR #3205, opened 2020-08-06
Use Eigen SparseMatrix to store knn indices and distances in LOF filter

chambbj:
Avoiding the repeated knnSearch in each separate pass results in a 3x speed up.

Add several Eigen headers that have been relicensed from LGPL to MPL2 (see https://gitlab.com/libeigen/eigen/-/commit/1dc1677d525b1df3cc32b3df5cd816e150b07185) since our last refresh.


hobu:
IIRC, I had removed the LGPL headers in our vendored Eigen. It is fine to put them back of the licensing was cleaned up.

chambbj:
Unclear to me how extensive it was, but certainly applied to these three.

abellgithub:
Why did you choose SparseMatrix instead of std::map? I don't see any functionality being used beyond what would be in std::map.

chambbj:
No compelling reason. I can share more of the thought process via Slack, but it's a bit meandering. Do you feel strongly that std::map would be more performant or easier (more familiar) to read?

chambbj:
@abellgithub Latest commit uses std::map. Perhaps I'm doing something inefficiently here. It's only about 2x as fast (as opposed to 3x). Let me know what you think.

PR #3206, opened 2020-08-06
Bump eigen to 3.3.7

chambbj:


abellgithub:
As you might expect, I'm not much of a fan of this unless there is a reason. I have in the past spent a bunch of time patching Eigen for various systems and I'd rather not do it again. It may be the case that the fixes I've made have been put into this release (they weren't in the latest code when I last checked), but without needing features missing from our version or addressing bugs that we care about, I'm not sure we should mess with this. I'd be much more of a fan for putting in right AFTER we do a release, if we really think we need this.

hobu:
I suggested making this PR so we could see if tests passed given a new release. If you have made a lot of Eigen patches, have they all be upstreamed? 

The reason to check this is @chambbj's LOF PR adds back some Eigen pieces that were licensed as LGPL and were removed from our tree for being incompatible with our licensing. Adding them back makes our vendored copy a bit of a frankenstein, and it would be best to have our vendored copy be a snapshot instead of a messy, hacked up fork.

PR #3207, opened 2020-08-07
readers.obj

rjpals:
Reads data from obj format. See the docs for more info.


pdal/Dimension.json
```diff
@@ -385,6 +385,18 @@
     "description": "Distance metric related to a point's nearest neighbors."
     },
     {
+    "name": "TextureX",
+    "alt_names": ["tx", "texture_x", "texturex" ],
```
abellgithub, line 389: Just curious if these alternate names came from real OBJ files.

hobu: yes I saw them in outputs.

PR #3211, opened 2020-08-13
Set classification changes only after computation is complete.

abellgithub:


chambbj:
Did we used to allow specification of multiple DimRanges and now it's a single one? Not sure I'm against this, just seems like a change in behavior (that might catch some off guard) and beyond the issue at hand.

abellgithub:
The behavior now matches that of other filters. Multiple `DimRange`s were handled before, just not always correctly.  But I can revert.

chambbj:
I guess the only thing I'm thinking, which is beyond the scope of what I think we're trying to achieve prior to release, is that if we are going to change something, it might be nice to use the new syntax introduced with `where`. I don't know that anybody is needing complex `domain` specification, but I could imagine it might be useful.

PR #3216, opened 2020-08-18
Add max_dist and init parameters to ICP

### LINKED ISSUE #3213: Provide option to reject gross outliers in ICP correspondences
https://github.com/PDAL/PDAL/blob/aaf90bafde6eb0f0d646939412628410e6820d49/filters/IterativeClosestPoint.cpp#L156-L157

This can be setup to only be tested if the user actually provides a value.

### LINKED ISSUE #3214: Provide option for user to provide initial guess at ICP transformation
https://github.com/PDAL/PDAL/blob/aaf90bafde6eb0f0d646939412628410e6820d49/filters/IterativeClosestPoint.cpp#L118-L119

ICP is sensitive to the initial guess, and while identity may work for small transformations, users with the luxury of having some a priori knowledge of what the transformation might be will have better luck if they start the process there.

chambbj:
`max_dist` will reject correspondences that exceed the provided maximum distance.

`init` is a user-provided initial guess at the transformation matrix.

Closes #3213 and closes #3214

### DIFF
--- a/filters/IterativeClosestPoint.hpp
+++ b/filters/IterativeClosestPoint.hpp
@@ -53,8 +53,14 @@ private:
     double m_rotation_threshold;
     double m_translation_threshold;
     double m_mse_abs;
+    Arg *m_maxdistArg;
+    double m_maxdist;
+    Arg *m_matrixArg;
+    std::string m_matrixStr;
+    std::vector<double> m_vec;
 
     virtual void addArgs(ProgramArgs& args);
+    virtual void prepared(PointTableRef table);
     virtual PointViewSet run(PointViewPtr view);
     virtual void done(PointTableRef _);
     PointViewPtr icp(PointViewPtr fixed, PointViewPtr moving) const;
--- a/filters/IterativeClosestPoint.cpp
+++ b/filters/IterativeClosestPoint.cpp
@@ -140,7 +140,7 @@ PointViewPtr IterativeClosestPoint::icp(PointViewPtr fixed,
     // be reasonable to alternately accept an initial guess.
     Matrix4d final_transformation;
     if (m_matrixArg->set())
-        final_transformation = Map<const Matrix4d>(m_vec.data());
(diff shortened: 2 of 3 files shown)


filters/IterativeClosestPoint.hpp
```diff
@@ -53,8 +53,14 @@ class PDAL_DLL IterativeClosestPoint : public Filter
     double m_rotation_threshold;
     double m_translation_threshold;
     double m_mse_abs;
+    Arg *m_maxdistArg;
+    double m_maxdist;
+    Arg *m_matrixArg;
+    std::string m_matrixStr;
```
abellgithub, line 59: I haven't looked at the Eigen code, but I would think that we could map the argument to Matrix4d or some such.  I think it has the necessary operators.  Or perhaps we should pull TransformationFilter::Transform out, rename and use that? Seems using eigen directly would be best, but a mapping also seems OK.

chambbj:
> the Eigen:: namespace qualifier is helpful

It just gets to be a lot to parse. I think Matrix and Vector types are common enough to do without, but I could be convinced to put it back in for Map, IOFormat, and umeyama, which without any context could be quite confusing.

filters/IterativeClosestPoint.hpp
chambbj: I thought I'd run into issues trying to take Matrix4d directly via ProgramArgs in the past. And while I looked at TransformationFilter, it just seemed a bit overkill to copy/paste for this usage. Yes, if it had been generally available, I probably would've gone that route. Let me look again at using Eigen and I'll let you know where I get stuck.

abellgithub:
Fair enough

filters/IterativeClosestPoint.hpp
abellgithub: I wouldn't copy/paste. I'd just extract it into some private space, but better to use Eigen directly if that works. Ultimately, probably better to eigen-ize the transformation filter.

PR #3217, opened 2020-08-20
Filters/relaxation dart throwing

chambbj:
Relaxation dart throwing is similar in nature to the Poisson disk sampling provided by SampleFilter. The key difference is that it will continue to loop, decreasing the minimum radius at each iteration, until the desired number of output points is met.

SampleFilter no longer requires Classification dimension. It also adds the option to shuffle points prior to sampling (the user can seed for more deterministic behavior, default = on).


filters/RelaxationDartThrowing.cpp
```diff
@@ -0,0 +1,157 @@
...
+        std::shuffle(shuffledIds.begin(), shuffledIds.end(),
+                     std::mt19937(m_seed));
+    }
+
+    while (finalIds.size() < m_maxSize)
+    {
+        // All points are marked as kept (1) by default. As they are masked by
+        // neighbors within the user-specified radius, their value is changed to
+        // 0.
+        std::vector<int> keep(np, 1);
```
abellgithub, line 105: Is this supposed to be inside the `if`? Maybe I don't understand.

filters/RelaxationDartThrowing.cpp
```diff
@@ -0,0 +1,157 @@
...
+        // Start by masking all points within radius of finalIds.
+        for (PointId i : finalIds)
+        {
+            PointIdList ids = index.radius(i, radius);
+            for (PointId const& id : ids)
+                keep[id] = 0;
+        }
+
+        // We are able to subsample in a single pass over the shuffled indices.
+        for (PointId const& i : shuffledIds)
```
abellgithub, line 116: Could you maybe get into an infinite loop here? If you don't hit the number of points in `finalIds` that you want, but you have eliminated all the nearby points by setting keep to 0?

chambbj: If you do not hit the target number of points, the mask is reset and the radius decreased. I think the only situations that could get you in trouble are 1) requesting more output points than are in the input, and 2) if there are duplicate points such that shrinking the radius never discriminates them. Both of these feel like edge cases, but we could add some checks on them. Otherwise, I think the worst case is that you (rightly) get back the input point cloud.

abellgithub: I guess I misunderstood what was happening.

This seems very expensive, as a single pass with the `radius` will give you all the information you need: You know all the points within the radius of the target point. The distances are available from nanoflann, though we don't seem to expose that right now. I guess the tradeoff is memory, but if you're actually going to end up relaxing (which is the point of the filter), I would think you're going to overwhelm with CPU, as nanoflann searching is slow.

Also, maybe when you've iterated so many times, you just give up and throw everything in? I get that with most real-life data and good parameters, this probably won't happen, but it also seems like something that someone will file an issue for.

chambbj: You are absolutely right that we could just stash the distances and recall those in subsequent iterations. I'll take a look at this.

abellgithub: I'm fine with merging this. If you want to enhance later, that's fine.

I just wonder about the name change for "filters.farthestpointsampling"

chambbj: I just pushed an update. Let me know what you think. It could probably be prettier or more succinct, but it seems to get the job done.

The name change for FPS should have been a correction in the docs only to match the filter name as implemented.

hobu:
I tried to bounce this to get CI to run, but for some reason it didn't. Anyway, 👍 to merge?

chambbj:
I'm ready, unless @abellgithub still had any unresolved issues with it.

abellgithub:
I never looked at the last changes. I'll do that now.

PR #3220, opened 2020-08-26
Obj texture fixes

rjpals:
Fixes crash on when reading OBJ files with 2 texture vertices. Adds option W dimension for obj files.


io/ObjReader.cpp
```diff
@@ -191,32 +212,66 @@ bool ObjReader::readFace(FACE& face, PointViewPtr view)
...
-                std::stringstream ss;
-                ss << "Could not convert vertex specification to double on line #"
+                std::stringstream errorMessage;
+                errorMessage << "Could not convert vertex specification to double on line #"
                     << lineOfFile << ": '" << line << "'" << std::endl;
-                throwError(ss.str());
+                throwError(errorMessage.str()); 
+            };
+
+            if(numDims == 3) {
```
abellgithub, line 225: This code is repetitive.

io/ObjReader.cpp
```diff
@@ -155,17 +159,34 @@ PointId ObjReader::addPoint(PointViewPtr view, VTN vertex)
...
+
+void ObjReader::newTextureVertex(double x, double y)
+{
+    m_textureVertices.push_back({x, y, 0, 0});
 }
 
 void ObjReader::newTextureVertex(double x, double y, double z)
 {
-    m_textureVertices.push_back({x, y, z});
+    m_textureVertices.push_back({x, y, z, 0});
```
hobu, line 184: default `w` values should probably be 1.0 instead of 0.

hobu:
Since we are not streaming anyway, can we sample a few rows of `v` and `vt` to develop a proper schema in a `prepare()` call? Users want this to just work.

io/ObjReader.cpp
abellgithub: This doesn't get used.

rjpals: [The Wikipedia article on obj](https://en.wikipedia.org/wiki/Wavefront_.obj_file#File_format) says it should be 0. Do you want it to be 1?

```
# List of texture coordinates, in (u, [,v ,w]) coordinates, these will vary between 0 and 1. v, w are optional and default to 0.
vt 0.500 1 [0]
```

Also, in this specific case, it is the fourth element of the texture, which isn't really defined, and is never read.

abellgithub:
I vote "no" on this change. I think it's not a big deal for users to tell us what they want loaded, especially when that data will probably get dropped on the floor, even if it exists. I can't think why someone would want this data in a PDAL context.

pdal-docs:
Are there other reader examples where we make the user tell us what to read via config before we read anything? The user doesn’t know what’s in the file or if they need it. 

The case for having this useless data is if you’re using pdal as a data access library, which many do, you are reading up the data and passing it along to the next thing. It isn’t pdal’s job to decide if the data is useless.

PR #3225, opened 2020-08-30
Add test file garbage_nVariableLength.las

phma:
This file has a floating-point number in nVariableLength, which should
be the integer 0, since there are no variable-length records. It's a
test file to find out why a program I'm writing, which reads a point cloud
and builds an octree, loses points. The code to write LAS files wasn't
working right yet. I read this file into CloudCompare, which allocated
all RAM and was killed; I reported the bug to CloudCompare, but was told
that the bug is in PDAL.


abellgithub:
If you think there's a bug, please submit an issue describing the problem.

abellgithub:
See #3226

PR #3227, opened 2020-08-31
Fix for #3204, Issues with Commas in writers.text GeoJSON output.

### LINKED ISSUE #3204: Invalid Geojson + Feature Request OGR Dimensions
**Describe the bug**
* Issue - PDAL writers.text doesn't write valid geojson. When using "headers":"true" the resulting geojson has an errant comma at the start of the feature array, and is missing commas in between features. "headers":"false" returns features, but without commas between them.
* Enhancement - PDAL writers.ogr has no option to include dimensions. This would potentially solve the issue by providing an alternate method for producing geojson with attributes(dimensions). It would also be better as the coordinate precision would be independent of the other dimension precision without having to list every dimension.

```

$ pdal pipeline ept-hdr.json
```

<details>
  
```
# Using sample data in the PDAL source
# ept-hdr.json
{
  "pipeline": [
    {
        "type": "readers.ept",
        "filename": "/data/ept/ellipsoid-zstandard/ept.json"
    },
    {
        "type":"filters.head",
        "count":100
    },
    {
        "type":"filters.decimation",
        "step": 10
    },
    {
      "type":"filters.reprojection",
      "out_srs":"EPSG:4326"
    },
    {
      "type":"writers.text",
      "format":"geojson",
      "write_header":"true",
      "keep_unspecified":"false",
      "precision":"8",
      "filename": "/data/ept-text.geojson"
    }
  ]
}
```

</details>

**Expected behavior**
Valid geojson. Checked with geojson-validation and QGIS.
Attaching what comes out of pdal and a fix version. You'll see the incorrect version has a leading comma, and then is missing commas.
[ept-text.zip](https://github.com/PDAL/PDAL/files/5031654/ept-text.zip)

The extra comma appear to come from https://github.com/PDAL/PDAL/blob/95be31eb320f6af0eb4d605cbbe33e50fcdfc770/io/TextWriter.cpp#L244
The missing commas, probably from this if statement not passing correctly https://github.com/PDAL/PDAL/blob/95be31eb320f6af0eb4d605cbbe33e50fcdfc770/io/TextWriter.cpp#L279
Another data set I tested with did have commas between records, but it's too big to provide here.

Please provide information on your PDAL version (`pdal --version`) and system (e.g., `uname -a` or `ver`).

```
docker pdal/pdal:latest
```

```
$ pdal --version
pdal 2.1.0 (git-version: c99ac0)
```

bitner:
Fixes #3204

- Removes comma that is inserted prior to the first feature.
- Adds comma before all but the first feature.


abellgithub:
Thanks for this. Can you please provide a test to go with it?

bitner:
@abellgithub Yep, that's why it's marked in progress :-)

abellgithub:
Thanks. Let me know when you think it's complete.

bitner:
@abellgithub should be good to go now. test pushed and passing

io/TextWriter.cpp
```diff
@@ -276,7 +275,7 @@ void TextWriter::processOneCSV(PointRef& point)
 
 void TextWriter::processOneGeoJSON(PointRef& point)
 {
-    if (m_idx)
+    if (m_idx > 0)
```
abellgithub, line 278: I don't understand this change.

PR #3230, opened 2020-09-01
RasterWriter

abellgithub:


doc/stages/writers.raster.rst
```diff
@@ -0,0 +1,91 @@
+.. _writers.raster:
```
hobu, line 1: We need to add writers.raster to the doc/stages/writers.rst

PR #3237, opened 2020-09-08
Check for bad LAS header values where possible.

abellgithub:


io/LasHeader.cpp
```diff
@@ -381,12 +388,22 @@ ILeStream& operator>>(ILeStream& in, LasHeader& h)
             in >> h.m_pointCountByReturn[i];
     }
 
+    if (!h.compressed() && h.m_pointOffset > h.m_fileSize)
+        throw LasHeader::error("Invalid point offset - exceeds file size.");
+    if (!h.compressed() && h.m_pointOffset + h.m_pointCount * h.m_pointLen > h.m_fileSize)
+        throw LasHeader::error("Invalid point count. Number of points exceeds file size.");
```
hobu, line 394: Can we add a switch to suppress this that's off-by-default? In the past we could still read files with bad point counts that were uncompressed. They are more rare than they used to be, but they are still out there.

io/LasReader.cpp
```diff
@@ -157,6 +158,20 @@ void LasReader::handleCompressionOption()
     m_compression = compression;
 }
 
+void LasReader::createStream()
+{
+    if (m_streamIf)
+        std::cerr << "Attempt to create stream twice!\n";
```
hobu, line 164: lint?

io/LasHeader.cpp
abellgithub: I suppose. :(

PR #3241, opened 2020-09-14
Add minimal test for LloydKMeans filter and switch back to RowPointTable

chambbj:


abellgithub:
I misspoke earlier when I said the layout should be finalized. The table itself should be finalized. Perhaps the finalize() call in PointLayout should be hidden.

PR #3250, opened 2020-09-21
Remove double loop in filters.csf

abellgithub:


chambbj (approved):
Oh, that's embarrassing.

PR #3251, opened 2020-09-22
Add Alpine and Docker GHA builders

hobu:


hobu:
We should have a discussion about automated pushing of Docker containers to both DockerHub and ghcr.io.

PR #3263, opened 2020-10-15
Update CMakeLists.txt and add missing includes

diddledan:
Compiling PDAL as part of OpenDroneMap on Ubuntu 18.04 I found the following were required changes:

* Add missing `#include` for required external headers in:
  - `io/Ilvis2MetadataReader.cpp` (`cpl_conv.h`)
  - `pdal/Geometry.cpp` (`cpl_conv.h` and `cpl_string.h`)
  - `pdal/private/gdl/GDALUtils.cpp` (`ogr_p.h`)

Signed-off-by: Daniel Llewellyn <daniel@snapcraft.ninja>

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -23,7 +23,6 @@ target_link_libraries(${PDAL_APP}
         ${PDAL_UTIL_LIB_NAME}
         ${GDAL_LIBRARY}
         ${WINSOCK_LIBRARY}
-        ${LASZIP_LIBRARIES}
 )
 target_include_directories(${PDAL_APP} PRIVATE
     ${NLOHMANN_INCLUDE_DIR}
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -23,6 +23,7 @@ target_link_libraries(${PDAL_APP}
         ${PDAL_UTIL_LIB_NAME}
         ${GDAL_LIBRARY}
         ${WINSOCK_LIBRARY}
+        ${LASZIP_LIBRARIES}
 )
 target_include_directories(${PDAL_APP} PRIVATE
     ${NLOHMANN_INCLUDE_DIR}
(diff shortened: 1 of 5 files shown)


abellgithub:
I'm not sure what you're doing, but these changes, in general, shouldn't be necessary. It would be helpful to have the link line and failures.

diddledan:
I'm just rebuilding in a clean environment to re-verify that those changes are all required for me. I'm not 100% certain, so I'm going to take the time to double check. Thanks for poking me quickly :-)

diddledan:
OK, from a clean build without the added header `#includes` on Ubuntu 18.04, I'm getting the following undefined entities reported:

- Function `CPLFree()` from missing `<cpl_conv.h>` in:
  - `io/Ilvis2MetadataReader.cpp`
  - `pdal/Geometry.cpp`
- Type `CPLStringList` from missing `#include <cpl_string.h>` in:
  - `pdal/Geometry.cpp`
- Function `OGRWktReadToken()` from missing `#include <ogr_p.h>` in:
  - `pdal/private/gdal/GDALUtils.cpp`

I'm rebuilding now in another clean environment without the linker changes to see if that definitely fails, and to see if I can get the linker command it used if it does still fail.

diddledan:
OK. the linker stuff (the changes to CMakeLists.txt) are not required. I'll reroll the PR to just the headers

abellgithub:
We build PDAL all the time on 18.04 with no problem, so we'd need to understand why you're having an issue. In particular, it would be important to know the version/installation of GDAL you're trying to use.

abellgithub:
I'll add that on the conda installation of GDAL 3.1.2, ogr_geometry.h includes cpl_conv.h

diddledan:
between you and the ODM folks I think you've both hit the nail - I was using GDAL 2.2.3 from the default Ubuntu 18.04 repository instead of using the Ubuntu-GIS team's unstable PPA/repository at https://launchpad.net/~ubuntugis/+archive/ubuntu/ubuntugis-unstable I'll rebuild with this PPA which appears to have 3.0.4 for Ubuntu 18.04 and report back, or close the issue if that fixes things.

diddledan:
Yes, that was the problem - the UbuntuGIS PPA resolved the issue.

PR #3266, opened 2020-10-15
default to unordered reads with TileDB

normanb:
Default to not ordering the result data on read with TileDB to avoid unnecessary overhead, ordering should be done with a subsequent pipeline stage.


abellgithub:
Please recreate with a pull request to the master branch.

PR #3267, opened 2020-10-15
Default to unordered reads with TileDB plugin

normanb:
@abellgithub opened this against master as asked.

Default to not ordering the result data on read with TileDB to avoid unnecessary overhead, ordering should be done with a subsequent pipeline stage.


abellgithub:
@normanb : I'll merge this when we get CI working.

Normally, we only backport bug fixes to the maintenance branches. Does this count as a bug that needs to be backported?

PR #3270, opened 2020-10-21
Minor updates to several filters

chambbj:
Initial focus is on updating static filters to always register official Dimensions.

* LiTreeFilter now registers existing ClusterID instead of custom TreeID.

* LOFFilter now registers existing NNDistance instead of custom KDistance.

* OptimalNeighborhoodFilter now registers existing OptimalRadius and OptimalKNN.

* Register the following as "official" Dimensions: Coplanar, Eigenvalue0, Eigenvalue1, Eigenvalue2, LocalOutlierFactor, LocalReachabilityDistance, Miniball, PlaneFit, RadialDensity, Rank, Reciprocity

While updating Dimensions, also took the opportunity to:

* Be more explicit about datatypes, instead of relying on auto, to improve readability.

* Use PointView range-based for loops where it make sense. Generally makes code a little more succinct.

* Do not pass KDIndex as frequently, when it can typically be grabbed off the view at no cost.

### DIFF
--- a/doc/stages/filters.litree.rst
+++ b/doc/stages/filters.litree.rst
@@ -17,8 +17,8 @@ given a ``ClusterID`` of 0.
 
 .. note::
 
-  In earlier PDAL releases, ``ClusterID`` was stored in the ``TreeID``
-  Dimemsion.
+  In earlier PDAL releases (up to v2.2.0), ``ClusterID`` was stored in the
+  ``TreeID`` Dimemsion.
 
 .. embed::
 
--- a/doc/stages/filters.litree.rst
+++ b/doc/stages/filters.litree.rst
@@ -6,14 +6,19 @@ filters.litree
 
 The purpose of the Li tree filter is to segment individual trees from an input
 ``PointView``. In the output ``PointView`` points that are deemed to be part of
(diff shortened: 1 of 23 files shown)


doc/stages/filters.litree.rst
```diff
@@ -6,15 +6,20 @@ filters.litree
...
 
 .. note::
 
   The filter differs only slightly from the paper in the addition of a few
   conditions on size of tree, minimum height above ground for tree seeding, and
   flexible radius for non-tree seed insertion.
 
+.. note::
+
+  In earlier PDAL releases, ``ClusterID`` was stored in the ``TreeID``
```
hobu, line 20: Adjust this to say versions less than or equal to 2.2.0 or something.

doc/stages/filters.lof.rst
```diff
@@ -32,6 +32,11 @@ users of this filter should find instructive.
   To inspect the newly created, non-standard dimensions, be sure to write to an
   output format that can support arbitrary dimensions, such as BPF.
 
+.. note::
+
+  In earlier PDAL releases, ``NNDistance`` was stored in the ``KDistance``
```
hobu, line 37: same as above

hobu (approved):

PR #3272, opened 2020-10-21
Set our standard to C++17

hobu:
We almost did this in 2.1.0, but OSX was lagging behind. The latest clang for OSX seems to build fine with C++ 17 standard being set.


abellgithub:
I don't see any real benefit to this.

PR #3275, opened 2020-10-27
Make Poisson sampling streamable

chambbj:
Still want to understand an artifact that I don't quite understand, but this is otherwise ready for review.

### DIFF
--- a/doc/stages/filters.sample.rst
+++ b/doc/stages/filters.sample.rst
@@ -8,39 +8,76 @@ The practice of performing Poisson sampling via "Dart Throwing" was introduced
 in the mid-1980's by [Cook1986]_ and [Dippe1985]_, and has been applied to
 point clouds in other software [Mesh2009]_.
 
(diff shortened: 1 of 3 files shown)


nicolas-chaulet (commented):
If I understand correctly, you maintain a set of occupied voxels that is growing linearly with the amount of data that you keep in the output point cloud. We had the same situation with the voxel filter that was turned into a semi streamable version. It might be worth mentioning that in the doc to avoid misunderstanding and people getting confused because the memory footprint of this stage could be quite large despite being streamable.

chambbj:
That is correct. The occupied voxels will grow as the space covered by the point cloud is enlarged while iterating. For massive datasets (or very small voxel size!), this could be quite a large memory footprint. I'll see if I can come up with some language to convey this.

abellgithub:
@chambbj : Can you please merge master into this if you still want the change merged? Thx.

chambbj:
Just rebased and squashed for a much cleaner commit @abellgithub

PR #3281, opened 2020-11-05
Fix bounds for faux test.

### LINKED ISSUE #3280: TEST(FauxReaderTest, uniform) broken
The test calls [1]

    ops.add("bounds", BOX3D(0, 100, 0, 100, 0, 100));

I suppose assuming that the `BOX3D` constructor takes the arguments `(minx, maxx, miny, maxy, minz, maxz)`, but in fact the `BOX3D` constructor takes the arguments in the order `(minx, miny, minz, maxx, maxy, maxz)`. This leads to the box being constructed with `miny=100, maxy=0`, which causes the test to fail:

    /usr/include/c++/10/bits/random.h:1751: std::uniform_real_distribution<_RealType>::param_type::param_type(_RealType, _RealType) [with _RealType = double]: Assertion '_M_a <= _M_b' failed.

[1] https://github.com/PDAL/PDAL/blob/master/test/unit/io/FauxReaderTest.cpp#L244

abellgithub:
Close #3280

### DIFF
--- a/test/unit/io/FauxReaderTest.cpp
+++ b/test/unit/io/FauxReaderTest.cpp
@@ -241,7 +241,7 @@ TEST(FauxReaderTest, uniform)
 {
     Options ops;
 
-    ops.add("bounds", BOX3D(0, 100, 0, 100, 0, 100));
+    ops.add("bounds", BOX3D(0, 0, 0, 100, 100, 100));
     ops.add("count", 1000);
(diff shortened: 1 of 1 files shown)


hobu (approved):
Curious why this hasn't bubbled up before.

abellgithub:
Because the algorithm is consistent and works on all platforms despite the fact that the standard says it is UB.

PR #3292, opened 2020-11-23
Fixes for VS 2019

abellgithub:


io/private/ept/Key.hpp
```diff
@@ -52,10 +52,10 @@ class PDAL_DLL Key
         const StringList tokens(Utils::split(s, '-'));
         if (tokens.size() != 4)
             throw pdal_error("Invalid EPT KEY: " + s);
-        d = std::stoull(tokens[0]);
-        x = std::stoull(tokens[1]);
-        y = std::stoull(tokens[2]);
-        z = std::stoull(tokens[3]);
+        d = std::stoi(tokens[0]);
```
hobu, line 55: @connormanning is this ok?

pdal/private/gdal/Raster.cpp
```diff
@@ -32,8 +32,11 @@
 * OF SUCH DAMAGE.
 ****************************************************************************/
 
+#pragma warning(push)
```
hobu, line 35: Rather than sprinkling this all over the place, maybe we should make our own private GDAL header that does this and whatever other junk we want.

hobu (approved):

chambbj (approved):

io/private/ept/Key.hpp
connormanning: Yeah.

connormanning (approved):

PR #3296, opened 2020-11-25
Add missing libc includes

theirix:
Add missing includes (reproduced on old clangs)

### DIFF
--- a/filters/private/expr/Lexer.cpp
+++ b/filters/private/expr/Lexer.cpp
@@ -1,6 +1,7 @@
 
 #include <cassert>
 #include <cctype>
+#include <cstdlib>
 
 #include "Lexer.hpp"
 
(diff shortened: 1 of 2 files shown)


abellgithub:
It would help to know the specific compiler/system where this was a problem for you.

theirix:
Sure!

cstdexcept is missing when building with gcc 4.9 on Linux x86_64:
```
pdal/util/Bounds.hpp:53:59: error: expected class-name before '(' token
         error(const std::string& err) : std::runtime_error(err)
```

cstdlib is missing for clang 4.0 on Linux x86_64:
```
filters/private/expr/Lexer.cpp:203:16: error: use of undeclared identifier 'strtod'; did you mean 'strtok'?
    double v = strtod(start, &end);
               ^~~~~~
               strtok
/usr/include/string.h:335:14: note: 'strtok' declared here
extern char *strtok (char *__restrict __s, const char *__restrict __delim)
```

abellgithub:
Thanks. Just note that we may not support such old compilers in the future if it takes much effort.

PR #3300, opened 2020-11-30
Don't use conda compiler

abellgithub:


hobu:
huh. 

Is the conda compiler stack too old or too new?

abellgithub:
The conda compilers are for use in building conda packages. We don't need them for CI.

hobu (approved):

PR #3303, opened 2020-12-01
Allow overlay filter to use 64 bit ints

### LINKED ISSUE #3302: filters.overlay can't deal with 64 bit integers
**Describe the bug**
I am reading a LAS 1.2 file (laz compressed), trying to overlay it with a geopackage with 64bit integer ids, and then save it out to a LAZ again. See also this [mailing list thread](https://lists.osgeo.org/pipermail/pdal/2020-November/002315.html).
 
```
$ pdal pipeline overlay.json
```

<details>
  
```
 [
     "/data/input.LAZ",
     {
       "type":"filters.range",
       "limits":"Classification[6:6]"
     },
     {
       "type":"filters.ferry",
       "dimensions":"=> BldId"
      },
     {
       "type":"filters.overlay",
       "dimension":"BldId",
       "datasource":"/data/all_tiles_extract.gpkg",
       "column":"bld_id", (gdal reads as Integer64)
       "layer":"layer_name" 
     },
     {
       "type":"filters.range",
       "limits":"BldId[1:18446744073709551615]"
     },
      {
         "type":"writers.las",
         "filename":"/data/blds_filter_then_overlaylaz",
         "extra_dims":"all",
         "minor_version":4
     }
 ]
```

</details>

**Expected behavior**
The int64 ids should transfer to the Dimension BldId. In fact the maximum of this dimension is shown as 2147483647, the signed 32 bit integer max.

PDAL 2.2 release

stev-0:
Fixes #3302
I noticed the minimum version of GDAL is 2.2, the [GetFieldAsInteger64](https://gdal.org/api/vector_c_api.html#_CPPv425OGR_F_GetFieldAsInteger6411OGRFeatureHi) is since 2.0

Discussion is [here](https://lists.osgeo.org/pipermail/pdal/2020-November/002315.html)

This is my first proper PR so excuse me if I don't know the ropes!


abellgithub:
This won't work in general, though it may work in your case because the PDAL dimension may not be a 64-bit integer. In order for this to work properly, the dimension needs to be created with the same type as the value in the database. This isn't currently happening. But it would be great if we could be flexible. Can we figure the type of the value from the database and register the dimension correspondingly?

A 64-bit value seems kind of strange, too, but who knows...

stev-0:
Ah, OK, thanks @abellgithub  I was assuming that since the default for new dimensions is a double (there is an issue somewhere that suggests making this configurable) then it would be OK (and it is in my case as you say). But there are two problems with my approach:

- Some bigints won't fit into doubles
- Some (most) people might want to use existing smaller int dimensions (although presumably this doesn't currently work for int8/16 fields?), I think I remember coming across that issue somehow

So yes this will need revising.

Any thoughts on the general test approach? And is there a test coverage gap if this is passing?

The use case by the way is to split out the point cloud by building, which often have these kind of high ids. OSM ids for example are int64, and have been for [7 years](https://wiki.openstreetmap.org/wiki/64-bit_Identifiers)

abellgithub:
Please re-open if issues are addressed.

Issue #3304, opened 2020-12-02
LasWriter: error autoscale with autooffset (0 variance data)


matyas-hollmann:
I have encountered a problem when trying to write LAS files using PDAL LasWriter when autoscale and autooffset are enabled. Problem is that the "scale" value computed for a spatial dimension with 0 variance in the data is 0, which causes problems when the points are converted (from a floating point to signed 32-bit integer) raising an error: 

> Unable to convert scaled value (NaN) to int32 for dimension 'Z' when writing LAS/LAZ file

The conversion is done using this formula: `(val - m_offset.m_val) / m_scale.m_val;` which in my case converts a perfectly valid value to `NaN`, which cannot be represented as a 32-bit signed integer.

matyas-hollmann:
Resolved in #3305.

PR #3305, opened 2020-12-02
LasWriter: proposed solution to autoscale with autooffset bug (0 variance data)

matyas-hollmann:
Proposed solution addressing https://github.com/PDAL/PDAL/issues/3304. I have also added 1 additional unit test which is addressing a case with 0 variance in the 'Z' spatial dimension.

Previously the calculation of (auto)"scale" value didn't take into account the "offset" which could cause a loss of precision. The calculated (auto)"scale" could also end up being 0 (e.g. maximum value in a spatial dimension was 0 or autooffset was enabled and there was 0 variance in a spatial dimension).

### DIFF
--- a/pdal/Scaling.cpp
+++ b/pdal/Scaling.cpp
@@ -92,26 +92,33 @@ void Scaling::setAutoXForm(const PointViewSet& pvSet)
     }
 
     if (m_xXform.m_offset.m_auto)
-    {
         m_xXform.m_offset.m_val = xmin;
-        xmax -= xmin;
-    }
     if (m_yXform.m_offset.m_auto)
-    {
         m_yXform.m_offset.m_val = ymin;
-        ymax -= ymin;
-    }
     if (m_zXform.m_offset.m_auto)
-    {
         m_zXform.m_offset.m_val = zmin;
-        zmax -= zmin;
-    }
+
     if (m_xXform.m_scale.m_auto)
-        m_xXform.m_scale.m_val = xmax / (std::numeric_limits<int>::max)();
+    {
+        xmax = (std::max)(std::abs(xmax - m_xXform.m_offset.m_val),
+                          std::abs(xmin - m_xXform.m_offset.m_val));
+        m_xXform.m_scale.m_val
+            = xmax ? xmax / (std::numeric_limits<int>::max)() : 1.;
+    }
     if (m_yXform.m_scale.m_auto)
-        m_yXform.m_scale.m_val = ymax / (std::numeric_limits<int>::max)();
+    {
+        ymax = (std::max)(std::abs(ymax - m_yXform.m_offset.m_val),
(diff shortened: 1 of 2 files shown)


abellgithub:
I don't understand this, exactly. If the issue is that the max could be 0, leading to a scale of 0, then isn't the check for that sufficient without the other change? What else are you trying to fix?

matyas-hollmann:
> I don't understand this, exactly. If the issue is that the max could be 0, leading to a scale of 0, then isn't the check for that sufficient without the other change? What else are you trying to fix?

Maybe I should have created a separate issue for that. The obvious problem of the previous calculation of 'scale' value is that it didn't take into account offset (in case offset was manually specified). Imagine 1D points: [-255, -128, 1] and user-provided 'offset' of let's say `-128`. If I didn't provide the additional changes the 'scale' would be set to `1/INT_MAX` which would raise an exception when writing the first point (-255) as it couldn't fit value `((-255 + 128) * INT_MAX)` into a 32-bit signed integer. 

> C++ exception with description "writers.las: Unable to convert scaled value (-2.727304232e+11) to int32 for dimension 'X' when writing LAS/LAZ file

matyas-hollmann:
My other concern which is not part of the proposed fix is that when using autoscale with autooffset spatial dimension data are stored only as positive integers. We could increase precision by utilizing the full range of `int` not shift everything into the positive half-plane.

abellgithub:
I'm with you here, but how would you do that last bit? Negative values are growing left and positive values are growing right.  What you've already done in this PR is made it so that it handles the offset reasonably. You're taking advantage of the full range to the left or right of 0, which seems all you can do with a factor.

(I noticed right away this morning that it wasn't messing with the non-auto offset and wondered why, but that code is years' old.)

abellgithub:
Perhaps you're talking about the auto offset case where we can choose our offset, in which case our data should be centered at 0, if I'm understanding your point.

matyas-hollmann:
Yes, I was talking about the auto offset feature where currently the offset value is set to the minimum which is from my POV suboptimal (possible loss of precision).

abellgithub:
Do you want to address the above by changing the assignment to auto offsets something like:
```
m_xXform.m_offset.m_val = -(xmax - xmin) / 2;
```

matyas-hollmann:
I did push another commit addressing auto offset. I think it's better to offset by the middle value. 
`m_xXform.m_offset.m_val = 0.5 * xmin + 0.5 * xmax;`
By splitting the calculation (first halve then add) we can avoid potentially getting `+-Inf`  double value as the offset.

abellgithub:
Not sure how you'd get `Inf` out of this math no matter, but this is fine. Thanks!

abellgithub:
Well, my wrong math wouldn't overflow 😆 , which is why I couldn't see worrying about it.

Thanks again.

PR #3326, opened 2021-01-07
Replace non-standard endian functions with alternative implementations

giordano:
`htonll` and `ntohll` aren't part of any standard and aren't provided by MinGW,
which prevents cross-compiling the library for Windows using this runtime.  For
the alternative implementations, see
https://stackoverflow.com/a/49167531/2442087.

### DIFF
--- a/pdal/util/portable_endian.hpp
+++ b/pdal/util/portable_endian.hpp
@@ -65,9 +65,9 @@
 #       define be32toh ntohl
 #       define le32toh(x) (x)
                         
-#       define htobe64 htonll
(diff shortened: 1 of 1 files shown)


abellgithub:
If you want to address this, you should make an entry specifically for mingw.

giordano:
The patch in #3328 was definitely wrong, but what's the problem with providing a consistent definition which would work with more compilers?

abellgithub:
I have no interest in changing working code for MinGW when this file specifically allows you to specialize for various targets.

abellgithub:
Fixed by #3532

PR #3327, opened 2021-01-07
Use consistent lowercase for Windows header files

giordano:
Windows header files are used with inconsistent cases in the codebase.  When
building on a case-insensitive file system (common when doing native compilation
on Windows) this doesn't matter, but when using case-sensitive file
system (e.g., when cross-compiling on Linux) this leads to errors.

### DIFF
--- a/pdal/DynamicLibrary.cpp
+++ b/pdal/DynamicLibrary.cpp
@@ -38,7 +38,7 @@
 // The original work was released under the Apache License v2.
 
 #ifdef _WIN32
-  #include <Windows.h>
+  #include <windows.h>
 #else
   #include <dlfcn.h>
 #endif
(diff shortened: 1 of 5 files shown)


hobu:
This PR fails on our MSVC builder with 
```
/FoCMakeFiles\pdalcpp.dir\filters\PoissonFilter.cpp.obj /FdCMakeFiles\pdalcpp.dir\ /FS -c ..\filters\PoissonFilter.cpp
2021-01-07T18:37:04.2757488Z D:\a\PDAL\PDAL\vendor\kazhdan\MemoryUsage.h(35): fatal error C1083: Cannot open include file: 'ssapi.h': No such file or directory
```

giordano:
Oops, that was a typo, I fixed that.  That file isn't apparently used in my build, so I didn't notice the typo when building.

PR #3298, opened 2020-11-29
Add vertical ellipsoid height wkt1

msmitherdc:
Source problem for LAS 1.4
```
(pdal pipeline writers.las Error) GDAL failure (1) PROJ: proj_as_wkt: Projected 3D CRS can only be exported since WKT2:2019
PDAL: Couldn't convert spatial reference to WKT version 1.
```

requires:
- PROJ 7.2.1
- GDAL 3.3.0

PR #3330, opened 2021-01-08
Update #3298 to fix docker GHA builder

hobu:


abellgithub:
@hobu: Is this good to go?

PR #3334, opened 2021-01-13
Strip trailing NULL bytes from WKT.

### LINKED ISSUE #3332: readers.las doesn't ignore excess NULL bytes in WKT
If a LAS file contains WKT padded with numerous NULL bytes, PDAL doesn't remove the trailing NULL bytes. This doesn't usually cause a processing problem, but those bytes may be displayed and seem "strange".

abellgithub:
Close #3332

### DIFF
--- a/io/LasHeader.cpp
+++ b/io/LasHeader.cpp
@@ -291,7 +291,10 @@ void LasHeader::setSrsFromWkt()
     const char *c = vlr->data() + len - 1;
     if (*c == 0)
         len--;
-    m_srs.set(std::string(vlr->data(), len));
+    std::string wkt(vlr->data(), len);
+    // Strip any excess NULL bytes from the WKT.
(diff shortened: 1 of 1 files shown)


io/LasHeader.cpp
```diff
@@ -291,7 +291,10 @@ void LasHeader::setSrsFromWkt()
     const char *c = vlr->data() + len - 1;
     if (*c == 0)
         len--;
-    m_srs.set(std::string(vlr->data(), len));
+    std::string wkt(vlr->data(), len);
+    // Strip any excess NULL bytes from the WKT.
+    wkt.erase(std::find(wkt.begin(), wkt.end(), '\0'), wkt.end());
```
hobu, line 296: Do we have a util that does this?

abellgithub: Don't think so. It's one line.

PR #3337, opened 2021-01-18
Draco Reader and Writer

kylemann16:
Adds reader and writer for the Draco compression library to PDAL. 

## Arguments
 - dimensions - Adds the dimension argument. Allows users to specify the data type they would like the PDAL dimension to be created with. These are in string form and are matched with pdal type naming. If one of X, Y, Z are specified all must be specified. The same follows for Textures, Normals, and RGB. These must also all be specified as the same type.
 - quantization - Specifies the quantization level for the given Draco geometry attribute to be used during compression.
```
{
        "type": "writers.draco",
        "filename": "draco.drc",
        "dimensions": {
            "X": "double",
            "Y": "double",
            "Z": "double",
            "ReturnNumber": "uint16"
        },
        "quantization": {
            "POSITION": 2,
            "NORMAL": 8,
            "TEX_COORD": 7,
            "COLOR": 6,
            "GENERIC": 2
        }
}
```

If `quantization` argument isn't used we use the default quantization levels specified by Draco in their readme.
```
    std::map<std::string, int> m_quant =
    {
        { "POSITION",  11 },
        { "NORMAL",     7 },
        { "TEX_COORD", 10 },
        { "COLOR",      8 },
        { "GENERIC",    8 }
    };
```

Draco knows some attributes, `POSITION`(XYZ), `NORMAL`(NormalXYZ), `COLOR`(RGB), `TEX_COORD`(TextureUVW). The reader and writer manage non-Draco specific types by making use of Draco's `GENERIC` type that applies a name for the attribute in the metadata along with its data type.


abellgithub:
I'm sure I'll get slapped, but using JSON as argument handling is non-standard and a PITA when invoking from the command line. In the case of dimensions, it's also different from similar arguments in other stages.

kylemann16:
@abellgithub I tried to copy the json usage word for word from the HdfReader because it was already implemented/had a working example.

hobu:
> I'm sure I'll get slapped, but using JSON as argument handling is non-standard and a PITA when invoking from the command line

PDAL's argument handling doesn't support complex trees of related information. It encourages every driver coming up with its own custom syntax to support such things. In my opinion, that's worse.

abellgithub:
Well, I disagree in principle, but also I think JSON sucks on the command line and we already have tons of stuff that doesn't use JSON. We're now making a mishmash which seems worse than either. But I'm getting tired of this discussion, so I'll be quiet.

CMakeLists.txt
```diff
@@ -274,6 +276,7 @@ target_include_directories(${PDAL_BASE_LIB_NAME}
         ${NLOHMANN_INCLUDE_DIR}
         ${GDAL_INCLUDE_DIR}
         ${LASZIP_INCLUDE_DIR}
+        ${draco_INCLUDE_DIR}
```
hobu, line 279: We don't want to leak this into our public interface. The target_include_directories should be set for every Draco target that needs this to `PRIVATE`.

plugins/draco/io/DracoReader.cpp
```diff
@@ -0,0 +1,363 @@
...
+        const AttributeMetadata* attr_metadata = m_pc->GetAttributeMetadataByAttributeId(at_id);
+        if (attr_metadata)
+        {
+            attr_metadata->GetEntryString("name", &name);
+        }
+
+        switch (at)
+        {
+            case GeometryAttribute::POSITION:
+                CopyAttributeData<double>(m_count, attr, m_positions);
```
hobu, line 277: Windows complains

```
2021-01-20T22:12:28.1465534Z ..\plugins\draco\io\DracoReader.cpp(277): warning C4244: 'argument': conversion from 'pdal::point_count_t' to 'int', possible loss of data
2021-01-20T22:12:28.7098803Z ..\plugins\draco\io\DracoReader.cpp(280): warning C4244: 'argument': conversion from 'pdal::point_count_t' to 'int', possible loss of data
2021-01-20T22:12:29.4432981Z ..\plugins\draco\io\DracoReader.cpp(283): warning C4244: 'argument': conversion from 'pdal::point_count_t' to 'int', possible loss of data
2021-01-20T22:12:30.0638888Z ..\plugins\draco\io\DracoReader.cpp(285): warning C4244: 'argument': conversion from 'pdal::point_count_t' to 'int', possible loss of data
2021-01-20T22:12:33.5287934Z [251/549] C:\PROGRA~2\MICROS~1\2019\ENTERP~1\VC\Tools\MSVC\1428~1.293\bin\Hostx64\x64\cl.exe  /nologo /TP -DNOMINMAX -DPDAL_DLL_EXPORT=1 -DWIN32
```

plugins/draco/io/DracoWriter.cpp
```diff
@@ -0,0 +1,455 @@
...
+    //create geometry attribute and add to pointcloud
+    draco::GeometryAttribute ga;
+    ga.Init(t, nullptr, n, dataType, false, draco::DataTypeLength(dataType) * n, 0);
+    m_attMap[t] = m_pc->AddAttribute(ga, true, m_pc->num_points());
+}
+
+void DracoWriter::initPointCloud(point_count_t size)
+{
+    //begin initialization of the point cloud with the point count
+    m_pc->set_num_points(size);
```
hobu, line 290: Windows complains:

```
2021-01-20T22:12:34.1171568Z ..\plugins\draco\io\DracoWriter.cpp(290): warning C4244: 'argument': conversion from 'pdal::point_count_t' to 'unsigned int', possible loss of data
```

plugins/draco/io/DracoWriter.cpp
```diff
@@ -0,0 +1,455 @@
...
+    //initialize pointcloud builder
+    initPointCloud(view->size());
+    //we don't know how many texture dimensions are available until here
+    parseDimensions();
+
+    PointRef point(*view, 0);
+    for (PointId idx = 0; idx < view->size(); ++idx)
+    {
+        point.setPointId(idx);
+        const auto pointId = draco::PointIndex(idx);
```
hobu, line 365: Windows complains:

```
2021-01-20T22:12:34.1172288Z ..\plugins\draco\io\DracoWriter.cpp(365): warning C4244: 'argument': conversion from 'pdal::PointId' to 'ValueTypeT', possible loss of data
2021-01-20T22:12:34.1173277Z         with
2021-01-20T22:12:34.1173545Z         [
2021-01-20T22:12:34.1174092Z             ValueTypeT=uint32_t
2021-01-20T22:12:34.1174387Z         ]
2021-01-20T22:12:34.1174912Z D:\a\PDAL\PDAL\plugins\draco\io\DracoWriter.hpp(114): warning C4244: 'argument': conversion from 'pdal::PointId' to 'ValueTypeT', possible loss of data
2021-01-20T22:12:34.1175366Z         with
2021-01-20T22:12:34.1175677Z         [
2021-01-20T22:12:34.1176082Z             ValueTypeT=uint32_t
2021-01-20T22:12:34.1176692Z         ]
2021-01-20T22:12:34.1177323Z ..\plugins\draco\io\DracoWriter.cpp(318): note: see reference to function template instantiation 'void pdal::DracoWriter::addToPointCloud<int8_t>(int,pdal::Dimension::IdList,pdal::PointRef &,pdal::PointId)' being compiled
```

cmake/draco.cmake
```diff
@@ -0,0 +1,20 @@
+#
+option(WITH_DRACO "Choose if Draco support should be built" TRUE)
+find_package(Draco QUIET 1.3.6)
+
+if (WITH_DRACO)
+    set_package_properties(Draco PROPERTIES TYPE RECOMMENDED
+        PURPOSE "Provides Draco compression")
+    if(Draco_FOUND)
+        include_directories(${draco_INCLUDE_DIR})
+        set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES}
```
hobu, line 10: I'm not sure we should be doing this.

cmake/draco.cmake
```diff
@@ -0,0 +1,20 @@
+#
+option(WITH_DRACO "Choose if Draco support should be built" TRUE)
+find_package(Draco QUIET 1.3.6)
```
hobu, line 3: We should set this to EXACT match for the moment. Once we get things working, we're going to have to go to the Draco team to fix up their CMake layout for 1.4.x

plugins/draco/io/DracoWriter.cpp
```diff
@@ -0,0 +1,455 @@
...
+#include "DracoWriter.hpp"
+
+
+namespace pdal {
+
+static PluginInfo const s_info
+{
+    "writers.draco",
+    "Write data using Draco.",
+    "http://pdal.io/stages/writers.draco.html"
```
hobu, line 50: Doc is missing

plugins/draco/io/DracoReader.cpp
```diff
@@ -0,0 +1,363 @@
...
+#include "DracoReader.hpp"
+
+#include <draco/io/point_cloud_io.h>
+#include <draco/compression/decode.h>
+
+namespace pdal {
+
+static PluginInfo const s_info
+{
+        "readers.draco",
```
hobu, line 47: Doc is missing

plugins/draco/io/DracoReader.cpp
```diff
@@ -0,0 +1,363 @@
...
+    draco::Decoder decoder;
+    auto geom_status = draco::Decoder::GetEncodedGeometryType(&m_draco_buffer);
+    if (!geom_status.ok())
+    {
+        return throwError(geom_status.status().error_msg());
+    }
+
+    const draco::EncodedGeometryType geom_type = geom_status.value();
+
+    draco::Decoder::GetEncodedGeometryType(&m_draco_buffer);
```
hobu, line 162: Does this method provide some kind of error status into `m_draco_buffer` or somewhere else to check the result of this operation?

plugins/draco/io/DracoReader.cpp
```diff
@@ -0,0 +1,363 @@
...
+{
+    using namespace draco;
+
+    log()->get(LogLevel::Debug) << "draco pc num_attributes: "
+                                << m_pc->num_attributes() << std::endl;;
+
+    const GeometryMetadata *metadata = m_pc->GetMetadata();
+    for (int i=0; i < m_pc->num_attributes(); ++i)
+    {
+        const PointAttribute* attr = m_pc->GetAttributeByUniqueId(i);
```
hobu, line 186: We need to check that `attr` exists before we try to use it.

plugins/draco/io/DracoReader.cpp
```diff
@@ -0,0 +1,363 @@
...
+        const AttributeMetadata* attr_metadata = m_pc->GetAttributeMetadataByAttributeId(i);
+        DataType dt = attr->data_type();
+        GeometryAttribute::Type at = attr->attribute_type();
+
+        int8_t nc = attr->num_components();
+        log()->get(LogLevel::Debug) << "named component: "
```
hobu, line 192: Tidy up the logs to output useful stuff to someone trying to figure out what's in their Draco file.

plugins/draco/io/DracoReader.cpp
```diff
@@ -0,0 +1,363 @@
...
+                    layout->registerDim(Dimension::Id::TextureW);
+                }
+                break;
+            }
+
+            case GeometryAttribute::GENERIC:
+            {
+                Dimension::Id id = pdal::Dimension::id(name);
+                if (id != Dimension::Id::Unknown) {
+                    m_generics[id] = std::vector<double>(m_pc->num_points(), 0.0);
```
hobu, line 244: Are generic dimensions always read into double, or can the user control this?

plugins/draco/io/DracoReader.cpp
```diff
@@ -0,0 +1,363 @@
...
+                CopyAttributeData<uint16_t>(m_count, attr, m_colors);
+            case GeometryAttribute::TEX_COORD:
+                CopyAttributeData<double>(m_count, attr, m_textures);
+                break;
+
+            case GeometryAttribute::GENERIC:
+                if (name.length() > 0)
+                {
+                    Dimension::Id id = Dimension::id(name);
+                    CopyAttributeData<double>(1, attr, m_generics.at(id));
```
hobu, line 292: Does Draco's metadata give you type information for generic GeometryAttributes? We want to read and write as the same types that were set, not as doubles, right?

plugins/draco/io/DracoReader.cpp
```diff
@@ -0,0 +1,363 @@
...
+                }
+
+                break;
+
+            default:
+                // Unsupported draco domain types
+                throw pdal_error("Unknown Geometry Attribute Type");
```
hobu, line 299: Tell the user more information why we failed here. What geometry type were we given that we can't read?

plugins/draco/io/DracoWriter.cpp
```diff
@@ -0,0 +1,455 @@
...
+        }
+        if ( tu != tv || tv != tw || tu != tw ) {
+            throw pdal_error("TextureU, V, and W dimensions must be of the same type");
+        }
+    }
+}
+
+void DracoWriter::ready(pdal::BasePointTable &table)
+{
+    *m_stream << std::fixed;
```
hobu, line 213: huh?

plugins/draco/test/DracoReaderTest.cpp
```diff
@@ -0,0 +1,167 @@
...
+    PointViewSet set = pipeline.views();
+    PointViewPtr v = *set.begin();
+
+    for (PointId i = 0; i < count; i += 1) {
+        double x = v->getFieldAs<double>(Dimension::Id::X, i);
+        double y = v->getFieldAs<double>(Dimension::Id::Y, i);
+        double z = v->getFieldAs<double>(Dimension::Id::Z, i);
+        EXPECT_TRUE(bounds.contains(x, y, z));
+        if (!bounds.contains(x,y,z)) {
+            std::cout << "X: " << x << std::endl;
```
hobu, line 159: lint

scripts/ci/linux/cmake.sh
```diff
@@ -8,8 +8,9 @@ LDFLAGS="$LDFLAGS -Wl,-rpath-link,$CONDA_PREFIX/lib" cmake .. \
       -DCMAKE_INSTALL_PREFIX=${CONDA_PREFIX} \
       -DBUILD_PLUGIN_I3S=ON \
       -DBUILD_PLUGIN_NITF=ON \
-      -DBUILD_PLUGIN_TILEDB=ON \
+      -DBUILD_PLUGIN_TILEDB=OFF \
```
hobu, line 11: oops, I turned this one off. Please turn back on.

hobu (changes requested):

cmake/draco.cmake
kylemann16: Why not? I didn't write this part, but it was working when I picked it up so I didn't mess with it.

hobu: It throws a cmake warning on newer cmake setups.

plugins/draco/io/DracoReader.cpp
kylemann16: User will not be setting this.

plugins/draco/io/DracoWriter.cpp
kylemann16: Unsure how I can avoid these. PointId is a uint64_t and ValueTypeT is a uint32_t. They chose to limit themselves there.

plugins/draco/io/DracoWriter.cpp
```diff
@@ -0,0 +1,457 @@
...
+    Dimension::Id dim = idList[0];
+    Dimension::Type type;
+
+    if (m_userDimMap.find(dim) != m_userDimMap.end()) {
+        std::string typeStr = m_userDimMap.at(dim);
+        type = Dimension::type(typeStr);
+    } else {
+        type = Dimension::defaultType(dim);
+    }
+
```
abellgithub, line 315: Just like in the reader, all this can be dispensed with by using the pdal function that will allow you to read data of the type you desire directly into a buffer.

plugins/draco/io/DracoWriter.cpp
```diff
@@ -0,0 +1,457 @@
...
+    encoder.SetAttributeQuantization(draco::GeometryAttribute::TEX_COORD, m_quant.at("COLOR"));
+    encoder.SetAttributeQuantization(draco::GeometryAttribute::GENERIC, m_quant.at("GENERIC"));
+    const auto status = encoder.EncodePointCloudToBuffer(*m_pc, &buffer);
+
+    if (!status.ok()) {
+        const std::string err = status.error_msg();
+        throw pdal_error("Error encoding draco pointcloud. " + err);
+    }
+
+    const auto bufferSize = buffer.size();
```
abellgithub, line 444: Unused

plugins/draco/io/DracoWriter.hpp
```diff
@@ -0,0 +1,159 @@
...
+        { "POSITION",  11 },
+        { "NORMAL",     7 },
+        { "TEX_COORD", 10 },
+        { "COLOR",      8 },
+        { "GENERIC",    8 }
+    };
+
+    FileStreamPtr m_stream;
+
+    std::map<draco::GeometryAttribute::Type, int> m_dims;
```
abellgithub, line 146: The purpose of all of these maps is unclear. At a minimum they should be documented here. I think a bunch of this can go away. I think all you're trying to do is the same thing as in the writer - build a single mapping:

```
struct
{
  Dimension::Id;
  Draco Dimension;
  Draco Position; // Red 0, Green 1, Blue 2, for example
  Draco Type;
  Quanitzation;
};
```
Once you've built this you can loop through it once to create the Draco dimensions and then use it when looping through the points to write them to Draco.

plugins/draco/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
+#
+# Draco plugin CMake configuration
+#
+if (NOT Draco_FOUND)
+    message(FATAL_ERROR "Can't find Draco support required.")
+endif()
+
+if (WIN32)
+    add_definitions(-DNOMINMAX)
```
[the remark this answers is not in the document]
hobu: I added this clobber because TileDB was failing too and it was causing test failure churn. I think something might have changed with MSVC 2019's behavior in regard to this because it hadn't been failing in the past.

plugins/draco/io/DracoWriter.cpp
```diff
@@ -0,0 +1,457 @@
...
+        {
+            //get draco type associated with the current dimension in loop
+            draco::GeometryAttribute::Type dracoType = it->second;
+            if (!m_dims.count(dracoType)) m_dims[dracoType] = 1;
+            else ++m_dims.at(dracoType);
+        }
+        else
+        {
+            m_genericDims.push_back(dim);
+        }
```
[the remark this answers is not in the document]
kylemann16: I unfortunately can't add the Draco attributes to the point cloud until it has been initialized, and the earliest I can do that is in the `write` method, so I create `m_genericDims` so that I can use it in `parseDimensions`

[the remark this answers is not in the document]
abellgithub: Can you not just delay calling `set_num_points()` until `write()` and do all the rest earlier?

plugins/draco/io/DracoWriter.cpp
```diff
@@ -0,0 +1,457 @@
...
+        if (std::string(m_userQuant.type_name()) == "null") return;
+        throw pdal_error("Option 'quantization' must be a JSON object, not a " +
+            std::string(m_userQuant.type_name()));
+    }
+    //Quantization levels are available for POSITION, NORMAL, TEX_COORD, COLOR,
+    //and GENERIC draco types
+    for (auto& entry : m_userQuant.items()) {
+        const std::string attribute = entry.key();
+        const int quant = entry.value().get<int>();
+        if (m_quant.find(attribute) == m_quant.end())
```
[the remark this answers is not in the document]
kylemann16: Yes, it's just meant to allow them to adjust the quant value away from the default, but if they try something outside of the possible Draco attributes then I think it should throw.

plugins/draco/io/DracoWriter.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+        if(!datasetName.is_string()) {
+            throw pdal_error("Every value in 'dimensions' must be a string. Key '"
+                + dimString + "' has value with type '"
+                + std::string(datasetName.type_name()) + "'");
+        } else {
+            log()->get(LogLevel::Info) << "Key: " << dimString << ", Value: "
+                << datasetName << std::endl;
+
+            Dimension::Id dimType = Dimension::id(dimString);
```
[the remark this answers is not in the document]
kylemann16: This is just poor naming. It's meant to be a data type, yes.

plugins/draco/io/DracoWriter.cpp
```diff
@@ -0,0 +1,457 @@
...
+        if ( nx != ny || ny != nz || nx != nz ) {
+            throw pdal_error("NormalX, Y, and Z dimensions must be of the same type");
+        }
+    }
+    //make sure textures u, v, w are all the same, but it's possible there are
+    //only u and v, so make sure we know how many dimensions are in the file first
+    if (m_userDimMap.find(Dimension::Id::TextureU) != m_userDimMap.end())
+    {
+        std::string tu, tv, tw;
+        int texNum = m_dims.at(draco::GeometryAttribute::TEX_COORD);
```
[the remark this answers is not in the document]
kylemann16: m_dims is the relation between a draco attribute and the number of dimensions associated with it. POSITION will be mapped to 3, for example. Textures can have 2 or 3 dimensions and we shouldn't be looking at TextureW if there are only 2.

plugins/draco/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+    target_link_directories(pdal_io_draco_writer_test PRIVATE ${draco_LIBRARY_DIR})
+
+    PDAL_ADD_TEST(pdal_io_draco_reader_test
+        FILES
+            test/DracoReaderTest.cpp
+        LINK_WITH
+            ${reader_libname}
+            ${writer_libname}
+            ${DRACO_LIBRARY}
+        INCLUDES
```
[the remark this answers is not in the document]
kylemann16: I tried taking this out and it worked fine on everything except the windows compilation. Seems like I have to have it in there to satisfy that.

plugins/draco/io/DracoWriter.cpp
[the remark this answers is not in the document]
kylemann16: I don't believe you can set attributes until after you've initialized the pointcloud. I was getting seg faults when I tried to do this.

hobu:
I don't understand. Why aren't we merging?

kylemann16:
I haven't finished writing the changes we wanted

PR #3339, opened 2021-01-20
Allow ability to start LAS reading at an arbitrary point

abellgithub:


pdal/pdal_config.cpp
```diff
@@ -93,6 +93,9 @@ bool hasFeature(Feature f)
         enabled = true;
 #endif
         break;
+    case Feature::LAS_START:
```
hobu, line 96: LAS_START? Would we not eventually add start capability to other drivers?

hobu (approved):

pdal/pdal_config.cpp
abellgithub: If you're building a client, you can use this to check if PDAL has this support without resorting to version numbers.

PR #3340, opened 2021-01-20
Allow pipeline to fail if a reprojection failure occurs

abellgithub:


filters/ReprojectionFilter.cpp
```diff
@@ -68,6 +68,8 @@ void ReprojectionFilter::addArgs(ProgramArgs& args)
     args.add("in_srs", "Input spatial reference", m_inSRS);
     args.add("in_axis_ordering", "Axis ordering override for in_srs", m_inAxisOrderingArg, {} );
     args.add("out_axis_ordering", "Axis ordering override for out_srs", m_outAxisOrderingArg, {} );
+    args.add("error_on_failure", "Throw an exception if we can't reproject any point",
+        m_errorOnFailure);
```
hobu, line 72: does `m_errorOnFailure` need a default value? I propose `true`

hobu (approved):

filters/ReprojectionFilter.cpp
abellgithub: The default is `false`, which maintains current behavior.

hobu: The current default behavior was changed. Past reprojection operations used to throw errors if points failed to reproject. It doesn't matter too much, the user has bad data and needs to deal with it.

abellgithub: Well, sure, but years ago. We drop points that can't be reprojected. I don't recall the discussion around that, but I don't see a good reason to change things on users when people are being provided an option to get this other behavior if they want it.

PR #3346, opened 2021-01-27
Add EvalKernel

chambbj:
Adds pdal application/subcommand to print a classification report when
given two input point clouds, one with predicted labels, another with
truth.

Contains a minor edit in Chamfer and Hausdorff kernel documentation,
where positional arguments were described as non-positional.

### DIFF
--- a/doc/apps/eval.rst
+++ b/doc/apps/eval.rst
@@ -1,0 +1,70 @@
+.. _eval_command:
+
+********************************************************************************
+eval
+********************************************************************************
+
+The ``eval`` command is used to compare the ``Classification`` dimensions of two
+point clouds.
+::
+
+    $ pdal eval predicted truth --labels LABELS
+
+::
+
+    --predicted arg     Positional argument specifying point cloud filename containing predicted labels.
+    --truth arg         Positional argument specifying point cloud filename containing truth labels.
+    --labels arg        Comma-separated list of classification labels to evaluate.
+
+The command returns 0 along with a JSON-formatted classification report
+summarizing various classification metrics.
+
+In the provided example below, the truth and predicted point clouds contain
+points classified as ground (class 2) and medium vegetation (class 4) in
+accordance with the LAS specification. Both point clouds also contain some
+number of classifications that are either unlabeled or do not fall into the
+specificied classes.
+
+::
+
+    $ pdal eval predicted.las truth.las --labels 2,4
+    {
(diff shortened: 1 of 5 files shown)


doc/apps/eval.rst
```diff
@@ -0,0 +1,70 @@
+.. _eval_command:
+
+********************************************************************************
+eval
+********************************************************************************
+
+The ``eval`` command is used to compare the ``Classification`` dimensions of two
```
hobu, line 7: Right now it is comparing only `Classification`, correct? We should update the note to that effect in case people want to compare other dimensions

kernels/EvalKernel.cpp
```diff
@@ -0,0 +1,169 @@
...
+        // guaranteed, so rather than using the same PointId, we search for the
+        // nearest neighbor.
+        PointId qid = kdi.neighbor(p);
+        PointRef q = truthView->point(qid);
+
+        // TODO (chambbj): We should perhaps look at the distance to the
+        // nearest point and reject or otherwise report distances greater than
+        // 0.0, indicating some sort of mismatch between files.
+
+        int pc = p.getFieldAs<int>(Id::Classification);
```
hobu, line 121: Maybe add an option to default to `Classification` so that a user could compare to their own dimensions other than `Classification` for whatever reason.

hobu (approved):

kernels/EvalKernel.cpp
chambbj: In my mind, I didn't want it to expand too far beyond comparing classifications, but I hadn't considered it could be useful to compare classifications in a custom dimension, which seems reasonable. I can make this change. (I also realize that it _could_ be handy to compare two separate dimensions in a single file, which even makes this streamable and cuts down on IO. I'll have to think about whether that is a tweak to this kernel or something different altogether. Also need to think about how useful that is as a general purpose tool, or if it's just too specialized.)

doc/apps/eval.rst
```diff
@@ -0,0 +1,70 @@
...
+point clouds.
+::
+
+    $ pdal eval predicted truth --labels LABELS
```
abellgithub, line 11: I think this would be better as:

`$ pdal eval <predicted> <truth> --labels <labels>`

Otherwise it looks like `predicted` and `truth` are literals.

hobu: Agree. Push this doc edit and it's ready to merge.

chambbj:
Docs fixed and rebased with master.

Issue #3358, opened 2021-02-15
Implement parallel processing for filters which use batch mode processing


whereskenneth:
Although not possible for streaming filters, the ability to utilize concurrency for batch processing point clouds is a much needed feature in end-user applications where quick turnaround is desired.

Using the STL in C+11, in particular the `<thread>` library, this can be easily implemented without adding additional dependencies to the codebase. 

There are two main classes of parallel problems that I've recognized - deterministic and non-deterministic time. The deterministic time problems are fairly easy to parallelize - the work is distributed among available threads evenly. In non-deterministic time, the time to perform the calculation is dependent on the dataset. This occurs notably in nearest neighbor searches, of which a sizeable amount of point cloud algorithms rely on. In this case, a worker queue is preferable for two reasons. 
  
1.  A worker thread can grab a new point to work on from a queue as it finishes it's calculations.
2. K-d tree index search functions are const, so no internal parameters are changed during thread execution, and should be thread safe (!!!)

One scheme which can be used is a queue - infinite loop scheme, where an infinite loop in a worker thread 

1. Locks a queue mutex, and pops a point to process from a deque, or breaks if none left.
2. Performs some expensive calculation.
3. Locks an output mutex, and stores the results of the expensive calcuation.

The general scheme in C++ would look like this.

```
#include <deque>
#include <numeric>
#include <thread>

size_t point_count;
std::deque<size_t> queue(point_count);
std::iota(queue.begin(), queue.end(), 0);

// Or some other initialization scheme.

std::mutex queue_lock, write_lock;

unsigned int num_threads;
std::vector<std::thread> thread_pool;
thread_pool.reserve(num_threads);

for (size_t i = 0; i < num_threads; i++) {
  thread_pool.emplace_back([&] {
    for (;;) {
      size_t point_to_process;
      {
        std::lock_guard<std::mutex> lock(queue_lock);
        if (queue.empty())
          break;
        point_to_process = queue.front();
        queue.pop_front();
      }
      
      Result r = expensive_calculation(point_to_process);

      {
        std::lock_guard<std::mutex> lock(write_lock);
        store_result(r);
      }
    }
  });
}
for (std::thread &t : thread_pool) {
  if (t.joinable())
    t.join();
}
```

[...]

PR #3359, opened 2021-02-15
#3358 Adds parallel processing to outlier filter.

whereskenneth:


test/unit/filters/OutlierFilterTest.cpp
```diff
@@ -0,0 +1,385 @@
...
+        outlier_filter.setInput(reader);
+
+        serial_radius_pipeline.prepare();
+        time_point<system_clock> tick = system_clock::now();
+        PointViewSet viewSet =
+            outlier_filter.execute(serial_radius_pipeline.pointTable());
+        time_point<system_clock> tock = system_clock::now();
+
+        auto diff = tock - tick;
+        std::cout
```
whereskenneth, line 241: Perhaps there is a better option than just dumping to stdout?

test/unit/filters/OutlierFilterTest.cpp
```diff
@@ -0,0 +1,385 @@
...
+    ASSERT_EQ(parallel_statistical_outliers->size(), serial_statistical_outliers->size());
+
+}
+
+TEST(OutlierFilterTest, thread_count_validation) {
+    using namespace Dimension;
+    using namespace std::chrono;
+
+    const size_t point_count = 100;
+    auto log = Log::makeLog("outlier_filter_test", "stdout");
```
whereskenneth, line 339: Is there a better way to check logging?

abellgithub:
This can be (and should) be done lock-free by grouping the points. You can simply pass a start/length to the threads.

whereskenneth:
> This can be (and should) be done lock-free by grouping the points. You can simply pass a start/length to the threads.

Fair enough, but the extra bookkeeping required to implement that isn't worth the readability cost, in my opinion, so I won't author it. Someone else can if they'd like to do it, though.

abellgithub:
? I think you're misunderstanding. I'll do one and if you're interested in helping with others that would be great.

abellgithub:
Maybe I'm missing something, but wouldn't this work without the locks?

```
int numThreads = 4;
ThreadPool pool(numThreads);

    double chunkSize = double(np) / numThreads;
    for (int t = 0; t < numThreads; ++t)
    {
        PointId start = std::floor(chunkSize * t);
        PointId end = std::floor(chunkSize * (t + 1));
        pool.add([&distances, count, &index, start, end]()
        {
            PointIdList indices(count);
            std::vector<double> sqr_dists(count);

            for (PointId i = start; i < end; ++i)
            {
                index.knnSearch(i, count, &indices, &sqr_dists);
                for (size_t j = 1; j < count; ++j)
                {
                    double delta = std::sqrt(sqr_dists[j]) - distances[i];
                    distances[i] += (delta / j);
                }
            }
        });
        start += chunkSize;
    }
    pool.join();
```

Would it be OK to at least take the tests that you've made?

whereskenneth:
Yeah, you can take anything you'd like out of there. 

Are you sure updating `distances[i]` is thread safe? Trying to research but can't find a clear answer.

whereskenneth:
23.2.2  Container data races[container.requirements.dataraces]1For purposes of avoiding data races (17.6.5.9),implementations shall consider the following functions to beconst:begin,end,rbegin,rend,front,back,data,find,lower_bound,upper_bound,equal_range,atand, except in associative or unordered associative containers,operator[].

I guess that means this is fine?

abellgithub:
Yes, direct access to a vector is thread-safe as long as you've not doing any modification to the vector itself (rather than the data, which is fine).

hobu:
Closing this one for now. Please reopen if you pick the effort back up again.

PR #3361, opened 2021-02-17
Refactor Hausdorff kernel and add modified Hausdorff metric as one of the outputs

chambbj:


kernels/HausdorffKernel.cpp
```diff
@@ -86,12 +86,13 @@ int HausdorffKernel::execute()
     ColumnPointTable candTable;
     PointViewPtr candView = loadSet(m_candidateFile, candTable);
 
-    double hausdorff = Utils::computeHausdorff(srcView, candView);
+    std::pair<double, double> result = Utils::computeHausdorff(srcView, candView);
```
hobu, line 89: Should we leave a Utils::computeHausdorff around that returns a double and make a new Utils::computeHausdorffPair method that does this? I'm unsure if there's any users of Utils::computeHausdorff though, so maybe it doesn't matter.

hobu (approved):

kernels/HausdorffKernel.cpp
chambbj: I initially thought we'd only be impacting ourselves via the call from the kernel, but I suppose since it's publicly available via Utils, we should still preserve the original. I can make this change.

hobu:
Can you rebase this one to catch up to the CI updates?

PR #3364, opened 2021-02-22
basic caching mechanism for readers.ept

gui2dev:
As discussed with @abellgithub on gitter, this is a simple implementation of a caching mechanism for readers.ept.

This could be enhanced in many ways, I am open to any suggestions.

### DIFF
--- a/io/private/ept/Connector.cpp
+++ b/io/private/ept/Connector.cpp
@@ -34,8 +34,11 @@
 
 #include "Connector.hpp"
 
-#include <pdal/pdal_types.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/system/error_code.hpp>
+
+#include <pdal/util/FileUtils.hpp>
+#include <pdal/pdal_types.hpp>
 #include <random>
 
 namespace pdal
@@ -73,30 +76,33 @@ StringMap s_cache;
 std::string s_uuid = generate_uuid();
 std::mutex s_cacheMutex;
 
-Connector::Connector() : m_arbiter(new arbiter::Arbiter())
-{}
+Connector::Connector() : m_arbiter(new arbiter::Arbiter()) {}
 
-Connector::Connector(const StringMap& headers, const StringMap& query, bool useCache) :
-    m_arbiter(new arbiter::Arbiter), m_headers(headers), m_query(query), m_useCache(useCache)
-{}
+Connector::Connector(const StringMap& headers, const StringMap& query,
+                     bool useCache)
+    : m_arbiter(new arbiter::Arbiter), m_headers(headers), m_query(query),
+      m_useCache(useCache)
+{
+}
 
-std::string Connector::getUniquePath() const
+std::string Connector::getUniquePath(const std::string& path) const
 {
     auto ret(pdalboost::filesystem::temp_directory_path());
     ret /= "pdalcache";
     ret /= s_uuid;
     makeDir(ret.string());
-    ret /= generate_uuid(); //should we keep extension ?
-    return ret.string(); //should check if exists or not
+    ret /= generate_uuid();
+    ret += FileUtils::extension(path);
+    return ret.string(); // should check if exists or not
 }
 
-std::string Connector::getCache(const std::string & path) const
+std::string Connector::getCache(const std::string& path) const
 {
     std::lock_guard<std::mutex> guard(s_cacheMutex);
     if (!s_cache.count(path))
     {
-        s_cache.insert(std::make_pair(path,getUniquePath()));
-        put(s_cache[path],m_arbiter->get(path,m_headers,m_query));
+        s_cache.insert(std::make_pair(path, getUniquePath(path)));
+        put(s_cache[path], m_arbiter->get(path, m_headers, m_query));
     }
     return s_cache[path];
 }
@@ -122,18 +128,18 @@ NL::json Connector::getJson(const std::string& path) const
     }
     catch (NL::json::parse_error& err)
     {
-        throw pdal_error("File '" + path + "' contained invalid JSON: " +
-            err.what());
+        throw pdal_error("File '" + path +
+                         "' contained invalid JSON: " + err.what());
     }
 }
 
-std::string Connector::getBinaryCache(const std::string & path) const
+std::string Connector::getBinaryCache(const std::string& path) const
 {
     std::lock_guard<std::mutex> guard(s_cacheMutex);
     if (!s_cache.count(path))
     {
-        s_cache.insert(std::make_pair(path,getUniquePath()));
-        put(s_cache[path],m_arbiter->getBinary(path,m_headers,m_query));
+        s_cache.insert(std::make_pair(path, getUniquePath(path)));
+        put(s_cache[path], m_arbiter->getBinary(path, m_headers, m_query));
     }
     return s_cache[path];
 }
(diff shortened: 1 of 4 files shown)


gui2dev:
Not sure what to use instead of pdalboost for windows...

gui2dev:
There also could be some cache in memory.

Plus a mechanism to control the cache volume.
Also something to clean up cache.

And why not share cache among pdal processes?

abellgithub:
A few things:
1) Please don't use boost::filesystem. Use FileUtils.
2) Please don't write your own UUID code. Use Utils::Uuid.

Maybe it would be good if you wrote the process you envision? I don't see the files actually getting cached, and I also don't understand why there is mutex-protected static data. A few sentences of explanation would help.

gui2dev:
> A few things:
> 
> 1. Please don't use boost::filesystem. Use FileUtils.
> 2. Please don't write your own UUID code. Use Utils::Uuid.
> 
> Maybe it would be good if you wrote the process you envision? I don't see the files actually getting cached, and I also don't understand why there is mutex-protected static data. A few sentences of explanation would help.

I could not find in `FileUtils` a method that could point me to a temporary directory, that's why. I'll then add one in `FileUtils` then.

I'll use `Utils::Uuid` as well.

The main idea is the following :

`Connector` has a static stringMap that will point cached (local) path for non local path.

When `Connector::get` (or `getBinary` or `getLocalHandle`) is called :
if `path` is local -> do nothing
else if caching is enabled (`m_useCache = true`) -> if the file is already cached, returns it's path, if not cache it and retrun it's path.

Caching is done in 
```cpp
std::string Connector::getCache(const std::string& path) const;
std::string Connector::getBinaryCache(const std::string& path) const;
arbiter::LocalHandle Connector::getLocalHandleCache(const std::string& path) const;
```
<details>
<summary>implementation instance</summary>

```cpp
std::string Connector::getCache(const std::string& path) const
{
    std::lock_guard<std::mutex> guard(s_cacheMutex);
    if (!s_cache.count(path))
    {
        s_cache.insert(std::make_pair(path, getUniquePath(path)));
        put(s_cache[path], m_arbiter->get(path, m_headers, m_query));
    }
    return s_cache[path];
}
```

</details>


Cache location is :
`temp_path\uuid\`

The static stringMap looks like this :
| Key | Value |
| ------------- |:-------------:|
| http:\\something\ept.json     | /tmp_path/first_uuid/uuid_1.json |
| http:\\something\foo.laz      | /tmp_path/first_uuid/uuid_2.laz | 
| http:\\something\bar.laz  | /tmp_path/first_uuid/uuid_3.laz      | 

I made the map static so it's not lost when `Connector` instances are destroyed (usefull if you'll have to run several readers.ept in the same pipeline).

Mutex is there to protect the map for concurrent access.

abellgithub:
I don't understand this, but maybe I didn't understand the initial desire. What is the case where a memory-based cache would be useful? I was thinking you wanted a disk-based cache that would allow the files to be fetched from disk rather than the network source from run to run, but when the cache mapping is in memory, you lose that functionality.

gui2dev:
The files are cached on the disk

gui2dev:
If that's ok with you so far, I'll put the map in a file and provide some file looking mechanism and I'll use Utils instead of boost.

abellgithub:
FYI, there are functions to make temp filenames.  The non-Windows version is kind-of illegal since it modifies a string in place (this comes from test code where I don't care.)

```
std::string makeTempFileName()
{
    std::string filename;
#ifdef _WIN32
    const char *fnTemplate = "fnXXXXXX";
    char name[9];
    strncpy(name, fnTemplate, 9);
    _mktemp_s(name, 9);
    char path[MAX_PATH];
    GetTempPath(MAX_PATH, path);
    filename = std::string(path) + name;
#else
    std::string s("/var/tmp/fnXXXXXX");
    mkstemp(const_cast<char *>(s.data()));
    filename = s;
#endif
    return filename;
}
```

gui2dev:
@abellgithub I made some changes.
I use a `LockFile` class to manage concurrent read/write on a `map.json` file that map remote -> local files.
`LockFile` class is in `FileUtils` but it could either be a private class for `readers.ept` or it could be on its own in `Utils` or `Vendor` as the code source is from another project.

abellgithub:
This is starting to seem complicated and I wonder if this would be better handled outside of PDAL. There are HTTP caching programs like [squid](http://www.squid-cache.org/).

gui2dev:
> This is starting to seem complicated and I wonder if this would be better handled outside of PDAL. There are HTTP caching programs like [squid](http://www.squid-cache.org/).

That's an alternative. Just let me know what would be best suited for PDAL.

hobu:
> That's an alternative. Just let me know what would be best suited for PDAL.

Can you get a local example working with squid? 

The old programmer koan applies here:

> There's three hard problems in computer science. Naming things, caching stuff, and off-by-one errors.

In my opinion, it is well out of scope for PDAL to worry about caching things. Squid would seem to be a natural fit with way more features and less code to maintain than anything internally provided by PDAL. For example, you are likely to want proxy support and a bunch of other more advanced caching features. A proper HTTP caching layer like squid has that, and adding it to PDAL is likely to put a not-sufficient and maybe-broken implementation out there that will be more code to maintain.

What is the problem that your caching layer needs to solve? What has you wanting to add this to PDAL in the first place?

gui2dev:
> What is the problem that your caching layer needs to solve? What has you wanting to add this to PDAL in the first place?

@hobu, in our application, we are making a lot of profiles from massive ept data.
For each profile, `readers.ept` would grab remote files, and from one profile to another, most of the files are in common. I was looking for a way to limit bandwidth and make the whole thing faster. So far, the tests I made with this basic implementation do limit bandwidth (by about 70%) and execution was a wee bit faster (10 - 20 %).

I was at first looking for a proxy-cache support sort of thing, didn't know squid even existed. I'll have to make a local example with that.

hobu:
Yes, I think squid would be a good fit for this scenario and provide you lots of levers and knobs to adjust as well. The best part is you don't have to write any code!

Please report back with any success you might have in terms of configuration so the next person who follows along behind has a breadcrumb trail to follow.

gui2dev:
So far I could, on the same (local) network :
-> install and configure squid on a given machine (ubuntu 20.04)
-> install and configure proxifier on an other one (windows 10)
-> run my program on this machine (windows 10)

Trafic would go through proxy, so I guess, configuring squid `cache_dir` would be ok.

I did not test if running squid and proxifier on the same machine is working.

I am not sure how to specifically say to pdal / my program to use the proxy (and leave the rest alone) without using some tool like proxifier.

hobu:
> I am not sure how to specifically say to pdal / my program to use the proxy (and leave the rest alone) without using some tool like proxifier.

Wouldn't you just modify the EPT url to point at your proxy endpoint?

gui2dev:
So, I started using squid and pdal on the same machine (wsl, ubuntu 18.04), I did not go further on a windows machine.
I configured squid to do some caching to dir and added some refresh rules for "specific files" (json / laz).

I could not directly specify endpoint in ept url (maybe you need a specific server over squid to handle that).
I tried using `"filename": "http:\\127.0.0.1:3128\http:\\some-remote-ept-host.com\ept.json"` along with `"header": "{"Host":"some-remote-ept-host.com"}` but with no success (looks like HEAD was `\http:\\some-remote-ept-host.com\ept.json` instead of `http:\\some-remote-ept-host.com\ept.json`).

Otherwise, using `http_proxy` env variable (that I setted to `http:\\127.0.0.1:3128`) allows me to have some effective caching ( `/var/log/squid/access.log` shows some `TCP_MEM_HIT/200`)

hobu:
> Maybe arbiter should allow proxy using CURLOPT_PROXY

Happy to merge that patch to Arbiter and upstream it into PDAL.

gui2dev:
> > Maybe arbiter should allow proxy using CURLOPT_PROXY
> 
> Happy to merge that patch to Arbiter and upstream it into PDAL.

I will work on that.

To go further on windows, I tried to set my squid instance (running in wsl) as my proxy in the network settings. It works fine so far, but it seems that `curl` just ignores this settings. Maybe an env variable could do the trick, but the arbiter options seems more robust and accurate to me.

gui2dev:
> Happy to merge that patch to Arbiter and upstream it into PDAL.

https://github.com/connormanning/arbiter/pull/40

gui2dev:
PR is merged on arbiter. I can close that one.

hobu:
please make a new ticket to vendor the latest arbiter in PDAL. We are gathering steam for a 2.3 release next month and will want to make sure it makes it.

gui2dev:
I'll work on that tomorrow.

abellgithub:
Glad you got something working without (seemingly) too much trouble.

PR #3366, opened 2021-02-24
ObjReader: Adds a virtual desctructor that closes m_istream if it is still open

schwehr:
Initialize pointers and m_index plus be sure that closing m_istream is safe in the destructor.

I didn't add any sort of warning to the case where the desctructor has to do the close.  I am assuming this case might come up if there is an error after `ready()` is called, but `done()` has not yet been called.

The initial fix was written by another Google employee based on PDAL 2.2.0.  However, that code didn't have `done()` 

A `done()` call without a preceding `ready()` (or two done's in a row) will pass a nullptr to `closeFile`, but that checks for a nullptr.  That means that I probably could have left out the `if` in the destructor.

### DIFF
--- a/io/ObjReader.cpp
+++ b/io/ObjReader.cpp
@@ -47,6 +47,12 @@ static StaticPluginInfo const s_info
 
 CREATE_STATIC_STAGE(ObjReader, s_info)
 
+ObjReader::~ObjReader()
+{
+    if (m_istream != nullptr)
+        Utils::closeFile(m_istream);
+}
+
 std::string ObjReader::getName() const { return s_info.name; }
 
 void ObjReader::addDimensions(PointLayoutPtr layout)
@@ -76,6 +82,7 @@ void ObjReader::ready(PointTableRef)
 void ObjReader::done(PointTableRef)
 {
     Utils::closeFile(m_istream);
+    m_istream = nullptr;
 }
 
 point_count_t ObjReader::read(PointViewPtr view, point_count_t cnt)
--- a/io/ObjReader.hpp
+++ b/io/ObjReader.hpp
@@ -47,6 +47,7 @@ namespace pdal
 class PDAL_DLL ObjReader : public Reader
 {
 public:
+    virtual ~ObjReader();
     std::string getName() const;
 
 private:
@@ -72,11 +73,11 @@ private:
     std::vector<XYZW> m_vertices;
     std::vector<XYZW> m_textureVertices;
     std::vector<XYZW> m_normalVertices;
-    TriangularMesh *m_mesh;
+    TriangularMesh *m_mesh = nullptr;
     using VTN = std::tuple<int64_t, int64_t, int64_t>;
     std::map<VTN, PointId> m_points;
-    std::istream *m_istream;
-    point_count_t m_index;
+    std::istream *m_istream = nullptr;
+    point_count_t m_index = 0;
 
     using TRI = std::array<VTN, 3>;
     using FACE = std::vector<VTN>;


abellgithub:
What is the use case for this? How are you taking a code path where this is a problem?

schwehr:
This is primarily in case of an error while reading a file.  Bad things sometimes happen outside of the process. e.g. a file is suddenly unavailable or unreadable.  e.g. permission change, file modification, unmounted or network timeout for remote filesystem.  Not all filesystems behave as one would hope.

Additionally:

What ensures that `done()` is always called?  

It's also good to make sure that  if there is an internal error to the class sometime after a future internal modification, that the close doesn't get called on an uninitialized value in the file pointer.

abellgithub:
You're talking in hypotheticals. I really would like to know the real code path that causes this problem. You should either throw an unrecoverable exception (in which case the file gets close), or `done` gets called.

schwehr:
Pardon not my being very familiar with C++ exceptions, but how does an unrecoverable exception cause the file to be closed? 

I don't see any tests of reading corrupt files in test/unit/io/ObjReaderTest.cpp.

I don't have a case.

abellgithub:
`done` is called because `ready` is called. The code that executes pipelines makes sure of this unless there's an exception. Because those functions are private, users have to call another function, `execute`, in order to invoke them.

I'm not sure how I would test for corrupt files. I'm not saying the code is correct, but I don't know what case I could invoke to create a corrupt file test and if I knew it I'd fix it. The cases that you mention (permissions, network timeout) are real, but closing the file isn't going to fix them.

IMO, the only real issue here is potential leakage of file handles. But there are *lots* of places where this could happen and if we're going to deal with this, doing it as a one-off in ObjReader doesn't seem the way (and also depends on the dtor being called, which we can't really know since we don't know the use case). There's definitely some improvement that can be made to file handling to make sure that things get closed, but this mostly is about the structure that has existed for files since the early days. I'm up for a discussion on this, but I don't think that this is easy and I haven't really heard of people having problems or creating applications using PDAL in some sort of a daemon/service where such things could become an issue. I'd think I'd rather create something like a file manager that you can destroy that makes sure everything is cleaned up rather than put this kind of thing in 100's of classes. Making sure that you've caught all the cases is hard.

I'm fine merging this, but I don't like the idea of propagating this all over.

schwehr:
Maybe it would help to know what happens if the code tries to read a truncated file?  Is there a test that covers that case?  And likely that would want to be a couple of test cases as often it matters very much where that truncation is.

aaronyonas:
Without this change, asan reports memory leaks in ObjReaderTest.cpp.

For example, in the NoVertex test:

==3866==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 336 byte(s) in 1 object(s) allocated from:
    #0 0x55acd39d521d in operator new(unsigned long) third_party/llvm/llvm-project/compiler-rt/lib/asan/asan_new_delete.cpp:99:3
    #1 0x7f53b9c250c7 in pdal::FileUtils::openFile(std::__u::basic_string<char, std::__u::char_traits<char>, std::__u::allocator<char> > const&, bool) third_party/pdal/pdal/util/FileUtils.cpp:122:11
    #2 0x7f53b9b3d128 in pdal::Utils::openFile(std::__u::basic_string<char, std::__u::char_traits<char>, std::__u::allocator<char> > const&, bool) third_party/pdal/pdal/PDALUtils.cpp:345:12
    #3 0x7f53b99fcc67 in pdal::ObjReader::ready(pdal::BasePointTable&) third_party/pdal/io/ObjReader.cpp:78:17
    #4 0x7f53b9ba72e4 in pdal::Stage::execute(pdal::BasePointTable&, std::__u::set<std::__u::shared_ptr<pdal::PointView>, pdal::PointViewLess, std::__u::allocator<std::__u::shared_ptr<pdal::PointView> > >&) third_party/pdal/pdal/Stage.cpp:254:5
    #5 0x7f53b9ba66bf in pdal::Stage::execute(pdal::BasePointTable&) third_party/pdal/pdal/Stage.cpp:216:32
    #6 0x7f53ba95ddba in pdal::ObjReader_NoVertex_Test::TestBody() third_party/pdal/test/unit/io/ObjReaderTest.cpp:104:35

abellgithub:
I have no idea why you're seeing the behavior you are. I'm having no problem with -fsanitize=address -fsanitize=leak on Ubuntu 20. Are you getting an exception?

aaronyonas:
I'm still on version 2.2.0, so maybe that explains the difference in behavior that we're seeing.  I'll check again once I'm at head.

aaronyonas:
I'm now at head and have confirmed that this bug has been fixed.
Let's cancel this PR.  Sorry for the false alarm!

PR #3367, opened 2021-02-24
HeaderVal.hpp: Initialize m_defVal in constructor

schwehr:


abellgithub:
m_defVal is never accessed in this case since m_valSet is false.

schwehr:
@abellgithub Thanks!

PR #3368, opened 2021-02-24
XMLSchema.cpp: Fix memory leak in loadMetadata

schwehr:
The strings returned from `xmlGetProp` and `xmlNodeGetContent` need to be freed and thus should also not be `const`


abellgithub:
`etype` appears not to even be used.

schwehr:
> `etype` appears not to even be used.

I will remove it.  Thanks for catching that.

PR #3369, opened 2021-02-24
TileTest.cpp: Only check read values if they are actually read

schwehr:


abellgithub:
I don't understand this. If the file isn't good, you want to know, which is what the tests will tell you.

aaronyonas:
The core issue here is that msan is reporting uninitialized memory reads.  That can happen, for example, if there's any trailing whitespace in the file: the loop iterates an extra time and x, y, and z are not set before they are used.

abellgithub:
Then please remove the test on the stream that's at the end of the loop as it becomes unnecessary.

schwehr:
Thanks!

Issue #3365, opened 2021-02-22
Small bug in iowa-entwine tutorial?


gsakkis:
Unless I am missing something, shouldn't [this line](https://github.com/PDAL/PDAL/blob/master/doc/tutorial/iowa.json#L42) be `"inputs": [ "classify" ]` in the [Reading data from EPT](https://pdal.io/tutorial/iowa-entwine.html) tutorial? Also the [gdal](https://github.com/PDAL/PDAL/blob/master/doc/images/pipeline-example-writers.gdal.png) and [las](https://github.com/PDAL/PDAL/blob/master/doc/images/pipeline-example-writers.las.png) writer images are not quite in sync with the pipeline json.

hobu:
Please make a pull request fixing these issues and we can merge it.

hobu:
Small fix added at ea0ff391723d9920ab768f7e461fab709909e1e9 but the fix in this PR is not correct. The LAS writer needs to connect to the ``writers.gdal`` for things to work. If the pictures are incorrect, please reopen.

PR #3378, opened 2021-02-26
Update tutorial's iowa.json pipeline to be in sync with the diagram

gsakkis:
Fixes https://github.com/PDAL/PDAL/issues/3365. 

Note: I didn't include the `0.01` and `auto` values in the [writers.las](https://pdal.io/_images/pipeline-example-writers.las.png) diagram as I'm not sure what options they correspond to.

### DIFF
--- a/doc/tutorial/iowa.json
+++ b/doc/tutorial/iowa.json
@@ -33,18 +33,18 @@
         {
             "filename": "iowa.laz",
             "inputs": [ "classify" ],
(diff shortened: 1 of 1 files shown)


abellgithub:
Have you tried this by actually running it? It doesn't seem like it would work.

gsakkis:
Not really, I assumed the diagrams are correct and changed the text to match them. Why it wouldn't work?

PR #3384, opened 2021-03-03
IStream.hpp: fixed read from uninitialized memory

aaronyonas:
Surfaced when trying to read an empty file.


abellgithub:
Can you please provide a test that fails that this change fixes?

aaronyonas:
I'll put together a unit test in a few days and get back to you.

abellgithub:
I don't mind the change, but it seems kind of pointless. What I'm more interested in is the case in practice where this comes up and should be dealt with before we get here. Initializing the string to zeros instead of junk really doesn't make anything correct, it just quiets the sanitizer.

I think most of these cases where we're reading from a file that's invalid can't reasonably be addressed without obtrusive tests that can also slow things down, but if there are easy one-time checks that can be done, I'd like to add them.

aaronyonas:
I'll look into adding some early validation as you say.

In the meantime, do you like this version better?  It's safe and (if resize() is well-implemented) should be faster, too.

```
        s.resize(size);
        m_stream->read(&s[0], size);
        s.resize(m_stream->gcount());
        size_t first_null = s.find('\0');
        if (first_null != std::string::npos)
            s.resize(first_null);
```

abellgithub:
I really don't care much. The issue is that this function is used to read a specific number of bytes from a stream into a string. If you can't read that number of bytes, something is wrong, so whether the string has garbage in it or zeros really doesn't matter. The key is that the bytes you were expecting weren't there. I think we generally catch this elsewhere by checking file status later (so we don't have to check all over the place during input handling), so whether we're reading uninitialized memory or not shouldn't really matter.

I think simply initializing things to 0 is the most clear.  Having a string that's fewer than `size` characters breaks the contract and might be an issue for some use cases.  I'll merge this as is, but if you have a real case where this leads to an error that is not *otherwise* caught, I'd like to fix that.

Thanks,

aaronyonas:
>  The issue is that this function is used to read a specific number of bytes from a stream into a string. If you can't read that number of bytes, something is wrong,

Ok, so it sounds like you're saying that it's the caller's responsibility to ensure the there are at least `size` bytes available to read in the stream, yes?  If so, you might want to update the function's documentation comment to say so.  It currently defines `size` as `Maximum number of bytes to extract.` which implies that it might extract fewer bytes.

Also maybe add `assert(size == m_stream->gcount());` after the read.

PR #3389, opened 2021-03-04
Fix spelling of 'ReturnNumber' in error messages

marcjansen:
This PR suggests to change three occurences of `ReternNumber` in error messages which most likely should be `ReturnNumber`.

Please review.


abellgithub:
Thanks

Issue #3391, opened 2021-03-15
filters.colorization keeps outside points in normal mode however doesn't keep outside points in streaming mode


dbrookes96:
**Describe the bug**
It might be nice if filters.colorization ignored points that are outside the input raster. For example if it wasn't correctly georeferenced to the point cloud. Or at least the user should be able to decide. I was playing around with it and noticed that there is actually a difference in the output between streaming mode (which is what I was using) and normal mode.

When running filters.colorization in streaming mode points that do not correspond to a position on the input raster are filtered out. `ColorizationFilter processOne` returns false when the raster read for a point X and Y position fail. However in normal mode points that do not correspond to a position on the input raster are ignored. `ColorizationFilter filter` ignores the return of `ColorizationFilter processOne`.

For now I can run in normal mode. However I like the benefits of streaming mode 😎.

This does bring up the question if any other filters have similar behaviour?

**Expected behavior**
```
{
	"pipeline":
	[
		{
		"type": "readers.las",
		"filename": "autzen-point-format-3.las"
		},
		{
		"type": "filters.transformation",
		"matrix": "1 0 0 1600 0 1 0 2200 0 0 1 60 0 0 0 1"
		},
		{
		"type": "filters.colorization",
		"raster": "autzen.jpg"
		},
		{
		"type": "writers.text",
		"filename": "autzen-point-format-3_colourised.txt"
		}
	]
}
```
This pipeline uses `autzen-point-format-3.las` from the unit test data. That file has 106 points. It purposefully translates the input points away by about half of the input point clouds bounding box as a test. When in streaming mode this pipeline outputs 23 points. When in normal mode this pipelines outputs 106 points.

I guess we need to decide whether the expected behaviour would be:

a) to by default skip away points that are outside (like streaming mode currently does).

or

b) to by default ignore points that are outside (like normal mode currently does).

Either way I think it being inconsistent is a bit of a problem. Would it then make sense to add a "keep_outside" argument to the filter to let the user decide.

**System/installation information**
Windows 10 Enterprise 19042.867 20H2
pdal 2.2-maintenance branch with 0f319a5 cherry-picked built with MSVC 14.28 (VS 2019 16.8).

hobu:
IMO, how normal mode works is how it should operate, and the streaming scenario right now is a bug in behavior. Colorizing points shouldn't toss them out.

abellgithub:
Fixed with 1a802af92a194e99cf551b56579ce21c07f02af8

PR #3393, opened 2021-03-16
filters.colorization keep points outside raster when streaming (#3391)

dbrookes96:
always return true from ColorizationFilter::processOne regardless of if
the raster read for the x, y position of the point did not succeed. the
output band(s) of points outside the raster are ignored.

the normal mode output and stream mode output should now be equal for
filters.colorization.

added a quick unit test to check this scenario.


dbrookes96:
fixes #3391

hobu:
The failures here have nothing to do with the PR, and this should be merged.

PR #3406, opened 2021-04-05
Allow 1.4 LAS to use geotiff for PDRF 0-5

### LINKED ISSUE #3242: readers.las won't use Geotiff for version 1.4 and point formats 0-5

abellgithub:
Close #3242


hobu (approved):
We should add to the docs to describe this flowchart of possibilities.

abellgithub:
> We should add to the docs to describe this flowchart of possibilities.

It's documented in the LAS spec. This was an oversight.

PR #3412, opened 2021-04-09
Remove unused gzip from arbiter

abellgithub:


hobu:
I don't think we can remove the Deflate and Gzip implementations from PDAL. Entwine uses them at least.

abellgithub:
I only removed the aribter implementation of gzip that was used to decompress gzip'd responses, which is now done automatically by curl.

connormanning (approved):

PR #3415, opened 2021-04-12
Write double-precision values with more precision

### LINKED ISSUE #3235: Option::add(key, double) loses precision
Option stores values as strings. Double-precision values stored as options may lose precision due to the use of default conversion operators.

abellgithub:
Close #3235


pdal/Options.hpp
```diff
@@ -63,15 +63,17 @@ class PDAL_DLL Option
...
     }
 
-    Option(std::string const& name, const std::string& value) :
-        m_name(name), m_value(value)
+    Option(std::string const& name, const std::string& value) : m_name(name), m_value(value)
     {}
 
+    Option(std::string const& name, const double& value) : m_name(name)
+    {
+        m_value = Utils::toString(value, 15);
```
hobu, line 74: What about providing a specialization ctor that 

```
Option(std::string const& name, 
              const double& value, 
              double precision=15)
```

abellgithub: I don't think there's any practical advantage.

hobu: The advantage is the ability for a user of Option to control it in the few cases it might matter, like the one that precipitated this change.

abellgithub: I don't think there's a good argument for less precision.

hobu: I agree. Is there an argument for *more* precision, however?

abellgithub: There may be, but I doubt a user will know this but it's probably better to avoid transit in strings if this isn't sufficient.

PR #3425, opened 2021-04-20
PCD Reader issues

dsmith8fr:
PCD Reader not seeking to correct location in Windows. 
Files were opened as text not binary causing issues with seekg and 
tellg methods. By setting the binanary argument to true when opening
the file, Utils::openFile(m_filename, true); this issue is resolved.


abellgithub:
After writing some test code on Windows, it just appears that file position on Windows with a UNIX formatted file is really, really strange. I can't find any very good explanation of the details and I can't intuit what's happening, but I've convinced myself that binary input is fine in both UNIX and Windows line ending cases.

PR #3427, opened 2021-04-22
Eliminate some WKT round trips and use WKT2

abellgithub:
Close #3427

### DIFF
--- a/pdal/Geometry.cpp
+++ b/pdal/Geometry.cpp
@@ -159,7 +159,8 @@ Utils::StatusWithReason Geometry::transform(SpatialReference out)
         return StatusWithReason(-2,
             "Geometry::transform() failed.  NULL target SRS.");
 
-    SrsTransform transform(getSpatialReference(), out);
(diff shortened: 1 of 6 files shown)


abellgithub:
This avoids the issue of needing to write WKT2-2019 by creating an SrsTransform directly from an OGRSpatialReference.

pdal/Geometry.cpp
```diff
@@ -187,7 +188,8 @@ SpatialReference Geometry::getSpatialReference() const
     if (srsValid())
     {
         char *buf;
-        m_geom->getSpatialReference()->exportToWkt(&buf);
+        const char *options[] = { "WKT2_2019", nullptr };
```
hobu, line 191: What version of GDAL was `WKT2_2019` introduced, and does this mean PDAL's minimum GDAL version would be limited to that?

hobu (changes requested):
This PR bumps the minimum GDAL version to at least 3.0, and the cmake config needs to be adjusted accordingly.

hobu (approved):

PR #3433, opened 2021-04-23
Support and test large file mapping

### LINKED ISSUE #3282: MapFile can fail on 32-bit Windows with large files
We're using size_t for file size, which is only 32-bit on Windows, so things will fail if file sizes are over the 4G limit.

abellgithub:
Close #3282


pdal/util/FileUtils.hpp
```diff
@@ -302,8 +313,8 @@ namespace FileUtils
       \return  MapContext.  addr() gets the mapped address.  what() gets
          any error message.  addr() returns nullptr on error.
     */
-    PDAL_DLL MapContext mapFile(const std::string& filename, bool readOnly,
-        size_t pos, size_t size);
+    PDAL_DLL MapContext mapFile(const std::string& filename, bool readOnly = true,
```
hobu, line 316: Is this API change going to bother anyone? Probably not, but we should include a note in the release notes.

hobu (approved):

pdal/util/FileUtils.hpp
abellgithub: The API is compatible. uintmax_t is at least as bit as size_t.

PR #3434, opened 2021-04-26
pin GDAL at 3.0.4

hobu:


hobu:
This was done on linux-only at 673dd087cba8cc51b163e346cf3a46dae5cd227e

PR #3436, opened 2021-04-26
Reset error handler when logging stops.

### LINKED ISSUE #3244: Crash on GDAL Error when logger is set to invalid stream
**Describe the bug**
Using the C++ API, the application can crash if the PDAL singleton logger has been set to a local stringstream.

Consider the following C++ application. (I voluntarily excluded the `#include`s)
```cpp
void dumpSStoCout(std::stringstream& ss)
{
    for(std::string line ; std::getline(ss, line);)
    {
        std::cout << line << std::endl;
        line.clear();
    }
}

void makeSomePDALWork() {
    std::stringstream output;

    auto log = pdal::Log::makeLog("PDAL Work", &output);
    log->setLevel(pdal::LogLevel::Debug);

    pdal::Options ro;
    ro.add("bounds", pdal::BOX3D(0, 0, 0, 99, 99, 99));
    ro.add("mode", "random");
    ro.add("count", 100);
    pdal::FauxReader r;
    r.setOptions(ro);

    pdal::Options wo;
    wo.add("filename", "test.txt");
    pdal::TextWriter w;

    w.setOptions(wo);

    pdal::PointTable t;
    t.layout()->registerDims({pdal::Dimension::Id::X, pdal::Dimension::Id::Y, pdal::Dimension::Id::Z});
    w.setLog(log);
    w.prepare(t);
    w.execute(t);

    dumpSStoCout(output);

    // Simulate a GDAL error that happens once the PDAL work is done.
    CPLError(CE_Fatal , 0, "This error doesn't make the app crash.");

    dumpSStoCout(output);
}

void main()
{
    // Some PDAL export triggered by an UI action,
    // ex: export a bunch of points to a LAS file.
    makeSomePDALWork();
        
    // Now the user trigger some other works that use GDAL directly
    // But... an error happens in this code.
    // Simulate a GDAL error that happens once the PDAL work is done.        
    CPLError(CE_Fatal , 0, "This error makes the app crash.");
    std::cout << "Bye" << std::endl;
}
```
Now executing this application displays:

```
(PDAL Work Debug) Executing pipeline in standard mode.
(PDAL Work writers.text Debug) Writing header to filename: test.txt
(Error) GDAL failure (0) This error doesn't make the app crash.
```
The PDAL stages logged properly in the `std::stringstream`.
The "GDAL error" in the `makeSomePDALWork` is also logged properly in the `std::stringstream`.
The "GDAL error" in the `main` isn't logged because the app crashed as it tried to log into a deallocated stream.

**Expected behavior**
I expect my application not to crash if a GDAL errors happens after I used some PDAL code.

**Personal investigation so far**
I see that PDAL use a singleton logger locked by a mutex (ErrorHandler).
When this singleton is created, [it sets itself as the GDAL Error Callback](https://github.com/PDAL/PDAL/blob/2.1.0/pdal/GDALUtils.cpp#L290).
I also see [it sets back this handler to the GDAL default handler when it is destroyed](https://github.com/PDAL/PDAL/blob/2.1.0/pdal/GDALUtils.cpp#L299).
When I saw this 2 lines, I had this question: _What is the reason behind having PDAL override the GDAL default error handler?_

To me, the real reason of why this crash occurs is that in the Stage, [when it starts logging, the global error handler's stream is set to be the stream of the Stage Log object](https://github.com/PDAL/PDAL/blob/2.1.0/pdal/Stage.cpp#L400) but [when it stops logging, there is nothing that unset this stream](https://github.com/PDAL/PDAL/blob/2.1.0/pdal/Stage.cpp#L404), making possible for the global error handler to attempt to log into an invalid stream.

To me there is something missing to [the stop logging method of a stage](https://github.com/PDAL/PDAL/blob/2.1.0/pdal/Stage.cpp#L404): if the start logging sets the header, log level and output stream of the global error handler, then the stop logging should reset all of these information.

**Suggestion of fix**
Now that this is identified, what do you think about setting PDAL global error handler as GDAL's error handler while stages are logging and reset the GDAL error handler to the previous handler (or the default one) when stages stop logging?

It happened on all platforms I tried on: MacOS, Linux, Windows.
The libraries I use are compiled inhouse, the dependency graph being:
- PDAL 2.1.0
- gdal 3.1.0
- libgeotiff 1.6.0
- proj 7.0.1
- LASzip 3.4.3
- laz-perf 1.4.0

Although using PDAL 2.1.0, I suspect this crash would also happens with 2.2.0.

abellgithub:
Close #3244

### DIFF
--- a/pdal/Stage.cpp
+++ b/pdal/Stage.cpp
@@ -440,6 +440,7 @@ void Stage::startLogging() const
 void Stage::stopLogging() const
 {
     m_log->popLeader();
+    gdal::ErrorHandler::getGlobalErrorHandler().clear();
 }
 
 std::istream& operator>>(std::istream& in, Stage::WhereMergeMode& mode)
--- a/pdal/private/gdal/ErrorHandler.cpp
+++ b/pdal/private/gdal/ErrorHandler.cpp
@@ -34,9 +34,6 @@
 
 #include <mutex>
 
-#include <cpl_conv.h>
-#include <cpl_error.h>
-
 #include "ErrorHandler.hpp"
 
 namespace pdal
@@ -56,10 +53,6 @@ ErrorHandler& ErrorHandler::getGlobalErrorHandler()
     return s_gdalErrorHandler;
 }
 
-namespace
-{
-}
-
 namespace {
 
 //ABELL - No idea why this is __stdcall
@@ -77,7 +70,7 @@ void trampoline(::CPLErr code, int num, char const* msg)
 /**
   Constructor for a GDAL error handler.
 */
-ErrorHandler::ErrorHandler() : m_errorNum(0)
+ErrorHandler::ErrorHandler() : m_errorNum(0), m_prevHandler(nullptr)
 {
     std::string value;
 
@@ -85,9 +78,6 @@ ErrorHandler::ErrorHandler() : m_errorNum(0)
     const char* set = CPLGetConfigOption("CPL_DEBUG", "");
     m_cplSet = (bool)set;
     m_debug = m_cplSet;
-
-    // Push on a thread-local error handler
-    CPLSetErrorHandler(&trampoline);
 }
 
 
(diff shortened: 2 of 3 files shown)


hobu (approved):
fun!

PR #3439, opened 2021-04-27
Forward query/headers properly during HEAD requests

connormanning:
For HTTP based drivers, in one spot query/header parameters were not forwarded to the endpoint.

Also fixes this warning:
```
../vendor/arbiter/arbiter.cpp:1287:14: note: use reference type 'const std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> &' to prevent copying
        for (const auto d : walk(pre)) dirs.push_back(d + post);
             ^~~~~~~~~~~~~~
                        &
```

### DIFF
--- a/vendor/arbiter/arbiter.cpp
+++ b/vendor/arbiter/arbiter.cpp
@@ -100,7 +100,7 @@ namespace
 
         return merge(in, config);
     }
-    
+
     inline bool iequals(const std::string& s, const std::string& s2)
     {
         if (s.length() != s2.length())
@@ -683,7 +683,7 @@ LocalHandle Endpoint::getLocalHandle(
         const std::string local(tmp + basename);
         if (isHttpDerived())
         {
-            if (auto fileSize = tryGetSize(subpath))
+            if (auto fileSize = tryGetSize(subpath, headers, query))
             {
                 std::ofstream stream(local, streamFlags);
                 if (!stream.good())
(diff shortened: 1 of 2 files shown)


vendor/arbiter/arbiter.cpp
```diff
@@ -1386,15 +1402,38 @@ std::unique_ptr<Http> Http::create(Pool& pool)
 }
 
 std::unique_ptr<std::size_t> Http::tryGetSize(std::string path) const
+{
+    return tryGetSize(path, http::Headers());
+}
+
+std::size_t Http::getSize(
```
abellgithub, line 1409: It seems like this is new. Is there some reason for it to be added that I'm missing?

connormanning: I'm not sure it's quite "new" as it's the same code body as before.  Just with missing overloads added (which already exist for the get/put functions, just not for the HEAD request calls) which forward the query/header params.  Otherwise you get a 401 error every time you hit [this line](https://github.com/PDAL/PDAL/pull/3439/files#diff-642075337fbe162193f18e3871532a17abb3e333a7d0bfedd57a66fb905b07dbR686) if your EPT endpoint is authenticated with a query or header parameter value.

abellgithub: Hmm. On my diff I don't see any `getSize()` in the old stuff, just `tryGetSize()`. I wouldn't care except this throws.

connormanning: Yes that's true, this specific overload of `getSize` is new - it's to bring the header/query-forwarding versions of get/getBinary/getSize/tryGet/tryGetBinary/tryGetSize to parity with the normal versions.

The generic `Driver` has [all of these](https://github.com/PDAL/PDAL/blob/master/vendor/arbiter/arbiter.hpp#L3751-L3781), which is what I mean by this function not being "new", per se.

The HTTP version is intended to have overloads of each of these which have the same functionality but also forward query/header params, but the `size` ones were missing from [here](https://github.com/PDAL/PDAL/blob/master/vendor/arbiter/arbiter.hpp#L4042-L4072).

So I added those overloads (the `getSize` one is not actually used, but now exists for consistency) and then updated the `getLocalHandle` function, which accepts query/header params, to use this new forwarding `tryGetSize` rather than omitting these params.

PR #3445, opened 2021-05-04
Add TileDB Hilbert configuration

normanb:
Added specifying hilbert ordering for TileDB sparse arrays and test that checks for this ordering.

### DIFF
--- a/plugins/tiledb/io/TileDBWriter.cpp
+++ b/plugins/tiledb/io/TileDBWriter.cpp
@@ -76,30 +76,6 @@ struct TileDBWriter::Args
     bool m_append;
 };
 
-std::string attributeDefaults(R"(
-{
-    "coords":[
-        {"compression": "bit-shuffle"},
-        {"compression": "gzip", "compression_level": 9}
-    ],
-    "Intensity":{"compression": "bzip2", "compression_level": 5},
-    "ReturnNumber": {"compression": "zstd", "compression_level": 7},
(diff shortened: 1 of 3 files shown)


plugins/tiledb/io/TileDBWriter.cpp
```diff
@@ -367,16 +367,18 @@ void TileDBWriter::ready(pdal::BasePointTable &table)
...
                 {{dimMin, dimMax}}))
                 .add_dimension(tiledb::Dimension::create<double>(*m_ctx, "Z",
                 {{dimMin, dimMax}}));
+            m_schema->set_cell_order(TILEDB_HILBERT);
+        }
     #endif
 #endif
-        m_schema->set_domain(domain).set_order(
-            {{TILEDB_ROW_MAJOR, TILEDB_ROW_MAJOR}});
+        m_schema->set_domain(domain);
```
abellgithub, line 381: Does this remove the ordering for TILEDB version < 2? Does this matter?

normanb: No, it uses the default ordering for TileDB version < 2.

normanb:
Ran tests against TileDB 1.7 and I now specify the tile sizes in the older test code as going forward we default to using hilbert ordering within TileDB and PDAL which doesn't require this.

This won't impact older builds of TileDB reading data. It is a requirement to specify tile sizes when writing data and not using the default ordering scheme.

PR #3453, opened 2021-05-13
Fix CI for Conda

abellgithub:


.github/workflows/conda.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         platform: ['ubuntu-latest','windows-latest','macos-latest']
 
     env:
-      PLATFORM: ${{ matrix.platform }}
```
hobu, line 16: I was just curious if the name `PLATFORM` was actually messing with anything. 👍 on the name change though.

hobu (approved):

.github/workflows/conda.yml
abellgithub: vcvarsall.bat sets that variable (using a different case arrangement), which was the root of our problem.

hobu: ah. brutal.

PR #3459, opened 2021-05-18
Lazperf

abellgithub:


hobu (approved):

pdal/compression/LazPerfVlrCompression.cpp
```diff
@@ -206,23 +205,61 @@ void LazPerfVlrCompressor::done()
...
+        uint64_t chunkTablePos;
+        in >> chunkTablePos;
+
+        in.seek(chunkTablePos);
+        uint32_t version;
+        uint32_t numChunks;
+        in >> version;
+        in >> numChunks;
+
+        assert(version == 0);
```
connormanning, line 225: Should this be a `pdal_error` instead?  I'm not quite clear on what the `version` means in this context so it could certainly be fine as-is.  But if version `1` could come out sometime in the future with a different expected behavior, then a runtime error would seem appropriate.

connormanning (approved):

pdal/compression/LazPerfVlrCompression.cpp
abellgithub: I might have put that in for my own sanity? It's Martin's versioning thing for the LASzip VLR.

I think your idea is a good one if I can find an easy way to do it.

PR #3467, opened 2021-05-31
Generalize filters.transformation to all homographies.

cffk:
Include the normalizing step in filters.transformation to allow general
homographies and update the documentation.  All the examples in the
documentation are unaffected (since the last row of the matrix is
[0,0,0,1]).  Homographies allow (painfully!), perspective images of a
point cloud to be created by following the transformation filter by
writers.gdal.  This setup is not ideal, because the effective camera
matrix isn't stored with the the output image.

### DIFF
--- a/doc/stages/filters.transformation.rst
+++ b/doc/stages/filters.transformation.rst
@@ -3,11 +3,8 @@
 filters.transformation
 ======================
 
-The transformation filter applies an arbitrary rotation+translation
+The transformation filter applies an arbitrary homography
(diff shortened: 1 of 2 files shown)


hobu:
> This setup is not ideal, because the effective camera
matrix isn't stored with the the output image.

A `forward` option a la `writers.las` metadata feature for `writers.gdal` that wrote the RPC matrix would seem quite useful to add.

hobu:
bouncing for CI churn

abellgithub:
Thanks for this.

PR #3470, opened 2021-06-04
ICP Tune-Up and Scale Option

pjhartzell:
Tune-Up:
- Moved the application of the initial transformation guess to before the "fixed" and "moving" clouds are demeaned by the "fixed" cloud centroid. This means the initial guess is now relative to the coordinate origin of the "moving" cloud, not to the "fixed" cloud centroid (as it was before). Open for debate on this. I can see how making an initial guess relative to the centroid location of the "fixed" cloud is intuitive if the "fixed" and "moving" clouds cover the same area (one is not much larger/smaller than the other in terms of areal extent and they are not significantly displaced). But if this is not the case, or if a user has a rough transformation matrix from a prior coarse registration (I stumbled over this issue for this use case), then I think it makes more sense for the initial guess to be a "true" transformation in terms of the coordinate values of the "moving" cloud.
- The MSE (mean square error) computation was an average distance error, rather than an average squared distance error. I updated the MSE computations to be an average squared distance error.
- Corrected the matrix multiplication order when updating the transformation between iterations.

Scale Option:
- Added an option to include a single scale factor (same scale factor applied in all three dimensions) in the solved transformation.

### DIFF
--- a/doc/stages/filters.icp.rst
+++ b/doc/stages/filters.icp.rst
@@ -120,21 +120,23 @@ max_iter
 max_similar
   Maximum number of similar transforms to consider converged. [Default: **0**]
 
-max_dist
-  Maximum distance between closest points to be included. [Default: no maximum]
-
-init
-  Initial transformation matrix guess. A whitespace-delimited matrix presented in row-major order. Only matrices with sixteen elements are allowed. [Default: no guess]
-
 mse_abs
-  Absolute change in MSE threshold. [Default: **1e-12**]
+  Absolute change in MSE convergence threshold. [Default: **1e-12**]
 
 rt
-  Rotation threshold. [Default: **0.99999**]
+  Rotation convergence threshold. [Default: **0.99999**]
 
 tt
-  Translation threshold. [Default: **9e-8**]
+  Translation convergence threshold. [Default: **9e-8**]
 
+max_dist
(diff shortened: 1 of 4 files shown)


abellgithub:
If the behavior was *wrong*, and we're fixing a bug, then that's fine, but I don't feel good about changing behavior that some might have come to expect. If this is a mixture of those two things, when we should separate it out -- one PR for bug fixes and one for behavior changes that we can discuss and perhaps drive with an option.

I don't have enough expertise in this to make specific recommendations without spending some time with it.

@chambbj ?

pjhartzell:
Yeah, the PR is a mixture of bugs/corrections and feature addition (scale option). Not sure which one changing the initial guess reference location falls under. I went into it with the thought that it was a bug (because it didn't fit with the way I wanted to use it), but unsure now. Regardless, I can separate out the bugs/corrections from the new features (scale option, initial guess reference point) into different PRs.

You can close this and I'll resubmit separate PRs.

chambbj:
It's fine to separate this into multiple PRs.

Maybe it's hard to call the initial guess transformation order a bug (unless we explicitly said one thing and did another), but I do think that @pjhartzell's suggestion makes the most sense. I do fear that users may expect that it's being applied first, only to find out later that the data is being centered first. At the same time, perhaps through trial and error, other users have adapted to the post-centering assumption and would be impacted by such a change.

PR #3479, opened 2021-06-21
support getting and setting dataset-level GDAL metadata in writers.gdal and readers.gdal

hobu:


io/GDALWriter.cpp
```diff
@@ -321,6 +326,21 @@ void GDALWriter::doneFile()
...
 
     getMetadata().addList("filename", m_filename);
+
+    std::vector<std::string> gdalitems = Utils::split(m_GDAL_metadata, ',');
+    for (auto& v: gdalitems)
+    {
+
+        std::vector<std::string> item = Utils::split(v, '=');
+        std::string name = item[0];
+        std::string value = item[1];
```
connormanning, line 336: Would it be reasonable to handle `=` characters in the `value`?  In particular I'm thinking of base64 data e.g. `foo=ABCD==,bar=WXYZ==`.  As long as keys may not contain `=` which certainly seems to be the case, this is unambiguous since we've already split on commas, and seems simple enough to implement.

hobu: It's GDAL that uses `=` as its split character, so I was just following the pattern there.

connormanning: This is UB for a specification without a right-hand side (e.g. `foo=,bar=`), right?  Maybe something of the form `std::string value = item.size() > 1 ? item[1] : ""`.

connormanning: > It's GDAL that uses = as its split character, so I was just following the pattern there.

Yeah that's fine - I'm saying that we could make sure we're using the *first* `=` as the split character.  As implemented, something like `expression=a+b=c` will store `expression` -> `a+b` in the GDAL metadata, since we seem to be discarding portions of the `value` following any `=` chars.

hobu: Patches graciously accepted.

connormanning: I think something like this would solve both of these issues, although I haven't compiled it.
```
// If no "=" character is found, then the name is the entire token and the value is an empty string.
const std::size_t pos = item.find_first_of("=");
const std::string name = item.substr(0, pos);
const std::string value = pos != std::string::npos ? item.substr(pos + 1) : "";
raster.addMetadata(name, value);
```

abellgithub: You can just check that `split` returns two things, no? Error otherwise.

connormanning: But why disallow `=` in the value, is what I was getting at.  Seems arbitrary.

abellgithub: Because it's most likely that someone is just doing something stupid. IMO simple is good unless proved otherwise.

io/GDALReader.hpp
```diff
@@ -70,6 +70,7 @@ class PDAL_DLL GDALReader : public Reader , public Streamable
     std::unique_ptr<gdal::Raster> m_raster;
     std::vector<Dimension::Type> m_bandTypes;
     std::vector<Dimension::Id> m_bandIds;
+    std::vector<std::string> m_GDAL_metadata;
```
abellgithub, line 73: Usually we call this `StringList`, but it doesn't matter.

io/GDALWriter.cpp
```diff
@@ -161,6 +166,9 @@ void GDALWriter::initialize()
     // don't expand by point if we're running in standard mode.  That's
     // set later in writeView.
     m_expandByPoint = !m_fixedGrid;
+
```
abellgithub, line 169: Unnecessary blank lines.

io/GDALWriter.cpp
```diff
@@ -321,6 +329,45 @@ void GDALWriter::doneFile()
         throwError(raster.errorMsg());
 
     getMetadata().addList("filename", m_filename);
+
+    std::vector<std::string> gdalitems = Utils::split(m_GDAL_metadata, ',');
+    for (auto& v: gdalitems)
+    {
+        const std::size_t pos = v.find_first_of("=");
```
abellgithub, line 336: I'm not sure why we'd accept '=' in a value. Does GDAL? Seems error-prone.

io/GDALWriter.cpp
```diff
@@ -321,6 +329,45 @@ void GDALWriter::doneFile()
...
+        if (pos != std::string::npos)
+        {
+            const std::string name = v.substr(0, pos);
+            const std::string value = pos != std::string::npos ? v.substr(pos + 1) : "";
```
abellgithub, line 340: Is it really OK to have empty metadata values?

io/GDALWriter.cpp
```diff
@@ -321,6 +329,45 @@ void GDALWriter::doneFile()
...
+            raster.addMetadata(name, value);
+        }
+    }
+
+    getMetadata().addList(raster.getMetadata());
```
abellgithub, line 345: I don't think you want arrays as output. The only reason `addList` exists is to create an array of one item.

io/GDALWriter.cpp
```diff
@@ -321,6 +329,45 @@ void GDALWriter::doneFile()
...
+
+void GDALWriter::readyTable(PointTableRef table)
+{
+    // Add these as GDAL metadata
+    if(m_writePDALMetadata)
+    {
+        MetadataNode m = table.metadata();
+        std::string json = Utils::toJSON(m);
+        std::vector<uint8_t> metadata(json.begin(), json.end());
+        std::string b64 = Utils::base64_encode(metadata.data(), metadata.size());
```
abellgithub, line 356: I don't understand why you'd be encoding a string to a string. Why not just save the string without encoding?

io/GDALWriter.cpp
```diff
@@ -321,6 +329,45 @@ void GDALWriter::doneFile()
...
+            m_GDAL_metadata += ",";
+
+        m_GDAL_metadata += "pdal_metadata=" + b64;
+
+        std::ostringstream ostr;
+        PipelineWriter::writePipeline(this, ostr);
+        json = ostr.str();
+        std::vector<uint8_t> pipeline(json.begin(), json.end());
+        b64 = Utils::base64_encode(pipeline.data(), pipeline.size());
+        m_GDAL_metadata += ",pdal_pipeline=" + b64;
```
abellgithub, line 367: Same here.

io/GDALWriter.cpp
hobu: someone giving us base64 to put into GDAL's metadata is going to be giving us '='.

io/GDALWriter.cpp
hobu: > Why not just save the string without encoding?

Because if they give us JSON or any other weird thing we don't have to worry about escaping it upon output anywhere – it's just a blob of crap that the downstream user is responsible for.

io/GDALWriter.cpp
hobu: a GDAL raster might have multiple metadata entries. This puts them all in the `raster` metadata entry as an array of key/value items, which is what we want.

io/GDALWriter.cpp
hobu: conflicted on this. It is tested in the tests, btw.

It is possible that GDAL gives us an empty item, and I wanted to be able to support it, or users might wonder where their keys went.

io/GDALWriter.cpp
abellgithub: I don't care, but isn't that just an object?

```
#include <iostream>

#include <pdal/Metadata.hpp>
#include <pdal/PDALUtils.hpp>

int main()
{
    using namespace pdal;

    MetadataNode raster("raster");

    raster.add("foo", "fooval");
    raster.add("bar", "barval");

    MetadataNode root("readers.gdal");
    root.add(raster);

    Utils::toJSON(root, std::cerr);

    return 0;
}
```
```
{
  "readers.gdal":
  {
    "raster":
    {
      "bar": "barval",
      "foo": "fooval"
    }
  }
}
```

hobu: updated as suggested.

PR #3485, opened 2021-06-24
Catch bounds exceptions

### LINKED ISSUE #3484: Exceptions thrown in attempt to parse bounds in Metadata::inferType()
Parsing bounds can throw an exception that is should be caught in Metadata::inferType().

abellgithub:
Close #3484

### DIFF
--- a/pdal/Metadata.cpp
+++ b/pdal/Metadata.cpp
@@ -75,15 +75,25 @@ std::string Metadata::inferType(const std::string& val)
 
     BOX2D b2d;
     std::istringstream iss1(val);
-    iss1 >> b2d;
-    if (iss1.good())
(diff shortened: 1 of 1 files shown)


hobu:
It isn't clear why we don't want these errors and why we want to now drop them on the floor.

abellgithub:
If there's an exception, the parsing failed and that means that the string isn't of the correct type (a BOX2D or BOX3D in this case). I was going to add a comment but when looking at the entire function, it seems pretty clear since most of the other cases are similar.

hobu (approved):

abellgithub:
Just for clarification, I don't believe that we used to throw an exception. This was new behavior where the exception handling got missed.

PR #3490, opened 2021-07-02
Eptwriter

abellgithub:
This only does standard mode.


io/EptWriter.hpp
```diff
@@ -0,0 +1,66 @@
+/******************************************************************************
+* Copyright (c) 2021, Hobu Inc. (info@hou.co)
```
hobu, line 2: hobu.co typo

PR #3491, opened 2021-07-03
Fix compilation error: E57XmlParser.cpp:149:39: error: ‘numeric_limits’ is not a member of ‘std’

nickmcummins:
This occurs when compiling with the E57 plugin enabled:
```
/home/nick/Code/Git/github.com/PDAL/plugins/e57/libE57Format/src/E57XmlParser.cpp: In member function ‘virtual XMLSize_t E57FileInputStream::readBytes(XMLByte*, XMLSize_t)’:
/home/nick/Code/Git/github.com/PDAL/plugins/e57/libE57Format/src/E57XmlParser.cpp:149:39: error: ‘numeric_limits’ is not a member of ‘std’
  149 |         const int64_t size_max = std::numeric_limits<size_t>::max();
      |                                       ^~~~~~~~~~~~~~
/home/nick/Code/Git/github.com/PDAL/plugins/e57/libE57Format/src/E57XmlParser.cpp:149:60: error: expected primary-expression before ‘>’ token
  149 |         const int64_t size_max = std::numeric_limits<size_t>::max();
      |                                                            ^
/home/nick/Code/Git/github.com/PDAL/plugins/e57/libE57Format/src/E57XmlParser.cpp:149:66: error: no matching function for call to ‘max()’
  149 |         const int64_t size_max = std::numeric_limits<size_t>::max();

```


abellgithub:
What compiler/version?

nickmcummins:
`CMAKE_C(XX)_COMPILER` is `/usr/bin/c++` and `/usr/bin/cc`:
```
➜  ~ pacman -Qo /usr/bin/cc       
/usr/bin/cc is owned by gcc 11.1.0-1
➜  ~ cc --version
cc (GCC) 11.1.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

This is just the default version of GCC from the distro package. I don't see anything in [Development > Compilation > Unix Compilation](https://pdal.io/development/compilation/unix.html) that states a particular version of GCC that should be used.

abellgithub:
This fix is fine. On other systems this header seems to come from another included file.
If we're missing a build error, I like to know what the circumstances are. What system is this?

abellgithub:
Ah, looking at the use seems you must be on a 32-bit system, which is unusual for us. On 64 bit systems the code that uses `limits` is ignored by the compiler.

PR #3498, opened 2021-07-23
TileDB time traveling

normanb:
This PR adds [TileDB time traveling](https://docs.tiledb.com/main/solutions/tiledb-embedded/internal-mechanics/reading#time-traveling) to both the PDAL TileDB reader and writer.

I have added a test that shows how the total array has 200 points but each TileDB fragment only contains 100 points.

### DIFF
--- a/plugins/tiledb/io/TileDBReader.cpp
+++ b/plugins/tiledb/io/TileDBReader.cpp
@@ -108,7 +108,7 @@ void TileDBReader::addArgs(ProgramArgs& args)
     args.add("end_timestamp", "TileDB array timestamp", m_endTimeStamp,
         point_count_t(0));
     args.addSynonym("end_timestamp", "timestamp");
-#if TILEDB_VERSION_MAJOR > 1 || TILEDB_VERSION_MINOR >= 3
+#if TILEDB_VERSION_MAJOR > 2 || (TILEDB_VERSION_MAJOR == 2 && TILEDB_VERSION_MINOR >= 3)
     args.add("start_timestamp", "TileDB array timestamp", m_startTimeStamp,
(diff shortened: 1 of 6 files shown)


plugins/tiledb/io/TileDBReader.cpp
```diff
@@ -105,6 +105,13 @@ void TileDBReader::addArgs(ProgramArgs& args)
     args.add("stats", "Dump TileDB query stats to stdout", m_stats, false);
     args.add("bbox3d", "Bounding box subarray to read from TileDB in format "
         "([minx, maxx], [miny, maxy], [minz, maxz])", m_bbox);
+    args.add("end_timestamp", "TileDB array timestamp", m_endTimeStamp,
+        point_count_t(0));
+    args.addSynonym("end_timestamp", "timestamp");
```
abellgithub, line 110: Here `timestamp` is made a synonym for `end_timestamp`, but the doc text makes it seem as if it's a synonym for `start_timestamp`, but this is also tied to the TileDB version somehow. Maybe the doc just needs some clarification?

normanb: Agreed, I went back and forth on how to name the time interval. I have updated the docs.

abellgithub:
I'll merge this after I fix CI.

abellgithub:
@normanb : Please see the compilation error here: https://github.com/PDAL/PDAL/runs/3144419444?check_suite_focus=true

abellgithub:
@normanb : This looks OK to me. Are you done with changes?

normanb:
Hi @abellgithub, yes, I am all done with the changes. Thanks for the review.

PR #3500, opened 2021-07-24
Fix the bug of 'fromNative' and 'toNative' when the input string containing "Chinese characters" in win32

RealityTracer:
To support the filename contain "Chinese characters" properly, the default constructor through 'std::wstring_convert<std::codecvt_utf8_utf16<unsigned short>, unsigned short>' should be replaced with 'std::codecvt_byname<wchar_t, char, std::mbstate_t>'

### DIFF
--- a/pdal/util/FileUtils.cpp
+++ b/pdal/util/FileUtils.cpp
@@ -76,17 +76,21 @@ std::string addTrailingSlash(std::string path)
 }
 
 #ifdef _WIN32
+class chs_codecvt : public std::codecvt_byname<wchar_t, char, std::mbstate_t>
+{
+public:
+    chs_codecvt() : codecvt_byname("CHS") {}
+};
 inline std::string fromNative(std::wstring const& in)
 {
     // TODO: C++11 define convert with static thread_local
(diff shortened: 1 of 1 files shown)


abellgithub:
This needs some doc. to explain what's going on and a test. Unfortunately, the online documentation of `codecvt_byname` is thin.

RealityTracer:
"codecvt_byname " is thin? Dude, please see [https://en.cppreference.com/w/cpp/locale/codecvt_byname](url), good luck! From my experiences, lots of open src libs(such as CloudCompare) often made the mistake that they could not load the files whose names contain some "Hanzi”, which are wide string!
"

RealityTracer:
To reproduce the bug:
(1)You should install the Chinese Language Supports 
(2) And install  a input software that  can type "Hanzi" by your keyboard.
(3)Then rename the laz/las file with some "Hanzi".
(4)load the renamed file by pdal

hobu:
This change is not going to be merged without 

1) tests
2) CI configuration that exercises these tests

hobu:
Please reopen this when there are tests that work with PDAL's [CI platform](https://pdal.io/project/integration.html)

PR #3505, opened 2021-08-04
Update EPT reader to handle origin query of EPT version > 1.0.0.

connormanning:
See [comment](https://github.com/PDAL/PDAL/commit/415f2b39d293fd338b13353f58c39ad1647eaebf#diff-ec87b1770b47916667540e9c2a9eb78701ffdf6c7ae12ba895ed1e1e73c40fbeR284-R299) for description of version changes.  Updated the origin query handler in the EPT reader, updated one of the test datasets to the new version, with unit tests for both versions.

One note: the search comparison changed from using `id` to using `path`, for *both* versions.  Even for the older version, the `path` key should have been used instead of `id` - as the `id` is just a pointer into the detailed metadata file and is more of an implementation detail, while the intent here is to search by source file path.  In most cases there is no practical difference because the `id` is (usually but not guaranteed to be) derived from the `path`, but searching based on the `path` is more accurate.


abellgithub:
I don't see this in the online spec.

PR #3328, opened 2021-01-07
Fix type of arguments of `ifstream`/`ofstream`

giordano:
The first argument of `ifstream`/`ofstream` must be `string`, not `wstring`:
https://en.cppreference.com/w/cpp/io/basic_ifstream.

PR #3507, opened 2021-08-06
FileUtils: MSVC-specific wstring fstream ctor

xantares:
the wide-string variant is an msvc-specific extension:
https://www.cplusplus.com/reference/fstream/ifstream/ifstream/

xref #3328

### DIFF
--- a/cmake/win32_compiler_options.cmake
+++ b/cmake/win32_compiler_options.cmake
@@ -4,6 +4,10 @@
 
 function(pdal_target_compile_settings target)
     set_property(TARGET ${target} PROPERTY CXX_STANDARD 11)
+    if (MINGW)
+      # for ostream/istream with wstring
+      set_property(TARGET ${target} PROPERTY CXX_STANDARD 17)
+    endif ()
     set_property(TARGET ${target} PROPERTY CXX_STANDARD_REQUIRED TRUE)
     target_compile_definitions(${target} PRIVATE
         -DWIN32_LEAN_AND_MEAN)
--- a/pdal/util/FileUtils.cpp
+++ b/pdal/util/FileUtils.cpp
@@ -45,7 +45,12 @@
 #include <codecvt>
 #endif
 
+#if __cplusplus >= 201703L
+#include <filesystem>
+namespace pdalboost { namespace filesystem = std::filesystem; namespace system = std; }
+#else
 #include <boost/filesystem.hpp>
+#endif
 
 #include <pdal/util/FileUtils.hpp>
 #include <pdal/util/Utils.hpp>
@@ -83,7 +88,7 @@ inline std::string fromNative(std::wstring const& in)
     auto p = reinterpret_cast<unsigned short const*>(in.data());
     return convert.to_bytes(p, p + in.size());
 }
-inline std::wstring toNative(std::string const& in)
+inline pdalboost::filesystem::path toNative(std::string const& in)
 {
     // TODO: C++11 define convert with static thread_local
     std::wstring_convert<std::codecvt_utf8_utf16<unsigned short>, unsigned short> convert;
@@ -119,7 +124,12 @@ std::istream *openFile(std::string const& filename, bool asBinary)
     if (asBinary)
         mode |= std::ios::binary;
 
+#if __cplusplus >= 201703L
+    std::filesystem::path native(toNative(name));
+    ifs = new std::ifstream(native, mode);
+#else
     ifs = new std::ifstream(toNative(name), mode);
+#endif
     if (!ifs->good())
     {
         delete ifs;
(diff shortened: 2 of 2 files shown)


hobu:
This will not be merged without tests that exercise it.

xantares:
The msvc code-path stays the same, this only enables build for mingw.

abellgithub:
I'm not inclined to make these changes as such. _MSC_VER is a *compiler version*, not an operating system tag. There has been *limited* interest in mingw before, but I'm not sure there's enough demand to take on the support burden. Rather than just submit a bunch of PRs, some discussion how/if this should be best handled.

xantares:
yes, these changes are meant to allow compilation with the mingw compiler which, like msvc also targets win32
and there are some differences between the two, mingw tends to stick to standards but also sometimes does not fully cover the win32 apis

I thought the PR path is ok as most changes are minimal, how do you propose we proceed then ?

hobu:
Are all of these changes in support of an open source project or something? While we can merge many of them, without a mingw CI platform, they're just going to rot again.

xantares:
yes I wanted to try building qgis
if this is accepted maybe I can setup a ci job

hobu:
Is mingw the supported build path for QGIS @wonder-sk? 

> if this is accepted maybe I can setup a ci job

It's the other way. For us to merge all this stuff, we want a mingw builder. Otherwise it is just going to rot, and we don't know how to test it or make it work. Without any of that stuff, it is just more code to break and maintain.

abellgithub:
Why would one want a mingw build of QGIS when native windows builds exist?

giordano:
Someone already [tried to answer](https://github.com/PDAL/PDAL/issues/3329#issuecomment-779357313).

abellgithub:
Right, but I think we're talking about a project that already has Windows and *nix CI and development support.

Keno:
The question here is not as much mingw vs MSVC or Windows vs non-Windows as it is libc++ or GNU libstdc++ vs Microsoft STL, since the latter has a proprietary extension here. gcc only works with the former, MSVC only with the latter, but there are compilers that's support both. Which one you want depends on the platform choice of your host application. For example over at julialang.org (which @giordano is supporting), our windows builds are standardized to libstdc++, so anybody wanting to use this library with Julia will require this patch or an equivalent (which we currently carry locally). Now, it would be nice if there was a preprocessor define to detect the std library independent of the compiler, but there currently isn't. Not that this isn't a problem in practice, since compilers that support both actually do set `_MSC_VER` when targeting the MS STL, even if they're not MSVC, so this patch is correct. We've in the past considered offering upstream CI with our build configurations, since we support a lot of them (> 30) that often upstream developers don't have access to, but unfortunately we're not yet in a position to do so. Still, merging this patch would be useful since it would mean support for libstdc++ standardized applications without extra patches.

hobu:
Our primary concern with all of these patches is being put on the hook to support a configuration we do not currently support on a platform that is not currently tested. That concern goes away if there's a CI that tests this configuration and platform.

I'm sympathetic to wanting to have these patches be upstreamed, and we're not trying to be intransigent, but a blizzard of drive-by PRs with little context, no promise of future support, no linkage to a downstream issue, or lack of demonstration of a bug doesn't engender enthusiasm to maintain this forever. If @xantares declares themselves forever the mingw maintainer and a CI worker is built that highlights when we should nag them to fix it 😄, we don't have any issue with this stuff. 

Blindly merging stuff that has the potential to break stuff we know to work in the configurations we currently support makes us nervous.

xantares:
For the configurations you already support I suppose it's already well tested, I see CI and everything, so it should be fine.
For mingw usually these platforms quirks are very low maintenance, and dependencies, flags etc dont change often.

hobu (changes requested):
Reviewing the actual content of the PR, this represents a behavior change. The patch should be to `toNative` instead of splatting an `#ifdef` right here. `toNative` is used in many other places.

wonder-sk:
> Is mingw the supported build path for QGIS @wonder-sk?

Apparently yes since ~Feb 2021, there is a mingw build in qgis CI:
https://github.com/qgis/QGIS/blob/master/.github/workflows/mingw64.yml
But that build does not include PDAL
https://github.com/qgis/QGIS/blob/master/ms-windows/mingw/build.sh

But on windows, only MSVC builds of QGIS are the ones shipped to users.

xantares:
I came with another solution: we enable c++17 on mingw with pdalboost::filesystem as an alias to std::filesystem.
Eventually it will allow to drop boost completely when the oldest supported version of msvc allows it (currently it seems to be vs2015/c++11).

abellgithub:
Believe this should be fixed with #3536

PR #3508, opened 2021-08-06
FileUtils: std::codecvt_utf8_utf16 unavailable on mingw

xantares:
fallback to boost::filesystem already needed

ref https://github.com/PDAL/PDAL/issues/3329

### DIFF
--- a/pdal/util/FileUtils.cpp
+++ b/pdal/util/FileUtils.cpp
@@ -79,14 +79,14 @@ std::string addTrailingSlash(std::string path)
 inline std::string fromNative(std::wstring const& in)
 {
     // TODO: C++11 define convert with static thread_local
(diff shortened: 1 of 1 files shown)


hobu (changes requested):
We have been working over the years to remove the vendored boost implementation and stuff like this makes that harder. Is there no system facility in mingw to do this? I get that it is easy to reach for boost here and be done with it, but we don't want that dependency going forward.

xantares:
turns out if I use char16_t instead of unsigned integer it builds ok

abellgithub:
I don't have any problem with this change, but how does this help MinGW? If `unsigned short` != `uint16_t` I would think there would be some other issues as well.

abellgithub:
Fixed by #3536

PR #3509, opened 2021-08-06
portable_endian.hpp: Add missing mingw definitions

xantares:
ref #3326

### DIFF
--- a/pdal/util/portable_endian.hpp
+++ b/pdal/util/portable_endian.hpp
@@ -54,6 +54,14 @@
 #   include <winsock2.h>
                      
 #   if BYTE_ORDER == LITTLE_ENDIAN
+
+// Add missing definitions for MinGW.
+#ifndef htonll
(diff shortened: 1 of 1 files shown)


pdal/util/portable_endian.hpp
```diff
@@ -54,6 +54,14 @@
 #   include <winsock2.h>
                      
 #   if BYTE_ORDER == LITTLE_ENDIAN
+
```
hobu, line 57: Where did the implementation for htonll and ntohll come from? Please add a comment.

hobu (changes requested):

pdal/util/portable_endian.hpp
xantares: i think from so:
https://stackoverflow.com/questions/16375340/c-htonll-and-back

abellgithub:
Fixed with #3528

PR #3511, opened 2021-08-06
GreedyProjection: Fix M_PI mingw define

xantares:
we must be sure to include cmath before eigen

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -295,7 +295,7 @@ target_link_libraries(${PDAL_BASE_LIB_NAME}
 )
 target_compile_definitions(${PDAL_BASE_LIB_NAME}
     PRIVATE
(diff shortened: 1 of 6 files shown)


hobu:
> we must be sure to include cmath before eigen

Does eigen not work correctly or something? Does it need its own mingw patch? Remembering the magical incantation to set _USE_MATH_DEFINES and `#include <cmath>` every time we might use eigen is going to be extremely brittle. We're going to forget to do so the very next time we use eigen.

xantares:
we can define that macro via cmake instead

abellgithub:
What is the real problem here? That isn't described or explained.

xantares:
@abellgithub on mingw M_PI was not defined because of some include order presumably

abellgithub:
Fixed by similar #3523

PR #3512, opened 2021-08-06
SparseMatrix: __pragma is msvc-specific

xantares:
we want to use the gnu variant on mingw


hobu (approved):
This should be upstreamed to kazhdan, but our experience is he doesn't merge stuff.

abellgithub:
There is usually a compiler option to ignore unknown pragmas. That should be used instead of this confusing stuff.

xantares:
we dont want to ignore omp pragmas

abellgithub:
Fixed by #3533

PR #3513, opened 2021-08-06
DynamicLibrary: Use dlfcn on mingw

xantares:
these apis exist on mingw too but cast to void* is an msvc extension, just assume dlfcn is available

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,6 +293,17 @@ target_link_libraries(${PDAL_BASE_LIB_NAME}
     INTERFACE
         ${PDAL_LIBDIR}
 )
+
+if (MINGW)
+  find_library (DL_LIBS NAMES dl)
+  set (CMAKE_DL_LIBS ${DL_LIBS})
+endif ()
+if (CMAKE_DL_LIBS)
+  target_link_libraries(${PDAL_BASE_LIB_NAME}
+      PRIVATE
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -293,6 +293,14 @@ target_link_libraries(${PDAL_BASE_LIB_NAME}
     INTERFACE
         ${PDAL_LIBDIR}
 )
+
+find_library (DL_LIBRARY NAMES dl)
```
hobu, line 297: Use CMAKE_DL_LIBS instead of this.

hobu (changes requested):

CMakeLists.txt
xantares: ok

hobu: You don't have to find_library with CMAKE_DL_LIBS. See https://trac.osgeo.org/qgis/ticket/1939 for an example usage.

xantares: this applies to unix only

hobu: The `CMAKE_DL_LIBS` variable should be populated with what you need for the dl functions. You shouldn't have to find_library anything. CMake does that for you for this particular library because it can be so wonky on the different platforms.

abellgithub:
Addressed by #3534

PR #3516, opened 2021-08-06
Set POSITION_INDEPENDENT_CODE for vendored Arbiter

xantares:


vendor/arbiter/CMakeLists.txt
```diff
@@ -20,19 +20,20 @@ target_compile_definitions(${PDAL_ARBITER_LIB_NAME}
     PRIVATE
         ARBITER_CURL)
 
-if (MSVC)
-    target_link_libraries(${PDAL_ARBITER_LIB_NAME} PRIVATE Shlwapi.lib)
+set_target_properties(${PDAL_ARBITER_LIB_NAME} PROPERTIES POSITION_INDEPENDENT_CODE ON)
```
hobu, line 23: Why was this change needed? Messing with our PIC settings often causes trouble. If this change is needed just for MINGW, it should be under a conditional, not change it for everything.

CMakeLists.txt
```diff
@@ -274,6 +274,7 @@ target_include_directories(${PDAL_BASE_LIB_NAME}
         ${GDAL_INCLUDE_DIR}
         ${LASZIP_INCLUDE_DIR}
 )
+
```
hobu, line 277: No whitespace changes please

vendor/arbiter/CMakeLists.txt
xantares: relocatable libs are needed on mingw too, so we enable it on every platform with the POSITION_INDEPENDENT_CODE option

then cmake knows to use the right flag depending on the compiler

hobu: CMake doesn't always work equally for all platforms, however. Especially older versions like those that are installed by older distributions. We are not ready to disrupt things here. Please put this under a MINGW conditional.

xantares: the main CMakeLists.txt requires version 3.5, I think this is enough : this was introduced in 2.8.9 dating from 2013

hobu: Please put it in the mingw conditional for now.

hobu (changes requested):

CMakeLists.txt
xantares: ok

vendor/arbiter/CMakeLists.txt
xantares: the thing is it should be also enabled for unix builds, that's why there's no condition

hobu (approved):

PR #3520, opened 2021-08-27
Improvements to filters.litree

### LINKED ISSUE #3518: filters.litree routine stuck on first tree
**Description**
I made a pipeline to 
1. read from an EPT source
2. get height above ground with `filters.hag_delauney`
3. apply the Li tree segmentation routine (`filters.litree`)
4. write a .las file
 
I'm trying to use this on small areas, but everywhere I try, the Li segmentation routine always gets stuck on the first tree. Can't figure out what is wrong here, looks like a bug? Saw another user on stack exchange who might have the [same issue](https://gis.stackexchange.com/questions/396400/using-litree-in-pdal#comment658362_403125).

<details>
**Pipeline**

```
[
  {
    "type": "readers.ept",
    "filename": "https://s3-us-west-2.amazonaws.com/usgs-lidar-public/USGS_LPC_CA_NoCAL_Wildfires_B1_2018/ept.json",
    "bounds": "([-13444124.366465904, -13443980.986961761], [4772585.959571438, 4772718.552748669])"
  },
  {
    "type": "filters.hag_delaunay"
  },
  {
    "type": "filters.litree",
    "min_points": 10,
    "min_height": 3,
    "radius": 100
  },
  {
    "type": "writers.las",
    "filename": "pdal_request_seg.las",
    "extra_dims": "all"
  }
]
```

**Command**
```
$ pdal pipeline C:/Users/Johnathan/Downloads/pdal_request_seg.json --debug 
```

**Result**
```
(pdal pipeline writers.las Info) writers.las: Writing dimension ClassFlags(uint8_t)  to LAS extra bytes.
(pdal pipeline writers.las Info) writers.las: Writing dimension HeightAboveGround(double)  to LAS extra bytes.
(pdal pipeline writers.las Info) writers.las: Writing dimension TreeID(int64_t)  to LAS extra bytes.
(pdal pipeline Debug) Executing pipeline in standard mode.
(pdal pipeline filters.litree Debug) Classifying points for tree 1 (|Ui| = 107543)
```

**System Info**

PDAL Version: 2.2.0 (git-version: Release), installed through Conda
System: Microsoft Windows [Version 10.0.18363.836]

chambbj:
- Roughly 8x speedup by eliminating repetitive KDtree builds and automatically assigning points to Ni if they are a set distance away (laterally) from the apex of the current tree Pi.
- Update docs -- descending sort of HeightAboveGround required.

Future improvements _could_ include:

- sorting by HeightAboveGround internally, and not permanently changing the point order
- using a fixed seed for the dummy point to Ni, that is outside the bounds of the point cloud, eliminating yet another search

Fixes #3518

### DIFF
--- a/filters/LiTreeFilter.cpp
+++ b/filters/LiTreeFilter.cpp
@@ -167,16 +167,16 @@ void LiTreeFilter::classifyPoint(PointId ui, PointView& view, PointIdList& Ni,
         double ux = view.getFieldAs<double>(Id::X, ui);
         double uy = view.getFieldAs<double>(Id::Y, ui);
 
-        std::vector<double> dists;
+        double mindist = std::numeric_limits<double>::max();
         for (PointId const& i : idx)
         {
             double nx = view.getFieldAs<double>(Id::X, i);
             double ny = view.getFieldAs<double>(Id::Y, i);
-            double dist =
-                std::sqrt((nx - ux) * (nx - ux) + (ny - uy) * (ny - uy));
-            dists.emplace_back(dist);
+            double dist = (nx - ux) * (nx - ux) + (ny - uy) * (ny - uy);
+            if (dist < mindist)
+                mindist = dist;
         }
-        return *std::min_element(dists.begin(), dists.end());
+        return std::sqrt(mindist);
     };
     double dmin1 = mindist(Pi);
     double dmin2 = mindist(Ni);
@@ -242,8 +242,8 @@ void LiTreeFilter::segmentTree(PointView& view, PointIdList& Ui,
     {
         double ux = view.getFieldAs<double>(Id::X, ui);
         double uy = view.getFieldAs<double>(Id::Y, ui);
-        double d = std::sqrt((ux - tx) * (ux - tx) + (uy - ty) * (uy - ty));
-        if (d < 10.0)
+        double d = (ux - tx) * (ux - tx) + (uy - ty) * (uy - ty);
+        if (d < 100.0)
             classifyPoint(ui, view, Ni, Pi);
         else
             Ni.push_back(ui);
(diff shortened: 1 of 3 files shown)


filters/LiTreeFilter.cpp
```diff
@@ -162,29 +162,30 @@ void LiTreeFilter::classifyPoint(PointId ui, PointView& view, PointIdList& Ni,
...
+    auto mindist = [&ui, &view](PointIdList idx) {
+        double ux = view.getFieldAs<double>(Id::X, ui);
+        double uy = view.getFieldAs<double>(Id::Y, ui);
 
-    PointViewPtr Pview = view.makeNew();
-    for (PointId const& p : Pi)
-        Pview->appendPoint(view, p);
-    KD2Index& Ptree = Pview->build2dIndex();
+        std::vector<double> dists;
+        for (PointId const& i : idx)
```
abellgithub, line 171: I must be missing something here, but I don't understand why we're building a list and then extracting the minimum value, rather than just checking the current `dist` against the minimum that we've found so far.

filters/LiTreeFilter.cpp
```diff
@@ -244,8 +236,18 @@ void LiTreeFilter::segmentTree(PointView& view, PointIdList& Ui,
...
 
     // "For iteration i, we classify each point in Ui as Pi or Ni."
+    double tx = view.getFieldAs<double>(Id::X, t0);
+    double ty = view.getFieldAs<double>(Id::Y, t0);
     for (PointId const& ui : Ui)
-        classifyPoint(ui, view, Ni, Pi);
+    {
+        double ux = view.getFieldAs<double>(Id::X, ui);
+        double uy = view.getFieldAs<double>(Id::Y, ui);
+        double d = std::sqrt((ux - tx) * (ux - tx) + (uy - ty) * (uy - ty));
```
abellgithub, line 245: Not sure why we're taking square roots just for a comparison since we seem to throw the value away after that.

chambbj: Actually meant to do this, both here and above.

filters/LiTreeFilter.cpp
chambbj: You aren't missing anything per se. It's just the way I happened to implement it. I can switch.

PR #3526, opened 2021-09-02
Add more meaningful capability macros

abellgithub:
@xantares: Do you see any issues with this?


xantares:
I dont see the point of these, but this is correct code.

abellgithub:
This point is that for us, who don't do MinGW, the difference between _MSC_VER and _WIN32 isn't obvious. I'm trying to make sure that changes that get made for MinGW don't just get removed because someone (me, perhaps), didn't realize that _MSC_VER means something different from _WIN32 and that it's actually about MinGW.

Thanks.

PR #3534, opened 2021-09-08
Use dlopen and such for mingw

abellgithub:


CMakeLists.txt
```diff
@@ -277,6 +277,7 @@ target_include_directories(${PDAL_BASE_LIB_NAME}
 target_link_libraries(${PDAL_BASE_LIB_NAME}
     PRIVATE
         ${CMAKE_THREAD_LIBS_INIT}
+        ${CMAKE_DL_LIBS}
```
hobu, line 280: Why did this move from the util lib to the base lib?

abellgithub: Because it wasn't used there, AFAICT.

PR #3538, opened 2021-09-09
More mingw fixes

xantares:


abellgithub:
I'm confused by the dynamiclibrary changes. Wasn't the point to use the unix functions with mingw?

xantares:
Indeed I meant to use dlopen previously, but its detection with cmake was complicated.
It turns out a simple cast to void* made it build on mingw so I decided to switch to the windows api for simplicity.

abellgithub:
OK. Perhaps you should PR cmake for MinGW so that you can use CMAKE_DL_LIBS without thinking about it, but it makes no difference to me.

abellgithub:
Is this still WIP? Do you want it merged?

xantares:
@abellgithub This can be merged

xantares:
@abellgithub can you review it ?

pdal/util/FileUtils.cpp
```diff
@@ -431,6 +436,7 @@ std::vector<std::string> glob(std::string path)
     if (path[0] == '~')
         throw pdal::pdal_error("PDAL does not support shell expansion");
 
+#ifdef _WIN32
```
abellgithub, line 439: What is this bit doing?

abellgithub:
Sorry. I think we were having CI problems and then I kind of forgot about it. Can you help me understand the one bit I noted? Thx.

pdal/util/FileUtils.cpp
xantares: I dropped this part at it was incomplete

xantares:
> Sorry. I think we were having CI problems and then I kind of forgot about it. Can you help me understand the one bit I noted? Thx.

No problem, I dropped that part. I also added a patch to silmplify the dimbuilder part when cross-compiling.

pdal/util/FileUtils.cpp
dg0yt: FTR I did some independent work as part of an attempt to update pdal in vcpkg. For `glob` in mingw, I switch to the A functions:

https://github.com/dg0yt/vcpkg/blob/ce74d9653848718f1044b3a5a10d17a7a8f004e6/ports/pdal/fix-mingw.patch#L104-L149

cmake/dimension.cmake
```diff
@@ -5,15 +5,9 @@
...
 
-# In cross compiling scenarios, the location of the native DIMBUILDER_EXECUTABLE
-# must be provided to the CMake invocation.
-IF (CMAKE_CROSSCOMPILING)
-    SET(DIMBUILDER "${DIMBUILDER_EXECUTABLE}")
-ELSE (CMAKE_CROSSCOMPILING)
-    SET(DIMBUILDER dimbuilder)
-ENDIF (CMAKE_CROSSCOMPILING)
-
+# In cross compiling scenarios, you might want to set CMAKE_CROSSCOMPILING_EMULATOR
```
abellgithub, line 8: @hobu: You just did this, right? I don't know what CMAKE_CROSSCOMPILING_EMULATOR is even after reading the doc (no surprise, given this is cmake doc). Or @xantares, can you explain?

hobu: I do know that my use of `CMAKE_CROSSCOMPILING` worked for me, and I had seen usage of it around. IDK if that's the prescribed way to do this though. Happy to have a patch by anyone that does a lot more cross compilation stuff.

xantares: @abellgithub for example when I cross-compile from a linux host to windows target I just set CMAKE_CROSSCOMPILING_EMULATOR to the wine executable, then cmake automagically knows that it has to run the dimbuilder.exe rule with wine, that simplifies the cmake logic.

xantares:
@dg0yt I integrated your FindFirstFileA code, thanks

cmake/dimension.cmake
abellgithub: @xantares: I'm not sure how this will work with general X-compiling, but I confess that I usually run from cmake magic.

@hobu: I know if this will address your conda M1 issue. Can you check before we commit this? I don't want to break what you just got going.

dg0yt: `wine` may run from Linux magic. 

For general cross-build, I would like to know if I could use a host `dimbuilder`?

hobu: The conda-forge cross build scenario is building on a osx-x86_64 host for osx-arm64. In that case, we need to build and execute `dimbuilder` on the host machine's architecture (x86_64), run that, and then continue compiling for arm64. We're not running in an emulator scenario, so I'm not sure if `CMAKE_CROSSCOMPILING_EMULATOR` applies.

xantares: in that case we should give the choice to use either DIMBUILDDER_EXECUTABLE or CMAKE_CROSSCOMPILING_EMULATOR, which wasnt possible previously.

xantares: in that case we could offer to choose either DIMBUILDER_EXECUTABLE or CMAKE_CROSSCOMPILING_EMULATOR:
use

abellgithub:
Can we maybe just move the cross-compiler thing to another PR and I can commit this one?

xantares:
if you want

PR #3552, opened 2021-09-30
Add cached ept plugin

gui2dev:
When using pdal in some applications that are intensely reading entwine dataset, it appears some already fetched data could be cached.

This could be done in many way, this implementation is one amongst many that could be improved in many ways.

The idea is to use a [thread safe LRU cache](https://github.com/tstarling/thread-safe-lru), build around TBB.

The plugin is merely a copy of `readers.ept` but caches `TilesContents` and all the hierarchy fetched as `NL::json`.

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -64,6 +64,11 @@ option(BUILD_PLUGIN_RIVLIB
 add_feature_info("RiVLib plugin" BUILD_PLUGIN_RIVLIB
     "read data in the RXP format")
 
+option(BUILD_PLUGIN_CACHEDEPT
+    "Choose if cached ept support should be built" FALSE)
+add_feature_info("CachedEpt plugin" BUILD_PLUGIN_CACHEDEPT
+    "read and cache entwine data")
+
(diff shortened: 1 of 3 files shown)


hobu:
ok, this is really large and duplicates the entire `readers.ept` driver. We're not going to merge it as-is due to the fact that it duplicates everything.  We would be happy to merge it to https://github.com/PDAL/unsuppported-plugins as-is, however.

I'm curious what use case had you spending time on this scenario. Do you have a long running application?

gui2dev:
Making it a static stage and not a plugin would duplicate less code, but it felt wrong to force a link to tbb.

There may be other options as it exists a [boost version of lru cache](https://www.boost.org/doc/libs/1_77_0/boost/compute/detail/lru_cache.hpp) that might not be thread safe though.

We are dealing with massive point clouds, that we use to digitize assets (mainly railway infrastructure). We are making a lot of sections to achieve that.
I'll be more than happy to discuss this in greater detail if you wish.

gui2dev:
Moved to [unsuported](https://github.com/PDAL/unsuppported-plugins/pull/3)

PR #3553, opened 2021-09-30
Add smrmsg reader

gui2dev:
Applanix POSPac MMS software produces a file that describes the accuracy of the post-processed solution that contains the position, orientation and velocity RMS after smoothing.

This reader aims to allow pdal user to read that file.


hobu:
Does this need to exist as its own driver, or would it be sufficient to update `readers.sbet` with these capabilities? It would be less to maintain going forward if so.

gui2dev:
That was my first intend  but I could not find a proper way to have both files `GPSTime` to match.
It seems that there are fewer records in smrmsg than in sbet.

hobu:
Thanks for this contribution, but we need tests before we will merge it. If you are able to add some and catch up to latest PDAL master, it can still make the 2.4 release window. Otherwise, it will be kicked ahead to 2.5.

gui2dev:
Will do.

gui2dev:
@hobu it's done.

hobu (approved):

PR #3554, opened 2021-09-30
add Polygon intersection method

gui2dev:
Having the possibility to build the intersection of two polygons might come handy.

This could also be an override of a virtual method from `pdal::Geometry` if prefered.


abellgithub:
I would suggest moving this function into your own code. There are public functions to fetch/set the OGR geometry. I think we're resisting the temptation to recreate an OGR interface in PDAL.

gui2dev:
Seems legit. Well it'll always be there if needed one day.

PR #3556, opened 2021-10-04
Add changes for Time Dim in TileDBReader and Writer

lex-c:


pdal/Stage.cpp
```diff
@@ -1,4 +1,4 @@
-/******************************************************************************
+ /******************************************************************************
```
abellgithub, line 1: This should be reverted.

pdal/util/Bounds.hpp
```diff
@@ -307,7 +307,7 @@ class PDAL_DLL BOX2D
 /**
   BOX3D represents a three-dimensional box with double-precision bounds.
 */
-class PDAL_DLL BOX3D : private BOX2D
+class PDAL_DLL BOX3D : public BOX2D
```
abellgithub, line 310: I tend to thing that all these BOX4D changes should ride with TileDB since I don't see them being used generically. @hobu?

plugins/tiledb/io/TileDBWriter.cpp
```diff
@@ -526,16 +558,27 @@ void TileDBWriter::ready(pdal::BasePointTable &table)
 
 bool TileDBWriter::processOne(PointRef& point)
 {
+    m_array->close();
```
abellgithub, line 561: I have lost track of the development of this writer, but it seems strange to do all this open/close of the array each time you want to write a point.

pdal/util/Bounds.hpp
hobu: IDK. Are we ever going to add time indexing to PDAL? I kind of thing not.

abellgithub: I think that when there is more need, it can move. I don't see any reason for this in a public interface.

normanb:
@abellgithub @hobu apologies, this was meant to be a PR within our fork. Good feedback though as I had already made a box4d implementation, I will move it within the TileDB plugin.

hobu:
No worries!

Issue #3555, opened 2021-10-02
Failing tests with PROJ 8.1.1 


gsakkis:
The "floating" CI jobs have started [failing](https://github.com/TileDB-Inc/PDAL/runs/3777084360?check_suite_focus=true) for two tests, `pdal_io_ept_reader_test` and `pdal_filters_crop_test`. My guess is this due to the PROJ upgrade from 8.0.1 to 8.1.1.  

Conda package diff between the latest successful floating job and the failing one:
```diff
-  geotiff            conda-forge/linux-64::geotiff-1.6.0-h4f31c25_6
+  geotiff            conda-forge/linux-64::geotiff-1.7.0-hcfb7246_3

-  libgdal            conda-forge/linux-64::libgdal-3.3.2-h399a533_1
+  libgdal            conda-forge/linux-64::libgdal-3.3.2-h9c9eb65_4

-  libspatialite      conda-forge/linux-64::libspatialite-5.0.1-h8694cbe_6
+  libspatialite      conda-forge/linux-64::libspatialite-5.0.1-h8796b1e_9

-  proj               conda-forge/linux-64::proj-8.0.1-h277dcde_0
+  proj               conda-forge/linux-64::proj-8.1.1-h277dcde_2
```

hobu:
```
2021-10-02T15:46:00.1418775Z 45: Expected equality of these values:
2021-10-02T15:46:00.1419841Z 45:   eptNp
2021-10-02T15:46:00.1420827Z 45:     Which is: 45
2021-10-02T15:46:00.1421386Z 45:   47u
2021-10-02T15:46:00.1421922Z 45:     Which is: 47
2021-10-02T15:46:00.1422752Z 45: /home/runner/work/PDAL/PDAL/test/unit/io/EptReaderTest.cpp:773: Failure
2021-10-02T15:46:00.1423941Z 45: Expected equality of these values:
2021-10-02T15:46:00.1424677Z 45:   sourceNp
2021-10-02T15:46:00.1425264Z 45:     Which is: 45
2021-10-02T15:46:00.1425804Z 45:   47u
2021-10-02T15:46:00.1426351Z 45:     Which is: 47
2021-10-02T15:46:00.1427422Z 45: [  FAILED  ] EptReaderTest.boundedCropReprojection (112 ms)
2021-10-02T15:46:00.1428578Z 45: [ RUN      ] EptReaderTest.ogrCrop
[...]

PR #3557, opened 2021-10-05
Streamable iterator

gsakkis:
This PR adds `PipelineStreamableExecutor` and `StreamableIterator` classes to enable streamable execution in PDAL Python (related PR coming soon). The `Streamable` and `PipelineExecutor` classes were also refactored extensively, along with a few minor necessary changes to `PipelineManager` and `PointView`.

### DIFF
--- a/pdal/PipelineExecutor.cpp
+++ b/pdal/PipelineExecutor.cpp
@@ -140,4 +140,28 @@ std::string PipelineExecutor::getLog() const
 }
 
 
+point_count_t PipelineStreamableExecutor::execute()
+{
+    point_count_t count = 0;
+    while (PointViewPtr view = executeNext())
+        count += view->size();
+    return count;
+}
+
+
+PointViewPtr PipelineStreamableExecutor::executeNext()
+{
+    if (!m_itPtr)
+        m_itPtr = std::unique_ptr<StreamableIterator>(m_managerPtr->executeStream());
+
+    StreamableIterator& it = *m_itPtr;
+    if (!it) {
+        m_executed = true;
+        return nullptr;
+    }
+    ++it;
+    return *it;
+}
+
+
 } //namespace pdal
--- a/pdal/PipelineExecutor.cpp
+++ b/pdal/PipelineExecutor.cpp
@@ -39,11 +39,13 @@ namespace pdal
 {
 
 
-PipelineExecutor::PipelineExecutor(std::string const& json)
-    : m_json(json)
+PipelineExecutor::PipelineExecutor(std::string const& json, point_count_t streamLimit)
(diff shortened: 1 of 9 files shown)


abellgithub:
This is what I was trying to discuss and perhaps we had a disconnect. I'm hesitant to modify the Streamable/Stage interfaces without good reason and understanding the intent. What are we trying to achieve with these changes, exactly?

gsakkis:
> This is what I was trying to discuss and perhaps we had a disconnect. I'm hesitant to modify the Streamable/Stage interfaces without good reason and understanding the intent. 

There is no change in the Streamable/Stage interfaces; all tests are passing (except for 2 unrelated failures also [failing on master](https://github.com/PDAL/PDAL/issues/3555)).

> What are we trying to achieve with these changes, exactly?

Running streamable pipelines from Python in streaming mode (i.e. prevent running out of memory), coupled with a Pythonic [iterator API](https://github.com/TileDB-Inc/python/tree/PipelineIterator#iterator-api). More examples/tests [here](https://github.com/TileDB-Inc/python/blob/PipelineIterator/test/test_pipeline.py#L203).

gsakkis:
> Maybe I'm missing something but I think this can be done without changing any PDAL code by adding a stage to extract the data at the end of the pipeline. 

I'm not quite sure what you're proposing. Do you mean without changing the PDAL `Streamable` code and adding something like a `writers.numpy_memoryview` plugin instead?

> When all points in the point table have been processed, clear() is called, which provides a hook to do any processing you would want to do in Python.

`clear()` is a Table method, not a `Stage` method; did you mean `done()`? Either way this is called by `Streamable::execute()` deep inside a loop. How would I yield back to Python the current filled NDArray while `Streamable::execute()` is still in progress?

abellgithub:
> I'm not quite sure what you're proposing. Do you mean without changing the PDAL Streamable code and adding something like a writers.numpy_memoryview plugin instead?

Pretty much, yes.

After I wrote last night I couldn't really see a reason why we couldn't just create a StreamPointTable that was backed by an NDArray and avoid the copying altogether when working in stream mode.

But even with this see two things missing that weren't really anticipated:

1) Allow a user to pass a table into the PipelineManager. I think this is a simple change.
2) Provide some interruption when `clear()` is called. It seems to beg for coroutines in this use case (first time I've ever seen a real use for coroutines in practice), but we could do this by executing PDAL in a thread. It would be even more slick to yield the active NDArray to Python while you continue processing with PDAL with another NDArray, but this may be too much.

Maybe some combination of this stuff with what you've done would work? I would love to keep base code change to a minimum. Let me know your thoughts or if I should try to clarify things with some code or more words.

abellgithub:
Closing after much discussion and alternate implementation.

PR #3559, opened 2021-10-06
fix box3d equal and clip

lex-c:
I think the equal in BOX3D was accidentally referencing the contains rather than equal from BOX2D for x and y. I also put in a little edit for the clip in BOX3D, where I think the signs were reversed (minz and maxz fixed to less and more than other.minz and other.maxz, respectively).

### DIFF
--- a/pdal/util/Bounds.hpp
+++ b/pdal/util/Bounds.hpp
@@ -424,7 +424,7 @@ public:
     */
     bool equal(const BOX3D& other) const
     {
-        return  BOX2D::contains(other) &&
+        return  BOX2D::equal(other) &&
             minz == other.minz && maxz == other.maxz;
     }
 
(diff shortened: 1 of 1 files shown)


abellgithub:
Ah, the problem with writing boilerplate code that you don't use and don't test.

This looks OK, but if you could add a line or two to the tests that would be great.

test/unit/BoundsTest.cpp
```diff
@@ -114,19 +114,39 @@ TEST(BoundsTest, test_clip)
...
-    // BUG: seems wrong -- need to better define semantics of clip, etc
-    // .clip() can make an invalid bounds, this should be fixed.
-    BOX2D r6(20,6, 40,8);
+    EXPECT_TRUE(r1 == r4);
 
-	EXPECT_DOUBLE_EQ(r1.minx, 20);
-	EXPECT_DOUBLE_EQ(r1.maxx, 6);
-	EXPECT_DOUBLE_EQ(r1.miny, 40);
-	EXPECT_DOUBLE_EQ(r1.maxy, 8);
+    BOX2D r6(8, 5, 5, 2);
```
abellgithub, line 122: This seems unused.

abellgithub (commented):
If you merge master into your branch now, it should all work and I'll merge these changes.

Thanks again.

test/unit/BoundsTest.cpp
lex-c: I think that's all, Andrew. Thank you!

lex-c:
All done now.

PR #3562, opened 2021-10-13
Add time dim in TileDB Reader and Writer plugins

calexat-123:
@abellgithub, this is ready for review. We added support for 4D reads/writes with extensions to Bounds and BOX3D in the plugin and extra args and some more tests for the changes.

### DIFF
--- a/plugins/tiledb/test/TileDBWriterTimeDimTest.cpp
+++ b/plugins/tiledb/test/TileDBWriterTimeDimTest.cpp
@@ -58,6 +58,10 @@ Options getTileDBOptions()
     options.add("z_domain_end", 10);
     options.add("time_domain_st", 0);
     options.add("time_domain_end", 10);
+    options.add("x_tile_size", 10);
+    options.add("y_tile_size", 10);
+    options.add("z_tile_size", 10);
+    options.add("time_tile_size", 10);
 
     return options;
 }


plugins/tiledb/io/Bounds4D.hpp
```diff
@@ -0,0 +1,464 @@
...
+****************************************************************************/
+
+#pragma once
+
+#include <pdal/util/Bounds.hpp>
+
+namespace pdal
+{
+
+class BOX2D_ : public BOX2D
```
abellgithub, line 42: It looks like this class exists just for the constructor that takes BOX3D, right? I guess I'd just make a note.

FYI, this was explicitly disabled because it lead to problems elsewhere, but if this meets your needs, that's fine.

plugins/tiledb/io/Bounds4D.hpp
```diff
@@ -0,0 +1,464 @@
...
+
+        oss.precision(precision);
+        oss.setf(std::ios_base::fixed, std::ios_base::floatfield);
+
+        oss << "box4d(" << minx << " " << miny << " " << minz << " " << mintm << ", " <<
+        maxx << " " << maxy << " " << maxz << " " << maxtm << ")";
+        return oss.str();
+    }
+
+    std::string toWKT(uint32_t precision = 8) const
```
abellgithub, line 178: Oh my, this is awful, isn't it? (Not your doing, just the nature of the beast.) Is it necessary? If it's just here for completeness maybe just drop it?

plugins/tiledb/io/TileDBReader.cpp
```diff
@@ -312,6 +318,10 @@ void TileDBReader::localReady()
         if (numDims == 2)
             m_query->set_subarray({m_bbox.minx, m_bbox.maxx,
                 m_bbox.miny, m_bbox.maxy});
+        else if (numDims == 4 && m_has_time)
```
abellgithub, line 321: This is a little weird. If you have 4 dims and no time, do you want it to fall into the case below? Should there be an explicit test for the case below rather than `else`?

plugins/tiledb/test/XYZTmUtils.cpp
```diff
@@ -0,0 +1,220 @@
...
+                                        1., 1., 1., 1));
+    args.add("xyz_mode", "mode for distribution of xyz dimension values", m_xyz_mode, Mode::Uniform);
+    args.add("time_mode", "mode for distribution of time dimension values", m_tm_mode, Mode::Ramp);
+    args.add("use_time", "Add a time dimension in addition to XYZ (default true)", m_use_time, true);
+    args.add("density", "Double value to set density dimension in points", m_density, 1.0);
+}
+
+void XYZTimeFauxReader::prepared(PointTableRef table)
+{
+    if (!m_countArg->set())
```
abellgithub, line 69: Maybe I'm missing something but this doesn't seem to inherit from FauxReader and I don't see `m_count` or `m_countArg` defined.

plugins/tiledb/io/Bounds4D.hpp
lex-c: That's right, Andrew; and I guess this may not be the best way even locally; I think virtual inheritance is probably the right way for us to implement this, so I'll try to change it.

plugins/tiledb/test/XYZTmUtils.cpp
lex-c: They're inherited from the Reader class. I saw the FauxReader set up like that so I just kind of used that pattern.

PR #3565, opened 2021-10-18
Export beam origin and direction

gui2dev:
rivlib `pointcloud` class allow access to `double[3] beam_direction` and `double[3] beam_origin` which are respectively the beam direction and origin of the current laster shot.

Then each taget from `std::vector<target> targets` contains :
> `float vertex;`
Cartesian coordinates in scanners own coordinate system. vertex[i] = beam_origin[i] + echo_range * beam_direction[i]
`double echo_range;`
The distance from the laser origin to the target. Note that this is not the same as sqrt(vertex[0]^2 + vertex[1]^2 + vertex[2]^2).

Exposing these values, enables further MTA correction.

### DIFF
--- a/pdal/Dimension.json
+++ b/pdal/Dimension.json
@@ -538,17 +538,17 @@
     "description": "Beam origin Z coordinate of current laser shot"
     },
     {
-    "name": "BeamDirectoyX",
+    "name": "BeamDirectionX",
     "type": "double",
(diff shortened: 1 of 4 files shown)


abellgithub:
This all looks fine, but won't this break for earlier versions of the rivlib software?

PR #3567, opened 2021-10-20
PipelineExecutor: read pipeline once

### LINKED ISSUE #3566: PipelineExecutor should read pipeline once
[PipelineExecutor::validate](https://github.com/PDAL/PDAL/blob/master/pdal/PipelineExecutor.cpp#L89) and [PipelineExecutor::execute](https://github.com/PDAL/PDAL/blob/master/pdal/PipelineExecutor.cpp#L99) call `m_manager.readPipeline` even if the pipeline has been read already. This used to be undetected before https://github.com/PDAL/PDAL/commit/f469b0fe52c66911c58c3552eee42a9d746f1c89 but now it is (correctly) logged as a ["Pipeline has multiple leaf nodes"](https://github.com/PDAL/PDAL/blob/master/pdal/PipelineReaderJSON.cpp#L142) error. To fix this, `readPipeline` should not be called more than once.

**Note**: although it is tempting to move the `readPipeline` call to the constructor, it should stay in `validate` and `execute` (as it is now) in order to allow the user to call `PipelineExecutor::setLogLevel` before reading the pipeline. Calling `setLogLevel` after the pipeline is read has no effect (by the way this could be considered a separate bug to be fixed, or at least trigger a warning/error/exception that it has no effect).

gsakkis:
Fixes #3566 (+ minor `PipelineExecutor` refactoring).

### DIFF
--- a/pdal/PipelineExecutor.cpp
+++ b/pdal/PipelineExecutor.cpp
@@ -41,6 +41,7 @@ namespace pdal
 
 PipelineExecutor::PipelineExecutor(std::string const& json)
     : m_json(json)
+    , m_read(false)
     , m_executed(false)
     , m_logLevel(pdal::LogLevel::Error)
 {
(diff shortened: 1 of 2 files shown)


abellgithub:
I think the reported issue can be fixed by simply adding:

`m_manager.destroyStage()` before the pipeline is read.

If there is an issue with logging (and I think there is), please open a ticket describing the problem. If you want to submit a PR that fixes only that issue, that would be fine.

abellgithub:
Closing. Fixed by #3587

PR #3569, opened 2021-10-20
FindTileDB.cmake: improvements

rouault:
Get changes from GDAL's https://github.com/rouault/gdal/blob/import_cmake4gdal/cmake/modules/packages/FindTileDB.cmake

- Remove hardcoded paths to /usr and /usr/local
- use TILEDB_ROOT variable
- change the name of the package to TileDB (instead of TILEDB) in
  FIND_PACKAGE_HANDLE_STANDARD_ARGS() to fix a warning emitted by CMake
  3.19 regarding the different in case with the FindTileDB.cmake
  file name
- Add a TileDB::TileDB import target

### DIFF
--- a/cmake/modules/FindTileDB.cmake
+++ b/cmake/modules/FindTileDB.cmake
@@ -10,6 +10,9 @@
 # TILEDB_LIBRARY     = where to find the hexer library.
 # TILEDB_VERSION     = version of library which was found, e.g. "1.4.1"
 #
+# IMPORTED targets
+# This module defines the following IMPORTED target: ``TileDB::TileDB``
+#
 # Copyright (c) 2019 TileDB, Inc
 #
 # Redistribution and use is allowed according to the terms of the BSD license.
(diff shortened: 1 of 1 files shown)


abellgithub:
@normanb : I'm going to let you deal with this. Let me know if/when you want it merged.

normanb:
@rouault TileDB uses cmake targets, you should be able to use `find_package` https://cmake.org/cmake/help/latest/command/find_package.html to find TileDB. We provide `TileDB::tiledb_shared` and `TileDB::tiledb_static`. Include directories and things available in the "standard" way: $<TARGET_PROPERTY:TileDB::tiledb_shared,INTERFACE_INCLUDE_DIRECTORIES>.

I should update this file and see if we can do the same in GDAL. Thanks for the changes, I suggest we merge this in and look to improve this later.

rouault:
> TileDB uses cmake targets, you should be able to use `find_package` https://cmake.org/cmake/help/latest/command/find_package.html to find TileDB. We provide `TileDB::tiledb_shared` and `TileDB::tiledb_static`. Include directories and things available in the "standard" way: $<TARGET_PROPERTY:TileDB::tiledb_shared,INTERFACE_INCLUDE_DIRECTORIES>.

indeed, that works. So I guess this PR isn't really needed.

PR #3578, opened 2021-11-02
Custom PointTable for PipelineManager/PipelineExecutor

### LINKED ISSUE #3564: Can't use custom point table with PipelineManager
It's a bit of a pain that you can't use a custom point table with the PipelineManager and would be easy to support.

gsakkis:
Fixes #3564. Based on top of #3567

### DIFF
--- a/pdal/PipelineExecutor.cpp
+++ b/pdal/PipelineExecutor.cpp
@@ -39,14 +39,19 @@ namespace pdal
 {
 
 
-PipelineExecutor::PipelineExecutor(std::string const& json)
-    : m_manager(0)
+PipelineExecutor::PipelineExecutor(std::string const& json, SimplePointTable* tablePtr)
+    : m_manager(tablePtr, new FixedPointTable(0))
     , m_executed(false)
     , m_json(json)
     , m_read(false)
 {
     setLogLevel(pdal::LogLevel::Error);
 }
+
+
+PipelineExecutor::PipelineExecutor(std::string const& json) :
+    PipelineExecutor(json, new ColumnPointTable())
+{}
 
 
 std::string PipelineExecutor::getPipeline() const
--- a/pdal/PipelineExecutor.cpp
+++ b/pdal/PipelineExecutor.cpp
@@ -41,6 +41,7 @@ namespace pdal
 
 PipelineExecutor::PipelineExecutor(std::string const& json)
     : m_json(json)
+    , m_read(false)
     , m_executed(false)
     , m_logLevel(pdal::LogLevel::Error)
 {
@@ -84,24 +85,29 @@ std::string PipelineExecutor::getSchema() const
 
 bool PipelineExecutor::validate()
 {
-    std::stringstream strm;
-    strm << m_json;
-    m_manager.readPipeline(strm);
+    read();
     m_manager.prepare();
-
     return true;
 }
 
-int64_t PipelineExecutor::execute()
+point_count_t PipelineExecutor::execute()
 {
-    std::stringstream strm;
-    strm << m_json;
-    m_manager.readPipeline(strm);
+    read();
     point_count_t count = m_manager.execute();
-
     m_executed = true;
-
     return count;
+}
+
+
+void PipelineExecutor::read()
+{
+    if (!m_read)
+    {
+        std::stringstream strm;
+        strm << m_json;
+        m_manager.readPipeline(strm);
+        m_read = true;
+    }
 }
 
 
@@ -138,4 +144,3 @@ std::string PipelineExecutor::getLog() const
 
 
 } //namespace pdal
-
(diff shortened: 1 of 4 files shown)


abellgithub:
George,

I've tried to express this a couple of times before, but we're not just going to change the PDAL API. We *might* extend it, but we're not going to change the existing API without really good reason and I don't see one here. There are certainly some things that we would like to change in retrospect, but we don't always have the foresight to do things perfectly.

I have no issue with adding a `read()` function to `PipelineExecutor` and the internal changes surrounding that, but otherwise, I just don't see any of this as necessary. If there are bits here without which your code can't be made to work, let me know. I thought previously that a way to pass a stream point table to PipelineManager was necessary, but it isn't as the custom point table interface I did as a sample for the python streaming code didn't require it.

gsakkis:
Andrew, unless I am missing something [this change](https://github.com/PDAL/PDAL/pull/3578/commits/00223438da464dd19fe875870833586c128eac29) only _extends_ the API with two new constructors, it doesn't change it.

Strictly speaking it is not necessary. I mean I had to [copy and paste PipelineExecutor::getSchema](https://github.com/PDAL/python/pull/94/commits/5b588ac6dac3a70932c3e7021f64f5c32cfc6983#diff-e8b64dc502626643eb48fb214a8c266b66df677b2f668404955600c45eef06a4R232) just to change the table reference but I can live with that. 

The only reason for this PR was that you had opened #3564 so I assumed you wanted to fix it at some point. If not feel free to close both.

abellgithub:
Hmm. I missed a bit of what was there.  Sorry. I don't think a new constructor is a problem, but I'm not sure that what's there is clear. I think the logic that you and I were both thinking of is is "use this point table", and a constructor that takes two tables seems confusing.

Talking through this... if you're going to call executeStream(), you pass the point table, so you already have that in hand, so I guess I'm not sure what good it does to pass it to the constructor, which is why I was backing away from the issue I created earlier. But maybe you're seeing something I'm missing.  I guess I also don't understand how this is related to `StreamableExecutor::getSchema()`.

gsakkis:
> I think the logic that you and I were both thinking of is is "use this point table", and a constructor that takes two tables seems confusing.

Indeed, I think it's unfortunate that `PipelineManager` takes two tables but we can't change this now. 

> if you're going to call executeStream(), you pass the point table, so you already have that in hand, so I guess I'm not sure what good it does to pass it to the constructor, which is why I was backing away from the issue I created earlier. But maybe you're seeing something I'm missing. I guess I also don't understand how this is related to StreamableExecutor::getSchema().

I guess we could replace `executeStream()` with `execute(ExecMode::Stream)` and pass the point table via the constructor as `m_streamTable`. The problem is that `PipelineManager` exposes publically only `m_table` via [pointTable()](https://github.com/PDAL/PDAL/blob/master/pdal/PipelineManager.hpp#L141) and this is used in several places, among which [PipelineExecutor.getSchema](https://github.com/PDAL/PDAL/blob/master/pdal/PipelineExecutor.cpp#L79). That's why I am passing the custom table as `m_table` and not `m_streamTable`.

abellgithub:
I don't remember all the details but I'm sure I wouldn't have dumped two tables in the PipelineManager unless it seemed the most reasonable way to meet the requirements at the time.  There was some strange stuff with not being able to determine whether a pipeline was streamable until after having processed some options :(

I'm still thinking that the answer to the issue you're having with `PipelineExecutor` is to either
- Ignore pdal::PipelineExecutor altogether and create whatever you want in the python interface. I think that's where I was going when I did the python custom point table, mostly because I was lazy and copy/paste of so little code didn't seem a big deal. I suppose you could also subclass it and override what you want, but there's so little there that I don't know it's worth it.
- Make a new pdal::PipelineInterface or whatever that's a good version of pdal::PipelineExecutor and use that. This is a long-term win, but it ties your new Python interface to a new PDAL release, which seems unfortunate. 🤷

gsakkis:
Are you referring to this PR or #3567 too? As I mentioned this PR would be a nice to have, the main/only benefit would be saving the override for `StreamableExecutor::getSchema()`. #3567 however fixes a general (potential) `PipelineExecutor` bug, in addition to making it easier to subclass. 

>  I suppose you could also subclass it and override what you want, but there's so little there that I don't know it's worth it.

Sure it's not a big complex class but it's not a one-liner either. I guess if it didn't exist (and wasn't already used by python-pdal since 2018) I'd use `PipelineManager` directly, but since it exists and does 95% of the job already why not tweak it a bit and reuse it. 

Having said that, if you'd rather keep `PipelineExecutor` as is, bugs and all, I'm ok with dumping it from python-pdal.

gsakkis:
Closing this as python-pdal doesn't use `PipelineExecutor` anymore.

PR #3594, opened 2021-11-15
CMake: Allow to use CMAKE_CROSSCOMPILING_EMULATOR

xantares:
For cross-compiling scenarios, we can either choose to use DIMBUILDER_EXECUTABLE
or run through CMAKE_CROSSCOMPILING_EMULATOR.


xantares:
it will need https://github.com/conda-forge/pdal-feedstock/pull/147

hobu:
This doesn't seem to be working in

xantares:
it seems it must be a STRING not PATH

PR #3597, opened 2021-11-16
set CMP0068 for OSX

hobu:


abellgithub:
Why?

hobu:
we are getting duplicate install_name_tool complaints on osx-arm64. I think recent cmake's set this by default anyway, so maybe it doesn't matter.

PR #3613, opened 2021-12-02
tests passing against tiledb 2.5.x

normanb:
TileDB 2.5 removed `max_buffer_elements`. I have updated the tests to use larger buffers and then adding a check on the result buffer size. The plugin itself was not using `max_buffer_elements`.

I also took the opportunity to do a minor refactor on the tests by setting the `count` and making the test names match.

### DIFF
--- a/plugins/tiledb/test/TileDBReaderTest.cpp
+++ b/plugins/tiledb/test/TileDBReaderTest.cpp
@@ -50,6 +50,8 @@ namespace pdal
 
 const SpatialReference utm16("EPSG:26916");
 
+const size_t count = 100;
+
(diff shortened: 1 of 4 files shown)


normanb:
This is quite a large PR now. I will wait for the CI but locally this works against 1.x and 2.x version of TileDB. We will remove 1.x support for TileDB as a separate PR which will simplify the code overall.

I was looking through the CI for PDAL which has changed, are you pulling in TileDB through the GDAL dependency?

abellgithub:
We don't get tiledb from gdal. Is there a particular workflow you're
looking at?

PR #3614, opened 2021-12-02
readers.ply support face element

### LINKED ISSUE #3307: readers.ply does not retain face data
**Describe the bug**
Reading a ply file (either binary or ascii) does not retain the face data. Using `pdal pipeline` with the file below outputs a ply file without face data. I was trying to convert a binary ply with faces to ascii but then the faces disappeared.

```
[
  "withface.ply",
  {
    "type": "writers.ply",
    "precision": 3,
    "filename": "withoutface.ply",
    "faces": true
  }
]
```

**Expected behavior**
I think the face data should be retained. I may be missing some filter but I can't find it in the manual.

```
$ uname -a
Linux localhost.localdomain 5.9.11-200.fc33.x86_64 #1 SMP Tue Nov 24 18:18:01 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

```
$ pdal --version
pdal 2.2.0 (git-version: Release)
```

longhuan2018:
fix #3307

### DIFF
--- a/io/PlyReader.cpp
+++ b/io/PlyReader.cpp
@@ -54,7 +54,7 @@ static StaticPluginInfo const s_info
 CREATE_STATIC_STAGE(PlyReader, s_info)
 
 
-PlyReader::PlyReader() : m_vertexElt(nullptr)
+PlyReader::PlyReader() : m_vertexElt(nullptr), m_faceElt(nullptr)
 {}
 
 
@@ -279,6 +279,8 @@ void PlyReader::extractHeader()
     for (Element& elt : m_elements)
         if (elt.m_name == "vertex")
             m_vertexElt = &elt;
+        else if (elt.m_name == "face")
+            m_faceElt = &elt;
     if (!m_vertexElt)
         throwError("Can't read PLY file without a 'vertex' element.");
 }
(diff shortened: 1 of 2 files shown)


hobu:
Thanks for the contribution, but we need some tests before we can merge this.

abellgithub:
There are also some issue with the code. I can go through it more, but know that PDAL point IDs don't necessarily map to PLY vertex indices.

abellgithub:
I'm also not sure there's a use case for this. It would get us into translating meshes. I don't really care, but it's kind of out-of-scope.

hobu:
> I'm also not sure there's a use case for this. It would get us into translating meshes. I don't really care, but it's kind of out-of-scope.

I mostly agree, but having a little convenience here prevents people from having to jump out of PDAL for simple stuff. With the new Python bindings, maybe this isn't as strong of an issue (the new Python bindings have explicit `meshio` support).

abellgithub:
The reader needs to use the vertex indices in the input file when loading the input file. I think this requires a local store of the data to map the local (file) vertex indices to the point ID provided when the vertex is created. There is no after-the-fact way to recover the PLY vertex indices that I can think of.

Jean-Roc:
hi, I don't know if this PR would directly solves it but here an example of faces use we have :

to build a pipeline working on a datasets to generate tin ground mesh for CAD users as the end result, it is divided into tiles with withheld points (used as buffer) and running in x individual parallel processes. For now with PDAL, a delaunay followed by a crop (to remove the buffer) would save only vertices, meaning we have to *"jump out"* for these steps.

abellgithub:
@Jean-Roc : I'm sorry but I don't understand. You can write faces with the PLY writer. Why do you want to read them?

hobu:
Closing this one for now. Please reopen if you can pick it back up with tests and some more enhancement.

PR #3615, opened 2021-12-02
get rid of quotes in argument descriptions

hobu:


abellgithub:
This is not the way to fix this.

abellgithub:
You can do this, but you should also do this:

https://github.com/PDAL/PDAL/tree/issue-3608

PR #3617, opened 2021-12-03
writers.ply support multiple outputs

longhuan2018:
Switch to FlexWriter and support multiple outputs.


hobu:
Please add some tests to this so we can merge.

test/unit/io/PlyWriterTest.cpp
```diff
@@ -259,4 +260,132 @@ TEST(PlyWriter, dimtypes)
     }
 }
 
+// Test that data from three input views gets written to separate output files.
+TEST(PlyWriter, flex)
```
abellgithub, line 264: Can you please put different data in the files so that the test can check that files were actually written correctly? Probably should also add a couple of faces in at least one of the files.

abellgithub:
This looks quite good.

PR #3618, opened 2021-12-03
fix writers.ogr

### LINKED ISSUE #3611: Erratic output of writers.ogr
I'm trying to export part of a point cloud using `writers.ogr`. Most of the time the file ends up being empty, without warnings or errors.
Implementing the same pipeline with the c++ API works every time.

```
$ pdal pipeline cut.json
```

<details>
  
```
{
  "pipeline":[
    "in/ept.json",
     {
        "type": "writers.ogr",
        "filename": "cut.dxf",
        "ogrdriver": "DXF"
    }
  ]
}
```

</details>

**Expected behavior**
File should be filled with proper values.

Windows, pdal 2.3.0, custom build

Any idea what might cause the issue ?

gui2dev:
fixes #3611

### DIFF
--- a/io/OGRWriter.cpp
+++ b/io/OGRWriter.cpp
@@ -122,6 +122,7 @@ void OGRWriter::readyTable(PointTableRef table)
 void OGRWriter::readyFile(const std::string& filename,
     const SpatialReference& srs)
 {
+    m_curCount = 0;
     m_outputFilename = filename;
     m_ds = m_driver->Create(filename.data(), 0, 0, 0, GDT_Unknown, nullptr);
     if (!m_ds)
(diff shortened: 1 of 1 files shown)


gui2dev:
Maybe rather than setting a value for `m_measureDim` in ctor, `m_measureDimName` could have a default value, as stated in the documentation

io/OGRWriter.cpp
```diff
@@ -60,7 +60,7 @@ static StaticPluginInfo const s_info
 CREATE_STATIC_STAGE(OGRWriter, s_info)
 
 OGRWriter::OGRWriter() : m_driver(nullptr), m_ds(nullptr), m_layer(nullptr),
-    m_feature(nullptr)
+    m_feature(nullptr), m_curCount(0), m_measureDim(Dimension::Id::Unknown)
```
abellgithub, line 63: `m_curCount` should be initialized in `readyFile()`, not in the ctor. This makes sure that it's properly set if the writer is used more than once.

io/OGRWriter.cpp
```diff
@@ -154,11 +154,13 @@ bool OGRWriter::processOne(PointRef& point)
     double x = point.getFieldAs<double>(Dimension::Id::X);
     double y = point.getFieldAs<double>(Dimension::Id::Y);
     double z = point.getFieldAs<double>(Dimension::Id::Z);
-    double m = point.getFieldAs<double>(m_measureDim);
 
     OGRPoint pt(x, y, z);
     if (m_measureDim != Dimension::Id::Unknown)
+    {
```
abellgithub, line 160: Not that it matters, but I think this change can go away. If `m_measureDim` is Unknown, `getFieldAs` will return 0, which is how OGRPoint initializes the measure dimension. You might test this, but I'm reasonably certain.

gui2dev: I agree with that.

abellgithub:
Are you done with this? It solves your issues?

gui2dev:
Before the last commit it was ok.
After I guess it's still ok as it was not a major modification.

PR #3622, opened 2021-12-03
Testing for TileDB 1.x and 2.x

normanb:
@abellgithub We can merge this into your branch and then into master.

`set_buffer` is deprecated. After all the tests pass and we merged master I will create a separate PR that makes CI only work with TileDB 2.5+.


abellgithub:
Great. Thx. I'll merge when CI is done.

PR #3625, opened 2021-12-06
Win warnings

abellgithub:


pdal/compression/LazPerfVlrCompression.cpp
```diff
@@ -317,7 +317,7 @@ class LazPerfVlrDecompressorImpl
         if (chunk == m_chunks.end() || nextChunk == m_chunks.end())
             return false;
 
-        m_chunkPointsTotal = nextChunk->count - chunk->count;
+        m_chunkPointsTotal = (int)(nextChunk->count - chunk->count);
```
hobu, line 320: shouldn't this be an explicitly sized type like `int32_t`? I suppose that requires much more refactoring for little value though.

hobu (approved):

pdal/compression/LazPerfVlrCompression.cpp
abellgithub: `m_chunkPointsTotal` is an int. But I don't think this matters anyway. On any system we're targeting an int is at least 32 bits.

PR #3627, opened 2021-12-09
Fix filepath UTF-8 encoding conversion problem

### LINKED ISSUE #3619: pdal.exe reports an error in the Chinese directory
Host Environment
OS : windows 10 Chinese version

When pdal.exe is in a directory containing Chinese, run ”pdal.exe --drivers“ and the program crashes immediately.
I debugged the code and found an exception when running FileUtils::toNative
```
#ifdef PDAL_WIN32_STL
std::wstring toNative(std::string const& in)
{
    // TODO: C++11 define convert with static thread_local
    std::wstring_convert<std::codecvt_utf8_utf16<uint16_t>, uint16_t> convert;
    **auto s = convert.from_bytes(in);**   #An exception is thrown here
    auto p = reinterpret_cast<wchar_t const*>(s.data());
    return std::wstring(p, p + s.size());
}
```

I think there is a problem with this code. The return value of GetModuleFileNameA is a local codepage, not always UTF-8 codepage. FileUtils:: getdirectory will call FileUtils::toNative to convert UTF-8 code to utf-16 code, which leads to an exception
```
std::string dllDir()
{
    std::string s;

#ifdef _WIN32
    HMODULE hm = NULL;

    if (GetModuleHandleEx(
        GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
        GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
        (LPCSTR)&dllDir, &hm))
    {
        char path[MAX_PATH];
        DWORD cnt = GetModuleFileNameA(hm, path, sizeof(path));
        if (cnt > 0 && cnt < MAX_PATH)
            s = path;
    }
#else
    Dl_info info;
    if (dladdr((const void *)dllDir, &info))
        s = info.dli_fname;
#endif
    return FileUtils::getDirectory(s);
}
```
However, even if the value of GetModuleFileNameA is modified to UTF-8 encoding, errors will also be reported in other places due to encoding problems.
I think all places where pdalboost::filesystem::path::string() is used need to return UTF-8 encoding.

longhuan2018:
fix #3619 
The default file path code in the program is UTF-8. Some cases are ignored when processing file path conversion under windows. Some patches are provided here.
For example:
 pdalboost::filesystem::path::string returns the current encoding of the operating system. We need to convert it to UTF-8 encoding.
 The Win32 API returns the current encoding of the operating system, which we need to convert to UTF-8 code.


hobu:
We cannot merge this without tests that demonstrates it works for your scenario and still doesn't break things.

pdal/DynamicLibrary.cpp
```diff
@@ -71,6 +72,16 @@ void DynamicLibrary::clear()
     m_handle = NULL;
 }
 
+#ifdef _WIN32
+std::wstring toNative(std::string const& in)
```
abellgithub, line 76: Even if this is the only place this is used, it should probably be in PDALUtils for symmetry and so that people know where to find it.

What's the with "static thread local" comment? Why would it matter?

pdal/DynamicLibrary.cpp
```diff
@@ -84,7 +95,7 @@ DynamicLibrary *DynamicLibrary::load(const std::string &name,
     void *handle = NULL;
 
 #ifdef _WIN32
-    handle = ::LoadLibraryA(name.c_str());
+    handle = ::LoadLibraryW(toNative(name).c_str());
```
abellgithub, line 98: I guess I don't understand the case where this would be an issue.

longhuan2018: A UTF-8 encoded file path is passed in. When calling Win32 API, it needs to be converted into system string, just like in FileUtil::getDirectory

pdal/DynamicLibrary.cpp
longhuan2018: I just copied it from PDALUtils::FileUtil

longhuan2018:
I found that there are Unicode coded test cases in the project pdal_file_utils_test, and I modified some contents in test_file_ops_with_unicode_paths,  to make it more effective to test the functions in FileUtils

pdal/DynamicLibrary.cpp
longhuan2018: If the file path contains Unicode encoding, it will cause problems

hobu:
Please grab the latest master and merge to catch up to a number of CI fixes.

hobu:
@xantares this PR has the potential to break mingw builds. Can you please test this and make any suggestions or updates as needed so we do not regress the patches you've added?

pdal/util/FileUtils.cpp
```diff
@@ -305,9 +323,9 @@ std::string toCanonicalPath(std::string filename)
 
 #ifdef _WIN32
     filename = addTrailingSlash(filename);
-    char buf[MAX_PATH];
-    if (GetFullPathName(filename.c_str(), MAX_PATH, buf, NULL))
-        result = buf;
+    wchar_t buf[MAX_PATH];
+    if (GetFullPathNameW(toNative(filename).c_str(), MAX_PATH, buf, NULL))
```
xantares, line 327: yes that breaks mingw build because toNative returns char* instead of wchar* there

pdal/util/FileUtils.cpp
```diff
@@ -524,7 +535,7 @@ MapContext mapFile(const std::string& filename, bool readOnly, uintmax_t pos, ui
 #ifndef _WIN32
     ctx.m_fd = ::open(filename.c_str(), readOnly ? O_RDONLY : O_RDWR);
 #else
-    ctx.m_fd = ::_open(filename.c_str(), readOnly ? O_RDONLY : O_RDWR);
+    ctx.m_fd = ::_wopen(toNative(filename).c_str(), readOnly ? O_RDONLY : O_RDWR);
```
xantares, line 538: there too, we may want a separate case too

hobu:
A couple of points:

* This PR cannot be merged without tests that test the behavior change/improvement
* PDAL recently migrated to C++17 #3643 which may be used to alleviate these issues

I am kicking this PR ahead to the next milestone, as it is too late to address it in this release window.

hobu:
I have made an attempt to incorporate this stuff in my `wmain` branch. dd62ca7c94be9e7db1816433461315fa3376cd60 Please fetch that and run through any unicode filename tests you might think of to trigger issues. This effort is now incorporated in #3824 and this PR is being closed.

PR #3628, opened 2021-12-09
change the output path of gitsha.cpp

longhuan2018:
In vcpkg, files with configure_file are not allowed to be output to PROJECT_SOURCE_DIR.
If it is modified here, there is no need to make a patch for this in vcpkg.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,7 +150,7 @@ get_git_head_revision(GIT_REFSPEC GIT_SHA1)
 
 configure_file(
   "${PROJECT_SOURCE_DIR}/gitsha.cpp.in"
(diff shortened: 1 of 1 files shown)


abellgithub:
@hobu This seems fine to me. You?

hobu:
This patch breaks ninja

> ninja: error: build.ninja:1709: multiple rules generate CMakeFiles/pdalcpp.dir/pdal/gitsha.cpp.o [-w dupbuild=err]

Is there someplace else we can stick gitsha.cpp other than `CMAKE_CURRENT_BINARY_DIR`?

hobu:
This was a result of my dirty tree.

Issue #3645, opened 2021-12-30
Huge Gaps/Holes in DTM


shubham9436:
I have generated DTM from a point cloud. There are huge gaps/holes in the DTM. It has not interpolated the gaps left in the point cloud.
![Capture](https://user-images.githubusercontent.com/19433739/147747194-57a29cc0-0b4b-4a5b-ac94-c863a6a3018b.PNG)

I have used the following json:
`
{
    "pipeline": [
        "ground.las",
        {
            "filename":"dtm_btif",
            "gdaldriver":"GTiff",
            "output_type":"idw",
            "data_type":"float",
            "resolution":"0.5",
            "type": "writers.gdal"
        }
    ]
}
`
I have zipped the DTM file
[dtm-b.zip](https://github.com/PDAL/PDAL/files/7792657/dtm-b.zip)

abellgithub:
You can try using the `window_size` option, but otherwise, there is currently no support for filling missing data.

shubham9436:
> You can try using the `window_size` option, but otherwise, there is currently no support for filling missing data.

What could be the optimal `window_size` that generally fills the missing data?

abellgithub:
It depends on your data. You might use something like `gdal_fillnodata` or another raster processing program.

Whether the algorithm is reasonable depends totally on your data. One would have to understand *why* the data was missing in order to evaluate an algorithm.

shubham9436:
The reason behind missing the data in between is because I have separated the ground and non ground point cloud. So, buildings, trees have been removed from the point cloud (in ground point cloud) and hence they create gaps. Hence, it should not be filled with no data. I just want it to be interpolated.

PR #3648, opened 2022-01-06
Add metadata support for readers.copc

abellgithub:


io/LasWriter.cpp
```diff
@@ -1185,45 +1206,28 @@ void LasWriter::finishOutput()
...
-
-    out.seek(0);
-    out << m_lasHeader;
-    out.seek(m_lasHeader.pointOffset());
+    las::setSummary(d->header, d->summary);
 
+    std::vector<char> buf = d->header.data();
+    m_ostream->seekp(0);
+    m_ostream->write(buf.data(), buf.size());
+    m_ostream->seekp(d->header.pointOffset);  //ABELL ?
```
hobu, line 1230: I think this might be holdover from LASzip's API behavior, where we were passing in our ostream and it was moving things around and not leaving it where we expected.

io/private/las/Vlr.cpp
```diff
@@ -0,0 +1,412 @@
...
+    std::string error;
+    return parseIgnoreVlrs(ignored, error);
+}
+
+VlrList parseIgnoreVlrs(const StringList& ignored, std::string& error)
+{
+    error.clear();
+
+    // Always ignore COPC VLRs and superseded VLRs.
+    VlrList ignoredVlrs { {"copc", 1000}, {"LASF_Spec", 7} };
```
hobu, line 58: We should ignore ALL COPC VLRs #3645 3649

io/private/las/Vlr.cpp
```diff
@@ -0,0 +1,412 @@
...
+        }
+        else if (el.key() == "data")
+        {
+            if (data.size())
+                throw pdal_error("Can't specify both 'data' and 'filename' "
+                    "in VLR specification.");
+            if (!el.value().is_string())
+                throw pdal_error("LAS VLR data must be specified as "
+                    "a base64-encoded string.");
+            //ABELL - Fix this.
```
hobu, line 221: convert to using `char` too?

io/private/las/Vlr.cpp
```diff
@@ -0,0 +1,412 @@
...
+            offset = e.offset;
+            length = (uint32_t)e.length;
+            break;
+        }
+    if (length == 0)
+        return vlrdata;
+
+    // Load the data plus the description.
+    const int DescripLen = 32;
+    assert(offset > DescripLen);
```
hobu, line 391: This should throw an error in case someone gave us a junk VLR definition, which happens.

hobu (approved):

io/CopcReader.cpp
```diff
@@ -121,8 +124,8 @@ struct CopcReader::Private
     Scaling scaling;
     uint64_t tileCount;
     int32_t tilePointNum;
-    lazperf::header14 header;
-    lazperf::copc_info_vlr copc_info;
+    las::Header header;
+    copc::Info copc_info;
```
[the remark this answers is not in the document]
abellgithub: Didn't do what I wanted and I was trying to exorcise lazperf other than compression/decompression.

io/LasWriter.cpp
abellgithub: That sounds right. I'll add a comment.

io/private/las/Vlr.cpp
abellgithub: You said that you wanted to have the info VLR in metadata so that you could know the input was a COPC file. It's fine if we don't *write* the COPC VLRs in output with LAS.

io/private/las/Vlr.cpp
abellgithub: It's code that people might be using. Didn't want to break.

io/CopcReader.cpp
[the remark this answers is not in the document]
hobu: Should we remove them from laz-perf?

io/private/las/Vlr.cpp
hobu: Upon conservation of the worry of advertising files that are written with PDAL that appear to be COPC that aren't actually clustered, I think we should ignore all COPC VLRs on read.

PR #3654, opened 2022-01-13
Add read/write fbi format

alavenant:
Code for read and write Fbi format. 
TerrasScan use fbi format internaly to manipulate points clouds.


io/FbiReader.cpp
```diff
@@ -0,0 +1,417 @@
...
+        for (size_t i(0); i<hdr->FastCnt; i++)
+        {
+            BYTE classif;
+            ifFbi->read(reinterpret_cast<char *>(&classif), hdr->BitsClass/8);
+            view->setField(Dimension::Id::Classification, i , int(classif));
+        }
+    }
+        
+    /*
+     Other Fbi contents - Not found equivalent in pdal model for now
```
hobu, line 340: It would be reasonable to make add PDAL dimensions for these.

io/FbiReader.cpp
```diff
@@ -0,0 +1,417 @@
...
+using namespace pdal::fbi;
+
+namespace pdal
+{
+
+static StaticPluginInfo const s_info
+{
+    "readers.fbi",
+    "Bbi Reader",
+    "http://pdal.io/stages/readers.fbi.html", // _ToDo_
```
hobu, line 51: We will need a doc to merge this.

io/FbiWriter.cpp
```diff
@@ -0,0 +1,424 @@
...
+using namespace pdal;
+using namespace pdal::fbi;
+
+namespace
+{
+    const StaticPluginInfo s_info
+    {
+        "writers.fbi",
+        "FBI Writer",
+        "http://pdal.io/stages/writers.fbi.html"
```
hobu, line 46: need a doc

test/unit/io/FbiReaderTest.cpp
```diff
@@ -0,0 +1,114 @@
...
+namespace pdal
+{
+
+
+namespace
+{
+
+std::string getTestfilePath()
+{
+    return Support::datapath("fbi/test.fbi");
```
hobu, line 50: Can you write the `1.2-with-color.las` file to fbi, add it to the test suite, and add a simple test that checks coordinates or intensity or something? This provides commonly tested values for various formats so we have some context when looking at transformed data.

hobu (changes requested):
This is great! Thanks for the contribution! A few small comments, it will need some docs, and a small tweak to the test config will help it fit with other format drivers. We will plan to include this in PDAL 2.4 if it can be cleaned up in the next few weeks.

abellgithub (commented):
General:

All this seems to assume you're running on a little-endian machine. This is pretty-much always the case these days, but if you're not going to support big-endian, you should probably throw an error if someone tries to use it on BE.

hobu:
Moved over to #3822 so I can push and finish this up. Thanks for the contribution!

PR #3665, opened 2022-01-25
Fix #3488, filters.csf leaving outputs in uncontrollable directories

### LINKED ISSUE #3488: filters.csf saves files in the current directory
Any time you run filters.csf, files (cloth_nodes.txt, collision-notes.txt and others) are created in the current directory that look like they are not needed. Their size can be very large. The code for saving these files to Cloth.cpp does not show how they are used later.
The suggestion is to remove the saving of these files.

hobu:
@chambbj can you please review this?

### DIFF
--- a/filters/CSFilter.cpp
+++ b/filters/CSFilter.cpp
@@ -79,6 +79,7 @@ struct CSArgs
     int m_iterations;
     std::vector<DimRange> m_ignored;
     StringList m_returns;
+    std::string m_dir;
 };
 
 CSFilter::CSFilter() : m_args(new CSArgs)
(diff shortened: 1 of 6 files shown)


chambbj:
In other filters (maybe PMF?) we make the directory optional, and if empty, skip the outputs. That would be nice here too if the CSF implementation doesn't rely on the intermediate files. I don't know off the top of my head if it does.

Otherwise, I'm fine with this.

hobu:
Feel free to push the updates you want. I was just trying to address the ticket to shut off the file noise usage of the filter causes.

chambbj (approved):

PR #3666, opened 2022-01-25
Add custom point table support to PipelineManager 

hobu:
address #3564

### DIFF
--- a/pdal/PipelineManager.cpp
+++ b/pdal/PipelineManager.cpp
@@ -53,6 +53,14 @@ PipelineManager::PipelineManager(point_count_t streamLimit) :
     m_progressFd(-1), m_input(nullptr)
 {}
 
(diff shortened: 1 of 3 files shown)


pdal/PipelineManager.cpp
```diff
@@ -53,6 +53,14 @@ PipelineManager::PipelineManager(point_count_t streamLimit) :
     m_progressFd(-1), m_input(nullptr)
 {}
 
+PipelineManager::PipelineManager(SimplePointTable& pointTable,
+                                FixedPointTable& fixedPointTable) :
+    m_factory(new StageFactory),
```
hobu, line 58: Gee, while we're doing this, I wonder if we shouldn't let the user provide their own StageFactory too.

hobu:
It was decided this was not that useful and a bad idea.

PR #3669, opened 2022-01-25
un-vendor gtest

hobu:
* bump minimum cmake version to 3.14
* Remove vendored googletest code
* Build google test based on google/googletest@f45d5865ed0b2b8912244627cdf508a24cc6ccb4 using CMake FetchContent_Declare


hobu:
I think we should default to WITH_TESTS=FALSE with this change, but I would like feedback before doing so.

hobu:
Let's not do this for now.
* hard to make local changes
* need the network to do anything

Issue #3659, opened 2022-01-20
PDAL built with only lazperf can't read empty LAZ files


klassenjs:
empty.txt:
```
X,Y,Z

```

Setup (built with laszip + lazperf):
```
$ pdal translate empty.txt empty.las
$ pdal translate empty.txt empty.lz.laz --writers.las.compression=laszip
$ pdal translate empty.txt empty.lp.laz --writers.las.compression=lazperf
```

Reading (built with laszip + lazperf):
```
$ pdal info empty.lz.laz 
{
  "file_size": 349,
  "filename": "empty.lz.laz",
  "now": "2022-01-20T14:49:24-0600",
  "pdal_version": "2.3.0 (git-version: 5f3619)",
  "reader": "readers.las",
  "stats":
  {
    "statistic":
    [
      {
        "average": 0,
        "count": 0,
        "maximum": -1.797693135e+308,
        "minimum": 1.797693135e+308,
        "name": "X",
        "position": 0,
        "stddev": 0,
        "variance": 0
      },
...
```

```
$ pdal info empty.lp.laz 
{
  "file_size": 345,
  "filename": "empty.lp.laz",
  "now": "2022-01-20T14:50:19-0600",
  "pdal_version": "2.3.0 (git-version: 5f3619)",
  "reader": "readers.las",
  "stats":
  {
    "statistic":
    [
      {
        "average": 0,
        "count": 0,
        "maximum": -1.797693135e+308,
        "minimum": 1.797693135e+308,
        "name": "X",
        "position": 0,
        "stddev": 0,
        "variance": 0
      },
...
```

Reading (built only with lazperf)
```
$ pdal info empty.las

{
  "file_size": 227,
  "filename": "empty.las",
  "now": "2022-01-20T14:38:52-0600",
  "pdal_version": "2.3.0 (git-version: 5f3619)",
  "reader": "readers.las",
  "stats":
  {
    "statistic":
    [
      {
        "average": 0,
        "count": 0,
        "maximum": -1.797693135e+308,
        "minimum": 1.797693135e+308,
        "name": "X",
        "position": 0,
        "stddev": 0,
        "variance": 0
      },
...
```

```
$ pdal info empty.lz.laz
PDAL: Unexpected end of file.

```

```
$ pdal info empty.lp.laz
PDAL: Invalid version 1179861324 found in LAZ VLR.

```

Built with the following versions
```
liblaszip: 3.4.1-6-gc7b67ca
[...]

PR #3673, opened 2022-01-27
Make sure we write the correct chunk table position.

abellgithub:
Don't decompress an empty chunk table.


hobu:
Does this fix #3659

abellgithub:
Yes. This is why it's the branch `issue-3659` :)

Issue #3671, opened 2022-01-26
Remove LASzip support in favor of laz-perf


hobu:
As mentioned on the mailing list https://lists.osgeo.org/pipermail/pdal/2021-October/002483.html we are removing LASzip support for PDAL in favor of laz-perf.

Additionally, with the relicense activity of [LASzip](https://github.com/LASzip/LASzip/pull/80) and [laz-perf](https://github.com/hobu/laz-perf/pull/114), we will be able to vendor and embed a copy of the laz-perf library for always-on LAZ support in PDAL.

hobu:
Completed in #3675

PR #3675, opened 2022-01-28
Replace LASzip implementation with laz-perf and embed it in PDAL

abellgithub:


hobu:
This PR is dependent upon https://github.com/LASzip/LASzip/pull/80 being merged

cmake/lazperf.cmake
```diff
@@ -1,12 +1,5 @@
...
-    else()
-        set(LAZPERF_LIBRARY "")
-        set(PDAL_HAVE_LAZPERF 0)
-    endif()
-endif()
+#
+# lazperf support
+#
+
+set(PDAL_LAZPERF_LIB_NAME pdal_lazperf)
```
dg0yt, line 5: I'm surprised to see that laz-perf is now unconditionally vendored into pdal. I looked over https://github.com/PDAL/PDAL/issues/3671 and the mailing list posts mentioned there, but I didn't see this particular aspect discussed anywhere. How is this redundancy maintained in the future? How do applications access both PDAL and laz-perf?

hobu: There was no significant pushback. More importantly, no one to stepped forward to support maintaining both the LASzip and laz-perf code paths. Given what we've done with COPC and EPT, LASzip was getting challenging, especially for things like multithreaded use. With Martin's death, there was no opportunity for it to get fixed without doing all the work ourselves.

The other cases for laz-perf are that it actually has code tests, we're stepping forward to maintain it going forward, it has a bit more convenience as a standalone library, and its API is much more pleasant to use. It is a better library, we know what is going on with it without spending lots of time tracing through it, and with the relicensing of LASzip/laz-perf, it gets to be vendored.

hobu: Does the changeover impact you?

dg0yt: I'm not concerned by the change-over or by the lack of laszip. I'm updating the vcpkg port of PDAL, and any "vendored dependencies" are unpopular in a wider library packaging ecosystem. 
vcpkg didn't have or use laz-perf, but a separate laszip port. Now I prepared a laz-perf port, only to find that I can't use it for PDAL without patching. For now I simply didn't submit the laz-perf port, because I didn't want to deal with versioning issues.

(FTR I'm also impact by the lack of std::filesystem in Ubuntu 18.04's g++. std::experimental::filesystem seems good enough to build, but I didn't test if the binaries work. [Patch](https://github.com/microsoft/vcpkg/pull/24355/files#diff-c855e918cb5fcca30982b4e36e741ad9d75bf9abf27a80a30aaa8afde4baeaf5)).

abellgithub:
On std::filesystem...you shouldn't use PDAL 2.4 if you don't have C++17
support.

cmake/lazperf.cmake
dg0yt: > This PR is dependent upon https://github.com/LASzip/LASzip/pull/80 being merged

The files in `vendor/lazperf/detail` still carry GNU Lesser General Licence headers.

hobu: grr.. There will be a 2.4.1 release in a couple of weeks that will fix this and a few other small things.

dg0yt: > The files in `vendor/lazperf/detail` still carry GNU Lesser General Licence headers.

This is still unchanged in 2.4.2.

PR #3676, opened 2022-01-28
Compute pointCount in CopcReader::inspect()

abellgithub:


hobu:
It seems the hierarchy isn't queried if we're not passing a bounds but we are passing a resolution. The behavior is a bit unexpected that you're required to pass a bounds in to get a resolution-limited `preview()` too.

PR #3682, opened 2022-02-10
Add back deleted LAS header/VLR API

abellgithub:


rcoup:
Building Entwine (which is set to C++11) produces a pile of these warnings (clang 13.0.0/macOS) against this, which it didn't beforehand:
```
In file included from /usr/local/include/pdal/io/LasHeader.hpp:45:
/usr/local/include/pdal/io/LasVLR.hpp:54:3: warning: use of the 'deprecated' attribute is a C++14 extension [-Wc++14-extensions]
[[deprecated]] OLeStream& operator<<(OLeStream& out, const LasVLR& v);
  ^
```

PDAL builds itself with C++17, but I'm not sure what C++ version the headers are meant to require?

hobu:
We haven't yet bumped Entwine to c++17. We were waiting for PDAL to do that and release before addressing Entwine.

PR #3686, opened 2022-02-15
Update arbiter bundle

connormanning:
- Add `az://` to the list of PDAL-recognized remote protocols
- Fix S3 globbing over HTTPS for bucket names which contain dots
- Add Azure SAS token support
- Some Azure env vars were aliased as both `AZ_ABC` and `AZURE_ABC`, made these aliases consistent throughout

### DIFF
--- a/vendor/arbiter/arbiter.cpp
+++ b/vendor/arbiter/arbiter.cpp
@@ -2183,8 +2183,8 @@ std::vector<std::string> S3::glob(std::string path, bool verbose) const
 
     // https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html
     const Resource resource(m_config->baseUrl(), path);
-    const std::string& bucket(resource.bucket());
-    const std::string& object(resource.object());
+    const std::string bucket(resource.bucket());
+    const std::string object(resource.object());
 
     Query query;
 
@@ -2197,9 +2197,9 @@ std::vector<std::string> S3::glob(std::string path, bool verbose) const
     {
         if (verbose) std::cout << "." << std::flush;
 
-        if (!get(resource.bucket() + "/", data, Headers(), query))
+        if (!get(bucket + "/", data, Headers(), query))
         {
-            throw ArbiterError("Couldn't S3 GET " + resource.bucket());
+            throw ArbiterError("Couldn't S3 GET " + bucket);
         }
 
         data.push_back('\0');
@@ -2357,7 +2357,7 @@ std::string S3::ApiV4::buildCanonicalRequest(
         const Query& query,
         const std::vector<char>& data) const
 {
-    const std::string canonicalUri("/" + resource.object());
+    const std::string canonicalUri = resource.canonicalUri();
 
     auto canonicalizeQuery([](const std::string& s, const Query::value_type& q)
     {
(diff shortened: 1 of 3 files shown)


hobu (approved):

vendor/arbiter/arbiter.cpp
```diff
@@ -2655,6 +2665,17 @@ AZ::Config::Config(const std::string s, const std::string profile)
...
     , m_baseUrl(extractBaseUrl(s, m_service, m_endpoint, m_storageAccount))
 {
+    const std::string sasString = extractSasToken(s);
+    if (!sasString.empty())
+    {
+        const auto params = split(sasString, '&');
+        for (const auto& param : params)
+        {
+            const auto kv = split(param, '=');
+            m_sasToken[kv.at(0)] = kv.at(1);
```
abellgithub, line 2675: This needs some hardening. I'd rather see the result of split() checked for size and then use `[]`, but with `at`  you have to catch an exception.

vendor/arbiter/arbiter.cpp
```diff
@@ -2693,6 +2714,10 @@ std::string AZ::Config::extractStorageAccount(
     {
         return *p;
     }
+    else if (auto p = env("AZ_STORAGE_ACCOUNT"))
```
abellgithub, line 2717: On line 2707, will `parse()` throw if the string isn't valid JSON?

vendor/arbiter/arbiter.cpp
```diff
@@ -2720,6 +2749,25 @@ std::string AZ::Config::extractStorageAccessKey(
     return "";
 }
 
+std::string AZ::Config::extractSasToken(const std::string s)
+{
+    const json c(s.size() ? json::parse(s) : json());
```
abellgithub, line 2754: Same question here...

vendor/arbiter/arbiter.cpp
```diff
@@ -2819,35 +2888,43 @@ bool AZ::get(
...
+
+    if (res->ok())
+    {
+        data = res->data();
         return true;
     }
     else
     {
-        std::cout << res.code() << ": " << res.str() << std::endl;
+        std::cout << res->code() << ": " << res->str() << std::endl;
```
abellgithub, line 2927: This seems like it should throw, log, or something other than just random output.

vendor/arbiter/arbiter.cpp
```diff
@@ -2738,6 +2786,14 @@ std::string AZ::Config::extractService(
     {
         return *p;
     }
+    else if (auto p = env("AZ_SERVICE"))
```
abellgithub, line 2789: Someone please tell me why `auto` is better than `const char *`.

vendor/arbiter/arbiter.cpp
```diff
@@ -2720,6 +2749,25 @@ std::string AZ::Config::extractStorageAccessKey(
     return "";
 }
 
+std::string AZ::Config::extractSasToken(const std::string s)
```
abellgithub, line 2752: Is there a reason this isn't a reference, given that it's already const? Some threading thing?

hobu:
oops, I thought this was ready to merge. Please push to the branch again with your additional updates @connormanning and reopen a PR.

PR #3687, opened 2022-02-20
fix: PDALConfig.cmake install location

t0b3:
PDALConfig{,Version}.cmake shall be installed at the same location
as PDALTargets.cmake
Without this patch these files end up in lib, with the others end up in
lib64

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -6,7 +6,7 @@ include(CMakePackageConfigHelpers)
 configure_package_config_file(
   PDALConfig.cmake.in
   ${CMAKE_CURRENT_BINARY_DIR}/PDALConfig.cmake
-  INSTALL_DESTINATION ${LIB_INSTALL_DIR}/cmake/PDAL
+  INSTALL_DESTINATION ${PDAL_LIB_INSTALL_DIR}/cmake/PDAL
   PATH_VARS INCLUDE_INSTALL_DIR LIB_INSTALL_DIR)
 
 write_basic_package_version_file(
@@ -17,4 +17,4 @@ write_basic_package_version_file(
 install(FILES
   ${CMAKE_CURRENT_BINARY_DIR}/PDALConfig.cmake
   ${CMAKE_CURRENT_BINARY_DIR}/PDALConfigVersion.cmake
-  DESTINATION ${LIB_INSTALL_DIR}/cmake/PDAL)
+  DESTINATION ${PDAL_LIB_INSTALL_DIR}/cmake/PDAL)


hobu:
On which platform/configuration were you seeing this issue? I believe this code has been this way for quite a while.

t0b3:
integrating pdal for gentoo made the need for such a patch visible see https://github.com/gentoo/gentoo/pull/23526/commits/972948007a09c4ef49526481f7d1e7bf5456a1df#diff-8f402ea5a25b139b4cf86e8d4e5aa5eb035bbc0e87af1ca62231e119b3729243

hobu:
The cmake min version bump will be happening in 2.4 https://github.com/PDAL/PDAL/blob/master/CMakeLists.txt#L6

I'm anxious to change this without some input from the other packagers. Presumably they have patches that are accounting for this too.

t0b3:
As noted, this PR is not about cmake min versions, as this if fixed in master already. :+1: 

Good point to collect further input from other distros...

To give you more background on the reason and circumstances this issue arises as of now:

Because `PDALConfig{,Version}.cmake` is installed to `${LIB_INSTALL_DIR}/cmake/PDAL` as defined at https://github.com/PDAL/PDAL/blob/a00819d42cfd1ecd7ab85c6d77ec277fa278aac9/cmake/config.cmake#L17-L20
while `PDALTargets.cmake` is installed to `${PDAL_LIB_INSTALL_DIR}/cmake/PDAL` as defined at https://github.com/PDAL/PDAL/blob/6bdd2e2c281fe0313f8acd8b8b9e4ccf9b2a13c3/CMakeLists.txt#L367-L378

given that `PDAL_LIB_INSTALL_DIR` is defined as `${CMAKE_INSTALL_LIBDIR}` which may differ from `lib` (i.e. `lib64`) as defined in https://github.com/PDAL/PDAL/blob/a00819d42cfd1ecd7ab85c6d77ec277fa278aac9/cmake/install_dirs.cmake#L10
while `LIB_INSTALL_DIR` is hard to `lib` at https://github.com/PDAL/PDAL/blob/a00819d42cfd1ecd7ab85c6d77ec277fa278aac9/cmake/config.cmake#L2

This leads us to the situation where files installed to destiny `lib*` end up as follows
```
/usr/lib/cmake/PDAL/PDALConfig.cmake
/usr/lib/cmake/PDAL/PDALConfigVersion.cmake
/usr/lib64/cmake/PDAL
/usr/lib64/cmake/PDAL/PDALTargets-relwithdebinfo.cmake
/usr/lib64/cmake/PDAL/PDALTargets.cmake
/usr/lib64/libpdal_base.so.13.0.0
...
/usr/lib64/libpdal_util.so.13.0.0
/usr/lib64/libpdalcpp.so
/usr/lib64/pkgconfig/pdal.pc

```
This unforeseen separation of `cmake` files renders qgis and entwine packages unable to compile.

Therefore the proposed and working solution of this patch is to reuse `PDAL_LIB_INSTALL_DIR` in order to install all those files to the same destination.

For distros which install everything to `lib` (i.e. debian, ubuntu, ...) this PR isn't expected to change anything.

PS: This has been tested and works with qgis & entwine on gentoo.

t0b3:
@sebastic debian and ubuntu should be not affected by this patch, right? - thx for confirming...

@svenstaro how is arch affected?

svenstaro:
In Arch `/usr/lib64` is a symlink to `/usr/lib` and so we currently end up with:
```
usr/lib/cmake/PDAL/PDALConfig.cmake
usr/lib/cmake/PDAL/PDALConfigVersion.cmake
usr/lib/cmake/PDAL/PDALTargets-noconfig.cmake
usr/lib/cmake/PDAL/PDALTargets.cmake
```
You can also check this here: https://archlinux.org/packages/community/x86_64/pdal/
PKGBUILD is here: https://github.com/archlinux/svntogit-community/blob/packages/pdal/trunk/PKGBUILD

jschueller:
+1 I guess this only affects rpm distro where everything should be installed under /lib64

t0b3:
@neteler will this patch help fedora or is it unaffected? thanks for your comment on the impact of this PR...

neteler:
> @neteler will this patch help fedora or is it unaffected? thanks for your comment on the impact of this PR...

In Fedora 35 they are already in the same directory:

```
rpm -qil PDAL-devel-2.3.0-3.fc35.x86_64 | grep cmake/PDAL
/usr/lib64/cmake/PDAL
/usr/lib64/cmake/PDAL/PDALConfig.cmake
/usr/lib64/cmake/PDAL/PDALConfigVersion.cmake
/usr/lib64/cmake/PDAL/PDALTargets-noconfig.cmake
/usr/lib64/cmake/PDAL/PDALTargets.cmake
```
with the SPEC file here:
https://src.fedoraproject.org/rpms/PDAL/blob/rawhide/f/PDAL.spec

t0b3:
@neteler :+1:  but it won't hurt fedora neither? 
thanks for your comment on the impact of this upstream PR...

neteler:
> @neteler +1 but it won't hurt fedora neither? 

I cannot judge this without patching and compiling it on Fedora (don't have much time these days, perhaps someone could try, even in docker?).

t0b3:
checked this by patching and compiling on fedora :+1: no negative impact

hobu:
ok, merging. Thanks for all the work!

PR #3691, opened 2022-02-22
Add COPC writer

abellgithub:


doc/stages/writers.copc.rst
```diff
@@ -0,0 +1,151 @@
+.. _writers.copc:
+
+writers.copc
+============
+
+The **COPC Writer** supports writing to `COPC format`_ files.
+
```
hobu, line 7:
```
.. warning::

    PDAL's ``writer.copc`` driver does all processing using local memory, and it does not support streaming mode. If you are wanting to process larger files, or do so in a way that exchanges memory usage for disk usage, consider using `Untwine <https://github.com/hobu/untwine>`__ in ``--single_file`` mode to create COPC data.

hobu (approved):
I already reviewed most of this once and didn't really have much for findings. My only item is to add a note to the doc to point people at Untwine to head off memory usage complaints.

PR #3693, opened 2022-02-25
Hint readers.las if a file could be COPC

hobu:
This PR adds metadata to tell a user if a LAS might also be metadata.


abellgithub:
- This risks a crash as the buffer can't be safely converted to a string.
- There's no reason for the private data - metadata can be set when you've determined the condition to be true.
- "copc" must be lowercase in order to be valid COPC.

PR #3697, opened 2022-03-04
Make valid glb

### LINKED ISSUE #3694: writers.gltf: not valid result?
Hi,

I'm trying to write a glTF (source data https://figshare.com/articles/dataset/NEON_Teaching_Data_LiDAR_Point_Cloud_las_Data/4307750 - 180 MB) using the following pipeline:

```
{
    "pipeline":[
            "NEONDSSampleLiDARPointCloud.las",
            {
                "type": "filters.poisson"
            },
            {
                "type":"writers.gltf",
                "filename":"NEONDSSampleLiDARPointCloud.glb",
                "metallic": 0.5
            }
        ]
}
``` 
A glb (3.4 MB) is created, but it's not valid according to https://github.khronos.org/glTF-Validator/ and the glb is not visible in https://gltf-viewer.donmccurdy.com/

pdal 2.3.0

Any ideas how to create a valid glTF?

abellgithub:
Close #3694


filters/PoissonFilter.cpp
```diff
@@ -317,6 +317,14 @@ PointViewSet PoissonFilter::run(PointViewPtr view)
             transform[i++] = xform.coords[r][c];
 
     TransformationFilter().doFilter(*outView, transform);
+
+    // Rerun normals as they may be expected.
+    {
+        NormalFilter f;
```
hobu, line 323: Must we always do this even if we are not writing normals for output according to the `m_writeNormals` option that was set?

io/GltfWriter.cpp
```diff
@@ -75,55 +76,58 @@ void GltfWriter::addArgs(ProgramArgs& args)
     args.add("filename", "Output gltf filename", m_filename).setPositional();
     args.add("metallic", "Metallic factor [0-1]", m_metallic);
     args.add("roughness", "Roughness factor [0-1]", m_roughness);
-    args.add("red", "Red factor [0-1]", m_red);
-    args.add("green", "Green factor [0-1]", m_green);
-    args.add("blue", "Blue factor [0-1]", m_blue);
+    args.add("red", "Base red factor [0-1]", m_red);
```
hobu, line 79: What does "Base" mean?

io/GltfWriter.cpp
```diff
@@ -75,55 +76,58 @@ void GltfWriter::addArgs(ProgramArgs& args)
...
+        table.layout()->hasDim(Dimension::Id::NormalZ);
 
-    m_writeNormals = table.layout()->hasDim(Dimension::Id::NormalX)
-                     && table.layout()->hasDim(Dimension::Id::NormalY)
-                     && table.layout()->hasDim(Dimension::Id::NormalZ);
+    if (!hasNormals && m_writeNormals)
+    {
+        log()->get(LogLevel::Warning) << getName() << ": Option 'normals' is set to "
+            "true, but one or more of the normal dimensions are missing. Not writing vertex "
+            "nromals." << std::endl;
```
hobu, line 102: typo

io/GltfWriter.cpp
```diff
@@ -133,8 +137,7 @@ void GltfWriter::write(const PointViewPtr v)
     TriangularMesh *mesh = v->mesh();
     if (!mesh)
     {
-        log()->get(LogLevel::Warning) << "Attempt to write point view with "
-                                      "no mesh. Skipping.\n";
+        log()->get(LogLevel::Warning) << "Attempt to write point view with no mesh. Skipping.\n";
```
hobu, line 140: Unrelated to this PR, but are we supposed to be putting `\n`'s in log outputs?

hobu (changes requested):

io/GltfWriter.cpp
abellgithub: I don't know. I read it somewhere. I have no idea how the renderer figures out what colors to use for which faces.

io/GltfWriter.cpp
abellgithub: `log()-get()` returns an std::ostream, so you should add newlines if you want them.

io/GltfWriter.cpp
abellgithub: Fixed.

hobu (approved):

PR #3703, opened 2022-03-10
Change voxel grid size

### LINKED ISSUE #3702: COPC writer grid cell size too large
The COPC writer grid cell size was changed and now exceeds internal limits.

abellgithub:
Close #3702


olsen232:
FYI I just hit this bug also - I don't have much knowledge about the project yet, but in case there is a reason why int(300 * Sqrt3) is the best possible magic number here, there doesn't seem to be any obstacle to giving extra bits to each dimension of GridKey - 10 bits instead of 8 bits, for instance - so that's also a potential fix.

abellgithub:
@olsen232 : There isn't really anything special about the values. The logic originally came from keeping points some distance from one another, but we're no longer doing this. I did some playing with the numbers and didn't see much difference in final output, but it could have just been that I hit this sampling error when I resized things. The data structure that's used here should probably be replaced, but we're trying to get a release done and it works for now, so maybe I can spend some more time playing with it for next time.

abellgithub:
@olsen232 : I should add that whether this value would make a substantial difference (assuming the data structure was good) depends a lot on the data set. Many lidar data sets aren't well-distributed, so it's often hard to "hoist" points to less detailed voxels with a sampling scheme like this.

Issue #3704, opened 2022-03-11
Handle quick info inspect for ply reader


dbrookes96:
**Describe the bug**
```
$ pdal info --summary test/data/ply/simple_text.ply
```
produces output
```
PDAL: No summary data available for 'test/data/ply/simple_text.ply'.
```
I brought this up a few months ago on the [mailing list](https://lists.osgeo.org/pipermail/pdal/2021-October/002491.html) and finally am getting around to it. It should be fairly simple to do for this file format and how the reader is already implemented. We can exclude bounds since we would have to read the entire file to calculate it. We can also exclude spacial reference. 

**Expected behavior**
Summary output for the input PLY file.

**System/installation information:**
```
$ uname -a
Linux mm00937 4.4.0-19041-Microsoft #1237-Microsoft Sat Sep 11 14:32:00 PST 2021 x86_64 x86_64 x86_64 GNU/Linux
```
(WSL on Windows 10)
```
$ pdal --version
------------------------------------------------------------------------------------------------------------------------
pdal 2.0.1 (git-version: Release)
------------------------------------------------------------------------------------------------------------------------
```

PR #3705, opened 2022-03-11
Handle quick info inspect for ply reader (#3704)

dbrookes96:
- Added implementation for inspect() to PlyReader. We can get
  everything we need from the vertex element setup by initialize().

- We ignore bounds to avoid having to read the entire file. We also
  avoid spatial reference.

- Added inspect test to PlyReaderTest unit tests.

### DIFF
--- a/io/PlyReader.cpp
+++ b/io/PlyReader.cpp
@@ -289,6 +289,28 @@ std::string PlyReader::getName() const
     return s_info.name;
 }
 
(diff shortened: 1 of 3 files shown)


dbrookes96:
Fixes #3704

hobu:
Thanks for the contribution, but unfortunately, this came in our 2.4.0 release freeze window, so it will be scheduled for the 2.5.0 release.

PR #3706, opened 2022-03-15
Copc metadata notification

hobu:


abellgithub:
I don't quite get the point of this, but it's fine. But I didn't get the point of the bit in the las reader, either.

abellgithub:
Also, I don't *think* this was what Mike was complaining about. I *thought* Mike was saying that there was no { "copc" : "false" } for a standard LAS file, but looking at the code, it seems he was mistaken.

abellgithub:
I mean, sure, but

https://en.wikipedia.org/wiki/File_(command)

PR #3722, opened 2022-03-29
TileDB 2.3 required

normanb:
This updates the TileDB plugin to require 2.3+, as discussed this reduces the maintenance for this module and removes a lot of macros that were in place to support earlier versions.

This code also enforces default compression for coordinates unless disabled by setting the compressor to have a compression level of -1.


hobu:
Please rebase master to get CI that's happy.

normanb:
The CI is failing due to the test I added to demonstrate https://github.com/PDAL/PDAL/issues/3724 . How do I get access to the pipeline metadata from the writer?

Issue #3715, opened 2022-03-22
faceraster smooth interpolation


zdila:
Please implement option to interpolate faces smoothly to prevent hard triangle edges.

Sample with hard edges: 
![image](https://user-images.githubusercontent.com/636095/159441692-8f8b10d7-6fcc-4057-8e83-60f8351b3b68.png)

zdila:
> We can't magically make data from nothing.

But you have something - neighbouring triangles. I am no expert in this area but I think it could involve interpolating face normals or some bezier smoothing (which is used for smoothing polygons).

zdila:
@hobu would it be worth for me to create a PR for this feature using one of the method I described or will it be rejected without a discussion as was this issue?

abellgithub:
Such smoothing wouldn't belong in this filter. You want to "smooth" the mesh. Generally, PDAL isn't the place for mesh manipulation - there are other software that do mesh manipulation.

zdila:
> Such smoothing wouldn't belong in this filter. You want to smooth the mesh

@abellgithub Actually I don't want to smooth the mesh. I will make faceraster to render it smoothly. It is analogy to shader in OpenGL. It renders existing mesh (triangles) to smooth surface (if smooth shader is used).

Anyway I will implement it just in my fork for personal purposes. I guess it will require adding information about triangle neighbours  and using different interpolation at https://github.com/PDAL/PDAL/blob/26cc5e10d4381d6c689800279b35c780fa42c8d2/filters/FaceRasterFilter.cpp#L145.

abellgithub:
That's certainly a way. This is open source software -- you are free to change the source code for your purposes as you wish. I have no idea where you're getting your mesh from, but the PDAL delaunay code keeps track of adjacent triangles.

zdila:
> PDAL delaunay code keeps track of adjacent triangles

Nice. It should make it even simpler to implement. Thanks.

zdila:
https://github.com/wangbolun300/Scattered_Points_Interpolation

zdila:
I wish that smooth output were the default in all tools which would make DEM data generated from sparse point cloud more usable for nice hill-shading. For example TINITALY DEM data suffers with this flat shading on many areas where points were sparse:

![image](https://user-images.githubusercontent.com/636095/162978166-a7910aae-a093-4fab-95ac-541a69d4122c.png)

That was the reason I had to reduce the resolution which on the other hand lowered details also of parts where point density was good.

[...]

PR #3725, opened 2022-03-29
Add support for hillshading for filters.faceraster

zdila:
This PR adds support for `hillshading` mode for `filters.faceraster`. Benefits of creating shaded relief directly instead of additional post-processing (with eg. `gdaldem hillshade`) is that post-processing requires neighboring points to compute the shade which makes it slightly blurry. This patch computes shade directly from the triangle orientation thus making it sharp.

In the future additional modes could be added like `slope` or `aspect` (both should be very easy to implement).

### DIFF
--- a/filters/GreedyProjection.cpp
+++ b/filters/GreedyProjection.cpp
@@ -203,7 +203,7 @@ void GreedyProjection::filter(PointView& view)
   // Initializing
   PointId isFree = 0;
   bool done = false;
-  int nr_parts=0, increase_nnn4fn=0, increase_nnn4s=0, increase_dist=0, nr_touched = 0;
+  int nr_parts=0, increase_nnn4fn=0, increase_nnn4s=0, increase_dist=0;
   bool is_fringe;
   angles_.resize(nnn_);
   std::vector<Eigen::Vector2d, Eigen::aligned_allocator<Eigen::Vector2d> > uvn_nn (nnn_);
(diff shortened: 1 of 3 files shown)


hobu:
This is interesting, but I wonder if it belongs in this particular writer or in some other kind of stand-alone filter for doing morphological operations to triangulations. The feature certainly is useful. 

We might not merge it as-is, but thanks for the contribution. If you have some ideas on where or how this might fit more generally, I would be interested to hear them. It seems duplicative to reimplement the triangle version of `gdaldem` in this kind of obscure PDAL writer...

zdila:
I agree, it could be probably cleaner to implement it as separate writers like `filters.faceHillshadeRaster`,`filters.faceSlopeRaster`, `filters.faceAspectRaster`. And original `filters.faceraster` could be renamed to `filters.faceBarycentricElevationInterpolatigRaster` (what a long name). All these rasters would share most of the code so it could be extracted to some abstract class. But I am no C++ developer (I can just hack it a little bit) so I leave it for others.

I also plan to add some [smooth interpolation](https://github.com/PDAL/PDAL/issues/3715) option as an alternative to linear barycentric one. And this could be applicable to all 4 filters/raster mentioned here. In any case this will be harder to implement for me :-)

> It seems duplicative to reimplement the triangle version of gdaldem in this kind of obscure PDAL writer.

Do you mean duplicative to gdaldem? Or you mean by putting it all to single PDAL writer? If the first, then I see no other way of reusing gdaldem here or its algorythms in the later phase because it computes the result from DEM which has already some information loss compared to triangles. If the second, then as I proposed, it could be split to separate filters (writers).

Disclaimer: So far I have very little experience with pdal because I used it only for a [narrow usecase](https://github.com/FreemapSlovakia/freemap-operations/blob/master/tools/pdal2dem.mjs).

If nothing else, this PR can just serve at least as an inspiration for the future proper implementation.

hobu:
> Do you mean duplicate to gdaldem? 

A little bit, although I get why they're useful. I'm mostly worried about them being buried in this obscure filter. 

BTW, I don't have a problem at all with long'ish names. Naming things is hard, and as long as they're descriptive and conveniently google'able, good enough for me.

zdila:
> I'm mostly worried about them being buried in this obscure filter.

I see. That's why I propose to split it to separate `filters.faceSomethingRaster`s.

PR #3730, opened 2022-04-05
Specify release build type in ubuntu Dockerfile

davidcaron:
I'm actually not very familiar with cmake, but I think the default mode is probably debug. The performance is much better after adding Release mode.


abellgithub:
Heh. No software should ever be run for real in debug mode these days.

PR #3743, opened 2022-04-15
Add trajectory filter

abellgithub:


plugins/trajectory/filters/Args.hpp
```diff
@@ -0,0 +1,66 @@
...
+        std::cerr << "dang = " << dang << "!\n";
+        std::cerr << "pitchweight = " << pitchweight << "!\n";
+        std::cerr << "scanweightest = " << scanweightest << "!\n";
+        std::cerr << "fixedpitch = " << fixedpitch << "!\n";
+        std::cerr << "multiweight = " << multiweight << "!\n";
+        std::cerr << "scanweight = " << scanweight << "!\n";
+        std::cerr << "attaccelweight = " << attaccelweight << "!\n";
+        std::cerr << "attclampweight = " << attclampweight << "!\n";
+        std::cerr << "extrapitchclamp = " << extrapitchclamp << "!\n";
+    }
```
chambbj, line 62: Would be nice to use PDAL's logger for this.

doc/stages/filters.trajectory.rst
```diff
@@ -0,0 +1,52 @@
...
+--------
+
+.. code-block:: json
+
+  [
+      "input.las",
+      {
+          "type": "filters.trajectory"
+      },
+      "trajectory.las"
```
chambbj, line 28: We can write an LAS, but I wonder if demonstrating the SBET writer might be more relevant in this case?

abellgithub:
The only test I had passed (which I believe is included) was provided by SRI and passes with the plugin. Some code was changed to eliminate the need for geographic lib and such, but most of the math is identical.

PR #3744, opened 2022-04-15
add 'count' metadata to readers.bpf metadata output

hobu:


abellgithub:
I don't care, but `pdal info --summary` calls it `num_points`.

hobu:
> I don't care, but pdal info --summary calls it num_points.

`readers.las` adds it as `count`, so I followed suit with that. Also, yes, the count is available as `num_points` via `--summary`, but if you are doing something like `--pipeline-serialization` you might need that count pushed there.

abellgithub:
Aok. Just was letting you know that it was different. You could probably also do something for all readers if you want.

hobu:
> You could probably also do something for all readers if you want.

It would be a good idea to normalize these.

hobu:
Picked back to `2.4-maintenance` with 4604297a94b1e32d66734417556e8bbddfd05ebc

PR #3748, opened 2022-04-22
Refactor pdal/pdal Docker container

hobu:
This PR changes the PDAL docker container to be based on miniforge. It's not that much different than before, but there is less for us to have to maintain in our script.

It also removes a bunch of stale docker configurations that are no longer in use.


abellgithub (commented):
IME you shouldn't be installing conda compilers. They're only for generating conda packages. I think I've worked around problems with this.

hobu:
This installs the compilers into the base environment, and it doesn't include them in the final container.

Are we not making a compiled PDAL-package that is supposed to work with the rest of the conda environment here? I thought that's what the CondaForge compilers package is for.

abellgithub:
I don't think this is correct but you should ask the conda gods. I thought the purpose of the conda compilers was to create a conda package. If you're not building a conda package, you should use the compiler for the system you're targeting.

I could be wrong.

hobu:
picked to `2.4-maintenance` in 35f3800bda6156b3aff38f949dda83b47cb8b6fa

PR #3766, opened 2022-05-17
Add CI debug task after build and before test

j9ac9k:
All the debug task does is list all conda installed packages and their
associated versions, this will be helpful in cases where a change in a
dependency causes a breakage.


j9ac9k:
rebased with `master` as there was a CI fix pushed after this PR was submitted

j9ac9k:
Actually, will close this and merge this into #3767

PR #3770, opened 2022-05-19
Clean up a bunch of dead stuff

hobu:


hobu:
I think we should move the `writers.pgpointcloud` and `readers.pgpointcloud` to [unsupported.](https://github.com/PDAL/unsuppported-plugins)

PR #3778, opened 2022-06-11
FileUtils: Fix mingw build

xantares:


pdal/util/FileUtils.cpp
```diff
@@ -234,7 +234,11 @@ std::vector<std::string> directoryList(const std::string& dir)
         fs::directory_iterator end;
         while (it != end)
         {
+#ifndef _MSC_VER
```
abellgithub, line 237: This should probably be PDAL_WIN32_STL as well.

xantares: yes

abellgithub:
I'm pretty sure that u8string() won't do the conversion from UTF-16 to UTF-8. Also, PDAL isn't doing C++20 at this time.

mmuetzel:
> I'm pretty sure that u8string() won't do the conversion from UTF-16 to UTF-8. Also, PDAL isn't doing C++20 at this time.

According to this, it does the conversion and is part of C++17 (like the remainder of the filesystem API):
https://en.cppreference.com/w/cpp/filesystem/path/string

mmuetzel:
It also looks like `clang` (at least on Windows) doesn't implicitly convert from `std::filesystem::path` to `std::string`. So, `fromNative(it->path())` throws an error during compilation with it.

pdal/util/FileUtils.cpp
```diff
@@ -234,7 +234,11 @@ std::vector<std::string> directoryList(const std::string& dir)
         fs::directory_iterator end;
         while (it != end)
         {
+#ifndef PDAL_WIN32_STL 
+            files.push_back(fromNative(it->path().string()));
+#else
             files.push_back(fromNative(it->path()));
```
mmuetzel, line 240: Afaict, implicit conversion from `std::filesystem::path` to `std::string` isn't defined in the standard.
It would probably good to replace this with
``` diff
-            files.push_back(fromNative(it->path()));
+            files.push_back(it->path().u8string());
```

That should also be working on all platforms. (No need to be conditional on `PDAL_WIN32_STL`.)

Issue #3726, opened 2022-04-04
Adding support for TILEDB_DATETIME_* datatypes to the PDAL tiledb reader


rowanwins:
**Describe the bug**

I am trying to open a tiledb array however it's failing because one of my dimensions uses the `datetime64[ns]` data type (see the [TileDB doco on datetime fields](https://docs.tiledb.com/main/background/internal-mechanics/datetimes)). From what I understand these are pretty much int64 values which PDAL does support.

```
pdal info --driver readers.tiledb my.tiledb -p 0
=> PDAL: Invalid Dim type from TileDB
```

I'm far from a c++ expert but scratching through the `plugins/tiledb/io/TileDBReader.cpp` I can see a few relevant pieces of code
For getting fields
https://github.com/PDAL/PDAL/blob/e63f88829a15891689f50e123fa327c2a539af8b/plugins/tiledb/io/TileDBReader.cpp#L60-L96
Reading the value
https://github.com/PDAL/PDAL/blob/e63f88829a15891689f50e123fa327c2a539af8b/plugins/tiledb/io/TileDBReader.cpp#L389-L432

There are a few handling options I can think of
- Improve the tiledb reader so that it converts the datatype into something that PDAL supports
  - Either you could write the raw int64 value, or you could attempt to convert it a similar datetime value as is used elsewhere in PDAL
- Add an option to tiledb reader that allows you to skip `TILEDB_DATETIME_*` dimensions - this seems feasible given that the buffer length is known to be int64.

I tried to find a public dataset in the tileDB samples that used one of these fields but couldn't, it seems like they mostly use float64 instead when storing datetimes. But if you are interested in pursuing a fix I can probably whip up a small sample to share.

If you're interested in supporting those fields I'd be happy to do a bit more investigating and see if I could put together a PR.

**System/installation information:**
I'm running PDAL v2.3.0

Thanks,
Rowan

hobu:
Added in #3798.

PR #3798, opened 2022-06-21
Support skipping over unsupported TileDB attributes in reader

normanb:
It is possible that a user creates a TileDB array outside of PDAL with the correct dimensions to be read by the PDAL driver. They may have attribute types that are not supported by the PDAL driver. This PR allows a user to skip over unsupported PDAL dimension (attribute) types. 

The option `strict` toggles this behaviour on/off, the default being on,  unrecognized attributes will cause an error to be thrown.


normanb:
https://github.com/PDAL/PDAL/issues/3726 is related, I added the datetime support as well.

PR #3800, opened 2022-06-22
Fixed ci for TileDB attribute schema evolution

normanb:


hobu:
This change didn't fix the issue.  https://github.com/PDAL/PDAL/runs/7010289760?check_suite_focus=true

Issue #3811, opened 2022-07-12
Memory leak using pdal's python bindings with PDAL>= 2.4.1


spolloni:
**Describe the bug**

There is a memory leak when repeatedly running PDAL pipelines using [pdal-python](https://github.com/PDAL/python) on PDAL versions >= 2.4.x

I am able to reproduce this with both the `pdal/pdal:2.4.1` and `pdal/pdal:2.4.2` docker images, as well as with `pdal 2.4.2` installed locally in my mac (with homebrew).  

I can rule out the python bindings being the culprit since the leak **does not** happen when using the `pdal/pdal:2.3.0` image

To reproduce: 
```{bash}
docker run \
	-v /Users/spolloni-mac/Desktop/memleak_repro.py:/memleak_repro.py \
	--name memleak_container \
	pdal/pdal:2.4.1 \
	/bin/bash -c "\
		/opt/conda/condabin/conda install -p /opt/conda/envs/pdal/ python=3.9 -y && \
		/opt/conda/envs/pdal/bin/pip install pdal && \
		/opt/conda/envs/pdal/bin/python -u /memleak_repro.py
	"
```

where memleak_repro.py  is:
```
json = """
[
   {
      "type": "readers.ept",
      "filename": "https://s3-us-west-2.amazonaws.com/usgs-lidar-public/OH_Statewide_Phase1_1_2019/ept.json",
      "bounds": "([-9109652, -9109552],[5047225,5047425])"
   }
]"""

import pdal
for i in range(int(1E100)):
	pipeline = pdal.Pipeline(json)
	count = pipeline.execute()
	print(f'iter {i}:', count)
```

then monitor memory use with:
```
docker stats memleak_container
```

spolloni:
notes:

- docker image `pdal/pdal:2.4.1` is actually shipping `pdal 2.4.0 (git-version: d63a84)` so I suspect the leak was introduced with https://github.com/PDAL/PDAL/compare/2.3.0...2.4.0 
- this is doesn't seem caused by `reader.ept`. modifying the example above to read from  a local laz file with `readers.las` will also exhibit unbounded memory increase.

hobu:
Confirmed. 
![image](https://user-images.githubusercontent.com/87751/178593020-ad8efd19-79ed-42fe-a0c2-7498cd26ffa4.png)

hobu:
>  PDAL: readers.ept: Error reading tile: Unexpected end of file.

[...]

PR #3814, opened 2022-07-13
fix lazperf usage leak caused by bad #ifdef #3811

hobu:


abellgithub:
While Lazpef_excludes thing should go away

abellgithub:
👍 
I might feel guilty about this bug, but I really don't. Conditional includes are bad and if you read the code, it *looked* OK.

hobu:
easy enough to fix. Lots of laszip/lazperf legacy to unwind too. Not surprising there was an issue and maybe there will still be some more.

PR #3818, opened 2022-07-19
fix readers.las.nosrs

hobu:
`readers.las.nosrs` does not work if the global encoding bit is set for WKT but it was using GeoTIFF keys. This patch allows the user to 🔨 it and move on.

### DIFF
--- a/io/LasReader.cpp
+++ b/io/LasReader.cpp
@@ -224,7 +224,7 @@ void LasReader::initializeLocal(PointTableRef table, MetadataNode& m)
     d->header.fill(headerBuf, las::Header::Size14);
 
     uint64_t fileSize = Utils::fileSize(m_filename);
-    StringList errors = d->header.validate(fileSize);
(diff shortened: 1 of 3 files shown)


hobu:
cherry-picked back to `2.4-maintenance` at e0060cdb6ab2837b48651ba4dd7d63e1c47eec64

reiguy:
Would it be possible to add nosrs=true fix described above to the tindex reader? My input *.las have global encoding bit set for WKT but are GeoTIFF keys, and neither the readers.tindex.override_srs nor readers.tindex.t_srs will push past the error "PDAL: readers.las: Global encoding WKT flag not set for point format 6 - 10."

I suppose it would look like readers.tindex.nosrs=true

hobu:
> Would it be possible to add nosrs=true fix described above to the tindex reader?

Implemented in #4098 for the upcoming PDAL 2.6.0 release

PR #3821, opened 2022-07-20
catch up PDAL's vendored laz-perf to 3.1.0

hobu:
This should fix https://github.com/PDAL/PDAL/pull/3675#discussion_r922800902 and there were a few other missing small bits of laz-perf.


hobu:
Cherry-picked back to `2.4-maintenance` at 30e286ec531d638b430ffecfbce265faff07c568

PR #3824, opened 2022-07-26
Add wmain support to 'pdal' command for wide filename characters on Windows

hobu:
This follows along with what was done in the Untwine repository in regard to wide character filenames. Other tweaks as part of this PR include

* Dropping some old OSGeo4W CMake junk
* Converting to use `std::filesystem::canonical` instead of our homespun version
* Incorporate other windows unicode filename stuff from #3627


apps/CMakeLists.txt
```diff
@@ -30,6 +30,10 @@ target_include_directories(${PDAL_APP} PRIVATE
     ${GDAL_INCLUDE_DIR}
     ${PROJECT_BINARY_DIR}/include)
 
+IF(WIN32)
```
MehdiChinoune, line 33: This should be `MSVC`

hobu: addressed in #3856

MehdiChinoune: Thanks

PR #3837, opened 2022-08-09
writers.ogr: add option to populate feature attributes from dimensions

rcoup:
Adds the ability to write OGR datasets with point dimensions as feature attributes. Do some other tidying up of the writers.ogr driver along the way, and adds tests.

1. always create 3D/Z geometries. If there's a specific format that doesn't work with `wkbPointZM` maybe that's something we should document rather than hobbling the default output to 2D? **This is a backwards-incompatible change, previously you needed `measure_dim` to get 3D XYZ output.**
2.  fixes a bug using `multicount` where the final geometry (eg `multicount=10` with 15x points → the final 5x) didn't get its geometry set.
3. calls `SetFID()` for each feature with a sane value. Previously, creating a GeoPackage (at least) crashed because there was no primary key. The value either uses the pointId or when using `multipoint` synthesises a value.
5. adds a new `attr_dims` option:

   > List of dimensions to write as feature attributes. Separate multiple values with ``,`` or use ``all`` to write all dimensions. ``X``, ``Y``, ``Z``, and any `measure_dim` are never written as attributes. This option is incompatible with the `multicount` option. [Default: None]
```
  [
      "inputfile.las",
      {
          "type": "writers.ogr",
          "filename": "outfile.geojson",
          "measure_dim": "Intensity",
          "attr_dims": "Classification"
      }
  ]

  [
      "inputfile.las",
      {
          "type": "writers.ogr",
          "filename": "outfile.geojson",
          "attr_dims": "Classification,Red,Green,Blue"
      }
  ]

  [
      "inputfile.las",
      {
          "type": "writers.ogr",
          "filename": "outfile.geojson",
          "attr_dims": "all"
      }
  ]
```

6. Adds the ability to set OGR layer creation options via `ogr_options`:

   > List of OGR driver-specific layer creation options, formatted as an ``OPTION=VALUE`` string. Separate multiple values with ``,`` or repeat the option. [Default: None]
```
  [
      "inputfile.las",
      {
          "type": "writers.ogr",
          "filename": "outfile.shp",
          "ogrdriver": "ESRI Shapefile",
          "ogr_options": "2GB_LIMIT=YES"
      }
  ]
```
8. Improves OGR CRS initialisation and reports errors
9. Adds a pile of writers.ogr tests using `ogrinfo` output.
10. Refreshes the docs for writers.ogr and adds the `attr_dims` option

### DIFF
--- a/test/unit/io/OGRWriterTest.cpp
+++ b/test/unit/io/OGRWriterTest.cpp
@@ -1,0 +1,436 @@
+/******************************************************************************
+* Copyright (c) 2016, Hobu Inc. (info@hobu.co)
+*
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following
+* conditions are met:
+*
+*     * Redistributions of source code must retain the above copyright
+*       notice, this list of conditions and the following disclaimer.
+*     * Redistributions in binary form must reproduce the above copyright
+*       notice, this list of conditions and the following disclaimer in
+*       the documentation and/or other materials provided
+*       with the distribution.
+*     * Neither the name of Hobu, Inc. nor the
+*       names of its contributors may be used to endorse or promote
+*       products derived from this software without specific prior
+*       written permission.
+*
+* THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+* "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+* LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+* FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+* COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+* INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+* BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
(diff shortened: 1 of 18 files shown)


hobu:
Thanks @rcoup for the contribution! I wonder who is still making multipoint shapefiles though...

You'll have to iterate a bit to get everything ✅, but we will plan to include this in 2.5.0 once it is complete.

rcoup:
> I wonder who is still making multipoint shapefiles though...

They still work as-is, but i'll happily rip out multicount if you want.

> You'll have to iterate a bit to get everything ✅

Yeah, I wasn't seeing those warnings locally under clang, but I'll figure them out.

hobu:
> They still work as-is, but i'll happily rip out multicount if you want.

No, it is fine. Gross 😄, but fine...

rcoup:
@hobu can you trigger CI please? I've simplified some of the tests to make them more deterministic and ctest is green for me here under macOS & Linux.

rcoup:
@hobu one more for the Windows in the back? 😁

hobu:
I can't push to this branch but you need to export ErrorHandlerSuspender in `pdal/private/gdal/ErrorHandler.hpp`

rcoup:
@hobu run CI again? 🙏

test/unit/io/OGRWriterTest.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+    auto handle = FileUtils::createFile(outinfofile);
+    *handle << info;
+    FileUtils::closeFile(handle);
+
+    EXPECT_EQ(compare(infofile, outinfofile, 1), 0) \
+        << "  (" << infofile << " <-> " << outinfofile << ")";
+}
+
+uint32_t diff_geojson(const std::string& file1, const std::string& file2, int32_t ignoreLine1=-1)
```
rouault, line 125: Comparing output of ogrinfo seems fragile to me.
Did you consider adding a reference/expected geojson file, and using OGR API to compare the one generated by the test with the reference one ? OGR_F_Equal() should help for that

Issue #3215, opened 2020-08-18
Add numpts band to writers.gdal output


hobu:
``writers.gdal`` output has a ``count`` band, but that is the count of the points used for the IDW interpolation, and not a simple count of the number of points in the cell. I propose adding a ``numpts`` attribute to include this. 

Alternatively, we could add an option to switch ``writers.gdal``'s statistics mode to use only points in the cell rather than in the interpolation window (count, stddev, mean, max).

rphillips91:
Another use case needed for writers.gdal statsitics mode: 
To convert vertical datums in geotiffs which needs a cell to cell conversion not currently available in pdal.

hobu:
> To convert vertical datums in geotiffs which needs a cell to cell conversion not currently available in pdal.

I don't understand this scenario at all. Could you please explain more? You should be able to apply vertical datums using PROJ and GDAL with PDAL as-is right now.

rphillips91:
```
{
    "pipeline" :

[
    {
            "type": "readers.gdal",
            "filename": "input.tif"
    },
    {
            "type": "filters.reprojection",
            "in_srs": "+init=EPSG:2193 +geoidgrids=C:/dev/lidar-qc/las_qclegacy/vdatum_nzvd09_16/nzgeoid2009.gtx",
            "out_srs": "EPSG:2193"
    },
    {
            "type": "filters.reprojection",
            "in_srs": "EPSG:2193",
            "out_srs": "+init=EPSG:2193 +geoidgrids=C:/dev/lidar-qc/las_qclegacy/vdatum_nzvd09_16/nzgeoid2016.gtx"
    },       
    {
        "type": "writers.gdal",
        "filename": "outfile.tif",
        "resolution" : 1        
    }
]
}
```

adamsteer:
@rphillips91 is this a task where `gdal_translate` would be better used?

@hobu a `writers.gdal.cell_stats` (or similar) option would rather selfishly solve https://github.com/adamsteer/forestmetrics/issues/4 ;)

rphillips91:
@adamsteer I did try gdalwarp and curently attempting to figure out why it is erroring out, for the first grid transformation (of the above pipeline). Using gdalwarp/gdaltranslate would mean a two step process with an intermediate file (correct me if I am wrong).
This is where the pdal pipeline comes in really handy.

hobu:
Use the new `binmode` in `writers.gdal` to get what you need here.

PR #3840, opened 2022-08-10
Add 'binmode' to writers.gdal to support non-interpolated rasters #3215

hobu:


hobu:
Here's the distinction in `count` between "bin" mode and "distance" mode. In distance mode, `writers.gdal` interpolates to the range, and this includes counting points that influence that. In bin mode, it only counts points that are inside the bin. The latter is useful for stuff like [lasgrid](https://rapidlasso.com/lastools/lasgrid/) operations and rasterization of categorical data like `Classification`.

distance
bin

adamsteer:
this is fantastic - directly useful, especially thinking about some old forest canopy structure work. Thanks!

PR #3843, opened 2022-08-11
[skip ci] try to silence doxygen noise to only warning/log

hobu:


hobu:
These two updates dramatically reduce the doc build log noise so it is easier to find real errors/warnings that break the build.

PR #3845, opened 2022-08-15
Update Utils.cpp

bacek:
Fix typo in Utils::escapeJSON


hobu:
Picked back to `2.4-maintenance` with 991ff6e6d8b4032561f7ac0e0843d151fc8210dc 

I don't know if we are going to have another 2.4.x release, but if we do, it will be included.

PR #3865, opened 2022-09-17
writers.ogr: don't set OGR FID

rcoup:
OGR FIDs need to be unique, but if a pipeline includes a merge filter `pointId`s may not be. OGR will generate a unique sequence anyway, and the user can always pass `pointId` as a dimension to create an attribute if they want it explicitly.

An additional change on top of #3863, since they both edit the same testcase.

### DIFF
--- a/io/OGRWriter.cpp
+++ b/io/OGRWriter.cpp
@@ -270,7 +270,6 @@ bool OGRWriter::processOne(PointRef& point)
         else
         {
             m_feature->SetGeometry(&pt);
-            m_feature->SetFID(point.pointId());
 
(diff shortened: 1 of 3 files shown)


abellgithub:
Just to be clear, point IDs being non-unique has nothing to do with a merge filter.

rcoup:
>  Just to be clear, point IDs being non-unique has nothing to do with a merge filter.

Sure — is just an easy reproducer :-)

hobu:
Can you rebase master onto this? I think we have all of the CI stuff fixed now.

craigds:
I think this is actually buggy with GDAL < 3.5. I'll try and fix it up

craigds:
This should be good to go now

Issue #3870, opened 2022-10-05
ioctl returns an empty winsize object is running older version of mac and newer version of xcode.


kylemann16:
**Describe the bug**
When running `lldb` from the terminal and interacting with a locally built pdal I find this error:
```
(lldb) run bin/pdal --drivers
error: invalid target, create a target using the 'target create' command
(lldb) target create bin/pdal
Current executable set to '/Users/kmann/code/pdal/build/bin/pdal' (x86_64).
(lldb) run --drivers
Process 28654 launched: '/Users/kmann/code/pdal/build/bin/pdal' (x86_64)
libc++abi.dylib: terminating with uncaught exception of type std::bad_alloc: std::bad_alloc
Process 28654 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fff72e0833a libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff72e0833a <+10>: jae    0x7fff72e08344            ; <+20>
    0x7fff72e0833c <+12>: movq   %rax, %rdi
    0x7fff72e0833f <+15>: jmp    0x7fff72e02629            ; cerror_nocancel
    0x7fff72e08344 <+20>: retq   
Target 0: (pdal) stopped.
```
Which is created from [these lines](https://github.com/PDAL/PDAL/blob/master/pdal/util/Utils.cpp#L536-L541), where `ioctl` is called to determine the terminal window size. Because I am running an older version of macos, but xterm continues to receive updates, an unexpected bug occurred as above, where it returns 0 instead of the correct amount. [Here](https://github.com/PDAL/PDAL/blob/master/apps/pdal.cpp#L134-L137) operations are done on the returned value assuming it to be correct. The above bug comes from trying to form a negative number of `=` characters into a string.

**Expected behavior**
I expected `ioctl` to return the correct terminal column length, but instead it returned 0.

**System/installation information:**
Mac version 10.15.7
XCode version 12.4

```
$ uname -a
Darwin Kyles-MBP 19.6.0 Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64 x86_64
```

```
$ pdal --version
----------------------------------------------------------------------------------------------------------------
pdal 2.5.0 (git-version: db4257)
----------------------------------------------------------------------------------------------------------------
```

[...]

PR #3871, opened 2022-10-05
Return line length of 80 if terminal window column size is 0

kylemann16:
Fixes bug found in #3870 by checking size and returning 80 as a default.

### DIFF
--- a/pdal/util/Utils.cpp
+++ b/pdal/util/Utils.cpp
@@ -538,6 +538,9 @@ int Utils::screenWidth()
     if (ioctl(0, TIOCGWINSZ, &ws))
         return 80;
 
+    if (!ws.ws_col)
+      return 80;
+
     return ws.ws_col;
 #endif
 }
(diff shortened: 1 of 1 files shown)


pdal/util/Utils.cpp
```diff
@@ -538,6 +538,9 @@ int Utils::screenWidth()
     if (ioctl(0, TIOCGWINSZ, &ws))
         return 80;
 
+    if (!ws.ws_col)
```
hobu, line 541: We should check the return of `ioctl` and report an error if something is up.

hobu (changes requested):

pdal/util/Utils.cpp
hobu: I meant it would be nice to tell the user what is going on. That it would fail is weird and quite unexpected. Silently defaulting to 80 seems presumptuous. 

https://man7.org/linux/man-pages/man2/ioctl.2.html

```
       EBADF  fd is not a valid file descriptor.

       EFAULT argp references an inaccessible memory area.

       EINVAL request or argp is not valid.

       ENOTTY fd is not associated with a character special device.

       ENOTTY The specified request does not apply to the kind of object
              that the file descriptor fd references.
```

hobu:
Done in #3893

PR #3873, opened 2022-10-09
Make PDAL Locale safe

### LINKED ISSUE #3450: Locale unsafe
**Describe the bug**
PDAL library is not safe on different "locale".
I noticed the problem getting the error `X scale of 0.0 is invalid!` (when I wrote clearly 1e-3 for the scale) in my German computer, that was not the case in other computers.
The library is used in a program that sets the locale to the one defined in the system (German in my case)
Numbers in German and in other languages (like Spanish) are written with comma `,` as decimal separator, not the dot `.`

Detected in Ubuntu 20.04, but probably happening in other operating systems.

**Way to reproduce in the tests**
in the file `pdal/pdal_test_main.hpp`, add the line `std::locale::global(std::locale("de_DE.utf8"));` before calling `RUN_ALL_TESTS`. You need the locale to be installed in you OS.
You can also add `std::locale::global(std::locale(""));` to use the default locale from your system (German in my case).
(I do not know how to set it if the proper locale is not installed... it can be tricky for future tests. Maybe we need a docker with the locale installed).

**Possible solution**
`std::stringstream` family supports setting an specific locale with the method `imbue`. Something like `oss.imbue(std::locale::classic());` makes that object locale immune to the configuration done in the application. We want that PDAL library writes (and read) `1.2` as a number, not `1,2` when parsing a string. (it is the case, right?)
That means adding this `imbue` in many places.
Unfortunately `std::stod` and `str::strtod` do not support a different locale, and use the global one. That means that they shouldn't be used.

Example from a test: ([55,55],[55,55]) is a BOX2D with 4 numbers (55). However in German locale, reads 2 numbers (55.55)... and fails.

I started working in a branch (https://github.com/jjimenezshaw/PDAL/tree/locale-safe) to fix that, replacing `std::stod` with `stringstream` (so far the Bounds parsing) and setting `imbue(locale::classic())` wherever it is used. Once it is done, it will become a PR.

Before spending so much time on it, I want to know if it makes sense to you, or if you have a better suggestion.

Thanks.

PS: changing the locale does not change the language of the OS. It affects things like date format (another nightmare), currency, number format (the problem that we address here), etc.

jjimenezshaw:
Fixes #3450 

## Description of the problem
Many C/C++ functions like `atof`, `strtod`, `to_string`, `stringstream`, etc do take into consideration the C locale (in both directions: number to string and string to number). The C locale is a global setting that can be set with a function like `std::locale::global(std::locale("de_DE.utf8"));` (in this example for the German locale... that has to be installed in the OS at execution time). https://en.cppreference.com/w/cpp/locale/locale/global

The global locale can be set by any application or library, and affects globally the running program, including any linked library.

Some locales (like German, Spanish, French, Dutch, etc) have a different number format, using `,` (comma) as decimal separator, and `.` (period) as thousand separator. (Yes, the opposite as English). Not all similar locales behave the same way. For instance `de_DE` uses `,` as decimal separator, while `de_CH` (Swiss German) uses `.`

## Simple example
```C++
#include <locale>
#include <iostream>
#include <sstream>
int main()
{
    std::locale::global(std::locale("de_DE.utf8"));
    float a = atof("42.5");
    std::cout << a << std::endl;
    std::ostringstream oss;
    oss << 3.14 << " " << 65535;
    std::cout << oss.str() << std::endl;
    std::cout << std::to_string(1.23456789) << std::endl;
    std::istringstream iss("2.5");
    double d = 0;
    iss >> d;
    std::cout << d << std::endl;
    return 0;
}
```
produces this output (remember, you have to install that locale in your system)
```
42
3,14 65.535
1,234568
25
```
instead of what you may expect
```
42.5
3.14 65535
1.234568
2.5
```
## Consequences in PDAL
PDAL is using JSON format for many string parsing. That means that numbers MUST use `.` as decimal separator, regardless the locale set by anybody/anything. Currently the code is not safe against a change in the global locale, or even calling `set::locale`.

## Solution used
In C++, you can "imbue" a locale into a stream to modify the behavior of that single instance. Setting `std::locale::classic()` sets the standard US English behavior. Actually it was (only) already done in PDAL in the file `io/PcdHeader.cpp`. To avoid calling `imbue(std::locale::classic())` everywhere, a wrapper class `ClassicLocaleStream` is created that calls `imbue` in its constructor.

In this PR the call to `imbue` only appears twice: in `pdal/util/Utils.hpp` (where other functions to convert to and from string were) and in `plugins/e57/libE57Format/src/Common.h` as E57 code has the same problem.

Unfortunately `TileDB` has the same problem (easy to check setting the global locale in their file `test/src/unit.cc`). I did open [an issue in their repository](https://github.com/TileDB-Inc/TileDB/issues/3555).

### How to run test with any specific locale
In the first commit I modify `pdal/pdal_test_main.hpp` to understand the environment variable `TEST_SET_LOCALE`, and call `std::locale::global` with that value. That may help you to run the tests in any locale without recompiling.
### Test in GitHub Actions 
To test it I have created a new test run in Linux. In Linux it is easy to install a locale in the OS. This test `Test with German locale` is run _in addition_ to the normal tests (see the second commit).
### Disabling TileDB tests
4 tests for TileDB are currently failing. I guess it is due to the same problem in that library. While it is fixed in their repo, I have disabled them in that new test set.
### What about Windows?
I do not have a Windows environment to test it. And I do not know how to set it up in GitHub Actions. @wrenoud tested the problem some weeks ago and it was reproducible (thanks for your help).

### It happens also with integer values
The thousands separator may be used (see simple example above), that will break any future parsing.
### Finding "every" use case in the code
To fix this problem I have run the test in Linux. Many of them where failing. However, if a test is not using numbers that trigger the failure (for instance, floating point numbers without decimal values), it may not be detected in the tests... and not fixed. I have not replaced every `stringstream` but only those that I though were needed. For instance I have not changed the log messages... that may confuse the user showing a number in a format that is not the same as in the file.

### DIFF
--- a/pdal/XMLSchema.cpp
+++ b/pdal/XMLSchema.cpp
@@ -431,7 +431,7 @@ bool XMLSchema::load(xmlDocPtr doc)
                     std::cerr << "Unable to fetch minimum value.\n";
                     return false;
                 }
-                dim.m_min = std::atof((const char*)n);
+                Utils::fromString((const char*)n, dim.m_min);
                 xmlFree(n);
             }
             if (propName == "maximum")
@@ -442,7 +442,7 @@ bool XMLSchema::load(xmlDocPtr doc)
                     std::cerr << "Unable to fetch maximum value.\n";
                     return false;
                 }
-                dim.m_max = std::atof((const char*)n);
+                Utils::fromString((const char*)n, dim.m_max);
                 xmlFree(n);
             }
             if (propName == "position")
@@ -570,13 +570,13 @@ void XMLSchema::writeXml(xmlTextWriterPtr w) const
         xmlTextWriterStartElementNS(w, (const xmlChar*)"pc",
             (const xmlChar*)"dimension", NULL);
 
-        std::ostringstream position;
+        Utils::OStringStreamClassicLocale position;
         position << (pos + 1);
         xmlTextWriterWriteElementNS(w, (const xmlChar*)"pc",
             (const xmlChar*)"position", NULL,
             (const xmlChar*)position.str().c_str());
 
-        std::ostringstream size;
+        Utils::OStringStreamClassicLocale size;
         size << Dimension::size(di->m_dimType.m_type);
         xmlTextWriterWriteElementNS(w, (const xmlChar*)"pc",
             (const xmlChar*)"size", NULL, (const xmlChar*)size.str().c_str());
@@ -590,7 +590,7 @@ void XMLSchema::writeXml(xmlTextWriterPtr w) const
         XForm xform = di->m_dimType.m_xform;
         if (xform.nonstandard())
         {
-            std::ostringstream out;
+            Utils::OStringStreamClassicLocale out;
             out.precision(15);
 
             out << xform.m_scale.m_val;
--- a/plugins/e57/libE57Format/src/CheckedFile.cpp
+++ b/plugins/e57/libE57Format/src/CheckedFile.cpp
@@ -276,14 +276,14 @@ CheckedFile& CheckedFile::operator<<(const ustring& s)
 
 CheckedFile& CheckedFile::operator<<(int64_t i)
 {
-   ClassicLocaleStream<stringstream> ss;
+   stringstream ss;
    ss << i;
    return(*this << ss.str());
 }
 
 CheckedFile& CheckedFile::operator<<(uint64_t i)
 {
-   ClassicLocaleStream<stringstream> ss;
+   stringstream ss;
    ss << i;
    return(*this << ss.str());
 }
@@ -306,7 +306,7 @@ template<class FTYPE> CheckedFile& CheckedFile::writeFloatingPoint(FTYPE value, 
    cout << "CheckedFile::writeFloatingPoint, value=" << value << " precision=" << precision << endl;
 #endif
 
-   ClassicLocaleStream<stringstream> ss;
+   stringstream ss;
    ss << scientific << setprecision(precision) << value;
 
    /// Try to remove trailing zeroes and decimal point
--- a/plugins/e57/libE57Format/src/CheckedFile.cpp
+++ b/plugins/e57/libE57Format/src/CheckedFile.cpp
@@ -276,14 +276,14 @@ CheckedFile& CheckedFile::operator<<(const ustring& s)
 
 CheckedFile& CheckedFile::operator<<(int64_t i)
 {
-   stringstream ss;
+   ClassicLocaleStream<stringstream> ss;
    ss << i;
    return(*this << ss.str());
 }
 
 CheckedFile& CheckedFile::operator<<(uint64_t i)
 {
-   stringstream ss;
+   ClassicLocaleStream<stringstream> ss;
    ss << i;
    return(*this << ss.str());
 }
@@ -306,7 +306,7 @@ template<class FTYPE> CheckedFile& CheckedFile::writeFloatingPoint(FTYPE value, 
    cout << "CheckedFile::writeFloatingPoint, value=" << value << " precision=" << precision << endl;
 #endif
 
-   stringstream ss;
+   ClassicLocaleStream<stringstream> ss;
    ss << scientific << setprecision(precision) << value;
 
    /// Try to remove trailing zeroes and decimal point
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -65,3 +65,16 @@ jobs:
       run: |
           source ./scripts/ci/linux/examples.sh
 
+    - name: Install German locale
+      env:
+        DEBIAN_FRONTEND: noninteractive
+        TZ: Etc/UTC
+      run: sudo apt-get install -y tzdata locales && sudo locale-gen de_DE.UTF-8
+
+    - name: Test with German locale
+      env:
+        TEST_SET_LOCALE: de_DE.UTF-8
+      run: |
+          source ../scripts/ci/linux/test.sh
+      working-directory: ./build
+
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -71,10 +71,12 @@ jobs:
         TZ: Etc/UTC
       run: sudo apt-get install -y tzdata locales && sudo locale-gen de_DE.UTF-8
 
-    - name: Test with German locale
+    - name: Test with German locale (except io_tiledb)
       env:
         TEST_SET_LOCALE: de_DE.UTF-8
       run: |
-          source ../scripts/ci/linux/test.sh
+          # Exclude 4 tiledb io tests that fail due to TileDB library implementation
+          # please, remove the exception rule here when fixed
+          source ../scripts/ci/linux/test.sh -E 'pdal_io_tiledb_(time_writer|time_reader|writer|reader)_test'
       working-directory: ./build
 
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -76,7 +76,8 @@ jobs:
         TEST_SET_LOCALE: de_DE.UTF-8
       run: |
           # Exclude 4 tiledb io tests that fail due to TileDB library implementation
+          # and 2 e57 io tests for the same reason.
           # please, remove the exception rule here when fixed
-          source ../scripts/ci/linux/test.sh -E 'pdal_io_tiledb_(time_writer|time_reader|writer|reader)_test'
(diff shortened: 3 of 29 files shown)


jjimenezshaw:
> At first glance this strikes me as excessive change. The issue that you described is parsing floating point input. Is there some additional issue that has caused trouble?

The issue is both for input and output. And not only floating point, but also integers, as I explain in the description.

abellgithub:
Is the thousand's separator an issue?

jjimenezshaw:
> Is the thousand's separator an issue?

Well, I "suffered" it in a unittest. If you write the string `65.535`, how should it be parsed later? (I use that number in the example in the description).

abellgithub:
If you're typing an integer like "12.345", then clearly you're expecting the locale to apply. This is why I find this entire issue so fraught. I don't understand why a user doesn't simply disable the locale interpretation if that's what they want. PDAL CAN'T know what you expect, and when you set your local to German or Spanish or whatever, I would think that you should expect the translation that is specified by the locale. It seems bold of PDAL or any other library to override what the user has requested. I don't believe that PDAL does its own floating point interpretation from strings (though there may some odd case). I believe there are home-grown integer interpretations, but those could be replaced.

abellgithub:
I'll add that I believe that locale support in general was a bad thing, but we are where we are. It's only a little bit worse than wstring.

abellgithub:
Your initial issue said something like "1e-3" in some context didn't work. Should it have worked? If not why not?

Anyway, I'm not in charge, but I don't much like the idea of throwing the locale away when someone has set it. If 1e-3 is valid floating point in the german locale and wasn't accepted, I think we should examine that issue and deal with it, rather than try to force users into C/Classic locale despite them having set something else.

jjimenezshaw:
@abellgithub we already discussed this in https://github.com/PDAL/PDAL/issues/3450

When I said "if you write the string `65.535`" I wanted to say that PDAL is writing `65.535` in some cases in a formatted file. That is not what you want for an integer value, but `65535`. As @hobu said PDAL format should be like JSON, that does not care about the locale of the user. (You can ask your friends living in Europe how crazy is to import some text files in Excel, as @wrenoud said once).

PDAL currently may be writing and reading files/streams taking into consideration the user locale!

This PR tries to _fix_ PDAL to behave as you really expect.

Otherwise, why are that many unit test failing just setting a different locale? It should be "immune", as a proper JSON parser.

> PDAL CAN'T know what you expect, and when you set your local to German or Spanish or whatever, I would think that you should expect the translation that is specified by the locale.

When your library is reading or writing a file/stream in a specified format that says that decimal separator must be `.`, it _must_ ignore the locale of the running computer, and read/write numbers using `.` as decimal separator. Unfortunately C/C++ formatting functions are not doing that, so you have to take special care when your code is reading or writing formatted numbers. (I also hate this problem in C/C++, but "it is what it is". I cannot change the standard).

> Your initial issue said something like "1e-3" in some context didn't work. Should it have worked? If not why not?

The problem there is that PDAL in my German computer was writing somewhere internally the string `0,001` when I provided the C++ value `double(1e-3)`. PDAL was not working in my computer to create a LAS file (but I did for other colleagues) even when I was not providing any text string. That's how I found the problem initially.

jjimenezshaw:
https://github.com/TileDB-Inc/TileDB/issues/3555#issuecomment-1275469201

pdal/XMLSchema.cpp
```diff
@@ -431,7 +431,9 @@ bool XMLSchema::load(xmlDocPtr doc)
                     std::cerr << "Unable to fetch minimum value.\n";
                     return false;
                 }
-                dim.m_min = std::atof((const char*)n);
+                Utils::ClassicLocaleIStringStream iss((const char*)n);
+                iss >> dim.m_min;
+                iss.str("");
```
rouault, line 436: is ``iss.str("");`` really needed ? If so, a comment in the code would be welcome. Same at line 449

rouault:
Looks good to me. We had the same sort of issues in GDAL and PROJ that were solved by using std::locale::classic() at relevant places.

pdal/XMLSchema.cpp
jjimenezshaw: Thanks @rouault . I was afraid that the destructor of `istringstream` could conflict with `xmlFree` that is called before the constructor. Now reading carefully the documentation of `istringstream`, I see that it is created with a copy of `n`, so it is safe (if I understood correctly). I will remove that line.

jjimenezshaw:
> Please rebase current master against this

Already rebased. All actions are green.

hobu (changes requested):
I'll admit I'm anxious about this. I recognize the need, and this is going in the correct direction, but I want to see a bit more in two areas:

1) There's not enough tests that have been added. ClassicLocaleStringStream and friends should have some explicit tests that demonstrate behavior and usage. Also, the place where locale stuff is the most difficult to deal with, Windows, is not tested at all.

2) There's some spots where creating big fat streams all the time are going to hurt performance. I tried to identify a few of those, but there might be some more. Updating them probably means changing PDAL's code a bit more.

rouault:
Maybe the following can help
```
#include <string>
#include <locale>
#include <sstream>

double c_locale_stod(const std::string& s)
{
    std::istringstream iss(s);
    iss.imbue(std::locale::classic());
    double d;
    iss >> d;
    if (!iss.eof() || iss.fail()) {
        throw std::invalid_argument("non double value");
    }
    return d;
}


double c_locale_stod_thread_local(const std::string& s)
{
    static thread_local std::istringstream iss = []() {
        std::istringstream l_iss;
        l_iss.imbue(std::locale::classic());
        return l_iss;
    }();
    double d;
    iss.clear();
    iss.str(s);
    iss >> d;
    if (!iss.eof() || iss.fail()) {
        throw std::invalid_argument("non double value");
    }
    return d;
}

int main() {
    std::string s("1.23");
    for(int i = 0; i < 1000 * 1000; ++i)
    {
        //std::stod(s);
        //c_locale_stod_thread_local(s);
        //locale_stod(s);
    }
    return 0;
}
```
On my Linux box, above test code when built with -O2:
when using std::stdod(), timing is 77 ms
when using c_locale_stod_thread_local(), timing is 195 ms
when using c_locale_stod(), timing is 487 ms

plugins/e57/libE57Format/src/CheckedFile.cpp
```diff
@@ -276,14 +276,14 @@ CheckedFile& CheckedFile::operator<<(const ustring& s)
 
 CheckedFile& CheckedFile::operator<<(int64_t i)
 {
-   stringstream ss;
+   ClassicLocaleStream<stringstream> ss;
```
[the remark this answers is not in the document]
jjimenezshaw: Thanks @hobu I hope I addressed all your points.

> * There's not enough tests that have been added. ClassicLocaleStringStream and friends should have some explicit tests that demonstrate behavior and usage. Also, the place where locale stuff is the most difficult to deal with, Windows, is not tested at all.
> 

I added tests for ClassicLocale stringstreams. About Windows, unfortunately I do not have a computer to test, and neither the knowledge to configure the GitHub actions. @wrenoud tested some things in #3450 using `TEST_SET_LOCALE=nl_NL` in Windows.

> * There's some spots where creating big fat streams all the time are going to hurt performance. I tried to identify a few of those, but there might be some more. Updating them probably means changing PDAL's code a bit more.

I added @rouault's idea of a static thread_local stream to avoid the creation of the stream in `fromString` and a few more places that you pointed as critical.
Most of the changes tried to be minimal to not produce a big PR. But for performance reasons some parts of the code could be optimized as you say. I do not know which parts of the code are called massively, that could impact more the global performance.

jjimenezshaw:
Thanks @hobu I hope I addressed all your points.

> * There's not enough tests that have been added. ClassicLocaleStringStream and friends should have some explicit tests that demonstrate behavior and usage. Also, the place where locale stuff is the most difficult to deal with, Windows, is not tested at all.
> 

I added tests for ClassicLocale stringstreams. About Windows, unfortunately I do not have a computer to test, and neither the knowledge to configure the GitHub actions. @wrenoud tested some things in #3450 using `TEST_SET_LOCALE=nl_NL` in Windows.

> * There's some spots where creating big fat streams all the time are going to hurt performance. I tried to identify a few of those, but there might be some more. Updating them probably means changing PDAL's code a bit more.

I added @rouault's idea of a static thread_local stream to avoid the creation of the stream in `fromString` and a few more places that you pointed as critical.
Most of the changes tried to be minimal to not produce a big PR. But for performance reasons some parts of the code could be optimized as you say. I do not know which parts of the code are called massively, that could impact more the global performance.

PR #3879, opened 2022-10-17
Fix writers.las not preserving classFlags to PDRF3

### LINKED ISSUE #3878: writers.las doesn't preserve classification flags when dataformat_id < 6
**Describe the bug**

While researching a possible fix for #1701 I have discovered that writers.las does not preserve classification flags when writing to legacy PDRFs.

This was probably the actual root cause of #3092.

To demonstrate, here's a 1-point CSV file with Classification=7 and ClassFlags=2:

[input.csv](https://github.com/PDAL/PDAL/files/9796212/input.csv)

With this pipeline:

```json
[
{"type": "readers.text", "filename": "input.csv"},
{"type": "writers.las", "dataformat_id": 3, "filename": "pdrf3.laz"}
]
```

We should expect that this should produce a Classification value of 71 (`7 + 2*32`) in the output file.

However, this is not the case:

```
$ echo '[
{"type": "readers.las", "filename": "pdrf3.laz"},
{"type": "filters.stats", "dimensions": "Classification", "count": "Classification"}]' | pdal pipeline --stdin --metadata=STDOUT | jq -SC '.stages["filters.stats"].statistic'
[
  {
    "average": 7,
    "count": 1,
    "counts": [
      "7.000000/1"
    ],
    "maximum": 7,
    "minimum": 7,
    "name": "Classification",
    "position": 0
  }
]
```

pdal 2.5.0 (git-version: 896648)

built using cmake from latest master.
MacOS 12.5.1

craigds:
Fixes #3878.

When running a pipeline which reads from a PDRF 6+ source, and then writing to a PDRF < 6 file, PDAL was not preserving the flag bits correctly into the classification field.

This change fixes the issue. I also added a warning when writing any classification values over 31, since those are not supported by PDRF < 6.

Previously PDAL was preserving such values, so it is *possible* someone could be relying on that behaviour. I have added a warning if this happens and also set the classification value to 1 (unclassified) since that seemed the most sensible value of the choices available.

Since this is only an issue if the user is downgrading the PDRF and *also* using user-defined or reserved classification values which are >31, this situation should hopefully be very rare.

### DIFF
--- a/io/LasWriter.cpp
+++ b/io/LasWriter.cpp
@@ -883,17 +883,24 @@ bool LasWriter::fillPointBuf(PointRef& point, LeInserter& ostream)
     uint8_t scanDirectionFlag = point.getFieldAs<uint8_t>(Id::ScanDirectionFlag);
     uint8_t edgeOfFlightLine = point.getFieldAs<uint8_t>(Id::EdgeOfFlightLine);
     uint8_t classification = point.getFieldAs<uint8_t>(Id::Classification);
+    uint8_t classFlags;
+    if (point.hasDim(Id::ClassFlags))
+    {
+        // source file is PDRF >= 6
+        classFlags = point.getFieldAs<uint8_t>(Id::ClassFlags);
+    }
+    else
+    {
+        // source file is PDRF < 6
+        classFlags = classification >> 5;
+        classification &= 0x1F;
+    }
 
     if (has14PointFormat)
     {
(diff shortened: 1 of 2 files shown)


hobu:
I thought we previously had this behavior, so @craigds's questions about it being broken were a bit disorienting to me. Once this is ✅, I would certainly support merging it (note that the Conda builder is broken because of a Conda Forge windows compiler behavior change that I have not had time to investigate).

PR #3892, opened 2022-11-18
screenWidth for Windows and unix

hobu:
A more complete implementation of #3871


abellgithub:
Code doesn't match description.

hobu:
indeed. fixing.

PR #3895, opened 2022-11-18
STAC Reader

kylemann16:
This Pull Request adds the STAC reader and JSON schema validation as features to PDAL. JSON schema validation will be added under PDAL's vendor directory, and the STAC Reader will be an available plugin.

STAC Reader Notes:
 - Can take STAC Individual Items or Catalogs
 - Catalogs will read through each item available, with items pruned by the arguments you specify.
 - Available pruning options
     - Date ranges
     - Bounds
     - ids (list of regex strings)
     - properties (key value match with properties field of STAC Items)
 - JSON Schema validation will go match against publicly available STAC JSON schema and will iterate through extensions to make sure those match as well.


cmake/macros.cmake
```diff
@@ -202,7 +202,8 @@ macro(PDAL_ADD_TEST _name)
         ${PDAL_ADD_TEST_INCLUDES}
         ${PROJECT_SOURCE_DIR}/test/unit
         ${PROJECT_BINARY_DIR}/test/unit
-        ${PROJECT_BINARY_DIR}/include)
+        ${PROJECT_BINARY_DIR}/include
+        ${NLOHMANN_INCLUDE_DIR})
```
hobu, line 206: I think we only want to add `${NLOHMANN_INCLUDE_DIR}` to individual tests. Or at least that's what we did in the past.

cmake/nlohmann_jsonConfig.cmake
```diff
@@ -0,0 +1,20 @@
+#
```
hobu, line 1: what's this stuff all about, and isn't it going to find the system nlohmann? We definitely don't want that to happen.

pdal/QuickInfo.hpp
```diff
@@ -34,6 +34,9 @@
 
 #pragma once
 
+#include <pdal/JsonFwd.hpp>
+#include <nlohmann/json.hpp>
```
hobu, line 38: I think we just want the fwd here and the implementation in the QuickInfo.cpp file.

pdal/util/Utils.cpp
```diff
@@ -538,6 +538,9 @@ int Utils::screenWidth()
     if (ioctl(0, TIOCGWINSZ, &ws))
         return 80;
 
+    if (!ws.ws_col)
```
hobu, line 541: I have a more extensive PR in #3893 that should address this.

plugins/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ if (BUILD_PLUGIN_TRAJECTORY)
     add_subdirectory(trajectory)
 endif()
 
+if (BUILD_PLUGIN_STAC)
```
hobu, line 8: Should this be a plugin or an always-on driver? The dependencies and vendored code are such that it could be always-on.

plugins/pgpointcloud/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ PDAL_ADD_PLUGIN(reader_libname reader pgpointcloud
         ${LIBXML2_LIBRARIES}
     INCLUDES
         ${LIBXML2_INCLUDE_DIR}
+        ${NLOHMANN_INCLUDE_DIR}
```
hobu, line 17: this shouldn't be needed in pgpointcloud

plugins/stac/io/stacReader.hpp
```diff
@@ -0,0 +1,102 @@
...
+            NL::json m_readerArgs;
+            std::vector<std::string> m_idList;
+            std::vector<std::unique_ptr<Stage>> m_readerList;
+            std::mutex m_mutex;
+
+            StageFactory m_factory;
+            MergeFilter m_merge;
+            PointViewSet m_pvSet;
+
+            virtual void handleReaderArgs();
```
hobu, line 88: add `override` to these methods for more explicit statement of what they are.

cmake/macros.cmake
kylemann16: We could, but it's pretty much every reader test plus quite a few others. I had a lot of them written out, and thought it would be cleaner to do this. Could definitely go back to the other style.

plugins/pgpointcloud/CMakeLists.txt
kylemann16: These were both in the list of things that were erroring on compile when I added nlohmann to QuickInfo.hpp

hobu:
Moved over to #3899

PR #3899, opened 2022-11-25
readers.stac implementation

hobu:
A number of tweaks and updates of #3895 due to it not being editable.


io/StacReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+        {
+            log()->get(LogLevel::Debug) << "Processing extension " << extSchemaUrl << std::endl;
+            std::string schemaStr = m_arbiter->get(extSchemaUrl);
+            NL::json schemaJson = NL::json::parse(schemaStr);
+            val.set_root_schema(schemaJson);
+            val.validate(stacJson);
+        }
+    }
+    else if (type == "Catalog")
+        schemaUrl = "https://schemas.stacspec.org/v1.0.0/catalog-spec/json-schema/catalog.json";
```
hobu, line 210: @kylemann16 this URL should be a defaulted argument so the users can control which URL they want to try to apply schema validation too

io/StacReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+        NL::json args = m_readerArgs[driver].get<NL::json>();
+        for (auto& arg : args.items()) {
+            readerOptions.add(arg.key(), arg.value());
+        }
+    }
+
+    readerOptions.add("filename", dataUrl);
+    reader->setOptions(readerOptions);
+
+    if (m_readerList.size() > 0)
```
hobu, line 254: If the m_readerList size is 0, is that an error?

io/StacReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+        reader->setInput(*m_readerList.back());
+
+    if (reader)
+    {
+        std::lock_guard<std::mutex> lock(m_mutex);
```
hobu, line 259: Why are we locking here? Is this coming for the TIndex reader?

io/StacReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+        {
+            int value = properties[key].get<int>();
+            int desired = val.get<int>();
+            if (value != desired)
+                return false;
+            break;
+        }
+        case NL::detail::value_t::number_float:
+        {
+            int value = properties[key].get<int>();
```
hobu, line 312: Is this bad copypasta? You're in the float case but fetching and asking for ints...

doc/stages/readers.stac.rst
```diff
@@ -0,0 +1,78 @@
...
+Example
+--------------------------------------------------------------------------------
+
+.. code-block:: json
+[
+    {
+        "type": "readers.stac",
+        "filename": "https://s3-us-west-2.amazonaws.com/usgs-lidar-stac/ept/catalog.json",
+        "reader_args": [{"type": "readers.ept", "resolution": 100}],
+        "ids": ["MD_GoldenBeach_2012"],
```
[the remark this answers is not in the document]
hobu: Also, point to std::regex docs as to the *type* of regex that is consumed.

io/StacReader.cpp
kylemann16: Not in this spot, but maybe elsewhere. Here we're just making sure that there are readers in the list before applying the previous back to the current reader's input.

io/StacReader.cpp
kylemann16: This is coming from [StageFactory.cpp](https://github.com/PDAL/PDAL/blob/master/pdal/StageFactory.cpp#L104-L114), which I was modeling this off of. I was thinking this reader should be multi threaded at that point, but didn't go through the rest of that. So I should probably remove it.

io/StacReader.cpp
kylemann16: Whoops, yes.

hobu:
``readers.stac`` needs to be added to readers.rst list

doc/stages/readers.stac.rst
```diff
@@ -0,0 +1,96 @@
...
+            "ids": ["MD_GoldenBeach_2012"],
+            "properties": { "pc:type": ["lidar", "sonar"], "pc:encoding": "ept" },
+            "asset_name": "ept.json",
+            "date_ranges": [
+                [
+                    "2022-11-11T0:00:0Z",
+                    "2022-11-30T0:00:0Z"
+                ]
+            ],
+            "schema_validate": true
```
hobu, line 32: this got renamed

io/StacReader.cpp
```diff
@@ -0,0 +1,587 @@
...
+        m_readerArgs[driver] = { };
+
+        for (auto& arg: readerPipeline.items())
+        {
+            if (arg.key() == "type")
+                continue;
+            m_readerArgs[driver][arg.key()] = arg.value();
+        }
+    }
+    std::cout << m_readerArgs.dump() << std::endl;
```
hobu, line 100: clean up this lint

Issue #3880, opened 2022-10-25
Error when retrieving statistics


maboualidev:
LAZ file: 
https://rockyweb.usgs.gov/vdelivery/Datasets/Staged/Elevation/LPC/Projects/ARRA-NJ_3Counties_2010/laz/ARRA-NJ_3Counties_2010_000091.laz

```
json_str = '''[
    "CA_SanFranBayFEMA_2004_000398.laz",
    {
        "type": "filters.stats",
        "dimensions": "X, Y, Classification",
        "enumerate": "Classification"
    }
]'''

pipeline = pdal.Pipeline(pipeline_json_str)
pipeline.execute()

#The next line generates an error:
pipeline.metadata['metadata']
```
The error is:
```
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
Cell In [24], line 1
----> 1 pipeline.metadata['metadata']

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcc in position 10237: invalid continuation byte
```

any idea how to fix this?

maboualidev:
Thank @hobu

Issue #3772, opened 2022-05-20
escapeJSON can produce invalid JSON


abellgithub:
Utils::escapeJSON assumes that the input string contains only ASCII characters. If bytes not representing ASCII characters (bytes with the high bit set) are contained, they are left in place, which *may* result in output that isn't valid JSON as the bytes won't likely be valid UTF code points. Not sure what we should do. Perhaps a string should trim all data when it encounters an invalid UTF-8 code point. In any case, we should produce valid JSON.

hobu:
Addressed in #3906

PR #3906, opened 2022-11-28
UTF sanitation of JSON output

hobu:
Implement utfcpp from https://github.com/nemtrif/utfcpp to provide UTF sanitation of our JSON output. This should prevent issues like https://github.com/PDAL/PDAL/issues/3880 from happening.

Supersedes #3904, which had a bad branch point


j9ac9k:
My first thought is we probably don't want the `.circleci` directory with the CI config for utfcpp?

hobu:
Updated escapeJSON to address #3772 as well.

PR #3917, opened 2022-12-28
update googletest to 1.12.1

hobu:
We are two releases behind on google test. Release notes show some updates to C++17 and better Abseil integration.


hobu:
What prompted me updating this is I couldn't get my debugger to read my debugging symbols out of gtest on a gcc11.

PR #3920, opened 2023-01-04
EPT and COPC fixes

hobu:
* Add `done()` method to `readers.ept` and reset the connector with it
* Reset the connector in `readers.copc`
* Add exponential backoff for retries in arbiter (needs to be upstreamed)
* Release thread back to pool in case of exception in ThreadPool
* Log non-server Curl messages (like when you run out of sockets)

All this to fix running out of sockets 😄

### DIFF
--- a/io/CopcReader.cpp
+++ b/io/CopcReader.cpp
@@ -111,7 +111,7 @@ public:
     std::unique_ptr<ThreadPool> pool;
     std::unique_ptr<copc::Tile> currentTile;
 
-    copc::Connector connector;
+    std::unique_ptr<copc::Connector> connector;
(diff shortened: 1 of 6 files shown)


abellgithub:
I don't understand the thread pool changes. Throwing an exception out of a thread is bad news.

connormanning:
Yeah this was going to be in conjunction with other changes that we didn't end up having to make as we were debugging things.  I think what I've added is essentially a no-op in the current state and can be backed out.  Currently there's an implicit rule that the thread-pool tasks must not throw: probably that should be documented internally, since previously the thread-pool would swallow exceptions from its tasks.

PR #3922, opened 2023-01-05
Fix WKT export from HexGrid using low precision

### LINKED ISSUE #3921: filters.hexbin sometimes returns empty boundary polygons
**Describe the bug**

For some of my input data, filters.hexbin returns `POLYGON EMPTY` in the `boundary` metadata instead of the true boundary.

It turns out that there is a regression in HexGrid::toWKT() caused by #3873:
- hexbin filter configures the output stream to use fixed precision:
```
    polygon.setf(std::ios_base::fixed, std::ios_base::floatfield);
    polygon.precision(m_precision);
```
- but HexGrid::toWKT() implementation now constructs its own output stream, which defaults to lower precision
- which in turn will cause Polygon::simplify() remove the outer polygon ring in Polygon::removeSmallRings() because its area() is calculated as zero because the polygon is not valid.

It is possible to replicate with:
```
[
    "somedata.laz",
    {
        "type": "filters.hexbin",
        "threshold": 0,
        "edge_size": 5
    }
]
```

```
pdal pipeline --metadata meta.txt test-boundary.json
```
In the output you get `boundary` and some other derived data incorrect:
```
{
  "stages":
  {
    "filters.hexbin":
    {
      "area": -1,
      "avg_pt_per_sq_unit": 0.07179245701,
      "avg_pt_spacing": "NaN",
      "boundary": "POLYGON EMPTY",
      "boundary_json": { "type": "Polygon", "coordinates": [ ] },
      "density": -1,
      "edge_length": 5,
      "estimated_edge": 8.660254038,
      "hex_offsets": "MULTIPOINT (0 0, -2.5 4.33013, 0 8.66025, 5 8.66025, 7.5 4.33013, 5 0)",
      "sample_size": 5000,
      "threshold": 0
    },
   // ... snip ...
```

pdal master

wonder-sk:
Fixes #3921

This rolls back changes in HextGrid::toWKT from #3873. It seems safe not to construct another stream in HexGrid::toWKT and one can expect that the caller will provide stream that is actually `OStringStreamClassicLocale`

### DIFF
--- a/filters/private/hexer/HexGrid.cpp
+++ b/filters/private/hexer/HexGrid.cpp
@@ -372,24 +372,21 @@ void HexGrid::cleanPossibleRoot(Segment s, Path *p)
 
 void HexGrid::toWKT(std::ostream& out) const
 {
-    pdal::Utils::StringStreamClassicLocale ss;
-
-    auto writePath = [this, &ss](size_t pathNum)
+    auto writePath = [this, &out](size_t pathNum)
     {
-       rootPaths()[pathNum]->toWKT(ss);
+       rootPaths()[pathNum]->toWKT(out);
     };
 
-    ss << "MULTIPOLYGON (";
+    out << "MULTIPOLYGON (";
 
     if (rootPaths().size())
         writePath(0);
     for (size_t pi = 1; pi < rootPaths().size(); ++pi)
     {
-        ss << ",";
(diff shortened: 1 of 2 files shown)


wonder-sk:
Unrelated pdal_io_stac_reader_test failure...

Issue #3935, opened 2023-01-11
filters.delaunay: 32-bit indices instead of size_t ?


wonder-sk:
The delaunay triangulation with delaunator-cpp uses these major arrays for the computation (added comments on memory use of each vector):

```cpp
    std::vector<double> const& coords;   // 2 * double * N
    std::vector<std::size_t> triangles;  // 2 * 3 * size_t * N
    std::vector<std::size_t> halfedges;  // 2 * 3 * size_t * N
    std::vector<std::size_t> hull_prev;  // size_t * N
    std::vector<std::size_t> hull_next;  // size_t * N
    std::vector<std::size_t> hull_tri;   // size_t * N
```

On 64-bit machines, that will be 136 bytes per point.

I am wondering if it would make sense to switch to smaller sized values (e.g. 32-bit unsigned ints instead of size_t) for all those indices:

 - we could nearly halve the memory consumption (only 76 bytes per point instead of 136 bytes)
 - the execution could be faster - smaller amount of memory touched
 - it is probably unlikely that people want/need to run this on more than 4 billion points

hobu:
Does this ticket belong upstream? https://github.com/delfrrr/delaunator-cpp 

Defaulting to uint32_t for the array sizes seems reasonable to me. Please make a PR.

abellgithub:
This probably should be a typedef in delaunator, like we have point_count_t in PDAL.

wonder-sk:
> Does this ticket belong upstream? https://github.com/delfrrr/delaunator-cpp

Upstream seems to be unmaintained, Andrew's fork has some more fixes: https://github.com/abellgithub/delaunator-cpp

abellgithub:
I think I'm the owner of what's up to date.

hobu:
Leaving this open until it gets upstreamed, but what's merged now is good enough to go forward for the PDAL 2.5.0 release.

PR #3938, opened 2023-01-12
Use shorter values for indices in delaunator-cpp

wonder-sk:
I had a quick stab at #3935 just to see how it would work... It is a pure search & replace of `std::size_t` to a new typedef `delaunator::index_t` (which is now defined as `uint32_t`). Haven't checked the code thoroughly.

Maybe there should be some errors thrown in case someone tried using more than 4 billion points?

Some quick results from a test with ~10M point LAS dataset:

|  |  Mem use before (MB) | Mem use after (MB) | Mem `after-before` (MB) | Time (s) |
|-|-|-|-|-|
| size_t | 928.3 | 2409.3 | 1481.0 | 11.0 |
| uint32_t | 928.3 | 1865.9 | 937.6 | 10.3 |

Memory usage from massif (before delaunator starts and after it initializes all vectors), timing being average from 5 runs, runing on my laptop with 64-bit Intel CPU (linux).

Testing pipeline:
```json
[
    "test.las",
    {
        "type": "filters.delaunay"
    }
]
```

Overall, with uint32_t the memory consumption of delaunator was ~37% less compared to size_t and there's a modest speed improvement (~6%).

### DIFF
--- a/filters/HagDelaunayFilter.cpp
+++ b/filters/HagDelaunayFilter.cpp
@@ -66,7 +66,7 @@ double delaunay_interp_ground(double x0, double y0, PointViewPtr gView,
     }
 
     delaunator::Delaunator triangulation(neighbors);
-    const std::vector<size_t>& triangles(triangulation.triangles);
+    const std::vector<delaunator::index_t>& triangles(triangulation.triangles);
 
     for (size_t j = 0; j < triangles.size(); j += 3)
     {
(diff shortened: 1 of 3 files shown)


abellgithub:
People do have point clouds that exceed the size supported by the proposed change. Perhaps we should just check before use and report an error?

wonder-sk:
> Perhaps we should just check before use and report an error?

Yeah - or possibly we could have templated delaunator with variants for both uint32_t and size_t, and PDAL would use the right variant? It is not that much code I guess so it should not be a big deal to have it compiled twice in the binaries...

PR #3942, opened 2023-01-13
STAC Reader bug fixes

kylemann16:
- Fixed bug where dates were still being compared when there was no date filter option being specified. Planetary Computer datetime key is set to null, so checking for that as well now.
 - Fixed bug where reader args from pipeline weren't being read correctly.

### DIFF
--- a/filters/HagDelaunayFilter.cpp
+++ b/filters/HagDelaunayFilter.cpp
@@ -66,7 +66,7 @@ double delaunay_interp_ground(double x0, double y0, PointViewPtr gView,
     }
 
     delaunator::Delaunator triangulation(neighbors);
-    const std::vector<size_t>& triangles(triangulation.triangles);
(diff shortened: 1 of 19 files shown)


hobu:
Example how to use this with PC:

```
pdal info --debug --driver readers.stac \
    --input "https://planetarycomputer.microsoft.com/api/stac/v1/search?collections=3dep-lidar-copc&ids=USGS_LPC_UT_StatewideSouth_2020_A20_12SUH7021,USGS_LPC_UT_StatewideSouth_2020_A20_12SUH7020" \
    --readers.stac.reader_args '{ "type": "readers.copc", "query": { "sv":"2020-08-04", "si":"usgs-3dep-copc-landrush-rw", "sr":"c", "sig":"token" } }'
```

hobu:
picked back to `2.5-maintenance` at 9b810cb54322c59dc34b97b390dcfb9ec7a26197

Issue #3944, opened 2023-01-18
linker error: undefined reference to re2::RE2::RE2


t0b3:
**Describe the bug**
compiling PDAL v2.5.0 on gentoo I run into a linker error:
```
PointViewTest.cpp:(.text+0x2c2d): undefined reference to `re2::RE2::RE2(re2::StringPiece const&)'
```
<details>
```
$ ninja -v -l0
[...]
/usr/bin/x86_64-pc-linux-gnu-g++ -march=native -O2 -pipe -Wl,-O1 -Wl,--as-needed test/unit/CMakeFiles/pdal_test_support.dir/Support.cpp.o test/unit/CMakeFiles/pdal_point_view_test.dir/PointViewTest.cpp.o -o bin/pdal_point_view_test  -Wl,-rpath,/var/tmp/portage/sci-libs/pdal-2.5.0/work/PDAL-2.5.0-src_build/lib64:/var/tmp/portage/sci-libs/pdal-2.5.0/work/PDAL-2.5.0-src_build/lib  lib64/libpdal_base.so.15.0.0  lib64/libpdal_util.so.15.0.0  lib/libgtest.so.1.10.0
FAILED: bin/pdal_point_view_test 
: && /usr/bin/x86_64-pc-linux-gnu-g++ -march=native -O2 -pipe -Wl,-O1 -Wl,--as-needed test/unit/CMakeFiles/pdal_test_support.dir/Support.cpp.o test/unit/CMakeFiles/pdal_point_view_test.dir/PointViewTest.cpp.o -o bin/pdal_point_view_test  -Wl,-rpath,/var/tmp/portage/sci-libs/pdal-2.5.0/work/PDAL-2.5.0-src_build/lib64:/var/tmp/portage/sci-libs/pdal-2.5.0/work/PDAL-2.5.0-src_build/lib  lib64/libpdal_base.so.15.0.0  lib64/libpdal_util.so.15.0.0  lib/libgtest.so.1.10.0 && :
/usr/lib/gcc/x86_64-pc-linux-gnu/11/../../../../x86_64-pc-linux-gnu/bin/ld: test/unit/CMakeFiles/pdal_point_view_test.dir/PointViewTest.cpp.o: in function `PointViewDeathTest_out_of_bounds_Test::TestBody()':
PointViewTest.cpp:(.text+0x2c2d): undefined reference to `re2::RE2::RE2(re2::StringPiece const&)'
[...]

PR #3945, opened 2023-01-18
fix linker error: add reference to re2

t0b3:
Closes: https://github.com/PDAL/PDAL/issues/3944

### DIFF
--- a/test/unit/CMakeLists.txt
+++ b/test/unit/CMakeLists.txt
@@ -79,6 +79,7 @@ PDAL_ADD_TEST(pdal_point_view_test
     INCLUDES
         ${PDAL_VENDOR_DIR}/eigen
 )
+target_link_libraries(pdal_point_view_test PRIVATE re2::re2)
 PDAL_ADD_TEST(pdal_point_table_test FILES PointTableTest.cpp)
 
 PDAL_ADD_TEST(pdal_program_arg_test


abellgithub:
There isn't a good reason why this would have to be explicitly linked, and it certainly isn't something to link on 99% of systems.

t0b3:
@abellgithub I don't mind how you fix this. As long as you come up with another useful solution which applies to gentoo ebuild we are all happy, see https://github.com/gentoo/gentoo/tree/master/sci-libs/pdal :+1:

abellgithub:
You're welcome to do whatever you need locally to fix your issue, of course, but you'll have to figure out the detailed reason for the problem before you can expect anyone to address it in the repository.

hobu:
It came from abseil, which now seems to require it. We should turn off abseil unless explicitly asked for so as to not drag this stuff in. 

Additionally, the 2.5.0 release hid all symbols that were not explicitly  exported. This might have some side effects.

abellgithub:
If it were a true dependency, there should be an entry in the linked library that references it and it should be automatically linked. It doesn't make sense that optimization is happening with a debug build. I would call that an unsupported configuration, though this may not be the issue at hand. I would ask what's different about this system/configuration.

hobu:
The re2 requirement came from https://github.com/PDAL/PDAL/commit/366aa8442db3b5036fbe95ab019231a32c7c3777

The newer version of abseil I was running locally required it but I didn't account for the fact that older versions didn't. Something must not be quite right about older versions' cmake configuration as to automatically bring re2 or not as needed.

I think we should switch abseil to an explicitly opt-on option. It's really just for developer scenarios.

hobu:
>I think we should switch abseil to an explicitly opt-on option. It's really just for developer scenarios.

Done in #3948

Issue #3947, opened 2023-01-19
pdal info raises proj-related errors in docker image of PDAL 2.5.0


joostmeulenbeld:
Trying to call `pdal info filename.laz` within the official PDAL docker image returns an error. 

For example, running `pdal info` on the first example file of https://limon.eu/support/sample-data/:
```bash
$ docker run -v /path/to/folder/:/data pdal/pdal:2.5.0 pdal info /data/Palac_Moszna.laz
```
I get this error:
```
(pdal info filters.stats Error) GDAL failure (1) PROJ: proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
PDAL: Could not import coordinate system 'EPSG:4326': PROJ: proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed.
```

Calling `pdal info` within the docker image on a `.laz` file I'm unfortunately not allowed to share, yields more of the same errors:
```
proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
(pdal info readers.las Error) GDAL failure (1) PROJ: proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
(pdal info readers.las Error) GDAL failure (1) PROJ: proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
(pdal info readers.las Error) GDAL failure (1) PROJ: proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
(pdal info filters.stats Error) GDAL failure (1) PROJ: proj_create_from_database: Open of /opt/conda/envs/pdal/share/proj failed
[...]

PR #3950, opened 2023-01-19
set GDAL,PROJ,GEOTIFF env variables for our container

hobu:
This should address the issue in #3947


hobu:
cherry picked to `2.5-maintenance` at fc4f5076c0882b3b14d0f5aa782019121487256b

Issue #3951, opened 2023-01-19
filters.csf very slow


n4z4m3:
The filters.csf filter runs much more slowly than it needs to.

Reproducing this only requires running the CSF filter on any LAS file. However, it is more obvious with LAS files with larger numbers of points (something like 15 million would do). For example, I have a LAS file that takes 34 minutes to run through the CSF filter.

I have determined the reason for this slowness and it is due to the writing of debugging information to text files. It is related to the already resolved issue #3488 . That issue resolved the fact that these text files were always stored in the current working directory by adding a **_dir_** argument to the CSF filter which allows specifying a different directory. However, these files were still being written. In the fix for that issue, the code has documentation that states that if the directory is empty, then those files will not be written. But that does not work. If you don't use the  **_dir_** argument it still writes to the current working directory. And if you specify the empty string as a value it fails stating that the argument cannot be empty.

I already tested commenting out the lines in the Cloth.cpp file where it writes the debugging files and the tool speeds up quite a lot. The file I have that was taking 34 minutes to run, took less than 2 minutes with those lines commented out.

So there are a few possible ways to fix this:
1. Remove writing the debugging information by deleting those lines that I commented out. However, this would mean you would probably have to remove the **_dir_** argument as well otherwise it would end up doing nothing. Also, people might still want that debugging information to be written.
2. Fix the ProgramArgs class so that you can pass an empty string as a valid argument so that an empty **_dir_** argument will work to turn off debugging. And of course use this as the default because I assume we don't want debugging info written by default.
3. Add a **_debug_** argument to the CSF filter as a boolean flag which is false by default. But when set to true, enables writing the debugging info and utilizes the **_dir_** argument.

My preference would probably be for 1, but I assume the **_dir_** argument was added because people still want this information written for debugging purposes. I also like 2, but I am not sure how to implement that one myself.

I have already implemented 3 myself for testing purposes to try it out and it works fine, but I would like to get feedback to see whether it is a desired approach before adding it as a pull request. The files affected are filters/CSFFilter.cpp,  filters/private/csf/CSF.cpp, filters/private/csf/CSF.h, filters/private/csf/Cloth.cpp, filters/private/csf/Cloth.h

Hopefully I hear back with some feedback within a few days, otherwise, I will add the pull request with option 3.

hobu:
Option 3 sounds great. Please make a PR and target it for the 2.5.1 release.

Issue #3690, opened 2022-02-22
filters.csf extremely slow in linux


nmgeomatics:
Running filters.csf in python in linux appears to have an issue. I run a pipeline in a python script in both windows and linux with the filters.csf and the script runs in seconds in windows (windows 11), but it runs for hours in linux (ubuntu).

Snippet from the script

```
csfrigidness=2
csfthreshold=0.25

pipeline_json= f"""
[
  {{
      "type":"readers.las",
      "filename":"{infile}"
  }},
  {{
        "type":"filters.csf",
        "ignore":"Classification[7:7]",
        "rigidness":"{csfrigidness}",
        "threshold":"{csfthreshold}"
    }}
]
"""

pipeline = pdal.Pipeline(pipeline_json)
pipeline.execute()
```

chambbj:
There should be nothing platform-dependent about `filters.csf`. I'm wondering if you are running the same version on each system. There is an unreleased change (if you are building from source) that would allow you to skip writing some intermediate outputs, which can speed things up significantly.

You'll need to let us know how you are installing PDAL, which versions you are installing, and ideally provide some sample data.

abellgithub:
There was also this: https://github.com/PDAL/PDAL/issues/3248, but it's also not system-dependent, but if you had an older Linux version...

abellgithub:
I'll also add that in the past we have found libstdc++ to behave differently from other C++ standard libraries and there could be a performance issue in that library that you have hit. If you could do some further diagnosis and verify versions, that would be helpful.

nmgeomatics:
My installation is as follows:
Ubuntu 20.04.3 LTS installed in Windows Subsystem For Linux 2 on Windows 11
Using Anaconda 4.11.0 and python 3.9.10 in a anaconda virtual environment
The conda list contents are attached which will show you the pdal version installed.
[myconda.txt](https://github.com/PDAL/PDAL/files/8145311/myconda.txt)

I am not sure if I can provide the sample data since it is from a company I am doing work for. I would have to get permission for that.

However, I wonder if the intermediate file thing mentioned by @chambbj  is the issue here since I am running in WSL and I bet it is slower to write to the Windows file system which might explain why it is much quicker when running in Windows directly.
[...]

PR #3952, opened 2023-01-20
Add optional debug argument to CSF filter

n4z4m3:
Add a new optional boolean debug argument to the CSF filter to enable debug output to the location provided by the dir argument. The default value of this argument will be false so that this debug output is not written by default.

This will fix issue #3951 and issue #3690


hobu:
Cherry picked to `2.5-maintenance` at e6de5e651ec0e3e750c074387fdd5f5b8b8b0c59

PR #3957, opened 2023-01-25
Fix building with MinGW

mmuetzel:
PDAL 2.5.0 doesn't build with MinGW compilers. This is mainly because of the missing/incomplete filesystem API in the C++ standard libraries on that platform.

The proposed changes add the work-around of using ANSI Windows API functions in more places.

Additionally, it fixes an issue with the deprecated attribute for GNU compatible compilers on Windows.

See also this downstream PR for the binary package of PDAL in MSYS2: https://github.com/msys2/MINGW-packages/pull/15234

### DIFF
--- a/apps/pdal.cpp
+++ b/apps/pdal.cpp
@@ -301,7 +301,7 @@ namespace
     LogPtr logPtr(Log::makeLog("PDAL", "stderr"));
 }
 
-#ifdef _WIN32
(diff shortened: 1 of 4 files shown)


hobu:
cherry picked to `2.5-maintenance` at 07918308ff830c8c5d7d35d4e41abafe637e1b41

PR #3958, opened 2023-01-27
Fix broken FAQ link

jshrake:
The link https://pdal.io/faq no longer works. The correct URL is https://pdal.io/faq.html.


hobu:
Thanks!

Issue #3954, opened 2023-01-20
Add default User-Agent to readers.copc and readers.ept 


hobu:
Users can override and set User-Agent using the `headers` option of each of those, but we should provide a default one.

We should push in the version numbers, which may have some utility in debugging.

```
pdal (2.5.0) / curl (7.87.0)
```

PR #3962, opened 2023-01-31
consolidate ept:: and copc:: Connector classes into a single one

hobu:
I don't know why there are two very similar classes here. This PR addresses #3954, which would have required copypasta in a couple of places.


hobu:
Picked to `2.5-maintenance` at 659bbadf2c988904236beefef780d09ed39419d4

PR #3966, opened 2023-02-03
Guard against bad return numbers.

abellgithub:


io/private/las/Summary.cpp
```diff
@@ -57,7 +57,8 @@ void Summary::addPoint(double x, double y, double z, int returnNumber)
     m_bounds.grow(x, y, z);
 
     // Returns numbers are indexed from one, but the array indexes from 0.
-    m_returnCounts[returnNumber - 1]++;
+    if (returnNumber >= 1 && returnNumber <= m_returnCounts.size())
+        m_returnCounts[returnNumber - 1]++;
```
hobu, line 61: returnNumber of 0 happens in files even though it is invalid and is not supposed to happen. I understand we don't want to access the array OB, but we do want histogram counts of `returnNumber == 0`.

abellgithub:
This is writing. Has nothing to do with histograms.

hobu:
picked back to `2.5-maintenance` at 156bd9d227a31f7437f436ebbb746f6840a7c7aa

PR #3967, opened 2023-02-03
Fix handling of string HeaderVal containing spaces

### LINKED ISSUE #3964: writers.las improperly reads pipeline string option containing space
**Description of bug**
When a LAS file System Identifier or Generating Software contains spaces the writer strips everything after the space for some reason.
For example, if I have a Ssytem Identifier of "My System ID", it will be written out as "My". And if it starts with a space, it will end up with an empty string.

**Expected behavior**
The System Identifier and Generating Software values should not be modified by the writer except to maybe truncate them to 32 characters since that is the maximum allowed by the LAS file format.

abellgithub:
Close #3964

### DIFF
--- a/io/CopcWriter.cpp
+++ b/io/CopcWriter.cpp
@@ -35,6 +35,11 @@
 #include <pdal/util/Algorithm.hpp>
 #include <pdal/util/FileUtils.hpp>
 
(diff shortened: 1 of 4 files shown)


hobu:
Picked back to `2.5-maintenance` at cc93cf821511d911616e314275441a0eaf1fb1eb

PR #3968, opened 2023-02-04
output better information when LAS Classification values can't be wri…

hobu:
…tten into a particular LAS version'd file


hobu:
Picked to `2.5-maintenance` at 69c245a6f603d54f699b6d295ec54e25c0f3d9bf

PR #3982, opened 2023-02-21
CI Simplification

hobu:
* Use conda environment in `./scripts/ci/environment.yml` to define environment settings instead of manually installing packages in `./scripts/ci/platform/setup.sh`
* Require Draco 1.5.6+. The latest Draco has proper draco-targets.cmake, and by requiring this version or later, we can ditch the usage of pkg-config in conda and CI

I plan to merge this to `2.5-maintenance`. I doubt anyone is really using Draco with PDAL in a way that is going to break with this change.


hobu:
cherry picked back to `2.5-maintenance` at 036c40aaef3dedc0c903b1f5b8d240714dc34fea

Issue #3979, opened 2023-02-21
pdal-config has still -std=c++11


nilason:
**Describe the bug**
`pdal-config` has hardcoded cxxflag `-std=c++11`, which doesn't reflect necessary standard support:

https://github.com/PDAL/PDAL/blob/3d3db44b79a53a7f134bcdd90f8654fa537c8516/apps/pdal-config.in#L55-L56

**Expected behavior**

Either remove the flag or update it to `-std=c++17`.

**System/installation information:**

PDAL master branch.

hobu:
I think this bug shows that it is time to make `pdal-config` go away entirely. This was added before pkg-config was stable and before CMake's CONFIG approach had taken hold. The proper way to get these settings and link information is to use the pdal-targets.cmake file, not `pdal-config`.

nilason:
GRASS still uses it. Please hold on to it till we have it changed :-)!

nilason:
Thanks!
I started the move to use pkg-config with https://github.com/OSGeo/grass/pull/2851.

PR #3983, opened 2023-02-21
Do not hardcode -std=c++11 into pdal-config

hobu:
Address #3979


hobu:
cherry picked back to `2.5-maintenance` at 7e82c70ca03c0c3e3e70dfc6abcf82b4a3ce4fcc

PR #3984, opened 2023-02-21
add allow_empty option to writers.gdal to write empty GDAL output

hobu:


hobu:
cherry picked back to `2.5-maintenance` at 3b87ba950c8d0faecf333c5de1298eb8472c4ee1

zdila:
@hobu I've just tried  `allow_empty: true` and in case of no points it SIGSEGVs for me.

I've locally solved it with this code:
```c
void GDALWriter::doneFile()
{
    if (!m_grid) {
        if (m_allowEmpty)
            return;

        throw pdal_error("Unable to write GDAL data with no points "
            "for output.");

    }
    
    ...
```

PR #3989, opened 2023-02-25
Streamable readers.stac

hobu:
Refactor `readers.stac` to be streamable when selected data sources are also streamable.


hobu:
Picked back to `2.5-maintenance` at a97adc7bc4068c85ef93872ba84216972cbc5782

Issue #3995, opened 2023-03-07
Proposal: add in_coord_epoch and out_coord_epoch optional argments in filters.reprojection for time-based coordinate transformations


rouault:
There's currently no easy way to do coordinate transformations with PDAL's filters.reprojection.

For example in https://github.com/PDAL/PDAL/issues/3975, for static CRS -> dynamic CRS or dynamic CRS -> static CRS, the workaround is to write a PROJ pipeline using +set v_4= to set the source or target coordinate epoch.
In the future PROJ may also support dynamic CRS at epoch A -> dynamic CRS at epoch B transformations, that would make writing such pipelines even more complicated.

One possibility currenly would be to use the WKT2:2019 [COORDINATEMETADATA](https://docs.ogc.org/is/18-010r7/18-010r7.html#130) construct that is supported since PROJ 9.2 to fill the in_srs and out_srs arguments of filters.reprojection. But this is quite verbose. I would propose to add ``in_coord_epoch`` and ``out_coord_epoch`` optional arguments, pass them to PDAL's  SrsTransform, which will forward them to GDAL's OGRSpatialReference::SetCoordinateEpoch().

[...]

PR #3998, opened 2023-03-08
Add WKTv2 VLRs to LAS

### LINKED ISSUE #3502: Add WKTv2 VLRs to LAS and LAZ files
libLAS and PDAL had an option for added WKTv1 data to files via VLR with user_id=`liblas` and record_id=`2112` . We should consider adding a WKTv2 VLR that overrides the WKTv1 and GeoTIFF key information if seen in the file. 

* Add VLR with user_id=`pdal` and record_id=`4224` 
* Store WKTv2 data in it
* Write WKTv1 per normal rules
* Write GeoTIFF keys per normal rules
* Prefer the `pdal:4224` WKTv2 if seen.

jjimenezshaw:
This PR Closes #3502

When the user enables the option `write_wkt2` in `LasWriter`, it stores a VLR `PDAL:4224` with the WKT2 representation of the SRS.

`LasReader` prefers VLR `PDAL:4224` when present over the WKT1 representation.

This allows for instance the usage of `DERIVEDPROJCRS` systems (example got from PROJ tests), that are not allowed as WKT1.

When the code is verified, I can continue with COPC format (in this PR or a later PR?)


pdal/SpatialReference.cpp
```diff
@@ -475,40 +486,32 @@ std::string SpatialReference::prettyWkt(const std::string& wkt)
...
-
-    outWkt = buf;
-    CPLFree(buf);
+    const char* apszOptions[] = { "MULTILINE=YES", nullptr }; // equivalent to exportToPrettyWkt
+    outWkt = exportToWkt(srs.get(), apszOptions);
     return outWkt;
 }
 
-std::string SpatialReference::getWKT1() const
+std::string SpatialReference::getWKT1(bool throwIfEmpty) const
```
hobu, line 494: what precipitated this change?

hobu:
Thanks for this. @rouault and I were just discussing about this in relation to #3995 

I think we should take this all the way though. I would define that as:

* Call the option `enhanced_srs_vlrs` and write both a WKTv2 *and* a [PROJJSON](https://proj.org/specifications/projjson.html) VLR. Those of us doing web stuff with COPC files would really like to have PROJJSON.
* `readers.copc` and `readers.las` should have an option that defines the cascading order of preference for SRS to consume. I would propose `wkt2, projjson, wkt1, geotiff` as the default, but users should be able to change this ordering. It should apply to and override all LAS version sensitivities with respect to WKT/GeoTIFF tags
* Tests 😄

pdal/SpatialReference.cpp
jjimenezshaw: TLDR; "I tried to keep the same behavior when wkt2 is not enabled"

In the function `LasWriter::addWktVlr()`, `getWKT1()` was throwing an exception if the result of `exportToWkt` is empty. That stopped the whole process. In case the wkt2 is stored, then we do not want to throw. In that case it just not write the wkt1 and geotiff.

I do not like this solution so much. I can either:
- do a try-catch around, removing the parameter `throwIfEmpty`
- change the behaviour of `getWKT1()` and do not throw.

See that if no srs is provided, there is no exception. The exception only is thrown if the existing srs does not produce an output in `exportToWkt` with `FORMAT=WKT1_GDAL`

jjimenezshaw:
> Thanks for this. @rouault and I were just discussing about this in relation to #3995
> 
> I think we should take this all the way though. I would define that as:
> 
>     * Call the option `enhanced_srs_vlrs` and write both a WKTv2 _and_ a [PROJJSON](https://proj.org/specifications/projjson.html) VLR. Those of us doing web stuff with COPC files would really like to have PROJJSON.

Interesting. I  like the idea of complementing with PROJJSON. If I understand correctly `enhanced_srs_vlrs` is still a boolean. In case it is set to `true`, we store 2 VLRs (not just one as I did), one for WKT2 (`PDAL:4224`) and another for PROJJSON (what about `PDAL:4225`? Or do you have another suggestion?)
> 
>     * `readers.copc` and `readers.las` should have an option that defines the cascading order of preference for SRS to consume. I would propose `wkt2, projjson, wkt1, geotiff` as the default, but users should be able to change this ordering. It should apply to and override all LAS version sensitivities with respect to WKT/GeoTIFF tags

Any suggestion for the name of that option? I understand that it would be a string field, with comma separated values. Or space separated values? I may have more questions when I study the code for `It should apply to and override all LAS version sensitivities with respect to WKT/GeoTIFF tags`

> 
>     * Tests smile
Of course.


Again, I will go first with normal LAS. When we are happy with that, I will work on COPC.

hobu:
> (what about PDAL:4225)

Perfect.

>  `LASF_Projection:2112` and `liblas:2112`. They are used depending on the WktMask, LAS version and other details.

These are equivalent and they are both the same WKT1. We started writing WKT into VLRs before LAS 1.4 added WKT.

hobu:
>  I get that you might want a WKT2 representation, but that's not LAS.

* A stated goal of LAS 1.5 is to bring in WKT2 representation of the CRS. At GeoWeek last month, @esilvia stated the intention to have this moving forward by the end of 2023.
* For libLAS, we started writing and preferring WKT1 in VLRs a couple of years before LAS 1.4 was even discussed. This was because WKT1 could represent stuff that GeoTIFF keys couldn't. We have the same situation now with WKT1 vs WKT2
* Regardless of the **n** *x* **m** support matrix of LAS minor versions vs allowed CRS types, users' intention and desire is to have and communicate the most complete definitions for complex situations. 

Maybe the idea of a priority stack isn't exactly what we want, but it is my opinion that the "right" thing in almost all instances is to consume the most complete definition that's available.

hobu:
> you lobby the powers-that-be to set aside a LAS VLR for WKT2 (and any other CRS types) now .

Good point.

jjimenezshaw:
- `enhanced_srs_vlrs` is used now, writing two VLRs, one for WKT2 and another for PROJJSON
- `srs_consume_preference` is added to `LasReader`, with the default value `"wkt2, projjson, wkt1, geotiff"`. The `SpatialReference` is constructed from the VLRs taken in that order.

About using a different identifier than `PDAL:4224` and `4225`, I am totally fine with it. That is easy to change before the merge. Let me know if you decide any other value/userId. 

If you have a better name `srs_consume_preference`, please, let me know.

Ups, I see now that `SpatialReference` cannot be constructed with a PROJJSON. I will fix it later.

If you are happy with the current implementation, I will continue with the tests, and later with COPC.

BTW,  the method `SpatialReference::valid()` will return false with some exotic CRSs, when `OSRValidate()` return `OGRERR_UNSUPPORTED_SRS`. I should change that too, right?

io/LasWriter.cpp
```diff
@@ -485,6 +488,8 @@ void LasWriter::addSpatialRefVlrs()
     deleteVlr(las::TransformUserId, las::GeotiffAsciiRecordId);
     deleteVlr(las::TransformUserId, las::WktRecordId);
     deleteVlr(las::LiblasUserId, las::WktRecordId);
+    // deleteVlr(las::PdalUserId, las::PdalWkt2RecordId);
+    // deleteVlr(las::PdalUserId, las::PdalProjJsonRecordId);
```
jjimenezshaw, line 492: I commented them out for test `read_consume_order`. Is there a better way to do it?

jjimenezshaw:
> That must mean we have permission smile

Does it mean that I should use `LASF_Projection:4224` and `4225` instead of `PDAL`?

jjimenezshaw:
> "fixed" is a specific combo of GDAL/PROJ

I have been some time trying to find out why it was failing on "fixed" without any change on my side. I think I found the reason

There is a compatibility issue between `PROJ >= 9.2.0` and `GDAL <= 3.6.0`
fixed in https://github.com/OSGeo/gdal/pull/6800 (thanks @rouault ) that produces that `DERIVEDPROJCRS` wkts are not parsed properly.

To workaround it I made a function more robust (it was not checking that a pointer could be null), and tests using that type are disabled for GDAL <= 3.6.0 in my last commit.

io/LasWriter.cpp
```diff
@@ -485,6 +488,8 @@ void LasWriter::addSpatialRefVlrs()
     deleteVlr(las::TransformUserId, las::GeotiffAsciiRecordId);
     deleteVlr(las::TransformUserId, las::WktRecordId);
     deleteVlr(las::LiblasUserId, las::WktRecordId);
+    // deleteVlr(las::TransformUserId, las::PdalWkt2RecordId);
```
hobu, line 491: lint to remove

io/LasWriter.cpp
```diff
@@ -524,8 +529,35 @@ void LasWriter::addGeotiffVlrs()
...
+    // WKT2 and PROJJSON can be writen in PDAL VLRs
+    if (d->opts.enhancedSrsVlrs) {
+        const std::string wkt2 = m_srs.getWKT2();
+        if (!wkt2.empty()) {
+            std::vector<char> wktBytes(wkt2.begin(), wkt2.end());
+            wktBytes.resize(wktBytes.size() + 1, 0);
+            addVlr(las::TransformUserId, las::PdalWkt2RecordId, "PDAL WKT2 Record", wktBytes);
+        }
+
+        const std::string projjson = m_srs.getPROJJSON();
```
hobu, line 541: should we always write PROJJSON? IMO it isn't that much of a cost, but I could see where people wouldn't want them written to the files and would want to control that they get WKT2 but not PROJJSON.

io/LasWriter.cpp
```diff
@@ -524,8 +529,35 @@ void LasWriter::addGeotiffVlrs()
...
+            std::vector<char> wktBytes(projjson.begin(), projjson.end());
+            wktBytes.resize(wktBytes.size() + 1, 0);
+            addVlr(las::PdalUserId, las::PdalProjJsonRecordId, "PDAL PROJJSON Record", wktBytes);
+        }
+    }
+
     // LAS 1.4 requires WKTv1
-    std::string wkt = m_srs.getWKT1();
+    std::string wkt;
+    try
```
hobu, line 551: I don't understand this bit of logic here. Why would getWKT1() throw?

io/private/las/Vlr.hpp
```diff
@@ -57,6 +57,8 @@ static const uint16_t TextDescriptionRecordId = 3;
 static const uint16_t ExtraBytesRecordId = 4;
 static const uint16_t PdalMetadataRecordId = 12;
 static const uint16_t PdalPipelineRecordId = 13;
+static const uint16_t PdalWkt2RecordId = 4224;
```
hobu, line 60: As referenced in the ticket, we changed this to LASFWkt2recordId. Maybe we should change the code to reflect that too.

pdal/SpatialReference.cpp
```diff
@@ -490,25 +505,18 @@ std::string SpatialReference::getWKT1() const
...
-            { "FORMAT=WKT1_GDAL", "ALLOW_ELLIPSOIDAL_HEIGHT_AS_VERTICAL_CRS=YES", nullptr };
-
-        srs->exportToWkt(&buf, apszOptions);
-        if (buf)
-        {
-            wkt1 = buf;
-            CPLFree(buf);
-        }
-    }
+    std::string wkt1 = exportToWkt(srs.get(), {"FORMAT=WKT1_GDAL", "ALLOW_ELLIPSOIDAL_HEIGHT_AS_VERTICAL_CRS=YES"});
```
hobu, line 508: `"ALLOW_ELLIPSOIDAL_HEIGHT_AS_VERTICAL_CRS=YES"` should probably be controllable now that I look at it.

pdal/SpatialReference.cpp
```diff
@@ -490,25 +505,18 @@ std::string SpatialReference::getWKT1() const
...
-        throw pdal_error("Couldn't convert spatial reference to WKT version 1.");
+       throw pdal_error("Couldn't convert spatial reference to WKT version 1.");
     return wkt1;
 }
 
+std::string SpatialReference::getWKT2() const
+{
+    std::string wkt = getWKT();
+    OGRScopedSpatialReference srs = ogrCreateSrs(wkt);
+    return exportToWkt(srs.get(), {"FORMAT=WKT2"});
```
hobu, line 518: do we want to set this to `WKT2:2019` to be explicit and to be fixed to a single version until we want to update?

hobu (approved):
Thanks so much for doing the work to upgrade this stuff

io/LasWriter.cpp
jjimenezshaw: Done.
I was not sure about this. If I delete those new Vlrs the same way as las::WktRecordId, I cannot run the test that writes different content on each to checks the values properly. The user couldn't neither write something different like an Engineering in wkt1 and a derivedprojected in Wkt2.

io/LasWriter.cpp
jjimenezshaw: Right, it isn't that much cost. If the user really want that fine control, they can create it themselves, like in the test `read_consume_order`.

io/private/las/Vlr.hpp
jjimenezshaw: done

pdal/SpatialReference.cpp
hobu: This was mostly a note to future-hobu to add support somehow for user-controllability of overrides like this. No need to address it just now unless it matters to you.

jjimenezshaw:
Hi @hobu . Some questions

a) I know copc should have the same behavior. Do you want it here, or in a different PR?
b) Is there anything pending to be developed in this PR (apart from copc)?
c) Do you miss tests? if yes, please tell me what should be tested.

d) I think in GDAL, a GeoTIFF created with a WKT that does not fit in GeoTIFF tags is creating the CRS as an EngineeringCRS. At least that is shown in gdalinfo (I couldn't find exactly where it is done in the code, probably in frmts/gtiff/gt_wkt_srs.cpp). Do we want that behaviour? If yes, in this PR?

Thanks.

hobu:
> a) I know copc should have the same behavior. Do you want it here, or in a different PR?

Yes. Same is fine.

> b) Is there anything pending to be developed in this PR (apart from copc)?

Thank you for implementing it, but I'm conflicted about `srs_consume_preference`. It is likely to cause some subtle and frustrating version-to-version behavior differences. Given that it is coordinate system minutiae, maybe this was always going to be the case. That said, it is going to give a determined person the ability to control what happens in a way that isn't possible now. 

If it is shown to be too disruptive, we might have to change the default order to `wkt1, geotiff, wkt2, projjson` which mimics current behavior. 

> c) Do you miss tests? if yes, please tell me what should be tested.

More tests are great. 

> d) I think in GDAL, a GeoTIFF created with a WKT that does not fit in GeoTIFF tags is creating the CRS as an EngineeringCRS. At least that is shown in gdalinfo (I couldn't find exactly where it is done in the code, probably in frmts/gtiff/gt_wkt_srs.cpp). Do we want that behaviour? If yes, in this PR?

Ugh. Should PDAL throw or try to gracefully degrade if the CRS is not representable in GeoTIFF? I lean toward throw so the user can adjust it as needed to fit in a way that expresses intent rather than GDAL/GeoTIFF/PROJ making some guesses.

jjimenezshaw:
> If it is shown to be too disruptive, we might have to change the default order to `wkt1, geotiff, wkt2, projjson` which mimics current behavior.

I did a commit changing the default to this order.

hobu:
Can you rebase this with master? I fixed the failure

jjimenezshaw:
> Can you rebase this with master?

Done

jjimenezshaw:
@hobu , I am trying to do the same with copc (I have a branch, `wkt2-vlr-copc`, to not bother here with the commits)
a) I have not found any test for CopcWriter.
b) I am not sure how to do to write the "enhanced_srs_vlr".

Any help would be welcome.

hobu:
No there are not any CopcWriter tests at this time. We pretty much just use https://validate.copc.io and https://github.com/hobuinc/copcverify for verification, but that hasn't made it into tests code yet. If you're feeling motivated ... 😄 

It would be completely reasonable to me to do the obviously thing to CopcWriter.cpp however. Please make the PR and I'll take a look.

jjimenezshaw:
> to do the obviously thing to CopcWriter.cpp

Would it make sense to add something like this in `CopcWriter::write`? (pseudocode)
``` C++
if (b->opts.enhancedSrsVlrs) {
    auto wkt2Vlr = makeWkt2Vlr(b->srs);
    b->vlrs.push_back(wkt2Vlr);

    auto projjsonVlr = makeProjjsonVlr(b->srs);
    b->vlrs.push_back(projjsonVlr);
}
```

hobu:
> Would it make sense to add something like this in` CopcWriter::write`? (pseudocode)

In this PR would be fine too.

hobu:
@jjimenezshaw I merged this, so make a new PR for the `writers.copc` addition...

jjimenezshaw:
> I merged this, so make a new PR for the `writers.copc` addition...

Thanks! I will.

hobu:
Here's the current plan for LAS 1.5. https://github.com/ASPRSorg/LAS/issues/95#issuecomment-1847337625

In short, 1.5 would allow any flavor of *published* OGC WKT in the WKT VLR. This doesn't change the need for supplemental WKT2 VLRs in older versions of LAS, but, it means people won't have to branch specially for just 1.5.

If anyone has comments on the approach, please do so on that ticket.

Issue #3990, opened 2023-02-25
ColumnPointTable Destructor Slow


b-gutzmann:
https://github.com/PDAL/PDAL/blob/366475794646082b973fc5ce3e142aa3aaec8a23/pdal/PointTable.hpp#L186

The ColumnPointTable destructor is slow with large point clouds because of the amount of buffers it allocates in addPoint(). Changing m_blockPtCnt from 16384 to 65536 (what RowPointTable uses) helps. Maybe a more complex approach where the buffer size grows over time is warranted.

I'm testing with tiffs found [here](https://visibleearth.nasa.gov/images/57752/blue-marble-land-surface-shallow-water-and-shaded-topography) and a simple pipeline that just reads the data in standard mode. With the largest tiff (21600 x 21600), the destructor took 78.713 seconds. After changing m_blockPtCnt to 65536, the destructor took 18.106 seconds. With (8192 x 4096), 146.3 versus 69.4 ms.
Timings were taken on windows in release mode.

`[
    "C:/test_data/land_shallow_topo_east.tif"
]`
`pdal.exe pipeline --nostream "pipeline_basic.json"`

hobu:
What was the pipeline you were using?

b-gutzmann:
Just 
```
[ 
    "C:/test_data/land_shallow_topo_8192.tif" 
]
```
Same results with 
```
[
    "C:/test_data/land_shallow_topo_8192.tif",
    "C:/test_data/temp.laz"
]
````

abellgithub:
This seems reasonable. Since there's are blocks for each dimension, they should probably be pretty large.

abellgithub:
These timings seem really high (slow). I just haven't seen timings like that. Are things built in release mode? Something else going on?

b-gutzmann:
My memory usage with the large dataset (466 560 000 points) was getting quite high. However I had no such issue with the smaller dataset (33 554 432 points). Using Measure-Command against my build produces the same timings as the conda release of pdal, so I don't think the issue is my compiler/configuration.

In general the GDALReader is slow which is what I was originally looking into. This is a small piece of that. But that's a separate issue as this issue isn't GDALReader specific.

hobu:
> GDALReader is slow

`readers.gdal` is mostly just a toy. What were you using it for? It is going to be quite sensitive to `GDAL_CACHE_MAX` too, so you might look at bumping that for better performance.

b-gutzmann:
[...]

PR #3999, opened 2023-03-09
readers.gdal performance improvements

hobu:
This improves `readers.gdal` performance in a few ways:

* look-before-you-leap conditionals in `writers.las` to prevent asking heavy `PointRef::getFieldAs` calls that do nothing
* memoized data members so we aren't allocating `std::array<double, 2>` for every pixel in `readers.gdal`
* Increased `m_blockPtCnt = 65536` for ColumnPointTable

On my Mac, this improves speed from 80sec to 60sec for the example provided in #3990.

The biggest performance improvement could be obtained by reading decent sized chunks of raster data from GDAL's `RasterIO`. We're reading single pixel-at-a-time and taking full function overhead of everything for each pixel. This doesn't seem worth the trouble unless raster => point cloud was a commonly done thing.

### DIFF
--- a/filters/ColorizationFilter.cpp
+++ b/filters/ColorizationFilter.cpp
@@ -202,11 +202,12 @@ void ColorizationFilter::ready(PointTableRef table)
 bool ColorizationFilter::processOne(PointRef& point)
 {
     static std::vector<double> data;
+    static std::array<double, 2> pix;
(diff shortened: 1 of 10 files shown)


hobu:
This work was moved to #4001 4012

PR #4001, opened 2023-03-12
Added a BlockReader class to GDALReader that reads from the raster a …

b-gutzmann:
…block at a time.

### DIFF
--- a/io/GDALReader.hpp
+++ b/io/GDALReader.hpp
@@ -58,6 +58,38 @@ public:
     ~GDALReader();
 
 private:
+    class BlockReader
+    {
+    public:
+        BlockReader(GDALReader& reader);
+        void initialize();
+        point_count_t processBlock(PointViewPtr view);
+        bool processOne(PointRef& point);
+
+    private:
+        struct Block
+        {
+            std::vector<std::vector<double>> m_data;
+            int m_blockCol;
+            int m_blockRow;
+        };
+
+        bool readBlock();
+
+        GDALReader& m_reader;
+        int m_blockRow;
+        int m_blockCol;
+        int m_blockWidth;
+        int m_blockHeight;
+        int m_numBlocksX;
+        int m_numBlocksY;
+        
+        Block m_currentBlock;
+        bool m_needsRead;
+        int m_colInBlock;
+        int m_rowInBlock;
+    };
+
     virtual void initialize();
     virtual void addDimensions(PointLayoutPtr layout);
     virtual void ready(PointTableRef table);
@@ -75,9 +107,8 @@ private:
     int m_width;
     int m_height;
     bool m_useMemoryCopy;
-    point_count_t m_index;
-    int m_row;
-    int m_col;
+
+    BlockReader m_blockReader;
 
     BOX3D m_bounds;
     StringList m_dimNames;
--- a/io/GDALReader.cpp
+++ b/io/GDALReader.cpp
@@ -62,7 +62,7 @@ std::string GDALReader::getName() const
 
 
 GDALReader::GDALReader()
-    : m_index(0)
+    : m_blockReader(*this)
 {}
 
 GDALReader::~GDALReader()
@@ -84,7 +84,7 @@ void GDALReader::initialize()
     m_height = m_raster->height();
     m_bandTypes = m_raster->getPDALDimensionTypes();
     m_metadata.add(m_raster->getMetadata());
-
+    m_blockReader.initialize();
 
     m_dimNames.clear();
     if (m_header.size())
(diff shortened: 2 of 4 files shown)


abellgithub:
I wonder why we need all this. I would think that GDAL would cache at least one block per band. If the issue is that we force block re-reads because we process in strict row-major order, then it would seem we could solve that without adding more caching by just reading the pixels in a block before moving to the next. We just need to get the block size to do that.

b-gutzmann:
I agree if GDAL caching works that way. I'm very unfamiliar with how GDAL works. I'll play around with this some more.

b-gutzmann:
Here are some timings. The working changes were with the RelWithDebInfo configuration, so they are a little slower than they should be. The command used is `pdal.exe pipeline --nostream "pipeline_basic.json"` with same pipeline as in #3990 

| width:height:bands | Time PDAL v2.5.2  (seconds) | Time working changes (seconds) | # Runs (median taken) |
| ----------- | ----------- | ----------- | ----------- |
| 2048:1024:3 | 1.09573 | 0.3803916 | 10 |
| 21600:10800:3 | 175.750394 | 75.7814017 | 1 |

As a side note `pdal.exe pipeline --nostream -v 8 "pipeline_basic.json"` reports `(pdal pipeline readers.gdal Debug) GDAL debug: GDAL: GDAL_CACHEMAX = 1634 MB` which seems a little large... I have yet to mess around with it though.

hobu:
what does `gdalinfo` report about your test tiff? Is it blocked or striped?

b-gutzmann:
```
Size is 21600, 10800
Metadata:
  TIFFTAG_RESOLUTIONUNIT=2 (pixels/inch)
  TIFFTAG_XRESOLUTION=72
  TIFFTAG_YRESOLUTION=72
Image Structure Metadata:
  COMPRESSION=LZW
  INTERLEAVE=PIXEL
  PREDICTOR=2
Corner Coordinates:
Upper Left  (    0.0,    0.0)
Lower Left  (    0.0,10800.0)
Upper Right (21600.0,    0.0)
Lower Right (21600.0,10800.0)
Center      (10800.0, 5400.0)
Band 1 Block=21600x1 Type=Byte, ColorInterp=Red
  Min=0.000 Max=255.000
  Minimum=0.000, Maximum=255.000, Mean=48.700, StdDev=68.006
  Metadata:
    STATISTICS_APPROXIMATE=YES
    STATISTICS_MAXIMUM=255
    STATISTICS_MEAN=48.699839065256
    STATISTICS_MINIMUM=0
    STATISTICS_STDDEV=68.005669360942
    STATISTICS_VALID_PERCENT=100
Band 2 Block=21600x1 Type=Byte, ColorInterp=Green
Band 3 Block=21600x1 Type=Byte, ColorInterp=Blue
```

hobu:
That's a big tiff 😄 

> I wonder why we need all this. I would think that GDAL would cache at least one block per band.

Our default pixel-at-a-time read strategy is not friendly to GDAL's cache, and I doubt we're getting much benefit from it.

io/GDALReader.hpp
```diff
@@ -58,6 +58,38 @@ class PDAL_DLL GDALReader : public Reader , public Streamable
...
+        {
+            std::vector<std::vector<double>> m_data;
+            int m_blockCol;
+            int m_blockRow;
+        };
+
+        bool readBlock();
+
+        GDALReader& m_reader;
+        int m_blockRow;
```
hobu, line 80: Please add a TODO to hide these args like we do with some of the other drivers.

hobu (approved):

io/GDALReader.hpp
b-gutzmann: I'm having a hard time finding a similar TODO. Want to keep the comment consistent. Not exactly sure what hide means in this case.

b-gutzmann:
I've only been testing GeoTIFFs. Are there any other gdal raster drivers I should be looking at?

io/GDALReader.hpp
hobu: Mostly a note to myself. I'll take care of it

hobu:
> I've only been testing GeoTIFFs. Are there any other gdal raster drivers I should be looking at?

Oh there's plenty, but they're all degenerate in unpredictable ways. If we support GeoTIFFs through GDAL, that should be good enough for most scenarios people care about.

hobu:
Yeah, I'm seeing a good performance bump here.

`master`
Screenshot 2023-03-14 at 12 00 42 PM

`PDAL:gdal-reader-block-read`
Screenshot 2023-03-14 at 11 58 50 AM

hobu:
This work was moved to #4012 4012

PR #4007, opened 2023-03-22
check if directory exists before listing it

hobu:
This should address https://github.com/PDAL/python/issues/145 where `PDAL_DRIVER_PATH` is set to a bunk directory and it is trying to list it on Windows and throwing an error.


hobu:
picked back to `2.5-maintenance` at 1f19241bed5b1a236510819b19feec261ea758c6

Issue #4008, opened 2023-03-22
Change in behavior when copying a LasHeader object


AlexBass05:
I'm encountering a change in behavior when copying a `LasHeader` object.

Consider the following unit test
```
TEST(LasReaderTest, header_modified)
{
    auto getHeader = [](const std::string& path){
        PointTable table;
        Options ops;
        ops.add("filename", path);

        LasReader reader;
        reader.setOptions(ops);

        reader.prepare(table);
        return reader.header();
    };

    LasHeader h = getHeader(Support::datapath("las/simple.las"));

    EXPECT_EQ(h.softwareId(), "TerraScan");

   // Remaining checks omitted
}
```

This test is a modified version of the existing ["LasReaderTest, header"](https://github.com/PDAL/PDAL/blob/master/test/unit/io/LasReaderTest.cpp#L120) with the difference that the `LasReader` goes out of scope after creating a copy of the `LasHeader` and the added `EXPECT_EQ` check.

**Expected behavior**
This test passes. (That is at least my expectation)

**Actual behavior**
The actual behavior depends on the PDAL version:
* `PDAL 2.3.0`: The test passes.
* `PDAL 2.5.2/master`: The test fails with
```
Expected equality of these values:
  h.softwareId()
    Which is: "\f\xEDJ\xECtU\0\0M"
  "TerraScan"
```

To me this looks like we are accessing an invalid memory location because the copy-constructor of `LasHeader` is "only" making a shallow copy with some reference pointing to the `LasReader`-object.

I believe the change in behavior was introduced in https://github.com/PDAL/PDAL/commit/348ffbf4db8f46f3179e7ac46eaadb404cf1f06f (shortly before `PDAL 2.4.0`).
Note that one commit before (https://github.com/PDAL/PDAL/commit/2d846ac8d38020703379e40b548c6eab05586f8b) the above test does not compile because `LasHeader` is missing a copy-constructor.

As a client of this API it is not clear that the copy-constructor of `LasHeader` is not making a deep copy.

**System/installation information:**
Tested with building PDAL from source on the above mention git tags/commits on Ubuntu 22.04.

hobu:
This was a consequence of trying to preserve the old API https://lists.osgeo.org/pipermail/pdal/2022-January/002525.html

hobu:
https://github.com/PDAL/PDAL/pull/3682

abellgithub:
[...]

PR #4011, opened 2023-03-25
LasHeader copyable

jjimenezshaw:
This PR fixes the issue described in #4008

The changes in the code try to be minimal.

The idea is to move the ownership of `las::Header`, `las::Srs`, `las::VlrList` from `LasReader` to `LasHeader`. Because `LasReader` will always have the internal instance of `LasHeader`, the references are never pointing to invalid data.

### DIFF
--- a/io/LasHeader.cpp
+++ b/io/LasHeader.cpp
@@ -55,6 +55,9 @@ std::string GetDefaultSoftwareId()
 
 struct LasHeader::Private
 {
+    Private()
+    {}
+
     Private(las::Header& h, las::Srs& srs, las::VlrList& vlrs) : h(h), srs(srs), vlrs(vlrs)
     {}
 
@@ -62,15 +65,18 @@ struct LasHeader::Private
         interfaceVlrs(src.interfaceVlrs)
     {}
 
-    las::Header& h;
-    las::Srs& srs;
-    las::VlrList& vlrs;
+    las::Header h;
+    las::Srs srs;
+    las::VlrList vlrs;
 
     VlrList interfaceVlrs;
 };
 
 LasHeader::LasHeader(las::Header& h, las::Srs& srs, las::VlrList& vlrs) :
     d(std::make_unique<Private>(h, srs, vlrs))
+{}
+
+LasHeader::LasHeader() : d(std::make_unique<Private>())
 {}
 
 LasHeader::LasHeader(const LasHeader& src) : d(std::make_unique<Private>(*(src.d)))
@@ -93,6 +99,21 @@ LasHeader& LasHeader::operator=(LasHeader&& src)
 
 LasHeader::~LasHeader()
 {}
+
+las::Header& LasHeader::getHeader()
+{
+    return d->h;
+}
+
+las::Srs& LasHeader::getSrs()
+{
+    return d->srs;
+}
+
+las::VlrList& LasHeader::getVlrs()
+{
+    return d->vlrs;
+}
 
 std::string LasHeader::getSystemIdentifier() const
 {
--- a/io/LasHeader.hpp
+++ b/io/LasHeader.hpp
@@ -300,6 +300,13 @@ private:
 
     static void get(ILeStream& in, Uuid& uuid);
     static void put(OLeStream& in, Uuid uuid);
+
+    friend LasReader;
+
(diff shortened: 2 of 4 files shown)


jjimenezshaw:
The constructor `LasHeader(las::Header& h, las::Srs& srs, las::VlrList& vlrs);` is not used now in the code of PDAL (I do not know if it makes sense for external users). Should I remove it?

abellgithub:
This is essentially API-breaking as sets to the header will no longer update the reader, and gets will no longer be guaranteed to get current data (depending on use).

jjimenezshaw:
Well, that is the issue we are addressing here. In C++, if I make a _copy_ of a variable returned by reference, I do not expect that changes there affect the original object. If I want that I would make a reference, not a copy.

abellgithub:
You're just expecting the object to contain all the header data. It does
not. It provides an API to the data.

hobu:
@jjimenezshaw what if we were to go forward and finish deprecating and removing LasHeader and friends? Part of the problem is we didn't give enough time for a transition and we slapped it back in the codebase. We would much rather have code using the (newer, better) API.

jjimenezshaw:
> deprecating and removing LasHeader and friends

No problem. This PR was to solve a bug in the management of the data, that could happen to a user without noticing it (that happened to us. We were lucky and a unit test failed, but could not fail as well). I do not have special interest on keeping LasHeader.

hobu:
>We were lucky and a unit test failed 

Does that mean you are currently using LasHeader and friends in your software and would have to migrate to the newer api?

jjimenezshaw:
> Does that mean you are currently using LasHeader and friends in your software and would have to migrate to the newer api?

Yes, we are using it to get the bbox, wkt, pointCount and las version. We are doing it in a couple of places, so changing it should be easy. Not a problem.

hobu:
Ok, I'll close this. The next person coming along should see they should use the other non-deprecated API. It appears that both LasHeader.hpp and LasVLR.hpp are decorated as `[[deprecated]]` in master, so I don't think there's more to do here other that to add more to the release notes that it will be removed in 2.7.

loreilei:
Hi @hobu , I am a colleague of @jjimenezshaw.

Sorry for jumping in the topic LasHeader deprecation topic. I am not against it, as said before we can adapt to the new API.

I was just wondering what is the new API to read the information mentioned by @jjimenezshaw ? (bbox, wkt, point count, las version).

I saw there is a `las::Header` struct that is private the `LasReader`, and a public `LasHeader` accessor which will be deprecated if I understood correctly the previous conversation. If that understanding is correct, what is the API to read the header information?

Thanks in advance for your answers.

hobu:
I think the intention was to make `las::Header` and friends public. Is that correct @abellgithub ?

abellgithub:
No. `las::Header` is an implementation detail. It allows an internal representation of the data that does not map exactly to the bytes in the file. I left the old LasHeader API to allow for backward-compatible data access, but it's not a structure that you can pass around. It's an interface, if you will.

See this comment: https://github.com/PDAL/PDAL/issues/4008#issuecomment-1480305306

If a user wants a LAS header as raw bytes, it's very easy to read the first 227/235/375 bytes of the file. You can also use the lazperf code to easily read the file header:
```
ifstream in("filename.las");
lazperf::reader::basic_file f;
f.open(in);
lazperf::header14 h = f.header();
```

You can use this whether your file is 1.4 or not, just don't try to access fields that aren't part of the header of the actual file version as they will be uninitialized.

jjimenezshaw:
> to get the bbox, wkt, pointCount and las version.

How can we get this four parameters from a `LasReader` (without reading the whole file)?

hobu:
[`Stage::preview()`](https://github.com/PDAL/PDAL/blob/master/pdal/Stage.hpp#L115) is what you should use to get a `MetadataNode` with that info in it for drivers that support the method. `readers.las` provides this, but `readers.copc` does not. We should probably add that to enable streaming of header info over the network.

abellgithub:
I don't understand the issue. You can use the provided `LasHeader`. As Howard said, you can also call `preview()` and extract the information via. metadata. You can also use the technique I provided above. It's 4 lines of code.

Issue #3846, opened 2022-08-16
readers.las: multiple file opens for info


rcoup:
**Describe the bug**

Initially I was investigating HTTPS/S3 access to (non-COPC) LAS/LAZ files, and realised `pdal info` would end up doing two complete file downloads.

```
$ VERBOSE=1 build/bin/pdal -v 8 info --schema --metadata --driver=readers.las "https://media.githubusercontent.com/media/PDAL/data/master/autzen/autzen.laz"
...
> GET /media/PDAL/data/master/autzen/autzen.laz HTTP/2
> HEAD /media/PDAL/data/master/autzen/autzen.laz HTTP/2
> GET /media/PDAL/data/master/autzen/autzen.laz HTTP/2
> HEAD /media/PDAL/data/master/autzen/autzen.laz HTTP/2
```

So, readers.las doesn't have the ability to do partial range reads over HTTP, it ends up downloading the whole file. But that's a separate issue. 

The duplicate file opens happen locally too:
```
$ build/bin/pdal -v 8 info --schema --metadata --driver=readers.las test/data/autzen/autzen.las
$ sudo dtruss -n pdal 2>&1 | grep -P "(open.*autzen)|(close.*?\(0x4)|(read.*?\(0x4)"
 9940/0xd94b:  open_nocancel("test/data/autzen/autzen.las\0", 0x0, 0x0)		 = 4 0
 9940/0xd94b:  read_nocancel(0x4, "LASF\0", 0x1000)		 = 4096 0
 9940/0xd94b:  read_nocancel(0x4, "LASF\0", 0x1000)		 = 4096 0
 9940/0xd94b:  read_nocancel(0x4, "\314\003\334\343\023\005\364\242\0", 0x1000)		 = 866 0
 9940/0xd94b:  read_nocancel(0x4, "LASF\0", 0x1000)		 = 4096 0
 9940/0xd94b:  read_nocancel(0x4, "LASF\0", 0x1000)		 = 4096 0
 9940/0xd94b:  read_nocancel(0x4, "\314\003\334\343\023\005\364\242\0", 0x1000)		 = 866 0
 9940/0xd94b:  read_nocancel(0x4, "LASF\0", 0x1000)		 = 4096 0
 9940/0xd94b:  read_nocancel(0x4, "LASF\0", 0x1000)		 = 4096 0
[...]

PR #4017, opened 2023-04-10
Dont open readers.nitf stream multiple times if already open

hobu:
This is a consequence of #3846 that was not addressed in the previous release.


hobu:
picked back to `2.5-maintenance` at 4a5a75aa1dbc82eedaf507f1adec4b0a8dc29b06

Issue #4016, opened 2023-04-08
pdal::Geometry copy constructor may segfault


wonder-sk:
The following code crashes:

```cpp
    pdal::Geometry a;
    pdal::Geometry b = a;
```

The copy constructor expects that the source Geometry always has valid OGRGeometryH object, which is not the case here
https://github.com/PDAL/PDAL/blob/master/pdal/Geometry.cpp#L69

PDAL 2.5

hobu:
Should it clone the OGRGeometryH? I suspect sometimes it is a wrapper for geometries owned by OGR datasources, and other times it is our own geometry object.

abellgithub:
It just needs to be protected. I can do it.

PR #4018, opened 2023-04-10
harden pdal::Geometry constructors to protect against bad references 

hobu:
This hopefully addresses #4016

### DIFF
--- a/pdal/Geometry.cpp
+++ b/pdal/Geometry.cpp
@@ -67,21 +67,35 @@ Geometry::Geometry(const std::string& wkt_or_json, SpatialReference ref)
 
 
 Geometry::Geometry(const Geometry& input) : m_geom(input.m_geom->clone())
-{}
+{
+    if (input.m_geom)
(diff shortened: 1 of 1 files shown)


hobu:
picked back to `2.5-maintenance` at b898cc1ccd8270d7eced79f0da2260471ba4c6f2

pdal/Geometry.cpp
```diff
@@ -67,21 +67,35 @@ Geometry::Geometry(const std::string& wkt_or_json, SpatialReference ref)
 
 
 Geometry::Geometry(const Geometry& input) : m_geom(input.m_geom->clone())
-{}
+{
+    if (input.m_geom)
+        m_geom.reset(input.m_geom->clone());
+}
```
wonder-sk, line 73: @hobu I think you have forgotten to remove the `m_geom` init through the initializer list - it is duplicated now (and thus it would still segfault)

hobu: cleanup in #4020 ﻿

PR #4030, opened 2023-04-19
prevent invalid conversion from const compile error

jgrocha:
Compiling PDAL was breaking with:
```
/home/jgr/dev/cpp/PDAL/pdal/Geometry.cpp: In member function ‘pdal::Utils::StatusWithReason pdal::Geometry::transform(pdal::SpatialReference)’:
/home/jgr/dev/cpp/PDAL/pdal/Geometry.cpp:175:61: error: invalid conversion from ‘const OGRSpatialReference*’ to ‘OGRSpatialReference*’ [-fpermissive]
  175 |     OGRSpatialReference *inSrs = m_geom->getSpatialReference();
      |                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
      |                                                             |
      |                                                             const OGRSpatialReference*
```

I'm running on Ubuntu, with:
```
gcc --version
gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0

gdalinfo --version
GDAL 3.7.0dev-7cd1a9bb31, released 2023/04/17 (debug build)
```

Tests were passed after this change.

### DIFF
--- a/pdal/Geometry.cpp
+++ b/pdal/Geometry.cpp
@@ -153,7 +153,7 @@ Geometry& Geometry::operator=(const Geometry& input)
 
 bool Geometry::srsValid() const
 {
-    OGRSpatialReference *srs = m_geom->getSpatialReference();
+    OGRSpatialReference *srs = const_cast<OGRSpatialReference *>(m_geom->getSpatialReference());
     return srs && srs->GetRoot();
 }
 
@@ -172,7 +172,7 @@ Utils::StatusWithReason Geometry::transform(SpatialReference out)
         return StatusWithReason(-2,
             "Geometry::transform() failed.  NULL target SRS.");
 
-    OGRSpatialReference *inSrs = m_geom->getSpatialReference();
+    OGRSpatialReference *inSrs = const_cast<OGRSpatialReference *>(m_geom->getSpatialReference());
     SrsTransform transform(*inSrs, OGRSpatialReference(out.getWKT().data()));
     if (m_geom->transform(transform.get()) != OGRERR_NONE)
         return StatusWithReason(-1, "Geometry::transform() failed.");


hobu:
picked back to `2.5-maintenance` at 516e33ad7f40e54affd5754617fbe26d70ef2dd4

dg0yt:
> I think the type can just be made `const` and the cast can be dropped.

I already know that this works with gdal 3.7.0.  (I'm trying to patch this for the gdal port in vcpkg.)
The question older versions of gdals provide `GetRoot` etc. for pointers to `const`.

>  WFM, anyway.

IMO this is a bad guideline when it comes to `const_cast`. It is hiding a violating of API contracts Even if it works today with one compiler, results may be different with another update of gdal, another compiler, or just more aggressive optimization settings. And it will break at runtime, not buildtime.

abellgithub:
> IMO this is a bad guideline when it comes to const_cast. It is hiding a violating of API contracts. Even if it works today with one compiler, results may be different with another update of gdal, another compiler, or just more aggressive optimization settings. And it will break at runtime, not buildtime.

What is a bad guideline? What would break at runtime?

dg0yt:
> What is a bad guideline?

"WFM, anyway" is a bad guideline IMO when it comes to `const_cast`. Recommended guideline: ["Don't cast away `const`."](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es50-dont-cast-away-const)

> What would break at runtime?

Nobody knows. ["Modifying a const object through a non-const access path ... results in undefined behaviour."](https://en.cppreference.com/w/cpp/language/const_cast).

---

Repeating my questions in other words:
Was there a concern with regard to old gdal version for preferring `const_cast`?
Would pdal accept a PR which uses `auto`?

abellgithub:
You misundersood. Making the type const, rather than applying the cast should be fine. WFM meant that such a change compiles for me (the pointer to const is allowed in the contexts in which it is used).

PR #4033, opened 2023-04-23
rethrow arbiter::ArbiterError in situations where we can't open a file

hobu:


hobu:
pushed to `2.5-maintenance` at f5ed148d75a198cd935171580b941ce13b2d699a

PR #4039, opened 2023-05-04
Include <cstdint> to fix GCC 13 compatibility

cbrl:
The reduced standard library header coupling in GCC 13 causes PDAL to fail compilation due to a missing `#include <cstdint>` in `E57Format.h`.


abellgithub (approved):
What's not to like :)

hobu:
picked back to `2.5-maintenance` at 4f8dabc86499bcdd7e79fb387fcef8fb66b2194c

PR #4040, opened 2023-05-05
Allows mixing user set and auto-generate domains for the TileDB writer

jp-dark:
This updates the TileDB writer to use the user set domains when the domain is only set for some of the dimensions.

For example, if the user sets valid values for `x_domain_st` and `x_domain_end`, this will use the user set domain for the `X` dimension and the default domain (either from stats or max domain) for the remaining dimensions.

This also fixes a check that all required data is set to include data needed when using time as a dimension.

### DIFF
--- a/plugins/tiledb/io/TileDBWriter.cpp
+++ b/plugins/tiledb/io/TileDBWriter.cpp
@@ -374,10 +374,7 @@ void TileDBWriter::ready(pdal::BasePointTable& table)
 {
     auto layout = table.layout();
     auto all = layout->dims();
-    MetadataNode m = table.metadata();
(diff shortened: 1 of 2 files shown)


hobu:
Ignore the Alpine build but the OSX one needs to be cured before we can merge.

jp-dark:
I believe the OSX breakage is unrelated. After an initial investigation, it looks like nitro is using a deprecated C++ 11 function. I don't know why it is only triggering in OSX and only now, but I am able to reproduce the failure on another branch without the changes included here.

hobu:
I have initiated a Nitro release to deal with the error.

PR #4043, opened 2023-05-12
Don't use const_cast

dg0yt:
Amends #4030.


dg0yt:
Alternative: use `auto`.

hobu:
picked back to `2.5-maintenance` at 236fc1190dcf383af457f05d31293ff89d797ac9

PR #4047, opened 2023-05-15
shut off CPD on alpine builder

hobu:


hobu:
picked back to `2.5-maintenance` at c64e5be71de124abf997b76feb741d612536fdac

PR #4053, opened 2023-05-22
Workshop Edit Suggestions

TomTheTonk:
Edits to the workshop to fix simple spelling errors, duplicate images, and clarify instructions.


.gitignore
```diff
@@ -173,3 +173,23 @@ test/data/*.json
...
+*.html
+*.ai
+*.woff2
+*.css
+*.doctree
+doc/_build/html/.buildinfo
+*.jpg
+*.pickle
+doc/_build/html/_static/dummy
+*.inv
```
j9ac9k, line 195: I'm fairly certain we don't want to ignore most of these, but we do want to ignore the contents of `doc/_build` as that is what the documentation builds.

PR #4055, opened 2023-05-23
Improvetrim

abellgithub:


hobu:
hehe, what prompted this?

abellgithub:
I saw some other code that used erase. I kind of assumed that substr would compile to the same thing if you were essentially doing the same thing, but no.

abellgithub:
Compiler explorer is great for small stuff like this.

Issue #4029, opened 2023-04-19
Add STAC Collection as an option for readers.stac to ingest


kylemann16:
STAC Collections can be used in a similar fashion to Item Collections and Catalogs for storing references to STAC Items, and PDAL should allow users to ingest Collections in a similar way to how they ingest Catalogs.

[STAC Catalog spec](https://github.com/radiantearth/stac-spec/blob/master/catalog-spec/catalog-spec.md)
[STAC Collection spec](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md)

Issue #4028, opened 2023-04-19
Readers.STAC should support relative paths on assets


kylemann16:
Readers.stac currently requires absolute paths for assets, when STAC allows you to create catalogs with relative paths. Readers.stac should follow the usage of STAC here.

PR #4057, opened 2023-05-24
STAC Reader overhaul

kylemann16:
Adds relative pathing to STAC with the method `getPath`, which determines if the path is relative using `FileUtils::isAbsolute()`. If the path has a protocol present it is accepted as an absolute path. This PR allows Items and Catalogs within Catalogs to have pathing relative to the base Catalog, as well as for Item Assets to have pathing relative to the Item.

A local catalog and item test has been added, and some other tests have been altered to make use of relative paths to speed up test times.

Addresses #4028 

This Pull Request overhauls the STAC reader to make it more user friendly from a development and user standpoint:
  - adds specific filtering structs
  - splits up long code chunks
  - STAC Collections have been added as an input type - #4029 
  - Created separate classes for each STAC type
  - Added more specific tests
  - Test remote files for existence by heading them - #4083


io/StacReader.cpp
```diff
@@ -387,8 +389,23 @@ std::string StacReader::extractDriverFromItem(const NL::json& asset) const
...
 
+// Create link relative to the Catalog/Item/Collection that is referencing it
+const std::string getPath(std::string srcPath, std::string linkPath)
+{
+    //Make absolute path of current item's directory, then create relative path from that
+
+    //Get driectory of src item
+    const std::string baseDir = FileUtils::getDirectory(srcPath);
+    if (FileUtils::isAbsolutePath(linkPath) ||
+            linkPath.find("://") != std::string::npos)
```
hobu, line 400: Shouldn't this check actually go in `FileUtils::isAbsolutePath`? I'm confused why you are testing for a protocol.

io/StacReader.cpp
```diff
@@ -411,7 +428,7 @@ void StacReader::initializeItem(NL::json stacJson)
 
     NL::json asset = stacJson.at("assets").at(assetName);
 
-    std::string dataUrl = asset.at("href").get<std::string>();
+    std::string dataUrl = getPath(itemPath, asset.at("href").get<std::string>());
```
hobu, line 431: Please break this into two lines and add a comment about what we're doing. It reads a bit weird – we're getPath'ing an href – why?

io/StacReader.cpp
```diff
@@ -578,11 +595,11 @@ void StacReader::initialize()
 
     std::string stacType = stacJson.at("type");
     if (stacType == "Feature")
-        initializeItem(stacJson);
+        initializeItem(stacJson, m_filename);
```
hobu, line 598: These are member functions that are always going to have access to `m_filename`. Rather than adding arguments to functions, maybe it would be better to add to the `NL::json` any special members that you might possibly need. As it is now, it is confusing - why do I need both a path and a json member and what does the path need to be?

hobu (changes requested):

io/StacReader.cpp
kylemann16: Because `isAbsolutePath` is already an established check, and I thought it would be better to do any more explicit checks on top of it rather than inside of it. Though as far as I'm aware, it should be true that anything with a protocol would be an absolute path, so I could move it inside of the `isAbsolutePath` check.

io/StacReader.cpp
kylemann16: `m_filename` in this instance is the src path that everything is being related to. I could maybe add a `relative_to` key to the STAC JSON there? I don't know if that feels any better from a programmatic standpoint though

doc/stages/readers.stac.rst
```diff
@@ -20,8 +20,8 @@ Example
             "type": "readers.stac",
             "filename": "https://s3-us-west-2.amazonaws.com/usgs-lidar-stac/ept/catalog.json",
             "reader_args": [{"type": "readers.ept", "resolution": 100}],
-            "item_ids": ["MD_GoldenBeach_2012"],
```
hobu, line 23: Do you have synonyms for this change?

kylemann16: I don't, should there be synonyms? `catalog_ids` and `collection_ids` just started feeling too long so I removed the `ids` from them all

kylemann16: Ahhh, makes sense. Okay,  I'll add them.

io/StacReader.hpp
```diff
@@ -31,68 +31,63 @@
...
 
 #include <pdal/PointView.hpp>
 #include <pdal/Reader.hpp>
-
-
 #include <pdal/JsonFwd.hpp>
 #include <pdal/StageFactory.hpp>
 #include <filters/MergeFilter.hpp>
-
+#include "private/stac/Item.hpp"
```
connormanning, line 45: Shouldn't be including this from a public header.  I think you're just using it for the `Item` class but its only usage here is by reference, so it can be forward-declared.

io/StacReader.cpp
```diff
@@ -127,112 +123,206 @@ void StacReader::addArgs(ProgramArgs& args)
...
+    reader->setLog(log());
+    m_merge.setInput(*reader);
+    m_p->m_readerList.push_back(reader);
+}
 
+void StacReader::handleItem(NL::json stacJson, std::string itemPath)
+{
+    stac::Item item(stacJson, m_filename, *m_p->m_connector,
+        m_args->validateSchema);
+    if (item.init(*m_p->m_itemFilters, m_args->rawReaderArgs, m_args->schemaUrls))
```
connormanning, line 207: It looks like the missing `else` could represent a failure case - maybe add a comment here about why it's ok to no-op if `item.init()` returns false.

io/StacReader.cpp
```diff
@@ -127,112 +123,206 @@ void StacReader::addArgs(ProgramArgs& args)
...
+    if (catalog.init(*m_p->m_catFilters, m_args->rawReaderArgs,
+        m_args->schemaUrls, true))
     {
-        schemaUrl = m_args->featureSchemaUrl;
-        for (auto& extSchemaUrl: stacJson.at("stac_extensions"))
+        for (stac::Item& item: catalog.items())
+            addItem(item);
+    }
+    stac::ErrorList errors = catalog.errors();
+    if (errors.size())
```
connormanning, line 224: Looks like this is a "soft" sort of failure since we aren't `throw`ing in response to errors here.  A comment about why this failure is not fatal would help.

io/StacReader.cpp
```diff
@@ -127,112 +123,206 @@ void StacReader::addArgs(ProgramArgs& args)
...
-    else
-        throw pdal_error("Invalid STAC type for PDAL consumption");
-
-    NL::json schemaJson = m_p->m_connector->getJson(schemaUrl);
-    val.set_root_schema(schemaJson);
-    val.validate(stacJson);
 }
 
-void StacReader::handleReaderArgs()
+void StacReader::handleCollection(NL::json stacJson, std::string catPath, bool isRoot)
```
connormanning, line 234: `handleCollection` and `handleCatalog` both have `isRoot` arguments that look unused.  Remove?

connormanning, line 234: I assume this one should be `colPath` instead of `catPath` which was presumably copy/pasted from the `catalog` version.  Same is true in the header prototypes.  Seems reasonable to just call them `path` or `filename` in this context since they are scoped to a function whose name already describes them.

io/StacReader.cpp
```diff
@@ -127,112 +123,206 @@ void StacReader::addArgs(ProgramArgs& args)
...
+        m_p->m_itemFilters->collections = m_args->collections;
+        m_p->m_colFilters->ids = m_args->collections;
     }
 
+
     if (!m_args->dates.empty())
     {
-        //TODO validate supplied dates?
-        log()->get(LogLevel::Debug) << "Dates selected: " << m_args->dates  << std::endl;
+        stac::StacUtils stacUtils;
```
connormanning, line 297: `getStacTime` is `static` so I think we can skip instantiating this thing.

io/StacReader.cpp
```diff
@@ -127,112 +123,206 @@ void StacReader::addArgs(ProgramArgs& args)
...
+            }
+
+            try
+            {
+                std::string minDate(datepair[0].get<std::string>());
+                std::string maxDate(datepair[1].get<std::string>());
+
+                std::time_t minTime = stacUtils.getStacTime(minDate);
+                std::time_t maxTime = stacUtils.getStacTime(maxDate);
+                m_p->m_itemFilters->datePairs.push_back({ minTime, maxTime });
```
connormanning, line 317: Maybe worth a warning if `minTime >= maxTime`?

io/StacReader.cpp
```diff
@@ -344,227 +406,9 @@ void StacReader::setConnectionForwards(StringMap& headers, StringMap& query)
...
-                initializeItemCollection(nextJson);
-            }
-        }
-    }
-}
-
 void StacReader::initialize()
 {
-
+    stac::StacUtils u;
```
connormanning, line 411: Should be unneeded since the methods are static.

io/private/stac/Item.cpp
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    void Item::validate()
+    {
+
+        nlohmann::json_schema::json_validator val(
+            [this](const nlohmann::json_uri& json_uri, nlohmann::json& json) {
+                NL::json tempJson = m_connector.getJson(json_uri.url());
+
+                std::lock_guard<std::mutex> lock(m_mutex);
```
connormanning, line 224: It doesn't appear that we need `m_mutex` here, or at all as a member variable, of this class.  This is the only place that it's used and it doesn't look like there is any contention possible with this usage.

io/private/stac/Utils.hpp
```diff
@@ -0,0 +1,170 @@
...
+namespace pdal
+{
+
+namespace stac
+{
+    pdal_error stac_error(std::string id, std::string stacType, std::string const& msg);
+    pdal_error stac_error(std::string const& msg);
+
+
+class StacUtils
```
connormanning, line 49: Can this just be a namespace like PDALUtils/FileUtils for example?  Doesn't look like we ever need an instantiation of this thing.

io/private/stac/Catalog.cpp
```diff
@@ -0,0 +1,221 @@
...
+                {
+                    std::lock_guard<std::mutex> lock(m_mutex);
+                    StacError p {absLinkPath, "Unknown error"};
+                    m_errors.push_back(p);
+                }
+            });
+        }
+
+
+        if (isRoot)
```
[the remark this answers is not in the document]
kylemann16: `isRoot` would only be called as true from the `StacReader`, since that's the entry point into the STAC structure, if there are sub-catalogs or sub-collections they'll be created with `isRoot=false` like [here](https://github.com/PDAL/PDAL/blob/relative-stac/io/private/stac/Catalog.cpp#L105-L106). So we'll only be doing a join if we're handling the original collection/catalog that was created by `StacReader`

io/StacReader.cpp
kylemann16: `isRoot` is the signal to say that this collection/catalog should handle the pool joining

connormanning: Makes sense, but I think these particular arguments are still unused since the StacReader itself hard-sets this to `true` without regard for this arg.

kylemann16: Oh whoops, you're right

io/private/stac/Utils.hpp
kylemann16: I did this mostly to clear up "unused functions" warnings, I've tried it in a few different ways now and I can't seem to get it to work as just a namespace

io/private/stac/Catalog.cpp
```diff
@@ -0,0 +1,221 @@
...
+                try
+                {
+                    if (linkType == "item")
+                    {
+                        NL::json itemJson = m_connector.getJson(absLinkPath);
+                        Item item(itemJson, absLinkPath, m_connector, m_validate);
+
+                        bool valid = item.init(*filters.itemFilters,
+                            rawReaderArgs, m_schemaUrls);
+                        if (valid)
```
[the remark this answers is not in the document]
kylemann16: Maybe valid is the wrong word to use here. If it returns false, then the item doesn't pass filtering, same for collections and catalogs. If the `init` method returns false then either it has no items to process or the catalog/collection itself didn't pass the filtering phase.

hobu (approved):

PR #4058, opened 2023-05-24
Point view sorting stopgap

connormanning:
`PointRef`, which is the iterator type for a `PointView`, has some special constructor/assignment operators which are not intended for direct downstream usage, but instead exist for `std::sort` and friends to work.  However on some compilers (e.g. clang libcxx15+ on M1 hardware) these are not currently working correctly.

This PR leaves the broken operators present, but switches internal PDAL PointView sorting to use a new method `PointView::sort`.  This is a stopgap until we can get these functions working properly with all compilers.  So currently, mutating algorithms like `std::sort` should not be used on a `PointView`, although read-only operations like `std::accumulate` should be fine to use.

### DIFF
--- a/pdal/PointView.hpp
+++ b/pdal/PointView.hpp
@@ -305,12 +305,12 @@ public:
     KD2Index& build2dIndex();
 
     template <typename Compare>
-    void sort(Compare compare)
+    void stableSort(Compare compare)
     {
         auto order = m_index;
         std::stable_sort(
-            order.begin(), 
-            order.end(), 
+            order.begin(),
+            order.end(),
             [this, &compare](const PointId a, const PointId b)
             {
                 return compare(PointRef(*this, a), PointRef(*this, b));
--- a/pdal/PointView.hpp
+++ b/pdal/PointView.hpp
@@ -304,6 +304,20 @@ public:
     KD3Index& build3dIndex();
     KD2Index& build2dIndex();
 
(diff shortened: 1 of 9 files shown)


pdal/PointView.hpp
```diff
@@ -304,6 +304,20 @@ class PDAL_DLL PointView : public PointContainer
     KD3Index& build3dIndex();
     KD2Index& build2dIndex();
 
+    template <typename Compare>
+    void stableSort(Compare compare)
+    {
+        auto order = m_index;
```
abellgithub, line 310: Why do you not sort the index in place?

pdal/PointView.hpp
```diff
@@ -304,6 +304,20 @@ class PDAL_DLL PointView : public PointContainer
...
+        std::stable_sort(
+            order.begin(),
+            order.end(),
+            [this, &compare](const PointId a, const PointId b)
+            {
+                return compare(PointRef(*this, a), PointRef(*this, b));
```
abellgithub, line 316: Not a big deal, but you could avoid the PointRef altogether and accept PointId's. There's already a comparator for around for that, but it also takes a dimension, which you could std::bind if you wanted.

hobu:
Picked back to `2.5-maintenance` at 862c6188efab7cede643c2d5a3523c37e8561d3b

PR #4069, opened 2023-06-01
Refactor TileDB writer

jp-dark:
* Clean-up and isolate code for creating a new TileDB array.
* Inline helper function that is only used once.
* Remove duplicate buffer data for TileDB dimensions.
* Remove unneeded private variables from TileDBWriter class,
* Move all member variables for user provided data to Args struct.

### DIFF
--- a/plugins/tiledb/io/TileDBWriter.cpp
+++ b/plugins/tiledb/io/TileDBWriter.cpp
@@ -207,14 +207,12 @@ void TileDBWriter::ready(pdal::BasePointTable& table)
     auto layout = table.layout();
     for (const auto& dimId : layout->dims())
     {
-        // Get the name and type of the dim.
-        std::string dimName = layout->dimName(dimId);
-        Dimension::Type dimType = layout->dimType(dimId);
 
         // Allocate the buffer.
-        m_buffers.emplace_back(dimName, dimId, dimType);
-        m_buffers.back().m_buffer.resize(m_args->m_cache_size *
-                                         Dimension::size(dimType));
+        m_buffers.emplace_back(layout->dimName(dimId), dimId,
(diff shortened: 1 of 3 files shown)


plugins/tiledb/io/TileDBWriter.cpp
```diff
@@ -252,245 +203,186 @@ void TileDBWriter::initialize()
...
+    for (const auto& dimId : layout->dims())
+    {
+        // Get the name and type of the dim.
+        std::string dimName = layout->dimName(dimId);
+        Dimension::Type dimType = layout->dimType(dimId);
+
+        // Allocate the buffer.
+        m_buffers.emplace_back(dimName, dimId, dimType);
+        m_buffers.back().m_buffer.resize(m_args->m_cache_size *
+                                         Dimension::size(dimType));
```
hobu, line 217: You probably need to get the [layout->dimSize](https://github.com/PDAL/PDAL/blob/master/pdal/PointLayout.hpp#L203) because this might not be the same as a default `Dimension::size` if someone added it to the PointLayout differently than expected.

plugins/tiledb/io/TileDBWriter.cpp
```diff
@@ -499,21 +391,49 @@ void TileDBWriter::ready(pdal::BasePointTable& table)
...
+        default:
+            throwError("Unsupported attribute type for " + dimBuffer.m_name);
+        }
 
-    m_xs.push_back(x);
-    m_ys.push_back(y);
-    m_zs.push_back(z);
-    if (m_use_time)
-        m_tms.push_back(tm);
+        size_t size = Dimension::size(dimBuffer.m_type);
```
hobu, line 434: same as above. This might not be the size you expect.

hobu (approved):

PR #4070, opened 2023-06-02
Arbiter S3 IMDS fallback v2->v1

connormanning:
Fall back to IMDSv1 if v2 fails.  Remove exponential backoff/retry loop for these queries (which is most of the diff here).

### DIFF
--- a/vendor/arbiter/arbiter.cpp
+++ b/vendor/arbiter/arbiter.cpp
@@ -1742,6 +1742,15 @@ namespace
         if (s.size() && std::isspace(s.back())) s.pop_back();
         return s;
     }
+
+    bool isVerbose()
+    {
+        std::string verbose;
+        if (auto e = env("VERBOSE")) verbose = *e;
+        else if (auto e = env("CURL_VERBOSE")) verbose = *e;
+        else if (auto e = env("ARBITER_VERBOSE")) verbose = *e;
+        return (!verbose.empty()) && !!std::stol(verbose);
+    }
 }
 
 namespace drivers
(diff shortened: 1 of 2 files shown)


msmitherdc:
It does fallback, but it takes a while with multiple timeouts to do so
`
pdal info s3://grid-dev-publiclidar/nato/slovenia/copc/TM_374_130.copc.laz --verbose=8
(PDAL Debug) Debugging...
Curl failure: Timeout was reached
Curl failure: Timeout was reached
(pdal info Debug) Executing pipeline in stream mode.
Curl failure: Timeout was reached
Curl failure: Timeout was reached
{
  "file_size": 123637865,
  "filename": "s3://grid-dev-publiclidar/nato/slovenia/copc/TM_374_130.copc.laz",
  "now": "2023-06-03T16:30:54+0000",
  "pdal_version": "2.6.0 (git-version: a0141a)",
  "reader": "readers.copc",
`

msmitherdc:
for `time pdal info --metadata s3://grid-dev-publiclidar/nato/slovenia/copc/TM_374_130.copc.laz --verbose=8`

v2: `real	0m0.319s`
v1:  `real	0m5.126s`

connormanning:
Yeah that's about expected with the current code - I can get that quite a bit lower with a few small tweaks but there is always going to be a bit of a timeout for this fallback.  I just pushed a few tweaks that should lower it to about a second, and there is a bit more that can be done to get it down to an explicitly configured timeout of, say, 100ms.

hobu:
Picked back to `2.5-maintenance` at 751a64bcea9475e013a0293d941f9aa897ef0779

PR #4071, opened 2023-06-03
update plugin cmake macro to specify MSVC

Crghilardi:
This PR modifies the plugin section of the `cmake/macros.cmake` file to further specify `MSVC` instead of `WIN32`. I was cross-compiling using mingw on a linux host, this would produce `liblibpdal_plugin_kernel_fauxplugin.dll` (note double `lib` prefix). With this PR, the dll is correctly named `libpdal_plugin_kernel_fauxplugin.dll`

FYI @xantares and @dg0yt since they have helped with many of the other merged `MinGW` patches.

### DIFF
--- a/cmake/macros.cmake
+++ b/cmake/macros.cmake
@@ -126,7 +126,7 @@ macro(PDAL_ADD_PLUGIN _name _type _shortname)
     set(multiValueArgs FILES LINK_WITH INCLUDES SYSTEM_INCLUDES)
     cmake_parse_arguments(PDAL_ADD_PLUGIN "${options}" "${oneValueArgs}"
         "${multiValueArgs}" ${ARGN})
-    if(WIN32)
+    if(MSVC)
         set(${_name} "libpdal_plugin_${_type}_${_shortname}")
     else()
         set(${_name} "pdal_plugin_${_type}_${_shortname}")


xantares:
That works, but I also think using MODULE instead of SHARED in add_library would remove the lib prefix

dg0yt:
IMO it is already wrong to vary target names by platform/compiler. This makes it hard to use these targets.
What you really want is to control the `OUTPUT_NAME` property.

hobu:
> That works, but I also think using MODULE instead of SHARED in add_library would remove the lib prefix

That didn't work with older CMake versions across all platforms we support. If it now works with our current minimum CMake version, we should make the change.

> IMO it is already wrong to vary target names by platform/compiler. This makes it hard to use these targets. What you really want is to control the OUTPUT_NAME property.

Agree. Again, old CMake. Happy to merge a patch for 2.6.0 that does this.

Crghilardi:
> IMO it is already wrong to vary target names by platform/compiler. This makes it hard to use these targets. What you really want is to control the `OUTPUT_NAME` property.

@dg0yt I don't think I understand how to do that without re-working a lot of the `CMake` around the plugins? Or am I misreading things?

I originally ran into this while updating the binary packaging for [Julia](https://github.com/JuliaPackaging/Yggdrasil/pull/6822). I have this particular change  (`WIN32` -> `MSVC` ) vendored there and CI is passing.

If someone can help me with this I am happy to try some things but otherwise I don't have enough `CMake` to take this much further based on the suggestions. I am OK with closing this and maybe it gets revisited at a later time.

dg0yt:
@Crghilardi I think it is okay if this PR is merged without further changes to the CMake code.
However the key question is if the plugins are found when needed. This PR changes the output name for MinGW, so you need to double check that it (still) works. The code which is loading the plugins at runtime must use the same name patterns.

PR #4077, opened 2023-06-10
Fix FindTileDB.cmake to check version

jp-dark:
Find version was not correctly checking version due to mixed case name differences. This change updates the TileDB find package module to consistently use snakecase.

### DIFF
--- a/cmake/modules/FindTileDB.cmake
+++ b/cmake/modules/FindTileDB.cmake
@@ -3,12 +3,12 @@
 # CMake module to search for the TileDB library
 #
 # On success, the macro sets the following variables:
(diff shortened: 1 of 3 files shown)


hobu:
Can we ditch FindTileDB.cmake entirely at this point and go to CONFIG-based find?

jp-dark:
> Can we ditch FindTileDB.cmake entirely at this point and go to CONFIG-based find?

I initially attempted that, but I hit issues with the version compatibility. The `TileDBConfig.cmake` requires minor version match for a version to be compatible and CMake does not provide a way to check for ranges of versions with this type of version compatibility (at least not a way I could figure out).

PR #4089, opened 2023-06-28
Update arbiter bundle

connormanning:
Update arbiter bundle:
- Properly populate the response data for PUT requests
- Remove an extraneous log


hobu:
picked back to `2.5-maintenance` at 376ec59dd11bfcae4d9e8cce98e643cc05cba650

PR #4096, opened 2023-07-07
Workshop instructions edits

ibell13:
Minor improvements on usability for PDAL workshop instructions and example pipelines

### DIFF
--- a/doc/workshop/generation/dtm/gdal.json
+++ b/doc/workshop/generation/dtm/gdal.json
@@ -1,6 +1,6 @@
 {
     "pipeline": [
-        "./exercises/analysis/ground/denoised-ground-only.copc.laz",
+        "./exercises/analysis/ground/denoised-ground-only.laz",
         {
             "filename":"./exercises/analysis/dtm/dtm.tif",
             "gdaldriver":"GTiff",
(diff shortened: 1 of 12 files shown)


abellgithub:
This documentation gets the most love.

doc/workshop/generation/dtm/gdal.json
```diff
@@ -1,6 +1,6 @@
 {
     "pipeline": [
-        "./exercises/analysis/ground/denoised-ground-only.copc.laz",
```
j9ac9k, line 3: Is there no `copc` file with the exercises to work with?

doc/workshop/manipulation/clipping/clipping.json
```diff
@@ -9,8 +9,8 @@
             "type": "filters.overlay"
         },
         {
-            "expression": "Classification == 6",
-            "type": "filters.expression"
+            "limits": "Classification[6:6]",
+            "type": "filters.range"
```
j9ac9k, line 13: If memory serves, a significant chunk of the recent changes was going from `filters.range` to `filters.expression`, this change would undo one of those changes, is there a particular reason for this change?

(I could also be remembering wrong)

doc/workshop/generation/dtm/gdal.json
ibell13: Just looked, think I might have gotten confused there from an old version of the ground exercise that output laz only. Newest version makes copc

hobu:
picked to `2.5-maintenance` at 5cd490db21e9cb0e17de88e71bc05a5101ddccea

PR #4097, opened 2023-07-10
Fix simplify test

hobu:
GEOS' simplify functions are quite sensitive to versions. GEOS 1.12 was recently released that changed the behavior of how one of the PDAL test polygons simplified, but there is no convenient way to get the GEOS version because PDAL uses GEOS through GDAL interfaces alone.

Rather than change a bunch of stuff, I just changed the test to be a simple circle and simplification of that. GEOS shouldn't change its behavior of these.


hobu:
Picked back to `2.5-maintenance` at f028f5e783206e5fc460a31fe9968ed336734ed3

PR #4100, opened 2023-07-10
Fix TileDBWriter argument types

jp-dark:


hobu:
Is this also going to change some behavior and need a doc update?

jp-dark:
It does change behavior, but the new behavior matches what I would expect reading the current documentation. I consider the previous tiles in particular to be a bug. Both the tile and domain should be the same type as the dimensions.

hobu:
> It does change behavior, but the new behavior matches what I would expect reading the current documentation. 

Ok, maybe add a note in the docs or something. I suspect this change might surprise someone, but maybe it is not so consequential.

PR #4104, opened 2023-07-14
update GHA workers to use v3 of checkout and mamba-org/setup-micromamba

hobu:


j9ac9k:
Would you like me to configure dependabot to keep our GHA config updated?

scripts/ci/environment.yml
```diff
@@ -13,6 +13,9 @@ dependencies:
   - nitro
   - ceres-solver
   - python=3.10
+  - conda
+  - mamba
+  - conda-build
```
j9ac9k, line 18: instead of installing `conda-build`, we probably want to install `boa` instead, it uses mamba to create to build environments:

https://boa-build.readthedocs.io/en/latest/

I think we would need to change the compile commands from `mamba build ...` to `conda mambabuild ...`

They appear to have a new recipe spec that you can build using `boa build ...` but I have no idea how widespread it's adoption is, and I don't know if it would be even beneficial for our use-case, but you can read about it here if you're curious: https://boa-build.readthedocs.io/en/latest/recipe_spec.html

PR #4107, opened 2023-07-14
Create updated filter profiles for TileDB writer

jp-dark:
* Add 3 possible filter profiles: none, balanced, and aggressive
* Add scale and offset input parameters for float-scale compression of X, Y, and Z
* Add delta filter (requires TileDB 2.16)


hobu:
Docs not building due to tiledb doc warnings.

jp-dark:
> Docs not building due to tiledb doc warnings.

They should be correct now. Sorry for all the extra CI jobs!

PR #4108, opened 2023-07-17
straighten filter

gui2dev:
# filters.straigthen

Given a polyline `(x,y,z,m)` as input, the **straighten filter** will transform the point cloud in a new parametric coordinate system, each point in world coordinate `(X,Y,Z)` is being projected along closest poyline segment, and rotated along the segment accordingly to the average m/roll value.

The new coordinate system `(X', Y', Z')` could be understood as :
* `X'` : curvilinear abcissa (or meter point)
* `Y'` : orthogonal distance to segment (or orthogonal distance to polyline)
* `Z'` : orthogonal distance from (rolling) plane

This can simplify some analysis operation on linear assets.

> **_Streamable stage_**
> 
>  This stage supports streaming operations

## Options

* `polyline` : wkt or json definition of a 3D linestring with measurment (`LINESTRING ZM` in wkt) along which the cloud will be straighten. M is supposed to be `roll` expressed in `radians`. This is mandatory.
* `reverse` : whether to straighten or unstraighten the point cloud (default : `false`)
* `offset` : if you want to add an `X'` during straightening operation (or take an offset into account while unstraightening). This can be understood as a starting meter point.

## Illustrations

| Original cloud with polyline | Straighten cloud | Unstraightened cloud fused with original cloud |
|:--:|:--:|:--:|
| [](https://gdevaltametris.z6.web.core.windows.net/pdal/filters.straighten/img/origin.jpeg) | [](https://gdevaltametris.z6.web.core.windows.net/pdal/filters.straighten/img/straight.jpeg) | [](https://gdevaltametris.z6.web.core.windows.net/pdal/filters.straighten/img/unstraight.jpeg)


gui2dev:
> This doesn't need to be in `./plugins` because it doesn't require anything beyond the standard set of dependencies that PDAL always has available to it (vendored Eigen, GDAL, etc.). Please move this into `./filters` and your utility functions into `./filters/private/straighten`. Make sure to use `CREATE_STATIC_STAGE` to create the stage instead of `CREATE_SHARED_STAGE`.
> 
> Please do the same for #4109 now that you have removed the GeographicLib dependency.
> 
> Thanks so much for these contributions!

moved.

Issue #47, opened 2011-08-29
OCI driver unit tests


mpgerlek:
Need to make unit tests for Oracle driver that can run by default.

PR #4111, opened 2023-07-19
Fix az driver

gui2dev:
In `arbiter` `AZ` driver was missing `override` for `std::unique_ptr<std::size_t> tryGetSize(std::string path, http::Headers headers, http::Query query) const`.
`Protocol` was missing on init as well.

As a consequence, most `az://` requests were failing.

### DIFF
--- a/vendor/arbiter/arbiter.cpp
+++ b/vendor/arbiter/arbiter.cpp
@@ -2696,7 +2696,7 @@ AZ::AZ(
         Pool& pool,
         std::string profile,
         std::unique_ptr<Config> config)
-    : Http(pool, "az", profile == "default" ? "" : profile)
+    : Http(pool, "az", "http", profile == "default" ? "" : profile)
     , m_config(std::move(config))
 { }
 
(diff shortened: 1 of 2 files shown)


gui2dev:
> Can you make this PR upstream at https://github.com/connormanning/arbiter ?

done. Check out [#47](https://github.com/connormanning/arbiter/pull/47)

vendor/arbiter/arbiter.cpp
```diff
@@ -2886,7 +2886,10 @@ std::string AZ::Config::extractBaseUrl(
     return account + "." + service + "." + endpoint + "/";
 }
 
-std::unique_ptr<std::size_t> AZ::tryGetSize(std::string rawPath) const
+std::unique_ptr<std::size_t> AZ::tryGetSize(
+    const std::string rawPath,
```
abellgithub, line 2890: Is there some reason for `const` in these arguments without passing by reference?

gui2dev: I don't know. It's the signature of the function in `Http` class. It's how @connormanning wrote it I guess.

abellgithub: It's just weird to need a copy that you're not going to modify, but it's not a big deal.

gui2dev: Maybe some refactoring ahead in `arbiter`.

gui2dev:
Closing as it's replaced by #4112

vendor/arbiter/arbiter.cpp
connormanning: I just like making things `const` by default when possible so I've adopted the habit of const-ifying in this context.  Note that this is in the cpp implementation, the header interface does not have this `const` so if for some reason this parameter copy had to be changed to be mutable in the implementation, the caller wouldn't notice.

PR #4115, opened 2023-07-20
GeoParquet and Arrow IPC read/write support 

hobu:
This PR adds `writers.arrow` and `readers.arrow` to allow data interchange between the Arrow/Parquet formats and PDAL.


plugins/arrow/io/ArrowWriter.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+
+    NL::json column;
+    column["encoding"] = "WKB";
+    column["geometry_types"] = std::vector<std::string> {"Point"};
+
+    if (ref.empty())
+        ref = SpatialReference("EPSG:4326");
+    column["edges"] = ref.isGeographic() ? "spherical" : "planar";
+    column["crs"] = pdal::Utils::replaceAll(ref.getPROJJSON(), "\n", "");
```
rouault, line 216: This should be the JSON object itself, i.e. not serialized as a string

paleolimbot (commented):
This is so cool! I left a note about type IDs, which might help consolidate/speed up the conversion logic.

I didn't see the place where you deal with geometry columns so perhaps that isn't quite sorted yet...if you *do* want to support GeoArrow in its infancy, you could use https://github.com/geoarrow/geoarrow-c/blob/main/src/geoarrow/geoarrow.hpp , which would let you do:

```cpp
#include "geoarrow.hpp"
#include "nanoarrow.hpp"

nanoarrow::UniqueSchema schema;
nanoarrow::UniqueArray array;
arrow::ExportArray(*mArray, schema.get(), array.get());

auto type = geoarrow::VectorType::Make(schema.get());
if (!type.valid()) {
  // error
}

switch (type.id()) {
  case GEOARROW_TYPE_WKB:
  case GEOARROW_TYPE_WKT:
  case GEOARROW_TYPE_LARGE_WKB:
  case GEOARROW_TYPE_LARGE_WKT: {
    ArrowArrayView array_view;
    ArrowArrayViewInitFromSchema(&array_view, schema.get());
    for (int64_t i = 0; i < array_view.length; i++) {
      if (!ArrowArrayViewIsNull(&array_view, i)) {
        ArrowBufferView value = ArrowArrayViewGetBytesUnsafe(&array_view, i);
      } else {
         // Wherever NULL geometries go
      }
    }
  }
  default: {
    if (type.geometry_type() != GEOARROW_GEOMETRY_TYPE_POINT) {
      // error
    }
    GeoArrowArrayView array_view;
    GeoArrowArrayViewInitFromType(&array_view, type.id());
    for (int64_t i = 0; i < array_view.length[0]; i++) {
      double x = GEOARROW_COORD_VIEW_VALUE(
        array_view.coord_view, array_view.offset[0] + i, 0);
      double y = GEOARROW_COORD_VIEW_VALUE(
        array_view.coord_view, array_view.offset[0] + i, 1);
    }
  }
}
```

The logic behind `auto type = geoarrow::VectorType::Make(schema.get());` is pretty involved. It currently requires that the geoarrow extension types are registered with Arrow C++ (i.e., https://github.com/geoarrow/geoarrow-c/blob/main/src/geoarrow/geoarrow_arrow.hpp#L45-L50 ) and that the user *wrote* an extension type in the first place, although it will expand such that you can do `geoarrow::VectorType::Make(schema.get(), extension_name);` to support geoarrow encoding in GeoParquet without extension types (which we haven't formally proposed, but will next week sometime). There will also be at some point the ability to "sniff" the extension name (e.g., binary column -> geoarrow.wkb, `struct<x: double, y: double>` -> geoarrow.point, etc.).

You could also use nanoarrow for your type conversions ( https://arrow.apache.org/nanoarrow )...it won't help you with Parquet IO but would remove the Arrow C++ dependency for things like reading an `ArrowArrayStream` from GDAL or an ADBC driver. (Full disclaimer: I wrote nanoarrow and its IPC extension).

plugins/arrow/io/ArrowReader.cpp
```diff
@@ -0,0 +1,512 @@
...
+
+bool ArrowReader::fillPoint(PointRef& point)
+{
+
+
+    for(int columnNum = 0; columnNum < m_currentBatch->num_columns(); ++columnNum)
+    {
+        // https://arrow.apache.org/docs/cpp/api/array.html#_CPPv4N5arrow5ArrayE
+        std::shared_ptr<arrow::Array> array = m_currentBatch->column(columnNum);
+        arrow::DoubleArray* dArray = dynamic_cast<arrow::DoubleArray*>(array.get());
```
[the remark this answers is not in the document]
hobu: Thanks for the tip. The docs are pretty thin, and I just flung some stuff together 😄

plugins/arrow/io/ArrowReader.cpp
```diff
@@ -0,0 +1,532 @@
...
+            continue;
+        }
+        arrow::BinaryArray* binaryArray = dynamic_cast<arrow::BinaryArray*>(array.get());
+        if (binaryArray)
+        {
+            // Binary arrays are assumed to be WKT from GeoParquet
+            std::string_view wkb = binaryArray->Value(m_currentBatchPointIndex);
+            pdal::Geometry pt = pdal::Geometry(std::string(wkb));
+            OGRGeometry* g = (OGRGeometry*) pt.getOGRHandle();
+            OGRPoint* p = g->toPoint();
```
rouault, line 459: toPoint() is actually a static_cast, and thus you should only invoke it if you known that wkbFlatten(g->getGeometryType()) == wkbPoint. Or use a dynamic_cast<OGRPoint*>(g)

plugins/arrow/io/ArrowReader.cpp
```diff
@@ -0,0 +1,532 @@
...
+        arrow::UInt64Array* uint64Array = dynamic_cast<arrow::UInt64Array*>(array.get());
+        if (uint64Array)
+        {
+            point.setField<uint64_t>(m_arrayIds[columnNum], uint64Array->Value(m_currentBatchPointIndex));
+            continue;
+        }
+        arrow::BinaryArray* binaryArray = dynamic_cast<arrow::BinaryArray*>(array.get());
+        if (binaryArray)
+        {
+            // Binary arrays are assumed to be WKT from GeoParquet
```
rouault, line 455:
```suggestion
            // Binary arrays are assumed to be WKB from GeoParquet
```

plugins/arrow/io/ArrowReader.cpp
```diff
@@ -0,0 +1,532 @@
...
+            std::string_view wkb = binaryArray->Value(m_currentBatchPointIndex);
+            pdal::Geometry pt = pdal::Geometry(std::string(wkb));
```
rouault, line 457: but I'm puzzled here since it seems the Geometry constructors expects wkt or json, not wkb...

hobu: I added some kind of ugly stuff to take in wkb (by checking if the first character is 00 or 01) in the ctor and accepting that first if found.

kylebarron:
This is awesome to see!

The GeoParquet spec (as of 1.0, which [will come out soon](https://github.com/opengeospatial/geoparquet/pull/184)), only supports geometries encoded as WKB. If I load that Parquet file, I see table schema/data with:

```
xyz: list<element: double not null> not null
  child 0, element: double not null
xyz: [[[494428.61,4877455.58,131.57],[494392.76,4877472.61,136.06],...,[494517.87,4878739.62,129.26],[494490.24,4878741.67,129.21]]]
```

Since this isn't WKB-encoded, this isn't a valid GeoParquet file as of 1.0. It's possible that GeoParquet will, after 1.0, also allow native coordinate layouts as part of the spec, but this needs further discussion on broad interoperability vs raw performance.

On the GeoArrow side, it looks like the Arrow schema is missing extension metadata that declares the `xyz` column to be a `geoarrow.point` column:

```py
import pyarrow.parquet as pq
schema = pq.read_schema('autzen-utm.parquet')
schema.field('xyz').metadata
# None
```

To meet the current GeoArrow spec (see [`extension-types.md`](https://github.com/geoarrow/geoarrow/blob/main/extension-types.md)) the _field_ metadata should additionally be named `geoarrow.point`

hobu:
@kylebarron thanks. This is exactly the feedback I was looking for. 

https://github.com/PDAL/PDAL/raw/af854282caf2a643a068bf14a1c41c320a0fcacc/test/data/arrow/1.2-with-color.parquet has WKB encoding. I currently default to OFF for writing GeoParquet WKB when writing WKB output. I'm not sure what we should do there. It seems wasteful to write the struct AND the WKB, and it's a lot more convenient to get the data from the struct than from the WKB. What are you guys typically doing with your writers?

Additionally, I wonder what this writer should be doing for batch stats (like @rouault has added for GDAL). Because point cloud data typically aren't well sorted, it's doubtful you get much value from the stats for most point clouds (because they won't provide much batch discrimination).

plugins/arrow/io/ArrowWriter.cpp
```diff
@@ -228,7 +261,22 @@ void ArrowWriter::ready(PointTableRef table)
             auto dimType = layout->dimType(id);
             std::string name = layout->dimName(id);
             std::shared_ptr<arrow::Field> field = toArrowType(name, dimType);
+            auto kvMetadata = field->metadata()
+                                ? field->metadata()->Copy()
+                                : std::make_shared<arrow::KeyValueMetadata>();
+            if (Utils::iequals("wkb", name))
+            {
+                kvMetadata->Append("ARROW:extension:name", "WKB");
```
paleolimbot, line 269: I think this one should be `"geoarrow.wkb"` (I know this is confusing, since the Geo*Parquet* encoding is just `"WKB"`).

plugins/arrow/io/ArrowWriter.cpp
```diff
@@ -228,7 +261,22 @@ void ArrowWriter::ready(PointTableRef table)
...
+                kvMetadata->Append("ARROW:extension:name", "WKB");   
+            }
 
+            NL::json dimDetail;
+            dimDetail["name"] = layout->dimName(id);
+            dimDetail["description"] = pdal::Dimension::description(id);
+            dimDetail["interpretation"] = pdal::Dimension::interpretationName(layout->dimType(id));
+            dimDetail["size"] = layout->dimSize(id);
+            kvMetadata->Append("PDAL:dimension:metadata", dimDetail.dump(-1));   
+
```
paleolimbot, line 278: If you are writing extension names (`"geoarrow.wkb"` or `"geoarrow.point"`), you probably also want to write extension metadata:

```
kvMetadata->Append("ARROW:extension:metadata", "{\"crs\": <crs as projjson>}");
```

(I'm not sure exactly where the discussion landed on whether you do or don't want to do that...I think it's a good idea but no pressure either way!).

paleolimbot (commented):
> I wonder what this writer should be doing for batch stats (like rouault has added for GDAL).

If you're referring to column statistics in Parquet, I think you probably do want to do this unless there is a severe performance consequence to doing so (I don't think this is usually the case; pyarrow and Arrow/R both do this by default). If you are using the StructArray (i.e., "geoarrow.point"), you will get bounding boxes embedded in the column stats (which might not be useful, but also could be very useful if a user has this set up or is writing pre-chunked clouds).

> this needs further discussion on broad interoperability vs raw performance.

I opened https://github.com/opengeospatial/geoparquet/issues/185 to officially propose GeoArrow in GeoParquet. 

If I had to advise you to do something *today*, it would be to always write extension name (`"ARROW:extension:name"` is `"geoarrow.wkb"` or `"geoarrow.point"`) and extension metadata (`"ARROW:extension:metadata"` with the CRS key) and if writing `"geoarrow.point"`, *don't* write GeoParquet metadata (yet) since that's not a valid encoding (yet). You could also write `"encoding": "geoarrow.point"` in the GeoParquet metadata and issue a warning that it's not part of the current spec.

plugins/arrow/io/ArrowWriter.cpp
hobu: yep, we got that https://github.com/PDAL/PDAL/pull/4115/files/af854282caf2a643a068bf14a1c41c320a0fcacc..73e74c0bf6dd4d1a938e04d20f8e051344a35fb0#diff-2d62ed8713995e770bac59974177b9adcc1272de741fecd3716d04e004425012R242

plugins/arrow/io/ArrowWriter.cpp
hobu: ok, fixed.

plugins/arrow/io/ArrowReader.cpp
```diff
@@ -0,0 +1,625 @@
...
+            NL::json columns = metadata["columns"];
+
+            std::string primary_column = metadata["primary_column"];
+            NL::json column = metadata["columns"][primary_column];
+
+            log()->get(LogLevel::Info) << "primary column is " << primary_column << std::endl;
+
+            if (!column.contains("crs"))
+            {
+                log()->get(LogLevel::Warning) << "no 'crs' key available to fetch spatial reference information" << std::endl;
```
paleolimbot, line 190: I'm not sure it would ever be relevant for point clouds, but I'm pretty sure the spec says a missing `"crs"` key means `"OGC:CRS84"`.

plugins/arrow/io/ArrowReader.cpp
```diff
@@ -0,0 +1,625 @@
...
+            if (crs.is_object())
+            {
+                ref.set(crs.dump());
+            }
+            else if (crs.is_string())
+            {
+
+                ref.set(column["crs"].get<std::string>());
+            }
+
```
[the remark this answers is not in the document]
hobu: If there's no value in the `crs` column, it will be an empty SpatialReference.

PR #4116, opened 2023-07-21
Add TileDBWriter option to compress sub-byte fields 

jp-dark:
…ribute

### DIFF
--- a/cmake/tiledb.cmake
+++ b/cmake/tiledb.cmake
@@ -1,7 +1,7 @@
 #
 # TileDB support
 #
-find_package(TileDB QUIET 2.3.0 REQUIRED)
+find_package(TileDB QUIET 2.15.0 REQUIRED)
 
 set_package_properties(TileDB PROPERTIES
         TYPE OPTIONAL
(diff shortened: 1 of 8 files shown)


hobu:
`pdal_io_tiledb_bit_fields_test` is failing during the locale tests. You will either need to disable them (see linux.yaml) or make the TileDB stuff locale aware/safe.

jp-dark:
> `pdal_io_tiledb_bit_fields_test` is failing during the locale tests. You will either need to disable them (see linux.yaml) or make the TileDB stuff locale aware/safe.

I think it's actual an issue with the TileDB version conda is grabbing. The TileDB version is different in the different tests, and there is a bug that was fixed since the older version (2.13) that I think fixes the failing tests. I am going to close this issue for now, and look into updating conda first.

hobu:
>  The TileDB version is different in the different tests, and there is a bug that was fixed since the older version (2.13) that I think fixes the failing tests.

Is it possible to harden the drivers so they are resilient to different versions of TileDB? If things become too sensitive we will simply shut the tests off and stuff will begin to rot.

PR #4117, opened 2023-07-22
Add WKTv2 VLRs to COPC

jjimenezshaw:
Continuation of #3998 , same behaviour in COPC as in LAS/LAZ

It writes WKT2 and PROJJSON vlrs when set with `enhanced_srs_vlrs` in CopcWriter
Reads the SpatialReference from the vlr defined in the option `srs_consume_preference`, which defaults to `wkt1, wkt2, projjson`.

### DIFF
--- a/doc/stages/readers.copc.rst
+++ b/doc/stages/readers.copc.rst
@@ -130,3 +130,6 @@ keep_alive
 fix_dims
   Make invalid dimension names valid by converting disallowed characters to '_'. Only
   applies to names specified in an extra-bytes VLR. [Default: true]
+
+srs_consume_preference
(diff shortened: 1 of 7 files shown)


jjimenezshaw:
I could not test it (there is not CopcWriter test, and therefore no way to test the reader). I created the PR in case you see anything clearly wrong.

hobu:
Please add the docs to the `readers.copc` and `writers.copc` pages.

jjimenezshaw:
Hi. I added the documentation requested (and more tests). But now there is a problem with the GitHub Action for Conda.
Looks like it is failing for many branches:
https://github.com/PDAL/PDAL/actions/workflows/conda.yml
Is that an infrastructure problem?
Rebasing onto master will fix it?

PR #4134, opened 2023-08-14
Add bit field packing to TileDB reader/writer

jp-dark:


hobu:
> Exclude bit field test from locale tests

It would be nice to not have to do this, but I don't know how involved it would be. I also don't think anyone particularly cares except that it is something that has potential to rot.

jp-dark:
> > Exclude bit field test from locale tests
> 
> It would be nice to not have to do this, but I don't know how involved it would be. I also don't think anyone particularly cares except that it is something that has potential to rot.

I 100% agree. We have a task to look into locales in the TileDB backlog that has not yet been addressed.

PR #4150, opened 2023-08-29
add filters.stats.commonsrs to allow user to override the default 'co…

hobu:
…mmon' srs when writing box metadata


hobu:
If you were doing lidar on the moon, you might change the `commonsrs` from `EPSG:4326` to `IAU:30100'` like this:

```
pdal info moon-lidar.copc.laz --all --filters.stats.commonsrs="IAU:30100"
```

PR #4154, opened 2023-09-06
Enhancements/fixes related to coordinate epochs

### LINKED ISSUE #3995: Proposal: add in_coord_epoch and out_coord_epoch optional argments in filters.reprojection for time-based coordinate transformations

rouault:
Complements PR #4049, closes #3995 . To be fully effective, it requires https://github.com/OSGeo/PROJ/pull/3884 and https://github.com/OSGeo/gdal/pull/8340, but it can be applied independently as it doesn't use new API

-   SpatialReference::set(): store coordinate epoch from OGRSpatialReference built from SetFromUserInput()
    
    This is useful to get the epoch for a SpatialReference built from a
    string like 'urn:ogc:def:coordinateMetadata:AUTHORTIY::CODE code' (since
    https://github.com/OSGeo/gdal/pull/8340 and https://github.com/OSGeo/PROJ/pull/3884), which
    resolves to a CoordinateMetadata object with a CRS and potentially a
    coordinate epoch

-  SrsTransform: make sure to propagate coordinate epoch when rebuilding OGRSpatialReference objects

-  SpatialReference::set(): when building from SetFromUserInput(), get the WKT2 version
    
    This limits the loss of information, particularly for CRS like
    'urn:ogc:def:coordinateMetadata:NRCAN::NAD83_CSRS_1997_MTM7_HT2_1997'
    which include a GEOIDMODEL[] node, that would be lost otherwise

With that PR and above mentionned GDAL and PROJ PRs,
```
$ cat in.csv
X,Y,Z
268955,5540412,0

$ cat pipeline.json
[
    {
        "type":"readers.text",
        "filename":"in.csv"
    },
    {
        "type":"filters.reprojection",
        "in_srs":"urn:ogc:def:coordinateMetadata:NRCAN::NAD83_CSRS_1997_MTM7_HT2_1997",
        "out_srs":"urn:ogc:def:coordinateMetadata:NRCAN::NAD83_CSRS_2010_UTM19_CGVD2013_2010"
    },
    {
        "type":"writers.text",
        "filename":"outputfile.txt"
    }
]

$ PROJ_NETWORK=ON bin/pdal pipeline pipeline.json

$ cat outputfile.txt 
"X","Y","Z"
356670.104,5540546.584,-0.306
```

hence we get the same results as with gdaltransform and cs2cs with similar invokations

### DIFF
--- a/pdal/SpatialReference.cpp
+++ b/pdal/SpatialReference.cpp
@@ -237,6 +237,8 @@ void SpatialReference::set(std::string v)
         throw pdal_error(oss.str());
     }
 
+    m_epoch = srs.GetCoordinateEpoch();
+
     m_wkt = exportToWkt(&srs);
 }
 
--- a/pdal/SpatialReference.cpp
+++ b/pdal/SpatialReference.cpp
@@ -208,16 +208,26 @@ void SpatialReference::parse(const std::string& s, std::string::size_type& pos)
 
 void SpatialReference::set(std::string v)
 {
+    m_wkt.clear();
+    m_wkt2.clear();
     m_horizontalWkt.clear();
     if (v.empty())
     {
-        m_wkt.clear();
         return;
     }
 
-    if (isWKT(v))
+    if (isWKT2(v))
     {
         m_wkt = v;
+        m_wkt2 = v;
+        return;
+    }
+    else if (isWKT1(v))
+    {
+        m_wkt = v;
+        OGRScopedSpatialReference srs = ogrCreateSrs(m_wkt);
+        if (srs)
+            m_wkt2 = exportToWkt(srs.get(), {"FORMAT=WKT2_2018"});
         return;
     }
 
@@ -240,6 +250,8 @@ void SpatialReference::set(std::string v)
     m_epoch = srs.GetCoordinateEpoch();
 
     m_wkt = exportToWkt(&srs);
+
+    m_wkt2 = exportToWkt(&srs, {"FORMAT=WKT2_2018"});
 }
 
 
@@ -483,13 +495,9 @@ SpatialReference SpatialReference::wgs84FromZone(int zone)
 }
 
 
-bool SpatialReference::isWKT(const std::string& wkt)
+bool SpatialReference::isWKT2(const std::string& wkt)
 {
-    // List comes from GDAL.  WKT includes FITTED_CS, but this isn't
-    // included in GDAL list.  Not sure why.
-    StringList leaders { "PROJCS", "GEOGCS", "COMPD_CS", "GEOCCS",
-        "VERT_CS", "LOCAL_CS",
-    // New specification names.
+    StringList leaders {
         "GEODCRS", "GEODETICCRS",
         "PROJCRS", "PROJECTEDCRS",
         "VERTCRS", "VERITCALCRS",
@@ -505,6 +513,27 @@ bool SpatialReference::isWKT(const std::string& wkt)
         if (wkt.compare(0, s.size(), s) == 0)
             return true;
     return false;
+}
+
+
+bool SpatialReference::isWKT1(const std::string& wkt)
+{
+    // List comes from GDAL.  WKT includes FITTED_CS, but this isn't
+    // included in GDAL list.  Not sure why.
+    StringList leaders { "PROJCS", "GEOGCS", "COMPD_CS", "GEOCCS",
+        "VERT_CS", "LOCAL_CS"
+    };
+
+    for (const std::string& s : leaders)
+        if (wkt.compare(0, s.size(), s) == 0)
+            return true;
+    return false;
+}
+
+
+bool SpatialReference::isWKT(const std::string& wkt)
+{
+    return isWKT1(wkt) || isWKT2(wkt);
 }
 
 
@@ -535,9 +564,7 @@ std::string SpatialReference::getWKT1() const
 
 std::string SpatialReference::getWKT2() const
 {
-    std::string wkt = getWKT();
-    OGRScopedSpatialReference srs = ogrCreateSrs(wkt);
-    return exportToWkt(srs.get(), {"FORMAT=WKT2_2018"});
+    return m_wkt2;
 }
 
 int SpatialReference::getUTMZone() const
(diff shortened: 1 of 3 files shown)


pdal/SpatialReference.cpp
```diff
@@ -237,7 +237,9 @@ void SpatialReference::set(std::string v)
         throw pdal_error(oss.str());
     }
 
-    m_wkt = exportToWkt(&srs);
+    m_epoch = srs.GetCoordinateEpoch();
+
+    m_wkt = exportToWkt(&srs, {"FORMAT=WKT2_2018"});
```
hobu, line 242: Is `WKT2_2018` the default format for WKT export from GDAL right now already? What flavor would you get if you didn't request this?

hobu (approved):

hobu:
The one linux failure is something I'm working on fixing...

pdal/SpatialReference.cpp
rouault: > Is `WKT2_2018` the default format for WKT export from GDAL right now already?

no, it is still WKT1_GDAL

hobu: Hmm, that might break some stuff that expects WKT1.

rouault: > Hmm, that might break some stuff that expects WKT1.

shouldn't they use the getWKT1() method then ?

I could potentially limit the scope of that change to situations where I detect a GEOID_MODEL in the WKT, but that would be a bit ugly

hobu: I think I like this better. That way WKT2+ is opt-in and we don't disrupt anyone who might have been assuming WKT1

rouault: > I think I like this better. That way WKT2+ is opt-in and we don't disrupt anyone who might have been assuming WKT1

done in the last 3 commits

pdal/SpatialReference.hpp
```diff
@@ -182,12 +182,15 @@ class PDAL_DLL SpatialReference
 
 private:
     std::string m_wkt;
+    mutable std::string m_wkt2;
```
hobu, line 185: I don't understand why this is `mutable` and *m_wkt* isn't.

hobu (approved):

pdal/SpatialReference.hpp
rouault: > I don't understand why this is `mutable` and _m_wkt_ isn't.

m_wkt is always set in the set() method, whereas m_wkt2 is only set in set() if the WKT is WKT2. Otherwise it is computed on demand in getWKT2() which is const, hence the mutable.
The alternative would be that I always compute m_wkt2, and thus getWKT2() would just return it

hobu: I think I would rather always compute it in set() so it is clear what's happening, instead of having the scary `mutable` without the comment to tell us why it is mutable.

rouault: > I think I would rather always compute it in set() so it is clear what's happening, instead of having the scary `mutable` without the comment to tell us why it is mutable.

done

Issue #4157, opened 2023-09-07
[PCD Reader] incorrectly assumes that all float fields (F) are single precision


karlHolmquist:
Hi thanks for the great repo, I recently encountered a bug which most likely is not the most common one.
The current pcd reader cannot handle double fields in pcd fields which leads to missaligned reads of the point data.

The files I'm working with has a per point time field of double precision (see header below). However, PCL uses F as type for any floating type and specifies the precision further using the SIZE field.

From the pcd reader it seems that in all cases the field is interpreted as single precision which would lead to missalignment when reading half or double precision data.

The same behaviour would probably occur for signed/unsigned integer fields which are not 32bits due to the lines below: 
https://github.com/PDAL/PDAL/blob/54d017f5c11d6011c55399b88fb63de76ccd93d1/io/PcdReader.cpp#L103
https://github.com/PDAL/PDAL/blob/54d017f5c11d6011c55399b88fb63de76ccd93d1/io/PcdReader.cpp#L199

`# .PCD v0.7 - Point Cloud Data file format
VERSION 0.7
FIELDS x y z intensity ring time range scan_number
SIZE 4 4 4 4 1 8 4 2
TYPE F F F F U F U U
COUNT 1 1 1 1 1 1 1 1
WIDTH 5212658
HEIGHT 1
VIEWPOINT 0 0 0 1 0 0 0
POINTS 5212658
DATA binary
`

It would be great if the reader could support all precisions but an alternative is to validate the SIZE of each type and skip the ones that are not 4 bytes long.

abellgithub:
Thanks for the report. This seems obviously wrong and easy to fix.

hobu:
Is the pre-formatted text you posed a PCD file we can use to cook a test?

karlHolmquist:
Great to hear,
Unfortunately it's just the header of one and I cannot share the full one but here is another one.
It's a unit sphere with the polar coordinates as double and uint16 field with one for for the points where x, y and z are all positive.

[sphere.pcd.txt](https://github.com/PDAL/PDAL/files/12550371/sphere.pcd.txt)

hobu:
After doing 
```
pip install wind-pypcd
```

Here's a script that makes one of our test files:

```
data = "https://github.com/PDAL/PDAL/blob/master/test/data/autzen/autzen-utm.las?raw=true"


json = """
    {
      "pipeline": [
        {
            "type": "readers.las",
            "filename": "%s"
        },
        {
[...]

PR #4158, opened 2023-09-07
PCD reading of 64bit values does not work

hobu:
See #4157


abellgithub (commented):
This is fine, but that code also screams template.

hobu:
> This is fine, but that code also screams template.

I scream when I see templates. When I look at this I know exactly what's going on and what's handled.

PR #4166, opened 2023-09-11
Fix LasReader to not fetch whole file when only metadata is requested

### LINKED ISSUE #4165: readers.las always fetches the entire file, even if no points are requested
**Describe the bug**
Using readers.las with points=0 always fetches the entire file.
Eg the following command:
`pdal info https://some-large-las-file --driver readers.las --readers.las.count=0`
Or as a pipeline:
```
[
  {
    "type": "readers.las",
    "filename": "s3://some-large-las-file",
    "count": 0
  },
  {
    "type": "filters.info"
  }
]
```
Or by using the flag `--metadata` which only has output that doesn't depend on the points and requests no points:
`pdal info s3://some-large-las-file --driver readers.las --metadata`

**Expected behavior**
Ideally, this would only fetch the headers of the LAS file when no points are required, which could be 1000 times smaller.

```
$ uname -a
Darwin Kernel Version 22.6.0: Wed Jul  5 22:21:56 PDT 2023; root:xnu-8796.141.3~6/RELEASE_X86_64 x86_64
```

```
$ pdal --version
pdal 2.6.0
```

Bigger picture:
There may be other readers that fetch the entire file file regardless.
There may be other operations that would ideally fetch just the headers, or just the headers plus some sample of points.

olsen232:
Fixes #4165

This does fix the issue (as written up by myself) but is not necessarily complete - I haven't written any tests, and the reviewer may have opinions about this issue that require further changes - eg there may be other scenarios where doing range requests is better than fetching the whole file, that can be tacked on.

I'm happy to do those things - to address any comments to get it up to PDAL's standard - I'm just starting the discussion here, I don't want to do too much more work until I hear "sounds like a plan" (as opposed to "not something we're interested in") from PDAL team.

### DIFF
--- a/io/LasReader.cpp
+++ b/io/LasReader.cpp
@@ -36,6 +36,7 @@
 
 #include "LasHeader.hpp"
 #include "LasReader.hpp"
+#include "private/connector/Connector.hpp"
 #include "private/las/Header.hpp"
 #include "private/las/Srs.hpp"
 #include "private/las/Utils.hpp"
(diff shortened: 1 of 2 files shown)


io/LasReader.cpp
```diff
@@ -178,17 +238,38 @@ QuickInfo LasReader::inspect()
...
 
     return qi;
 }
 
 
 void LasReader::createStream()
 {
-    if (!m_streamIf)
-        m_streamIf.reset(new LasStreamIf(m_filename));
+    if (m_count == 0)
```
olsen232, line 249: This if-block here means that doing `pdal info --metadata s3://some-large-las-file` will also do range requests - if this if-block is removed then pdal info / pdal pipeline will always fetch the entire file, and only calling `QuickInfo LasReader::inspect()` will do the range requests. All of the other changes to this file are just there for the range requests codepath to work in general, whether that's used only for calls to LasReader::inspect or used more broadly.

olsen232:
Superceded by https://github.com/PDAL/PDAL/pull/4178

PR #4176, opened 2023-09-18
Stac info kernel

kylemann16:
Add `--stac` as an option for pdal info calls. This outputs a `stac` key in the console output that is a valid STAC Feature that includes the pointcloud and projection extension information.


kernels/InfoKernel.cpp
```diff
@@ -222,6 +239,20 @@ void InfoKernel::makePipeline()
...
+
+    if (m_stac)
+    {
+        // filters required for stac: metadata, stats,
+        if (stage != m_statsStage)
+        {
+            Options stacOps;
+            if (m_enumerate.size())
+                stacOps.add({"enumerate", m_enumerate});
+            m_stacStage = stage = &m_manager.makeFilter("filters.stats", *stage, stacOps);
```
hobu, line 251: It is possible we already have a `filters.stats` added to the pipeline at this point if we're doing `--all` as suggested above.

hobu (approved):

kernels/InfoKernel.cpp
kylemann16: Yeah, this should account for that with the stage boolean check above. Could definitely make it more obvious by just saying `if (m_showStats)` though.

connormanning (approved):

PR #4178, opened 2023-09-20
Modify pdal info --summary to not fetch the entire file...

### LINKED ISSUE #4165: readers.las always fetches the entire file, even if no points are requested

olsen232:
...but only when --driver=readers.las
Fixes #4165

Again, this is a proof-of-concept - this can be improved if the overall approach is deemed acceptable


hobu:
Conflicted about this. It fattens up the InfoKernel for just the LAS reader. Maybe we could just bury it its own single function/file so we don't have to think about it 😄

olsen232:
I can put it in a separate file where you don't have to look at it if that will get it over the line. Otherwise... well, you could see if you like the earlier attempt, #4166, any more by comparison. It's a similar amount of code - perhaps a bit more boilerplate in the form of interfaces and headers - but it keeps LAS specific code in the LAS reader, if that's a concern.

hobu:
Please put this stuff into its own file in `./kernels/private` so I don't have to look at it and we will merge it.

olsen232:
Moved into kernels/private/PointlessLas.cpp

olsen232:
(Ran astyle and force-pushed the astyle changes)

PR #4179, opened 2023-09-20
[WIP] Add support for GDAL 4.0 RFC95

rouault:
Changes needed to get PDAL to build against GDAL 4.0 RFC 95 : https://github.com/OSGeo/gdal/pull/8399  (not approve yet nor committed)


hobu:
should this be guarded with a GDAL version?

rouault:
> should this be guarded with a GDAL version?

this won't have any effect on GDAL < 3.8, so not strictly necessary

PR #4099, opened 2023-07-10
address scaling issues demonstrated in #4088 

hobu:
`writers.copc` potentially has the same issue described in #4088. This should address #4092

PR #4180, opened 2023-09-20
writers.copc: round X/Y/Z values to match writers.las

rcoup:
The COPC writer truncates scaled & offset coordinate values to `int32` during packing. This didn't match the LAS/LAZ writer which rounds them, which meant that point coordinates could needlessly change during conversion from LAS/LAZ to COPC.

Fix this by applying the same rounding approach that is used for LAS/LAZ. The approach [`pdal::LasWriter::fillPointBuf()`](https://github.com/PDAL/PDAL/blob/810b87646bf88e3c5250e86d5e501efa7bedee5c/io/LasWriter.cpp#L886-L907) uses is implemented/the same as for [`las::V10BaseLoader::pack()`](https://github.com/PDAL/PDAL/blob/810b87646bf88e3c5250e86d5e501efa7bedee5c/io/private/las/Utils.cpp#L544-L570), but not `las::V14BaseLoader::pack()`, which COPC uses.

And add a test case for good measure.

(I don't really understand why `LasWriter` has it's own implementation separate from the `las::*` stuff? But that's out of scope for here)

Before:
```
LAStools (by info@rapidlasso.de) version 221128
pdal 2.6.0 (git-version: 810b87)

Querying clip.las via lasinfo64...
  Time             X           Y           Z      RawX      RawY      RawZ  PointSourceID
  316404070.712554 1406018.497 4917487.174 62.276 406018497 -82512826 62276 1303
Querying clip.las via pdal info...
   GpsTime,    X,          Y,          Z
  [316404070.7,1406018.497,4917487.174,62.276]
Comparing...
  no differences

Converting clip.las to clip.copc.laz...

Converting clip.las to clip2.las...

Querying clip.copc.laz via lasinfo64...
  Time             X           Y           Z      RawX      RawY      RawZ  PointSourceID
  316404070.712554 1406018.496 4917487.174 62.276 406018496 -82512826 62276 1303
Querying clip.copc.laz via pdal info...
   GpsTime,    X,          Y,          Z
  [316404070.7,1406018.496,4917487.174,62.276]
Comparing...
  -316404070.712554 1406018.497 4917487.174 62.276 406018497 -82512826 62276 1303
  +316404070.712554 1406018.496 4917487.174 62.276 406018496 -82512826 62276 1303
  -[316404070.7,1406018.497,4917487.174,62.276]
  +[316404070.7,1406018.496,4917487.174,62.276]

Querying clip2.las via lasinfo64...
  Time             X           Y           Z      RawX      RawY      RawZ  PointSourceID
  316404070.712554 1406018.497 4917487.174 62.276 406018497 -82512826 62276 1303
Querying clip2.las via pdal info...
   GpsTime,    X,          Y,          Z
  [316404070.7,1406018.497,4917487.174,62.276]
Comparing...
  no differences
```
</details>

### DIFF
--- a/io/private/las/Utils.cpp
+++ b/io/private/las/Utils.cpp
@@ -682,11 +682,27 @@ void V14BaseLoader::load(PointRef& point, const char *buf, int bufsize)
 void V14BaseLoader::pack(const PointRef& point, char *buf, int bufsize)
 {
     LeInserter ostream(buf, bufsize);
-    int32_t xi = (int32_t)m_scaling.m_xXform.toScaled(point.getFieldAs<double>(Dimension::Id::X));
-    int32_t yi = (int32_t)m_scaling.m_yXform.toScaled(point.getFieldAs<double>(Dimension::Id::Y));
-    int32_t zi = (int32_t)m_scaling.m_zXform.toScaled(point.getFieldAs<double>(Dimension::Id::Z));
 
-    ostream << xi << yi << zi;
+    auto converter = [](double val, Dimension::Id dim) -> int32_t
+    {
+        int32_t i(0);
+
+        if (!Utils::numericCast(val, i))
+            throw std::runtime_error("Unable to convert scaled value (" +
+                Utils::toString(val) + ") to "
+                "int32 for dimension '" + Dimension::name(dim) );
+        return i;
+    };
+
+    double xOrig = point.getFieldAs<double>(Dimension::Id::X);
+    double yOrig = point.getFieldAs<double>(Dimension::Id::Y);
+    double zOrig = point.getFieldAs<double>(Dimension::Id::Z);
+
+    int32_t x = converter(m_scaling.m_xXform.toScaled(xOrig), Dimension::Id::X);
+    int32_t y = converter(m_scaling.m_yXform.toScaled(yOrig), Dimension::Id::Y);
+    int32_t z = converter(m_scaling.m_zXform.toScaled(zOrig), Dimension::Id::Z);
+
+    ostream << x << y << z;
 
     uint16_t intensity = point.getFieldAs<uint16_t>(Dimension::Id::Intensity);
     int returnNum = point.getFieldAs<int>(Dimension::Id::ReturnNumber);
--- a/test/unit/io/CopcWriterTest.cpp
+++ b/test/unit/io/CopcWriterTest.cpp
@@ -36,6 +36,8 @@
 
 #include <pdal/pdal_test_main.hpp>
 
+#include <pdal/util/FileUtils.hpp>
+#include <io/BufferReader.hpp>
 #include <io/CopcReader.hpp>
 #include <io/CopcWriter.hpp>
 #include <io/LasReader.hpp>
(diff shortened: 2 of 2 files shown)


hobu:
We just danced this in #4099

rcoup:
> We just danced this in https://github.com/PDAL/PDAL/pull/4099

Ah, that's why they don't match. But `V10BaseLoader` isn't used afaics 😄

rcoup:
Anything else I need to do here while I have it front-of-mind?

TBC, the changes in #4099 don't solve the issue because they only modified `V10BaseLoader`. COPC uses `V14BaseLoader` since it's all LAS1.4, PRDF>=6; and `V14BaseLoader` doesn't inherit from `V10BaseLoader`

io/private/las/Utils.cpp
```diff
@@ -682,11 +682,27 @@ void V14BaseLoader::load(PointRef& point, const char *buf, int bufsize)
 void V14BaseLoader::pack(const PointRef& point, char *buf, int bufsize)
 {
     LeInserter ostream(buf, bufsize);
-    int32_t xi = (int32_t)m_scaling.m_xXform.toScaled(point.getFieldAs<double>(Dimension::Id::X));
-    int32_t yi = (int32_t)m_scaling.m_yXform.toScaled(point.getFieldAs<double>(Dimension::Id::Y));
-    int32_t zi = (int32_t)m_scaling.m_zXform.toScaled(point.getFieldAs<double>(Dimension::Id::Z));
 
-    ostream << xi << yi << zi;
+    auto converter = [](double val, Dimension::Id dim) -> int32_t
```
abellgithub, line 686: I don't understand this change.

The change in #4088 was for a bad header value. The stored point value was correct. Can you explain?

hobu:
Reverted in 00b93d8bd283f4ded77ec8924ea73fe1b233b33b. I want to make sure we need this.

hobu:
Unreverted in db396008d6293c9f7cc55634a4059563cc8054d8 😄 

Thanks for the contribution @rcoup

io/private/las/Utils.cpp
rcoup: If reading a LAS file, it gets decoded from an integer to float via `offset + scale * (double)encodedValue`. When writing we do `(coordinate - offset) / scale`, and we should get the same encoded integer value back as we started with. But because the code truncates instead of rounds 

Currently it casts, but that means, for example:

```
offset = 1000000
scale = 0.001
encoded coordinate = 406018497

coordinate = 1000000 + 0.001 * (double)406018497 == 1406018.497

encoded = (1406018.497 - 1000000) / 0.001 == 406018496.99999994

int(406018496.99999994) == 406018496
round(406018496.99999994) == 406018497
```

Impact? `pdal translate input.las output.copc.laz` ends up with different encoded coordinate values. And when they're subsequently decoded the coordinate has moved.

rcoup: Implementation-wise, I would have used `std::round()` but the code already in V10BaseLoader (and missing from V14BaseLoader) achieved the same thing.

PR #4184, opened 2023-09-21
Terminate when input is smaller than requested output size or radius has grown too small

### LINKED ISSUE #4119: Infinitely-running relaxationdartthrowing filter
Hello, I am having issues with the relaxationdartthrowing filter used as below :

```
$ pdal translate in.laz out.laz relaxationdartthrowing <options>
```

I would expect the filter to throw errors or otherwise adapt to the following situations :

- the "count" parameter is more or equal to the starting number of points in in.laz
- several points are very close to eachother/have the same position

Instead, the current behaviour as far as I can see in those situations is that the filter simply gets stuck and runs "infinitely" (or at least, more than 10 minutes even for very small point clouds of <200 points).

This also happens is other cases when the parameter "count" is close but still strictly smaller than the number of points in in.laz, even after running the "sample" filter to remove very close points. However, I was not able to figure out the cause, nor a reason for which threshold for "count" would completely solve the issue.

```
$ uname -a
Linux L1AAS87 5.19.0-46-generic #47~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Jun 21 15:35:31 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

```
$ pdal --version
pdal 2.5.5 (git-version: 9d28a2)
```
(installed with conda)

chambbj:
Closes #4119


chambbj:
Will add some unit tests prior to merging, but this should do it.

Issue #4094, opened 2023-06-30
class flags are getting set when they shouldn't when converting text to las


kjwaters:
When reading a text file (readers.text) that contains points with classes over 31 and then writing the file to an LAS 1.4 record type 6 file, the class flags are getting set as if the class bytes were from LAS 1.2.

```
$ pdal pipeline pipe
```

input file points.txt:

```
X,Y,Z,Classification,Time
50000,10032,0,1,34.3
50001,10033,0,32,34.4
50002,10034,0,64,34.5
```

file pipe  

```
[
{
        "type":"readers.text",
        "filename":"points.txt",
        "default_srs":"EPSG:3717+5703"
},
{
        "type":"writers.las",
        "minor_version":4,
        "dataformat_id":6,
        "filename":"points_pdal.las",
        "forward":"all"
}
]


```

**Expected behavior**
pdal info should show classflags as all zeros. Instead it shows this:
```
{
        "average": 1,
        "count": 3,
        "maximum": 2,
        "minimum": 0,
        "name": "ClassFlags",
        "position": 14,
        "stddev": 1,
        "variance": 1
      }
```

**System/installation information:**

```
pdal 2.5.5 (git-version: 9d28a2)
```

kjwaters:
clearly messed up the formatting on this one. Sorry.

abellgithub:
I don't understand why you think class flags should be zero.

kjwaters:
They should be zero because I didn't set any flags. Zero would be no flags
set. If I set the class to 32, that should be class 32 without setting any
of the classification flags (synthetic, key-point, withheld, or overlap).
Instead, it sets class 32 and synthetic. I suspect classes 128 to 191 would
always be marked as withheld, which would not be intended. I haven't
checked to see if there is a way to set the classification flags from a
text file, but it shouldn't be based on the class number.

Kirk

abellgithub:
This has all become very complex. When receiving input from a text reader, there's no way to know if classification should be interpreted as it would be for LAS 1.3 or LAS 1.4. If the input is 1.3, people often want the flags embedded in the classification to be set in the classFlags for output. I blame this on the way that LAS changed between 1.3 and 1.4, but that can't be changed. Changes have been made to the associated logic to satisfy various demands over time, but I don't believe that we can make everyone's use case work without some sort of user-specified option.

Someone else can chime in WRT their take on the behavior.

LFraser:
[...]

Issue #4000, opened 2023-03-10
Better support for ClassificationFlags


wonder-sk:
The ClassificationFlags dimension has a bitwise combination of flags (Synthetic, Key-Point, Withheld, Overlap).

Working with these flags is somewhat complicated:
- it is pretty much undocumented how to access those flags (user needs to consult LAS spec and understand bitwise operations)
- it is hard to use expressions to filter them correctly due to missing bitwise ops (e.g. `ClassificationFlags == 2 || Classification == 3 || Classification == 6 || Classification == 10` for key point flag)

It would be useful to have some more direct way of working with them when needed - two solutions that come to mind:
1. have a new parameter in `las.reader` that would add extra true/false dimension(s) for these flags - at the expense of having each point using more memory
2. have "virtual" dimensions - with no extra memory usage. When a value from `KeyPoint` dimension would be requested, PDAL would internally do the bitwise check on ClassificationFlags and return 1/0

abellgithub:
Maybe the easiest thing from an implementation standpoint would be a special operator for the assignment syntax. Maybe something like "dimension[start bit, cnt = 1] to get a value from a bit. You could also use standard C bitwise operations, but currently, all the math currently happens as double-precision floating point, so it would be a bunch of work to do something where you have to know types.

I think maybe this would solve most problems. If you have a 1.3 file and you want to merge it with a 1.4 file, you could do this in an assign filter before merging with 1.4 data:

"value" : "ClassFlags = Classification[5, 3]" // Extract the flags bits
"value" : "Classification = Classification[0, 5]" // Strip the flags bits

Or if you just want to ignore synthetic points at some point, you could extract that bit:
"value" : "Synthetic = Classification[5]"

hobu:
Fixed (hopefully for good) in #4186

Issue #3092, opened 2020-05-23
Las 1.4 readers.las -> writers.las corrupts Classification


jrobichaud:
As requested by @abellgithub this is a simpler way to reproduce #3091 with a single point and a single pipeline.

**Describe the bug**

Download this point cloud: 

https://github.com/jrobichaud/pdal-las-bug/blob/master/output/bug_single_point/text_to_las.las

It has a single point to 0,0,0 with classification 71.

pipeline.json:
```json
{
  "pipeline": [
    {
      "type": "readers.las",
      "filename": "text_to_las.las"
    },
    {
      "type": "writers.las",
      "minor_version": "4",
      "extra_dims": "all",
      "compression": "laszip",
      "filename": "las_to_las.las"
    }
  ]
}
```

For the bug to happen it requires `"minor_version": "4"` and `"compression": "laszip"`.

```bash
$ pdal pipeline pipeline.json
```

Then if you run:
`pdal translate text_to_las.las text_to_las_translated_to_txt.txt`
`pdal translate las_to_las.las las_to_las_translated_to_txt.txt`
`diff text_to_las_translated_to_txt.txt las_to_las_translated_to_txt.txt`

You get this diff:

```diff
2c2
< 0.000,0.000,0.000,0.000,1.000,1.000,0.000,0.000,71.000,0.000,0.000,0.000,0.000,0.000,0.000,0.000,0.000,0.000
---
> 0.000,0.000,0.000,0.000,1.000,1.000,0.000,0.000,7.000,0.000,0.000,0.000,0.000,0.000,0.000,0.000,0.000,0.000

```

You can notice the classification change from 71 to 7. It will change any classification to  `classification % 32`. Ex: 255 -> 31. See table below.

**Expected behavior**
The classification should be preserved

| Input Classification | Expected Classification | Result Classification | Works |
|----|----|----|---|
| 0 | 0 | 0 | :white_check_mark: |
| 1 | 1 | 1 | :white_check_mark: |
| 2 | 2 | 2 | :white_check_mark: |
| 3 | 3 | 3 | :white_check_mark: |
| 4 | 4 | 4 | :white_check_mark: |
| 5 | 5 | 5 | :white_check_mark: |
| 6 | 6 | 6 | :white_check_mark: |
| 7 | 7 | 7 | :white_check_mark: |
| 9 | 9 | 9 | :white_check_mark: |
| 10 | 10 | 10 | :white_check_mark: |
[...]

PR #4186, opened 2023-09-22
Split out ClassFlags into its bitfields

connormanning:
Support for the bitfields associated with Classification, which are Synthetic/KeyPoint/Withheld/Overlap, have historically been a sticking point in PDAL: #4000, #4094, #1701, #3092, #1063, #3878.  Their behavior has been tweaked [several times](https://github.com/PDAL/PDAL/issues/1701#issuecomment-1280157073) and their initial implementation predicts the need to [change them in the future](https://github.com/PDAL/PDAL/issues/1063#issue-120887388).

Depending on what LAS version data is coming from, the behavior is different - from legacy PDRFs 0-5 you end up getting "incorrect" classifications because the flags are included in the classification value, and with the modern ones they are ClassFlags.  ClassFlags can also be awkward to work with since PDAL not having a first-class concept of handling bits.

As an abstraction library, it would be reasonable to abstract this away - giving the flags their own fields and *always* keeping them separate from Classification: logically the flags are independent from Classification\* (an exception to this is discussed below) and the fact that they may be bit-packed together in LAS is really an implementation detail.

This PR does the following:
- Add dimensions `Synthetic`, `KeyPoint`, `Withheld`, and `Overlap`
- Make `Classification` *never* include these flags in its value
- Use these new dimensions in the LAS reader/writer, removing the conditional in the LAS writer which handled the existence of the ClassFlags dimension differently
- Add the concept of "deprecation" for dimensions, emitting a warning if a pipeline is executed with a deprecated dimension
- Deprecate the `ClassFlags` dimension in favor of the new ones
- Update tests accordingly

\* The only potential ambiguity here is for legacy PDRFs with Classification=12.  This is now a reserved value, but in legacy PDRFs this indicated Overlap before Overlap had its own flag.  In this case:
- when reading a legacy PDRF and `Classification=12`, we set `Overlap = 1` and `Classification=0` (Never Classified)
- when writing a legacy PDRF, write the legacy `Classification=12` if `Overlap=1` and `Classification=0`

Potential questions:
- Can we change dimension behavior like this?  Since this behavior has already been changed over time and the initial ClassFlags implementation note mentions changing it later, I'd think yes.  This breaks pipelines which specifically target legacy PDRFs and rely on the bit fields being included in Classification
- What's the best way to purge ClassFlags?  Deprecation as implemented here, removing the dimension, something else?

Previous contributors to this conversation: @wonder-sk @kjwaters @LFraser @lc-guy @craigds @jrobichaud


abellgithub:
I haven't looked over the details, but I'm fine with this in concept, though this was the way it was mostly to save space. Perhaps this isn't a consideration anymore. I was thinking about a way to store bit flags in the current scheme. If this is something we're interested in, it could be pursued a bit.

hobu:
> though this was the way it was mostly to save space

As currently implemented it adds 3 bytes per point vs ClassFlags which was a single. The problem with the previous approach is it wasn't really convenient or possible in user or pipelines to control what happened with them. This implementation will allow people to simply set the flag and have it expressed as they expect.

Some follow on activity here might be to add bit shifting operators to the expression parser, but people can do that from Python right now if they wanted.

abellgithub:
It's fine by me. Probably not worth more thought or complication.

io/private/las/Utils.cpp
```diff
@@ -515,18 +518,32 @@ void V10BaseLoader::load(PointRef& point, const char *buf, int bufsize)
...
 
-    istream >> intensity >> flags >> classification >> scanAngleRank >> user >> pointSourceId;
+    istream >> intensity >> flags >> classificationWithFlags >> scanAngleRank >> user >> pointSourceId;
 
     uint8_t returnNum = flags & 0x07;
     uint8_t numReturns = (flags >> 3) & 0x07;
     uint8_t scanDirFlag = (flags >> 6) & 0x01;
     uint8_t flight = (flags >> 7) & 0x01;
 
+    uint8_t classification = classificationWithFlags & 0x05;
```
uclaros, line 533: Shouldn't this be `classificationWithFlags & 0x1F;` to mask the 5 LSBs?

connormanning: I think you're correct.

connormanning: @uclaros I fixed this in #4269.  However this error had no practical effect, since it's in unused code.  Good catch though.

PR #4189, opened 2023-09-25
add UserData as a synonym for user_data for both Stage and Dimension …

hobu:
…usage


hobu:
`UserData` is the AWS name for the same thing, and I couldn't remember without looking it up in the code.

Issue #4185, opened 2023-09-21
Options toMetadata method changes list to a joined string


kylemann16:
In [Options.hpp](https://github.com/PDAL/PDAL/blob/master/pdal/Options.hpp#L151-L170), PDAL is translating valid JSON lists into strings joined by a comma. This invalidates some key value pairs, in my case `filters.assign` where I am trying to apply multiple assignments at once, which is supported according to the docs. 

This
```
[
    "Classification = 0",
    "ReturnNumber = 1 WHERE ReturnNumber < 1",
    "NumberOfReturns = 1 WHERE NumberOfReturns < 1",
]
```

Is transformed into
```
"Classification = 0, ReturnNumber = 1 WHERE ReturnNumber < 1, NumberOfReturns = 1 WHERE NumberOfReturns < 1"
```

Minimal example: https://gist.github.com/kylemann16/31dd11192cf3c341272b268b6eacfb0a

The PDAL python library follows this path:

1. https://github.com/PDAL/python/blob/main/pdal/PyPipeline.cpp#L112-L117
2. https://github.com/PDAL/PDAL/blob/master/pdal/PipelineWriter.cpp#L98
3. https://github.com/PDAL/PDAL/blob/master/pdal/Stage.cpp#L100
4. https://github.com/PDAL/PDAL/blob/master/pdal/Options.hpp#L151

where it joins an iterable into a string.

hobu:
Fixed in #4190

PR #4190, opened 2023-09-25
do not concatinate lists into comma-separated strings in Options::toM…

hobu:
…etadata() #4185

### DIFF
--- a/pdal/Options.hpp
+++ b/pdal/Options.hpp
@@ -154,12 +154,12 @@ public:
         {
             StringList l = getValues(k);
             std::string vs;
-            for (auto vi = l.begin(); vi != l.end(); ++vi)
-            {
-               if (vi != l.begin())
-                   vs += ", ";
-               vs += *vi;
-            }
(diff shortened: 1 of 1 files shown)


pdal/Options.hpp
```diff
@@ -154,12 +154,12 @@ class PDAL_DLL Options
         {
             StringList l = getValues(k);
             std::string vs;
-            for (auto vi = l.begin(); vi != l.end(); ++vi)
-            {
-               if (vi != l.begin())
-                   vs += ", ";
-               vs += *vi;
-            }
+//             for (auto vi = l.begin(); vi != l.end(); ++vi)
```
abellgithub, line 157: If you don't want this, I'd suggest deleting rather than commenting -- there's always source control  history -- just came from a project with mountains of commented code and it's terrible.

hobu:
actually fixed in a3bbe98cd0dbc5b535fcbd74c4a6997d4bdb84ce 😄

PR #4199, opened 2023-09-30
Make backtrace optional

nilason:
Make it possible to turn off direct backtrace support, avoid dependency of libunwind/libexecinfo.

It is by default turned on, thus doesn't change present behaviour.


hobu:
Picked over to master in 95c5e378d6b31b88cbdbabcb75a71ffe9d6a1c10

PR #4201, opened 2023-10-02
Barycentric interpolation improvements 

abellgithub:
Improve performance of barycentric interpolation.
Improve accuracy of barycentric interpolation.

### DIFF
--- a/pdal/private/MathUtils.cpp
+++ b/pdal/private/MathUtils.cpp
@@ -34,6 +34,7 @@
 
 #include <array>
 #include <cfloat>
+#include <iomanip>
 #include <numeric>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


chambbj:
LGTM. Thanks for this @abellgithub

abellgithub:
Any issues before merge?

leavauchier:
@abellgithub : it looks like this modification in the code keeps only part of the fixes from https://github.com/PDAL/PDAL/pull/4155.
Your change gets rid of the imprecision on the determinant computation but keeps the issue with the sign of areas (that can be slightly negative due to numerical imprecision).

It looks like the test data I added was not complete enough to address completely both of the issues as tests passed but I still get this no-data issue on larger data. 
I'll try and generate a test sample today

leavauchier:
@abellgithub : it looks like this modification in the code keeps only part of the fixes from https://github.com/PDAL/PDAL/pull/4155.
Your change gets rid of the imprecision on the determinant computation but keeps the issue with the sign of areas (that can be slightly negative due to numerical imprecision).

It looks like the test data I added was not complete enough to address completely both of the issues as tests passed but I still get this no-data issue on larger data.
I'll try and generate a test sample today

abellgithub:
@leavauchier If there are issues with this implementation (it's possible), some example data would be great.

leavauchier:
Sorry for the undue ping, after double checking my data, it seems to work fine (the issue was somewhere else, that's why I ended up removing my message). I'm still a bit surprised that checking for the exact zero value for the areas and the sign check do not cause the same precision issues as with the previous implementation.
Anyway, thanks for having looked for a better fix than the one I had before.

abellgithub:
Eliminating the division when deciding in vs. Out makes the likelihood of a precision issue pretty small. It depends on the data, of course, but I think in the cases we're normally dealing with, this is likely to work.

abellgithub:
I don't think very small or very large is the issue, as such. Certainly, some real numbers are not representable as floating-point, but by the time the values come into this filter, that distinction is lost (we're only dealing with the base-2 representations). What typically causes problems is when you try to do math with values that are not of commensurate scale. In our case, assuming the scale of the X and Y values is relatively similar and the points are "close" to each other WRT that scale, I would think that the determinants could be computed exactly with the values as passed to the function, at least this is the hope.

In general, though, "fixing" this function isn't the best way to deal with the original faceraster filter issue. If one knows that a test point is within some bounding polygon, we need to place it in some triangle, and if we do that, we can compute an interpolated Z value. If there are further problems, I guess I'd try to do something different in the faceraster filter to force points into some triangle. Our current dumb scheme for storing triangles makes this a bit of a challenge, but I think it's probably best to wait until we have more issues to add complexity for an unusual occurrence.

abellgithub:
All we care about is the sign of the results being correct. Since subtraction of _floating point_ values that are within a factor of 2 of one another is exact, I'm thinking this should be fine. It is certainly the case that we're unable to represent some numbers exactly, but by the time we hit this math, that ship has sailed.

If this comes up again, I think the place to fix it is in the faceraster filter. If we know that a point is in some polygon, we can force it to be in some triangle of the triangulation. The simplistic way we store triangles makes this hard, but we can cross that bridge when necessary.

PR #4204, opened 2023-10-08
Eliminate ambiguity in code readability #3997

xiaodong2077:
#3997

Although error messages with GDAL3.0.4 are
```
CMake Error at /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find GDAL: Found unsuitable version "3.0.4", but required is at
  least "3.4" (found /usr/lib/libgdal.so)
```

I think the comparation part should stay with the new version.

In fact, as the `find_package(GDAL 3.0 REQUIRED)` line will generate error messages, is it necessary to keep the comparation part `if...endif()`?

### DIFF
--- a/cmake/gdal.cmake
+++ b/cmake/gdal.cmake
@@ -20,9 +20,9 @@ if (GDAL_FOUND)
     #
     # Older versions of FindGDAL.cmake don't properly set GDAL_VERSION
     #
-    if (GDAL_VERSION VERSION_LESS 3.0.0)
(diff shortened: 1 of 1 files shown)


hobu:
> In fact, as the find_package(GDAL 3.0 REQUIRED) line will generate error messages, is it necessary to keep the comparation part if...endif()?

No. This is very old CMake configuration that is no longer relevant.

PR #4205, opened 2023-10-10
CRS Priority refactoring

abellgithub:
Changes to fix 4191 and by default use WKT/projjson VLRs for LAS and COPC.


io/CopcReader.hpp
```diff
@@ -53,6 +53,7 @@ namespace las
     struct Header;
     struct Vlr;
     class VlrCatalog;
+    using VlrList = std::vector<Vlr>;
```
hobu, line 56: Shouldn't this be in somewhere more central? I know it's just a declaration here though.

io/private/las/Vlr.hpp
```diff
@@ -157,10 +157,21 @@ class VlrCatalog
...
+
+        Entry(const std::string& userId, uint16_t recordId, uint64_t offset, uint64_t length) :
+            userId(userId), recordId(recordId), offset(offset), length(length)
+        {}
+
         std::string userId;
-        uint16_t recordId;
-        uint64_t offset;
-        uint64_t length;
+        uint16_t recordId = 0;
```
hobu, line 172: were these not being initialized a problem?

io/private/las/Vlr.cpp
```diff
@@ -433,29 +433,30 @@ std::vector<char> VlrCatalog::fetch(const std::string& userId, uint16_t recordId
...
-            if (e.length > (std::numeric_limits<uint32_t>::max)())
-                return vlrdata;
+    const Entry& e = find(userId, recordId);
 
-            offset = e.offset;
-            length = (uint32_t)e.length;
-            break;
-        }
-    if (length == 0)
+    // We don't support VLRs with size > 4GB
```
hobu, line 453: Should we throw in this case? Maybe warning with a Debug message at least...

hobu (approved):

PR #4206, opened 2023-10-10
implement writers.copc.nosrs

hobu:


doc/stages/writers.copc.rst
```diff
@@ -159,6 +159,11 @@ threads
 enhanced_srs_vlrs
   Write WKT2 and PROJJSON as VLR [Default: false]
 
+nosrs
```
connormanning, line 162: Doc was added to the writer but this is a reader option.

PR #4207, opened 2023-10-10
Specialization for projected bounds query in COPC/EPT readers

connormanning:
The COPC/EPT readers have an implicit assumption, when supplied with `bounds` which are a different SRS from the data, that accumulating the mins/maxes of the cubic node corners reprojected into the query SRS is accurate for determining overlapping nodes.  This is generally true for projected SRSes but can be incorrect for combinations of geographic/geocentric/projected systems.

This PR only corrects a single combination which was incorrect before: where the data is in BCBF but the bounds query is in geographic.  There may need to be specializations for other combinations in the future.  For now, these implementations are duplicated across the COPC/EPT readers, and another future enhancement would be to combine a lot of the common functionality between these readers.  It is not just the new code, there is a lot of duplicated code there so that's out of scope for this PR.

This code also needs a unit test - this could be simple since prior to this specialization, data about 45 degrees north or south of the equator would be missed for full-body global data in BCBF when queried in lon/lat.

### DIFF
--- a/io/CopcReader.cpp
+++ b/io/CopcReader.cpp
@@ -72,6 +72,67 @@ const StaticPluginInfo s_info
     { "copc" }
 };
 
+void reprogrow(BOX3D& b, SrsTransform& xform, double x, double y, double z)
+{
+    xform.transform(x, y, z);
+    b.grow(x, y, z);
+}
+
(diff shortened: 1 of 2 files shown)


hobu (approved):

hobu:
Add a test?

connormanning:
> Add a test?

Added.

PR #4208, opened 2023-10-11
Remove unwanted include in georeference filter

gui2dev:
As @xiaodong2077 pointed out in #4109, removing extra include that are not used.


hobu:
@gui2dev what is the minimum PROJ version? A version filter should be put into https://github.com/PDAL/PDAL/blob/master/cmake/proj.cmake to reflect that.

gui2dev:
As far as I know, there is no minimum for the C implementation

Issue #4217, opened 2023-10-31
LiTree never terminates


fnands:
I'm trying to run the LiTree algorithm to create some tree canopy classifications. 

Btw, I did see #3810 , but as far as I can tell that was never really resolved. 

What seems to be happening is that the algorithm reaches a point where it has an amount of points left that are `< min_points`, but doesn't terminate. 

In my case, I set `min_points` to `50`, and get: 

```
(pdal pipeline filters.litree Debug) 		Max HAG value of 21.32
(pdal pipeline filters.litree Debug) Classifying points for tree 73 (|Ui| = 48)
(pdal pipeline filters.litree Debug) 			|Pi| = 1, |Ni| = 48
(pdal pipeline filters.litree Debug) Minimum cluster size not met; skipping tree.
```  

over and over again. 

Which (if I am reading this correctly), is telling me that the amount of points left is 48, meaning it will never reach a point where it can create the final tree. 

Doesn't it make sense to add some sort of exit criteria? 

The line I think is causing this is [here](https://github.com/PDAL/PDAL/blob/1492745b366afc6acea91e503f17aefb41c52c29/filters/LiTreeFilter.cpp#L282C7-L282C7). 

Example file on [gdrive](https://drive.google.com/file/d/1i2E4R6PUxXpxEikjDpix-x00INycxSn7/view?usp=sharing)

Pipeline: 
```
[
    {
        "type": "readers.las",
        "filename": "filtered_1m_tiny.laz",
        "extra_dims": "HeightAboveGround=float"
    },

    {
        "type":"filters.sort",
        "dimension":"HeightAboveGround",
        "order":"DESC"
    },
    {
        "type":"filters.litree",
        "min_points": 50,
        "min_height":10.0,
        "radius":15.0,
        "where": "Classification == 1"

    },
    {
        "type":"writers.las",
        "filename":"output_radius15_minp50.laz",
        "minor_version":1.4,
        "extra_dims":"all"
    }
]
```

PR #4218, opened 2023-10-31
Set while condition for LiTree to be larger than min points

fnands:
Set exit condition so LiTree doesn't get stuck when number of points left are fewer than `min_points`. 

I explain the meat if the issue in [#4217](https://github.com/PDAL/PDAL/issues/4217), but basically LiTree never terminates, and needs some kind of exit condition. 

I submitted this without testing, will get to that soon.

### DIFF
--- a/filters/LiTreeFilter.cpp
+++ b/filters/LiTreeFilter.cpp
@@ -279,7 +279,7 @@ void LiTreeFilter::filter(PointView& view)
 
     // "...stop when Ui is empty."
     int64_t tree_id(1);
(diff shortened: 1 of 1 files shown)


chambbj:
Seems like a reasonable solution to me. Good catch!

While you are at it, I think that we could save some processing and also check that the original PointView has at least the minimum number of points.

Unit tests would be great of course, although it looks like we didn't have any to start with for this particular filter.

abellgithub:
LGTM.

Waiting for a test.

PR #4220, opened 2023-11-01
Fix readers.stac doing head requests before necessary

kylemann16:
STAC Item class was doing a HEAD request before exhausting all other options when determining what PDAL driver to use, which was causing a massive slowdown, especially in particularly large `FeatureCollections`. 

Fixed by moving `m_factory.inferReaderDriver(dataUrl);` to before the head request.


hobu:
Picked back to `2.6-maintenance` at 597ab2dfec9d7407f347d659268df4a1afc0e069

PR #4232, opened 2023-11-09
Add Radius assign filter

leavauchier:
Hi, 
I've got a first implementation for a radius search filter (a filter that allows you update the value of a dimension (using
an assignment expression) for specific points depending on their neighbors in a given radius)

@abellgithub for the moment it is still quite slow (approx 10 minutes for a point cloud of 50 million points), I'll continue to look for some optimizations in my algorithm but I thought it was the easiest way to share where I'm at with this filter


leavauchier:
I had forgotten to re-build in release mode, it's rather 2 minutes for this same 50 millions points las (including reading it).

chambbj (commented):
- I might prefer `filters.radiusassign`. It feels like we are doing more than searching here and so that alone might be misleading.
- Is `Keypoint` being used simply because it is a similar Dimension type as `Classification`? Is the intent to ultimately overwrite `Classification` but you don't want to do this in-place?
- I have no problem with limiting this to a 2D search, but I think it's probably worth mentioning in the documentation.
- I have not thought about the logistics of making this happen, but this is quite similar to `filters.neighborclassifier`. Probably would be too much of a force fit to make that happen, as it is based on fixed number of KNN and a separate file for classifications.

leavauchier:
> * I might prefer `filters.radiusassign`. It feels like we are doing more than searching here and so that alone might be misleading.

You're right, I've changed the name

> * Is `Keypoint` being used simply because it is a similar Dimension type as `Classification`? Is the intent to ultimately overwrite `Classification` but you don't want to do this in-place?

The intent is to ultimately mark points with a custom dimension (for example mark points to use in a triangulation (in another pipeline a priori)  while keeping all the points in the point cloud file

> * I have no problem with limiting this to a 2D search, but I think it's probably worth mentioning in the documentation.

I can also update the filter to give a choice between 2D and 3D as it seems possible by using a KDFlexIndex

> * I have not thought about the logistics of making this happen, but this is quite similar to `filters.neighborclassifier`. Probably would be too much of a force fit to make that happen, as it is based on fixed number of KNN and a separate file for classifications.

My first attempt was to try to update the neighborclassifier, but dealing with too many options became cumbersome (especially as the neighborsclassifier only works on classification for now, and uses neighbors classes instead of assigning one given by the user)

leavauchier:
After some time, I've updated the code of this new feature and added better tests.
However there remains a failing test on the Windows VMs that does not seem to be related to this new filter (pdal_io_draco_reader_test is the only one that fails). Do you have any clue on what could be wrong with this test?

hobu:
> Do you have any clue on what could be wrong with this test?

Yeah, we have an issue with MSVC and an interaction with how our PointRef implementation works for sorting. @abellgithub is working to fix it and then we will have a 2.9.1 release.

PR #4235, opened 2023-11-10
Allow pdal to link with CURL target

Faroito:
Hello,

When I'm building pdal with curl 7.88, pdal is unable to link with curl. 
This MR tries to fix that, by allowing CURL::libcurl to be link instead of CURL_LIBRARIES (with a fallback if the target doesn't exist).


hobu:
Picked back to `2.6-maintenance` with 7b8f423753e2ab3b12b824f8f56bffb2160468ad

Issue #4234, opened 2023-11-10
How to read and write in streaming mode


oleg-alexandrov:
**Describe the bug**

The documentation could use an example of how to read and write from a las file in streaming mode. I've spent quite a lot of time reading the existing doc, looking at the tutorial, architecture overview, and unit tests, and I could not find one block of code that could be put to good use (there are some pieces here and there that maybe could be integrated somehow).

I think this is relevant since point cloud files are usually too large to store in memory, so streaming is likely the more usual approach.

abellgithub:
I don't the understand. What, exactly, do you want to know how to do? Are you using the API? If so, what is unclear to you?

oleg-alexandrov:
I want to read and write from a las file in streaming mode. There is no example of using the streaming API.

I've spent half a day reading through the code, and will do more today, and I am getting to the point I will cook up something. 

I hope you understand now. It is really a matter you using a new software and wishing those folks spelled out things some more for you.

hobu:
> There is no example of using the streaming API.

Plenty of those in the test suite https://github.com/PDAL/PDAL/blob/master/test/unit/StreamingTest.cpp#L275

oleg-alexandrov:
Thank you. I read this yesterday. The documentation says that the StreamPointTable class must be used. This example does not use that class. This example uses instead the Streamable class, which is not mentioned in the doc.  

So all the pieces are there, somehow, and by trial and error I've been making some progress.

hobu:
Another place for good examples is the kernels. TranslateKernel makes heavy use of streaming https://github.com/PDAL/PDAL/blob/master/kernels/TranslateKernel.cpp

oleg-alexandrov:
BTW, the doc mentions once StreamingPointTable, but the actual class is called StreamPointTable. 

I also worked with TranslateKernel yesterday and went 5 levels deep through the pipeline architecture before I thought there must be an easier way to figure this all out.

So you closed the issue, and that is fine, but the fact remains that there is no example of using the streaming code and the internals are inconsistent.

hobu:
> So you closed the issue, and that is fine, but the fact remains that there is no example of using the streaming code and the internals are inconsistent.

Please use the mailing list for discussion and questions.

> I also worked with TranslateKernel yesterday and went 5 levels deep through the pipeline architecture before I thought there must be an easier way to figure this all out.

[...]

PR #4238, opened 2023-11-11
Add streaming tutorial

oleg-alexandrov:
This adds an example to PDAL for how to create and stream to disk a cloud point-by-point. I think this is an important use case since clouds are often too big to fit in memory.  The background on this is in https://github.com/PDAL/PDAL/issues/4234.

I also made a small documentation fix replacing the mention of StreamingPointTable with FixedPointTable, as the latter is the real workhorse for streaming unless the user needs some custom logic.

This was tested with PDAL 2.6.0 as fetched with conda.

Happy to hear any feedback and iterate.

### DIFF
--- a/examples/writing-streamer/streamer_tutorial.cpp
+++ b/examples/writing-streamer/streamer_tutorial.cpp
@@ -1,0 +1,131 @@
+#include <pdal/PointView.hpp>
+#include <pdal/PointTable.hpp>
+#include <pdal/Options.hpp>
+#include <pdal/StageFactory.hpp>
+#include <pdal/Streamable.hpp>
+#include <pdal/Reader.hpp>
+
+#include <vector>
+
+namespace pdal {
+    
+// A class to produce a point cloud point-by-point, rather than
+// having it all in memory at the same time. It will be streamed to
+// disk. See the GDALReader class for how to add more fields
+// and read from disk.
+class PDAL_DLL StreamedPointCloud : public Reader, public Streamable
+{
+public:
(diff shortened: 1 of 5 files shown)


examples/writing-streamer/streamer_tutorial.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+int main(int argc, char* argv[])
+{
+    using namespace pdal;
+    
+    // A point cloud. Only one point at a time will be in memory
+    StreamedPointCloud stream_cloud;
+
+    // Will copy here each point and then stream it to disk    
+    FixedPointTable t(1);
```
hobu, line 112: A table size of 1 is going to create lots of function call churn. This may or may not be what you want, but a comment here explaining the consequences of you choice would be helpful.

abellgithub:
The [architecture overview](https://pdal.io/en/2.6.0/development/overview.html) does discuss streaming. Perhaps the heading should be something like ” development with the API" rather than just "development”.

abellgithub:
I will look this over.

examples/writing-streamer/streamer_tutorial.cpp
oleg-alexandrov: Good point. Yeah, normally one processes a tile at once, especially if stored on disk that way. I adjusted this.

hobu:
picked back to `2.6-maintenance` at cbb8026e460ce960348f8416dca110ab22e7316a

PR #4240, opened 2023-11-13
Fix position calculation when parsing bounds.

### LINKED ISSUE #4228: Error when specifying spatial reference in "bounds" option
**Describe the bug**
Using "bounds" option with COPC and EPT readers with a spatial reference causes the pipeline to fail.

```
$ pdal pipeline --input myPipeline.json
```

Here is the relevant part of my pipeline:
```
"bounds": "([-71.135695, -71.135491], [42.427430, 42.427709])/EPSG:4326"
```
I tried both with and without the 'EPSG'

**Expected behavior**
The documentation says that "The bounds can be followed by a slash (‘/’) and a spatial reference specification to apply to the bounds."

However, the pipeline terminates with this error:
```
"PDAL: Invalid character following valid bounds box."
```

I'm using PDAL 2.5.6.

abellgithub:
Close #4228

### DIFF
--- a/pdal/util/Bounds.cpp
+++ b/pdal/util/Bounds.cpp
@@ -252,17 +252,6 @@ bool discardSpacesBefore(std::istringstream& ss, char nextChar)
     return Utils::eatcharacter(ss, nextChar);
 }
 
-std::string::size_type countRemaining(std::istringstream& ss)
-{
-    std::string::size_type count = 0;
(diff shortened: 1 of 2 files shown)


hobu:
picked back to `2.6-maintenance` at d2f9fdf677ec8929314b3bfa88a325bf74544f5f

PR #4246, opened 2023-11-16
Use proper level for grid size.

### LINKED ISSUE #4230: readers.copc.bounds produces different results than filters.crop.bounds
Using the following COPC file, we get different filter results depending on our usage.

## Using `readers.copc.bounds`

```
(pdal26) pdal info test_data.copc.laz --readers.copc.bounds='([510.00,540.00],[300.00,360.00],[-1000,1000])' | jq '.stats.statistic[0]'
{
  "average": 525.0045977,
  "count": 435,
  "maximum": 539,
  "minimum": 511,
  "name": "X",
  "position": 0,
  "stddev": 8.989752333,
  "variance": 80.81564702
}
```

## Cropping with `filters.crop.bounds`

```
pdal translate data/test_data.copc.laz cropped.las --filter crop --filters.crop.bounds='([510.00,540.00],[300.00,360.00],[-1000,1000])'

pdal info cropped.las | jq '.stats.statistic[0]'
{
  "average": 525,
  "count": 1682,
  "maximum": 539,
  "minimum": 511,
  "name": "X",
  "position": 0,
  "stddev": 8.369088474,
  "variance": 70.04164188
}
```

The correct point filter is being produced by `filters.crop.bounds`. The following image shows the file, which is just a grid of points. There should be $29^2 == 841$ points in each cell and the bounds we are giving should be selecting two of them, which should give us $1682$ points. I don't understand where the 435 number is coming from.

Screenshot 2023-11-08 at 8 38 06 PM

[example.zip](https://github.com/PDAL/PDAL/files/13303124/example.zip)

abellgithub:
Close #4230

### DIFF
--- a/io/private/copcwriter/Grid.cpp
+++ b/io/private/copcwriter/Grid.cpp
@@ -96,7 +96,7 @@ void Grid::resetLevel(int level)
 {
     // We have to have at least level 1 or things break when sampling.
     m_maxLevel = (std::max)(level, 1);
-    m_gridSize = (int)std::pow(2, level);
+    m_gridSize = (int)std::pow(2, m_maxLevel);
 
     if (m_cubic)
     {


hobu:
Picked back to `2.6-maintenance` at 617036694e405d5df92cf50e372c826e2e192ab9

Issue #4245, opened 2023-11-16
Documentation for how to read a cloud point-by-point


oleg-alexandrov:
The developer documentation does not explain how to open an existing point cloud and read each point one by one for some user-specific purpose. I think this is a rather basic use case when integrating PDAL in a third-party application. 

This is different than https://github.com/PDAL/PDAL/pull/4238, where the goal was for user to create and write the cloud point by point.

I made this work by starting with GdalWriter, and stripping it to bare essentials. What is left is still a rather non-simple class that has to implement some functions from Streamable and also from FlexWriter (but for the latter as placeholders). 

If there is interest, I could provide a pull request with a self-contained example program, to be mentioned in the doc.

abellgithub:
Pdal is not a point-by-ponit reader. We intentionally do not offer such an API. Of course you're welcome to modify the source code for your own use as you choose.

oleg-alexandrov:
With LibLAS no longer developed and not having support for LAS 1.4, PDAL appears to be the only alternative, for reading LAS from C++. 

I did not modify any PDAL source code. Reading point-by-point is a matter of creating a little writer class that implements processing one point. 

Imagine saying GDAL is not in the business of reading Tif files.

hobu:
> With LibLAS no longer developed and not having support for LAS 1.4, PDAL appears to be the only alternative, for reading LAS from C++.

[LASlib](https://github.com/LAStools/LAStools/tree/master/LASlib) from LAStools is a libLAS alternative. The licensing may be problematic for people, however. If you wish to take over the maintenance of libLAS to add what's needed, I would be very happy to hand you the keys.

>> how to open an existing point cloud and read each point one by one for some user-specific purpose

> Pdal is not a point-by-ponit reader. We intentionally do not offer such an API.

PDAL's design and approach was a reaction to the limitations of the point-by-point approach of libLAS 😁 I thought at one time @wenzeslaus had written a libLAS=>PDAL migration guide, but I can't seem to find it just now.

My worry about a point-by-point tutorial for PDAL is the next developer would follow it and come along and say "PDAL sucks for reading point-by-point because it has all this function call overhead". They're right, of course, but that's not the intended normal usage.

All that to say if you made a PR with the doc, I'd merge it 😄

oleg-alexandrov:
> If you wish to take over the maintenance of libLAS to add what's needed, I would be very happy to hand you the keys.

I appreciate the highly generous offer!

[...]

PR #4247, opened 2023-11-16
Add reading streamer example

oleg-alexandrov:
This provides an example for how to stream points from a cloud. The background is in https://github.com/PDAL/PDAL/issues/4245#issuecomment-1814847080.

The previously added "streamer_tutorial" got renamed to "writing_streamer", and this new one is "reading_streamer".

Thoughts welcome.

BTW, concerns about overhead of using such an API are likely exaggerated. A lot more time is likely spent in disk access and compression.


abellgithub:
I'm sorry but I don't understand the point of this. There are MANY examples of filters and writers that use stream mode. This seems to just be boilerplate. Also, there is no reason NOT to support standard mode AND stream mode in most stages.

Perhaps we should just update the existing examples with processOne() and say something about using various PointTables.

oleg-alexandrov:
It took me three days of reading and compiling code and experimenting to derive the "boilerplate". That because architecture of PDAL is not well-documented, and has very few hands-on examples. 

PDAL also has a very strong bias against doing very simple things, such as reading a point cloud. It is not possible now to use PDAL without understanding it in great detail.

To you, as a developer, this all looks very obvious, yes. 

I am not interested in filters or creating PDAL extensions. I want to be able to read a point cloud on the terms of my application. I am asking for a very simple example a user can build upon.

oleg-alexandrov:
If you can write a compliable standalone example that illustrates the best practices for how to give the user access to each point, replacing my example, that is fine with me. It has to have a main() rather than be a filter. I don't call boilerplate either but could not come up with anything better.

oleg-alexandrov:
One advantage of the boilerplate code is that it illustates fully the PDAL architecture. When using PDAL for actual work all the member functions in this example need to be populated.

After working some more with PDAL, it looks that lack of a function to get and set a point is a serious limitation. PDAL can't be used in a larger setting unless when all processing must happen at once as controlled by a writer. LibLAS could be made to work in an existing application, rather than rewrite the application around PDAL.

abellgithub:
If you could provide some specific example of your desired use case, it
might allow others to help you.

If liblas serves your needs, there is no reason not to use that instead of
PDAL.

abellgithub:
You can also use lazperf. It allows for simple reading of all formats of LAS/LAZ points.

oleg-alexandrov:
PDAL can do streaming over a network (as I recall) and supports LAS 1.4, which libLAS can't. And libLAS is not developed.

My current sponsor wants PDAL, so that kind of settles it. Good to know about lazperf.

Mine was a general observation. I do understand that choices were made when PDAL was designed and that it functions well given what it is meant for. Looking forward though, a way of streaming in and out pointwise would be a nice touch.

Thank you for offering your help. The particular issue I have now is because my application uses another opinionated framework, which like PDAL, wants to be in driver's seat. That's why it is good to give the developer flexibility and not impose hard constraints on how people should do things. But, as before, I will make do with what there is. Thanks again for your help.

abellgithub:
Here is a complete program that reads points one at a time. I didn't deal with the termination case.

```
#include <thread>
#include <mutex>
#include <condition_variable>
#include <iostream>

#include <pdal/io/LasReader.hpp>
#include <pdal/filters/StreamCallbackFilter.hpp>

double x, y, z;
std::mutex mutex;
std::condition_variable produceCv, consumeCv;
bool consumed = true;

void makeProcessor()
{
    while (true)
    {
        {
            std::unique_lock<std::mutex> l(mutex);
            consumeCv.wait(l);
            std::cerr << "X / Y / Z = " << x << "/" << y << "/" << z << "!\n";
            consumed = true;
        }
        produceCv.notify_one();
    }
}

bool gotPoint(pdal::PointRef& p)
{
    static int cnt = 0;
    {
        std::unique_lock<std::mutex> l(mutex);
        produceCv.wait(l, []{ return consumed; });
        x = p.getFieldAs<double>(pdal::Dimension::Id::X);
        y = p.getFieldAs<double>(pdal::Dimension::Id::Y);
        z = p.getFieldAs<double>(pdal::Dimension::Id::Z);
        consumed = false;
    }
    consumeCv.notify_one();
    if (++cnt == 100)
        return false;
    return true;
}

void makeReader(const std::string& filename)
{
    using namespace pdal;

    FixedPointTable t(10000);

    Options o;
    o.add("filename", filename);

    LasReader r;
    r.setOptions(o);

    StreamCallbackFilter f;
    f.setCallback(gotPoint);
    f.setInput(r);
    f.prepare(t);
    f.execute(t);
}

int main()
{
    std::string filename = "/Users/acbell/pdal/test/data/laz/autzen_trim.laz";

    std::thread t(makeReader, filename);
    makeProcessor();
    t.join();
    return 0;
}
```

oleg-alexandrov:
Andrew, thank you.

This is a very valuable example of using a callback. It would actually be great if it could be added to the list of examples.

The example I provided looks at a different aspect, on how to precisely set all the pieces needed to implement a full writer. That is useful when you need a way to keep track of internal states and package everything in a class. Though presumably the callback mechanism is equivalent.

In my own experience of developing user-facing software one can never go wrong by providing the user with many well-documented examples, so they can choose what works best for them. My own suggestion would be for this to be an additional example in the doc.

But obviously PDAL is not my project to maintain, so anything which makes sense for you is good, and even one example is better than no example.

Both of these, of course, have the same limitation that the writer can't be told to pause and resume later, but this we already figured is a PDAL design decision.

abellgithub:
The provided.example pauses as each point is consumed. makeProcessor() doesn't need to run a loop.

oleg-alexandrov:
Thank you. It looks that this can serve the purpose of pleasing two workflows that do different things but one needs the second one periodically. One workflow can tell the other to collect some points, the second one does that and notifies the first one, the first one does whatever it needs to do and then reaches out again for more points.

Enough to make one's head spin. :) 

On my side I ended up rewriting my code to use PDAL's processOne() function and just make a periodic dump on disk. I will try later to incorporate the kind of logic you put in. This is clever and I learned something.

oleg-alexandrov:
Andrew, I tested your code. I was getting into odd race conditions on occasion. I must say I am not sure if it is something I did not do right.

After reading around more regarding the producer/consumer paradigm, I made some modifications to the code. I also added a stopping criterion.

Not sure if this is to your liking, but all I can say that I tested this code very thoroughly.

My suggestion here is to have two examples. One for a full writer class, and second, showing the callback mechanism (which can stop resume streaming).

This is what this pull request now, with documentation where I think made sense.

Happy as before to iterate.

abellgithub:
I fixed the issue in the program provided above. It now handles the case where a notify_one() could be called by the producer before the consumer was waiting. It also handles completion.

oleg-alexandrov:
If you can pick a version you like, put it on top of what I had in the pull request, and pull all this in, that is good with me.

oleg-alexandrov:
Ding ding, and happy new year to all. 

Last attempt at suggesting a merge. Counting on a developer's pride in knowing one's code is well-rounded and easy to use.

hobu:
> Any further thoughts?

Sorry for the ghosting, we just haven't had time to review and incorporate. I'll try to do so next time we roll up a release (beginning of next month'ish).

hobu:
Merging this for now and we can refactor it as we need.

PR #4249, opened 2023-11-20
Zero out point count fields when creating a LAS file that has no points.

connormanning:
For remote files, we create LAS files with only headers/VLRs/EVLRs ([added here](https://github.com/PDAL/PDAL/pull/4178)) to avoid downloading the full file just to peek at its metadata.  However the point counts were not zeroed, so [this check](https://github.com/PDAL/PDAL/blob/482688f3669677079536583190d766e31f54d2cf/io/private/las/Header.cpp#L130-L132) would trigger an error for uncompressed LAS data: the `fileSize` passed into this function would of course not accommodate the actual point data which was removed.  So zero those fields to make sure our frankensteined zero-point file is fully valid.

Fixes `pdal info http://some-uncompressed-las-file.las --metadata`


hobu:
Picked back to `2.6-maintenance` at 270867e599b09c49f0a75211380155a2f02b31dd

PR #4251, opened 2023-11-21
Correct metadata point counts for remote-las-file optimization

connormanning:
`pdal info` with `--metadata` and `--summary` flags was (correctly) emitting point counts of zero after #4249, which set the header point count values to 0 to keep our zero-point LAS file valid.  This PR returns the point count separately while it creates the zero-point LAS file so we can correct this value in our metadata.  This is all a little messy but with the special zero-point LAS file stuff it seems to be the only way to do it.


msmitherdc:
@connormanning works for me

hobu:
Picked back to `2.6-maintenance` at 1541a025d1fee8b10e21f9a913c78c99062adc8d

PR #4252, opened 2023-11-22
new filter : calculate 2d/3d  relative distance and height to target pointcloud

dezhi-l:
- using one point cloud as a reference, calculate the relative elevation and distance values of another point cloud
- As shown below, it is rendered according to the relative elevation dimension.
The red point is the reference point cloud (the ins trajectory point), and the other is the point cloud rendered according to the relative elevation.
image

### DIFF
--- a/filters/DisAndHeightToTarget.cpp
+++ b/filters/DisAndHeightToTarget.cpp
@@ -1,0 +1,90 @@
+#include "DisAndHeightToTarget.hpp"
+
+namespace pdal
+{
+
+    static PluginInfo const s_info{
+        "filters.disandheighttotarget",
+        "cal 2D or 3D distance and height to target pointcloud ,first input pointcloud as target pointcloud ",
+        "http://pdal.io/stages/filters.disandheighttotarget.html"};
+
+    CREATE_SHARED_STAGE(DisAndHeightToTarget, s_info)
+
+    std::string DisAndHeightToTarget::getName() const
+    {
+        return s_info.name;
+    }
+
+    void DisAndHeightToTarget::addArgs(ProgramArgs &args)
+    {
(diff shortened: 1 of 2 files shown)


filters/DisAndHeightToTarget.cpp
```diff
@@ -0,0 +1,90 @@
...
+    {
+        m_heightToTargetDim = layout->registerOrAssignDim(m_heightToTargetDimName, Dimension::Type::Double);
+        m_disToTargetDim = layout->registerOrAssignDim(m_disToTargetDimName, Dimension::Type::Double);
+    }
+
+    void DisAndHeightToTarget::prepared(PointTableRef table)
+    {
+        PointLayoutPtr layout(table.layout());
+
+        if (layout->findDim("X") == Dimension::Id::Unknown ||
```
abellgithub, line 35: This entire function is unnecessary. X, Y and Z are guaranteed to exist.

filters/DisAndHeightToTarget.cpp
```diff
@@ -0,0 +1,90 @@
...
+        if (!this->m_target)     
+        {
+            log()->get(LogLevel::Debug2) << "Adding target points\n";
+            this->m_target = view;
+        }
+      
+        else
+        {
+
+            KD3Index &kd_fixed3D = this->m_target->build3dIndex();
```
abellgithub, line 54: I don't think you want to build both indices.

abellgithub (approved):

hobu:
I don't understand what this is supposed to do. How is it different (besides being way less sophisticated) than `filters.icp`? Or `filters.geomdistance` (which is admittedly quite simple)?

hobu:
This will need some updates soon if it is to make the upcoming PDAL 2.7.0 release.

PR #4253, opened 2023-11-22
Update arbiter bundle

connormanning:
- remove some extraneous debug logging to cout
- allow unsigned AWS requests for public buckets from connormanning/arbiter#48


hobu:
picked back to `2.6-maintenance` at 79813404588ef3f7917203741b8905eb3915e849

PR #4254, opened 2023-11-22
Arbiter updates for S3 endpoints.json handling

connormanning:
The default DNS suffix was incorrectly used when endpoints.json lacked an explicit hostname, now the `region.service.suffix` should be built using the correct default suffix for a given region if its partition configuration is found.


hobu:
picked back to `2.6-maintenance` at 8e5cb1084ad417100c2a98b7f1400d67cdeb0514

hobu:
Picked back to `2.5-maintenance` at d0cb24bbb52930889e96ee98438547fd9a843f8e

Issue #4256, opened 2023-11-25
undefined reference to `_ULx86_64_step'


jiapei100:
Failed to build PDAL under Ubuntu 22.04.3 as follows:

```
[100%] Linking CXX executable ../bin/pdal
cd ....../PDAL/build/apps && /usr/bin/cmake -E cmake_link_script CMakeFiles/pdal.dir/link.txt --verbose=1
/usr/bin/c++ -std=c++17 -O3 -DNDEBUG CMakeFiles/pdal.dir/pdal.cpp.o -o ../bin/pdal  -Wl,-rpath,....../PDAL/build/lib:/usr/local/lib: ../lib/libpdalcpp.so.16.0.0 /usr/local/lib/libgdal.so 
/usr/bin/ld: ../lib/libpdalcpp.so.16.0.0: undefined reference to `_ULx86_64_step'
/usr/bin/ld: ../lib/libpdalcpp.so.16.0.0: undefined reference to `_Ux86_64_getcontext'
/usr/bin/ld: ../lib/libpdalcpp.so.16.0.0: undefined reference to `_ULx86_64_init_local'
/usr/bin/ld: ../lib/libpdalcpp.so.16.0.0: undefined reference to `_ULx86_64_get_proc_name'
/usr/bin/ld: ../lib/libpdalcpp.so.16.0.0: undefined reference to `_ULx86_64_get_reg'
```

Can anybody help?

Cheers
Pei

aimixsaka:
Similar issue: #4211

jiapei100:
Already solved by setting `-DWITH_BACKTRACE=OFF`. 
Please refer to [https://github.com/PDAL/PDAL/issues/4211](https://github.com/PDAL/PDAL/issues/4211)

PR #4262, opened 2023-11-28
default to WITH_BACKTRACE=OFF

hobu:
Prevent #4256 by defaulting BACKTRACE support to OFF.


abellgithub:
Do what you wish, but this is fixed by #4267

PR #4266, opened 2023-11-29
 Fixing seg fault referencing bounding box array entry. 

kylemann16:
STAC metadata creation was referencing a `MetadataNode` by array index, which fails the assumption when the source file doesn't have a CRS. Move to referencing by child name and failing to an error message in the `stac` key of the info dump if key `EPSG:4326` doesn't exist in key `stats.bbox`.

STAC users expect the geometry to represent the data in 4326, so the `stac` key will now carry a message if the STAC information fails to be made in this way.


abellgithub:
Test?

hobu:
picked back to `2.6-maintenance` at 0a58df7c9aff6e589bdfd4969d6c29b2c16119e2

PR #4269, opened 2023-11-29
Fix classification mask for LAS V10 loader utility

### LINKED ISSUE #4268: LAS utility masks classification incorrectly
The V10 LAS utility does not mask the classification value properly: https://github.com/PDAL/PDAL/pull/4186#discussion_r1409720235

This utility does not appear to be used at all within PDAL: the LAS readers/writers do not use it.  The COPC drivers use these utilities in general but would *not* use this specific V10 loader since COPC data is V14.  So I don't think any of PDAL's behavior has been affected by this.

I'll fix the utility and add a test for this.

connormanning:
A private utility function had an incorrect mask for extracting legacy LAS data.  However, this function was unused (and untested): these utilities are only used by the COPC reader, whose data must be LAS 1.**4**.  To make that clear, I've removed the legacy PDRF loaders from the `init()` function in these utilities for now.

At some point, we should probably make the LasReader use these utilities as well, since currently they are duplicated with slight changes between `las/Utils` and the `io/LasReader`.  That's why I haven't deleted them entirely.  The equivalent line in the LasReader is written correctly [here](https://github.com/PDAL/PDAL/blob/300bc6613f535d5df79c9589ecf6f6bc79247220/io/LasReader.cpp#L616) and is also sufficiently tested.

Because these code paths were completely unused and also not public API, this change shouldn't necessitate a re-release and should not have produced any incorrect data from PDAL in the meantime.

Closes #4268.

### DIFF
--- a/io/private/las/Utils.cpp
+++ b/io/private/las/Utils.cpp
@@ -455,6 +455,7 @@ void LoaderDriver::init(int pdrf, const Scaling& scaling, const ExtraDims& dims)
 {
     switch (pdrf)
     {
+    /*
     case 0:
         m_loaders.push_back(PointLoaderPtr(new V10BaseLoader(scaling)));
         break;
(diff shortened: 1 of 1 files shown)


hobu:
Picked back to `2.6-maintenance` at d21c0072a9ba7975d57f86bb5f0bab543ed05a0e

PR #4276, opened 2023-12-05
Fix time tile check for missing tile warning

jp-dark:
Warning was mistakenly being written when not setting tiles and not using time as a dimension.


hobu:
Picked back to `2.6-maintenance` at 9c244fb21ecd316fcc99972a8d54e49f3a230b8d

PR #4277, opened 2023-12-06
Update how TileDB domain is set when not provided by user

jp-dark:
Change:
* Allow both domain and tiles to be set automatically.

Fix:
* Fix getting the GpsTime domain from stats filter.


hobu:
picked back to `2.6-maintenance` at 8a037d820183c1ce83445119201132e40e84206d

PR #4282, opened 2023-12-14
Create dimensions in filters.assign

hobu:
We previously had to use a `filters.ferry` to create a dimension that didn't exist before a `filters.assign`. This removes that hack by allowing `filters.assign` to create dimensions with their expressions.

### DIFF
--- a/filters/private/expr/IdentExpression.cpp
+++ b/filters/private/expr/IdentExpression.cpp
@@ -13,6 +13,20 @@ Utils::StatusWithReason IdentExpression::prepare(PointLayoutPtr layout)
     return false;
 }
 
+
+std::string IdentExpression::name() const
+{
+    const VarNode *n = dynamic_cast<const VarNode *>(topNode());
+
+    n->name();
+    if (n)
+        return n->name();
+    else
+        return std::string("");
+
+}
+
+
 Dimension::Id IdentExpression::eval() const
 {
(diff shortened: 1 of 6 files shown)


test/unit/filters/AssignFilterTest.cpp
```diff
@@ -234,3 +234,40 @@ TEST(AssignFilterTest, test_condition)
...
+    PointTable t;
+    f.prepare(t);
+    PointViewSet s = f.execute(t);
+    PointViewPtr v = *s.begin();
+
+    PointLayoutPtr l = v->layout();
+    Dimension::Id xiDimensionId = l->findDim("xi");
+
+    int ielse = 0;
+    int i6 = 0;
```
rouault, line 266: unused variables:
```suggestion
```

filters/private/expr/IdentExpression.cpp
```diff
@@ -13,6 +13,20 @@ Utils::StatusWithReason IdentExpression::prepare(PointLayoutPtr layout)
     return false;
 }
 
+
+std::string IdentExpression::name() const
+{
+    const VarNode *n = dynamic_cast<const VarNode *>(topNode());
+
+    n->name();
```
rouault, line 21: dereference before nullptr check:
```suggestion
```

PR #4283, opened 2023-12-14
Use current PDAL for workshop

j9ac9k:
Workshop environment should use the current PDAL version.


hobu:
picked back to `2.6-maintenance` at fe67698e87c2c27e1ff29fe8e2c80e2943c8c3e4

PR #4286, opened 2023-12-19
Fix getting/setting the pointFormat of LAZ-files via LasHeader

AlexBass05:
Until now, 'LasHeader' was accessing the 'pointFormatBits' of the internal 'las::Header' directly, when getting/setting the 'pointFormat'. This prevented the 'CompressionMask' from being applied for compressed LAS-files, resulting in the 'pointFormat' being offset by '0x80'.

The fix simply forwards the getter/setter calls of 'LasHeader' to the corresponding functions in 'las::Header', which apply the 'CompressionMask' correctly.

### DIFF
--- a/io/LasHeader.cpp
+++ b/io/LasHeader.cpp
@@ -226,12 +226,12 @@ void LasHeader::setPointOffset(uint32_t offset)
 
 uint8_t LasHeader::pointFormat() const
 {
-    return d->h.pointFormatBits;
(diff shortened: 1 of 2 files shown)


abellgithub:
This seems fine. Thanks.

hobu:
picked back to `2.6-maintenance` at 8b8d1a40d68e21cf638cec77b5e09d0a8c872334

PR #4288, opened 2023-12-19
Add math unary functions to AssignmentStatement in `filters.expression` and `filters.assign`

hobu:


filters/private/expr/MathParser.cpp
```diff
@@ -167,9 +167,29 @@ bool MathParser::primary(Expression& expr)
 
 bool MathParser::function(Expression& expr)
 {
+    using namespace std;
```
abellgithub, line 170: This isn't necessary on OSX. Was there a problem on some other platform? Curious.

hobu: dropped.

PR #4291, opened 2023-12-26
Use standard CMake installation locations

hobu:
PDAL's install location stuff predates standardization of many variable locations by CMake. This PR attempts to align with current CMake installation locations so that packagers can control where things go with normally provided variable overrides as needed.


abellgithub (approved):
Cmake is a wonder

PR #4293, opened 2024-01-05
Added tiledb metadata fields to easily tag array type

normanb:
TileDB is a multi-modal array database with applications outside of geospatial, this PR adds tagging to the array so that the array can be easily recognized by other TileDB client libraries. Currently `raster`, `geometry` and `pointcloud` are used for geospatial array tags.

Related: https://github.com/OSGeo/gdal/pull/9034


hobu:
Picked back to `2.6-maintenance` at 44e8665dbaf06966b06357df1b16079cbe84ee67

PR #4299, opened 2024-01-12
Add ignore_missing_vlrs option to LasReader

wintal:
This allows the user to workaround slightly dodgy las importers and allows PDAL to read some las files already readable by other products


wintal:
This addresses a ticket I opened here: https://github.com/PDAL/PDAL/pull/4299

hobu:
We have precedent with `nosrs=True` in `readers.las` to skip over and ignore whatever baloney the LAS file is providing, so IMO we should merge this too. 

I would feel much better about an Actual Trimble Developer coming on this ticket and telling us the issue is fixed going forward, but given they don't participate in the [LAS Committee](https://github.com/ASPRSorg/LAS), I'm not going to hold out hope.

abellgithub:
I'm not sure how having to use an option to read a bad file is better than running a program to make the file good. And once the file is good, it will work in all contexts. To me it's dumb to depend on all the software around to support corrupt data, and at that, certain kinds of corruption. This option is really "--read_crap_file_created_by_trimble", because there is no knowing that a bad Vlr count couldn't just as well be a bad start of point data.

But I'm happy to lose this debate. It won't impact my retirement. 😀

hobu:
> This option is really "--read_crap_file_created_by_trimble", because there is no knowing that a bad Vlr count couldn't just as well be a bad start of point data.

Yes, and if the Trimble devs show up here for their public flogging it would be even better. The problem is users are stuck in the middle and just want to read their damn points.

hobu:
I don't want to merge this to a maintenance branch because it is a new feature. Merging to `master` for when 2.7 comes out later this year...

wintal:
Thanks @hobu, appreciate it. I'll let you know if we get a response from Trimble, but based on past experience we don't think we will.

PR #4307, opened 2024-01-24
Added threading functionality to filters.overlay

bramton:
There is still one issue which would require some attention. The function `Polygon::contains()` does some initialisation when it is first called. This can be seen here:

https://github.com/PDAL/PDAL/blob/98207ec4fa664f056295f6b5efa28c668f0e426a/pdal/Polygon.cpp#L300-L302

If threading is used, this will lead to a race condition. I have resolved this by first doing mock calls to the `Polygon::contains()` function to make sure the `m_pd->m_grids` is initialised prior to the threads making use of it. I feel this is not a very elegant solution, and I feel there should be a better solution. Any ideas?

### DIFF
--- a/filters/OverlayFilter.cpp
+++ b/filters/OverlayFilter.cpp
@@ -85,6 +85,8 @@ void OverlayFilter::prepared(PointTableRef table)
     m_dim = table.layout()->findDim(m_dimName);
     if (m_dim == Dimension::Id::Unknown)
         throwError("Dimension '" + m_dimName + "' not found.");
+    if (m_threads < 1)
+        throwError("Number of threads should be positive.");
 }
 
 
@@ -144,10 +146,11 @@ void OverlayFilter::ready(PointTableRef table)
     }
     while (feature);
 
-    // Required to initialise m_grids, otherwise this will lead to errors when using threading.
+    // Initialise m_grids, otherwise this will lead to a race condition when
+    // using threading.
     for (const auto& poly : m_polygons)
     {
-        poly.geom.contains(42.0, 42.0);
+        poly.geom.initGrids();
     }
 }
 
@@ -183,7 +186,9 @@ bool OverlayFilter::processOne(PointRef& point)
 
 void OverlayFilter::filter(PointView& view)
 {
-    point_count_t nloops = view.size();
+    point_count_t npoints = view.size();
+    unsigned int chunk_size = npoints / m_threads;
+    if (npoints % m_threads) chunk_size++;
     std::vector<std::thread> threadList(m_threads);
 
     for (int t = 0; t < m_threads; t++)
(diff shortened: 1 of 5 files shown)


abellgithub (commented):
I agree with your assessment about `Polygon::contains()`. I would hoist the first three lines into a separate function that you can call from this code (and call from `Polygon::contains()`).

hobu:
If `filters.crop` is going to have a `threads` option and `writers.las` already has a `threads` option, do we need to worry about sync'ing these to a PipelineManager's option instead of adding them all to individual Stages?

bramton:
> If `filters.crop` is going to have a `threads` option and `writers.las` already has a `threads` option, do we need to worry about sync'ing these to a PipelineManager's option instead of adding them all to individual Stages?

Agreed, a more unified approach towards threading would be a good idea. As a reference, these filter stages already support threading:
* filters.covariancefeatures
* filters.miniball (threading option not documented)
* filters.planefit
* filters.reciprocity (threading option not documented)

Personally, I have been eyeballing the `filters.groupby` stage as another candidate for some threading performance gain.

abellgithub:
We could stick a thread pool in some global space and use that for most things.

filters/OverlayFilter.cpp
```diff
@@ -175,13 +186,30 @@ bool OverlayFilter::processOne(PointRef& point)
...
+    point_count_t npoints = view.size();
+    unsigned int chunk_size = npoints / m_threads;
+    if (npoints % m_threads) chunk_size++;
+    std::vector<std::thread> threadList(m_threads);
+
+    for (int t = 0; t < m_threads; t++)
     {
-        point.setPointId(id);
-        processOne(point);
+        threadList[t] = std::thread(std::bind(
```
bramton, line 196: My C++ knowledge is limited, but is a `std::bind` really necessary here?

filters/OverlayFilter.cpp
```diff
@@ -183,7 +186,9 @@ bool OverlayFilter::processOne(PointRef& point)
 
 void OverlayFilter::filter(PointView& view)
 {
-    point_count_t nloops = view.size();
+    point_count_t npoints = view.size();
+    unsigned int chunk_size = npoints / m_threads;
```
abellgithub, line 190: This should be `point_count_t`, `unsigned int` may be too small.

abellgithub (approved):

filters/OverlayFilter.cpp
bramton: @abellgithub Would you mind sharing your expert view on this?

abellgithub: The `std::bind` is not necessary. The arguments are forwarded to the thread when started. There are a bunch of ways this could be written that would be just fine but I think what's there is fine and would be better if the unnecessary `bind` is removed.

hobu:
Should we also make `filters.crop` multithreaded too?

PR #4321, opened 2024-02-12
CMake: Add WITH_OPENSSL CMake-option

AlexBass05:
Linking against OpenSSL is optional and not necessary if Google storage IO support is not required (or desired). However, until now there was now way of disabling linking against OpenSSL in CMake.

The added WITH_OPENSSL CMake-option is set to TRUE to keep the current default behavior.

### DIFF
--- a/cmake/openssl.cmake
+++ b/cmake/openssl.cmake
@@ -1,7 +1,14 @@
-find_package(OpenSSL 1.1)
-if (OPENSSL_FOUND)
-    set(ARBITER_OPENSSL TRUE)
-else ()
-    message("OpenSSL NOT found - `export OPENSSL_ROOT_DIR=___`")
-    message("Google storage IO will not be available")
+#
+# OpenSSL support
+#
+option(WITH_OPENSSL
+    "Build with OpenSSL for Google storage IO support" TRUE)
+if (WITH_OPENSSL)
+    find_package(OpenSSL 1.1)
+    if (OPENSSL_FOUND)
+        set(ARBITER_OPENSSL TRUE)
+    else ()
+        message("OpenSSL NOT found - `export OPENSSL_ROOT_DIR=___`")
+        message("Google storage IO will not be available")
+    endif ()
 endif ()


abellgithub:
I think what you're saying is that you don't want Google storage support. If that's what you're after, having a way to explicitly disable that would make more sense.

AlexBass05:
The primary intention is to provide a way to avoid having to link against OpenSSL. 

If it is preferred to achieve this with something like `WITH_GOOGLE_STORAGE_SUPPORT`, I'm happy to adapt the PR.

AlexBass05:
How can we move forward with this?

To state my initial intention with this PR again: As someone building and using PDAL as a library, limiting the number of dependencies is very helpful. I don't want to link against OpenSSL. I'm aware that I'm losing some functionality (i.e. Google storage IO support), but that is acceptable for my use case.

I very much like [GDAL's approach](https://gdal.org/development/building_from_source.html#cmdoption-arg-GDAL_USE_EXTERNAL_LIBS-BOOL) for handling optional external dependencies by the way.

abellgithub:
You've already solved this issue for your own need, so this isn't pressing and it's not an error, as such.

There is a tradeoff between simplicity and control. For most users it's easier to fetch dependencies than it is to have to turn things on and off. For development, it can be a bug-inducing hassle to allow features to be turned on and off.

We're discussing the best way move forward with this.

hobu:
> We're discussing the best way move forward with this.

I'm fine to merge this without needing it to be specially tied to "Google support". TBH, I didn't realize we had to link in SSL specially for just that. 

I'm not so enthusiastic about a GDAL-like "No external dependencies" option (the one you linked to only matters with respect to library vendoring, not all external dependencies), because we've never signed up for such a thing. We vendor a number of libraries that matter to us, and then we get shit for doing that instead of externally linking them when the external ones are broken or out of date or don't accept upstreaming of fixes.

abellgithub:
My concern with this solution is that it's pretty opaque. Why would anyone specifically want to turn off SSL? I don't know @AlexBass05 's need here. But it's not clear what you're losing when you turn it off or why you would want to do it.

hobu:
> My concern with this solution is that it's pretty opaque.

so make it opt-in instead of opt-out? or make it "support google cloud storage" and have it require OpenSSL? I have no strong opinion, but I recognize the need for a switch of some kind.

jjimenezshaw:
Imagine that I am making an application that links with PDAL. I am distributing PDAL library like all the rest of my dependencies.
If my application is not using OpenSSL at all, I would like to avoid the compilation and link of a library that I do not need, and that is external to PDAL.
Currently if it is present on my building machine (because the guy that maintains CI installed it), then it is a dependency of PDAL, and therefore a transient dependency of my application. If it is no present in the machine for the final user of the application, the application cannot even start (and remember that I do not need OpenSSL functionality at runtime). So my only option is to care and include a library, just in case it appears in CI.
It is also extra work to make security updates... of something that I don't need at all.

The case of a dependency in CI is not fictional. It happened with GDAL and zstd (and several others). I don't now why but at some point zstd appeared in CI (probably due to another software/package installed in that machine). Later the application was not starting in the final user.

That's why I find useful "No external dependencies" from GDAL. First I set that, an later I enable what I really need for my application and take care. I know what are my dependencies, keep them up to date, and package them with my application. With that option I know I am safe, not including any dependency "by accident". Otherwise I have to disable manually all the dependencies from GDAL... that are many. And I am not safe against any new dependency added with a new development. It is something like "all off by default, regardless what is installed; enable what you really need".

hobu:
Active over here now #4323 4323

PR #4328, opened 2024-02-20
make filters.expression evaluate multiple expressions

hobu:
Stream mode cannot work for multiple expressions. If you pass multiple expressions and try to execute in stream mode it will throw.

### DIFF
--- a/filters/ExpressionFilter.cpp
+++ b/filters/ExpressionFilter.cpp
@@ -107,7 +107,7 @@ void ExpressionFilter::prepared(PointTableRef table)
 
 bool ExpressionFilter::processOne(PointRef& point)
 {
-    if (!(m_args->m_expressions.size() == 1))
+    if (m_args->m_expressions.size() != 1)
         throwError("Streaming of expressions only works with a single expression");
 
     bool status = m_args->m_expressions[0].eval(point);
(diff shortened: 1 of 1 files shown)


filters/ExpressionFilter.cpp
```diff
@@ -82,48 +82,70 @@ void ExpressionFilter::addArgs(ProgramArgs& args)
...
         if (!status)
             throwError("Invalid 'where': " + status.what());
+    }
 }
 
 
 bool ExpressionFilter::processOne(PointRef& point)
 {
-    bool status = m_args->m_expression.eval(point);
+    if (!(m_args->m_expressions.size() == 1))
```
abellgithub, line 110: I just think the style difference here are funny. (I really laughed.) I would write "!= 1" instead of "(!(==1)". Nothing wrong, but interesting how people think of these things differently.

And below, I'd never use a temporary just to return it.

Wonder why we end up doing the things we do. Some we may have reasons for. Other things, probably not.

filters/ExpressionFilter.cpp
```diff
@@ -82,48 +82,70 @@ void ExpressionFilter::addArgs(ProgramArgs& args)
...
-        if (processOne(p))
-            outView->appendPoint(*inView, p.pointId());
+        PointViewPtr outView = inView->makeNew();
+        views.insert(std::make_pair(&expr, outView));
     }
 
-    viewSet.insert(outView);
+    // eval our expression across each point for each view.
+    // TODO: make this threaded
+    for(auto& p: views)
```
abellgithub, line 134: Had a comment here about reversing order, but I'm not sure where it went. You don't want to loop points more than once and you want to avoid using a map if possible. Performance reasons.

hobu: addressed

filters/ExpressionFilter.cpp
hobu: addressed

PR #4341, opened 2024-02-29
Fix radius parameter in `erodeDiamond` for `SMRFilter`

mone27:
I was going through the SMRF paper and the implementation in PDAL and the [original matlab implementation](https://github.com/thomaspingel/smrf-matlab/tree/master) and I think that I found a bug in the SMRFilter implementation.

The radius parameter for the `erodeDiamond` function was always `1` and didn't change with the radius like in the `dilateDiamond` function.

### DIFF
--- a/filters/SMRFilter.cpp
+++ b/filters/SMRFilter.cpp
@@ -672,7 +672,7 @@ std::vector<int> SMRFilter::progressiveFilter(std::vector<double> const& ZImin,
     {
         // "On the first iteration, the minimum surface (ZImin) is opened using
         // a disk-shaped structuring element with a radius of one pixel."
-        math::erodeDiamond(erosion, m_rows, m_cols, 1);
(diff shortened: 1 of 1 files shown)


chambbj:
@mone27 Thanks for the detailed look at the code. Upon refreshing my own memory though, I feel confident that we have it right. The erosion ground surface is initialized to `ZImin` outside of the for loop, and then erodes itself (not `ZImin`) with radius one for each iteration. By the end, it has eroded the full `max_radius`. The dilation is applied differently because we have to essentially undo the **cumulative** number of erosions at each iteration of the loop. It's a change we made [seven years ago](https://github.com/PDAL/PDAL/pull/1541) to speed up the progressive filtering.

mone27:
Hi @chambbj thanks for your prompt reply!
Thanks for your explanation! I actually missed the fact that `erosion` was iteratively updated at every loop. Sorry for the useless PR and the wasted time.
Would be something useful to add a comments in the code?

PR #4344, opened 2024-03-01
Add Workshop Setup Instructions for ArcGIS Users

j9ac9k:
Added the lines to enter in the Python Command Prompt to get the workshop environment setup if the only version of conda available is the one that ESRI provides with ArcGIS


hobu:
picked back to `2.6-maintenance` at 04d5ac32af00e68e85f8c4d463648c6cbcac35d9

PR #4345, opened 2024-03-03
Add filters.h3 to add H3 index values to points

hobu:


abellgithub (commented):
Nice.

hobu:
> Nice.

I think I'm going to vendor the h3 code though. It is too finicky on different environments to depend on the headers being available and exported as expected.

PR #4353, opened 2024-03-12
Adding log pointer to stac class instantiation to connect to reader logs

kylemann16:
Adding log pointer to stac classes so we can better output logs during the processes, rather than waiting until after the whole thing is done to output logs.


hobu:
I don't think this is correct. 

`StacReader::initialize()` calls `StacReader::initializeArgs()` which grabs the `log()` at https://github.com/PDAL/PDAL/blob/62b1bb32762211fa21a32bbbfc3b65de9f8de5e1/io/StacReader.cpp#L329

It's always been my understanding that the log is not wired up and cannot be used in any stage's `initialize` function. @abellgithub can confirm, but I think we're just getting lucky here with the ✅

hobu:
picked to `2.7-maintenance` at 3fdaa2e54816b87ca04192790cc49ea24a89063f

PR #4361, opened 2024-03-19
Don't break strict aliasing rules.

abellgithub:


hobu:
picked back to `2.7-maintenance` at e7a179c963a53387a4a0eac0cf16a56c914fb738

PR #4364, opened 2024-03-21
Add documentation to readers.stac for curl environment settings

kylemann16:
Add section to Readers.STAC docs show show how to set curl verbosity and curl timeout from command line.


kylemann16:
Didn't see this was already done.

PR #4365, opened 2024-03-21
we do not need to bring in GDAL for the Arrow plugins

hobu:


hobu:
picked back to `2.7-maintenance` at e828c34dcf3ce34b41a4e48682879a0ef365c88d

Issue #4318, opened 2024-02-09
filters.litree never terminating


WhiteSte:
Hello everyone,
Inside details section you can find the pipeline. (Since the file is really dense i suggest you to keep the crop filter to speed up the execution)
<details>

```
 {
                "pipeline": [
                    {
                        "filename": "/Users/steve/Desktop/Lavoro/ArchiMethodSA/spacepix/projects/raw_data/SwissSurface3D/2700-1226.las",
                        "type": "readers.las",
                        "spatialreference": "EPSG:2056"
                    },
                    {
                        "distance": 150,
                        "type": "filters.crop",
                        "point": "POINT(2700774.3 1226782.9 774.0)"
                    },
                    {
                        "type": "filters.range",
                        "limits": "Classification[3:3]"
                    },
                    {
                        "raster": "/Users/steve/Desktop/Lavoro/ArchiMethodSA/spacepix/projects/raw_data/ground/2700-1226.tif",
                        "type": "filters.hag_dem"
                    },
                    {
                        "count": 30000,
                        "decay": 0.9,
                        "type": "filters.relaxationdartthrowing",
                        "radius": 0.9
                    },
                    {
                        "type": "filters.sort",
                        "dimension": "HeightAboveGround",
                        "order": "DESC"
                    },
                    {
                        "min_points": 10,
                        "min_height": 3,
                        "type": "filters.litree",
                        "radius": 5
                    }
                ]
            }
```

</details>

 the filter litree there goes into an infinite loop. 
The following is the last print

```
(javapipeline filters.litree Debug)             Max HAG value of 37.2397
(javapipeline filters.litree Debug) Classifying points for tree 242 (|Ui| = 22638)
(javapipeline filters.litree Debug)                     |Pi| = 1, |Ni| = 22638
(javapipeline filters.litree Debug) Minimum cluster size not met; skipping tree.
(javapipeline filters.litree Debug)             Max HAG value of 37.2397
(javapipeline filters.litree Debug) Classifying points for tree 242 (|Ui| = 22638)
```

As you see it get stuck with 22638 remaining points classyfing for tree 242. 

The problem is resolved and terminates correctly if i set for example ` "radius": 50 ` in the `litree.filter`

Idk if it's wise to add a check that if the number of points is not decreasing from the previous step it leaves (or at least if that happens a certain tolerance number of times)

Since the two files are 400mb of size i'm putting the link to download a .zip (this link will expire 10/03/2024)
[...]

PR #4366, opened 2024-03-22
[filter.litree]  Try to fix an infinite loop when the seed radius is too little

WhiteSte:
This PR follows from #4318. 

If the seed radius is too little the initial dummy point `n0` is set too near to the tree. The result will be in having |Pi| = 1 and |Ni| = |Ui| . Because of that the algorithm goes into an infinite loop. 

This PR is aimed to solve the above mentioned issue trying to expand automatically the radius whenever this condition occur. For now the radius is increased by 10  `  m_dummyRadius += 10; // decide here how to increase the radius` . It could be possible to change this behaviour scaling it by percentage for example. 

I'm not confident with C/C++, i have no idea on how to test the proposed code, if it could work or not.

### DIFF
--- a/filters/LiTreeFilter.cpp
+++ b/filters/LiTreeFilter.cpp
@@ -235,6 +235,8 @@ void LiTreeFilter::segmentTree(PointView& view, PointIdList& Ui,
     PointId n0 = locateDummyPoint(view, Ui, t0);
     Ni.push_back(n0);
 
+    double initialRadius = m_dummyRadius; // Remember the initial radius to reset if needed
+
     // "For iteration i, we classify each point in Ui as Pi or Ni."
     double tx = view.getFieldAs<double>(Id::X, t0);
     double ty = view.getFieldAs<double>(Id::Y, t0);
@@ -252,6 +254,34 @@ void LiTreeFilter::segmentTree(PointView& view, PointIdList& Ui,
     log()->get(LogLevel::Debug3)
         << "|Pi| = " << Pi.size() << ", |Ni| = " << Ni.size() << std::endl;
 
+   // Check if Ni is equal to Ui, indicating the radius might be too short
+    if (Ni.size() == Ui.size() && Pi.size() == 1)
+    {
+        log()->get(LogLevel::Warning) << "Expanding search radius as |Ni| == |Ui|.\n";
+        m_dummyRadius += 10; // decide here how to increase the radius
+
+        // Reset Ni and retry segmentation with the updated radius
+        Ni.clear();
+        
+         n0 = locateDummyPoint(view, Ui, t0); // Recompute n0 with the new radius
+      
+        Ni.push_back(n0); // Reinsert the dummy point with the updated radius
+
+        // Reclassify points with the updated radius
+        for (PointId const& ui : Ui)
+        {
+            double ux = view.getFieldAs<double>(Id::X, ui);
+            double uy = view.getFieldAs<double>(Id::Y, ui);
+            double d = (ux - tx) * (ux - tx) + (uy - ty) * (uy - ty);
+            if (d < 100.0)
+                classifyPoint(ui, view, Ni, Pi);
+            else
+                Ni.push_back(ui);
+        }
+
+        m_dummyRadius = initialRadius; // Reset the radius for future iterations
+    }
+  
     // Use Pi to assign current tree_id to TreeID dimension, only if minimum
     // size is met
     if (Pi.size() >= m_minSize)


chambbj:
Have you run this on your problem dataset? Did it fix the problem?

The rule for the "dummy radius" is simply that it be "far away" from the initial tree point. Did you try first running with a larger `radius` value (instead of forcing it to grow iteratively as shown here)?

I'm further confused because I thought that you said in #4318 that you solved the problem by actually using a smaller and not larger radius.

I certainly agree that this needs to be handled -- I'm just unsure of the proper solution.

WhiteSte:
> Have you run this on your problem dataset? Did it fix the problem?

I never coded in c/c++, i don't even know how to run it. I just tried a logical workaround. Should i compile myself the entire software in order to test it? 

>The rule for the "dummy radius" is simply that it be "far away" from the initial tree point. Did you try first running with a larger radius value (instead of forcing it to grow iteratively as shown here)?

Yes, running a larger radius solves the problem. The fact is that i saw a lower one gives better results, but at the same time i cannot find a general parameter that fits good for all the datasets are processed. (I'm not running it by hand, i've implemented an automatic solution)

>I'm further confused because I thought that you said in https://github.com/PDAL/PDAL/issues/4318 that you solved the problem by actually using a smaller and not larger radius.

nope in the issues i said the opposite

> I certainly agree that this needs to be handled -- I'm just unsure of the proper solution.

Me aswell :) Do you mean is not the proper solution the fact that the radius automatically grows or how i did it?

chambbj:
> > I'm further confused because I thought that you said in #4318 that you solved the problem by actually using a smaller and not larger radius.
> 
> nope in the issues i said the opposite

I see. I forgot that you had been running with a very small radius initially. Keep in mind that the default is 100 m, not 5 m (first attempt) or 50 m (fix). Interesting that you had better luck with 5 m. Depending on the tree, that seems like it would be quite close to be randomly selecting a point expected to not be part of the current tree segment.

We should keep in mind also that the dummy point isn't guaranteed to be radius distance away in the current implementation. We are simply searching for points within that radius and taking the farthest away. As the points available to the set `U` tend toward zero, this could in fact be much smaller than the requested radius (and subsequently, requesting a larger and larger radius may not actually give you a different answer). I remember thinking it would be nice to simply select the farthest available point at each iteration, but that's quite complex to recompute.

One possibly workaround that I remember considering when the issue was first posted is to skip the point `Pi` that was last attempted and to move on to the next highest available point to see if it had any luck generating a new tree segment. I just never got around to coding that logic.

WhiteSte:
>Interesting that you had better luck with 5 m.

worst luck with 5m. That was the parameter sending the algorithm into the loop.

> I remember thinking it would be nice to simply select the farthest available point at each iteration, but that's quite complex to recompute.

Here i have a dubt about the dummy point: Isn't it setting a constraint in the dataset? (any point farther to it is automatically excluded from being a cluster point). That's why i wanted to reduce the seed radius , in order to speed up the computation,  but maybe my idea is simply wrong?

> move on to the next highest available point 

Also here i'm not really confident with the algorithm, but isn't that creating big bugs? I mean: if you start skipping the highest point, jumping around, it can happen that then all of these points will never be clustered. The descending property seems to me being really important in the concept


> and subsequently, requesting a larger and larger radius may not actually give you a different answer

There could be a check that if the radius exceed the total radius of the dataset everything stops. 

----

Since we are on the topic i'm posting here another "problem". Let me know if it should go inside another issue.
I don't know if this was intentional in the original paper (That i've never read :D) but as a cluster reaches the min amount of points it will be closed and it will start clustering another tree. If you have trees with very different sizes/densities that could lead to the behaviour in the screenshot below . Basically a tree is decomposed in several clusters. 
<details>
  <summary>
    Click to view the screenshot
  </summary>
    Screenshot
</details>

chambbj:
> Isn't it setting a constraint in the dataset? (any point farther to it is automatically excluded from being a cluster point).

Not exactly. There is a separate check for this, which isn't really exposed to users at the moment. Instead, the classification of tree points segments all unclassified points into two sets. The first are those points belonging to the current tree segment, the rest are points that do not belong to the tree. The highest unclassified point is the seed for the tree cluster and the far away dummy point is the seed for the non-tree cluster. Some of the thresholds factor in the distance from the point in question to each of these growing clusters.

> I mean: if you start skipping the highest point, jumping around, it can happen that then all of these points will never be clustered. 

I've not thought through this carefully and certainly haven't testing anything, but my intuition was that if the next highest point (after the failed attempt) was part of the same tree, then the point we just skipped would likely just get grouped with the current attempt. Or maybe that particular point (maybe a tree?) never gets clustered, but at least the others still do. It would be interesting to know which point is actually causing this failure. Might it be that there is high noise in the dataset that is causing problems?

If you want to repost the link to the data, I can take a look. I tried last week and it appeared to be dead.

chambbj:
And for your unrelated issue...

> as a cluster reaches the min amount of points it will be closed and it will start clustering another tree

That isn't how the algorithm works. The minimum number of points doesn't terminate clustering. It ensures that any cluster is _at least_ that many points, but it can certainly continue growing.

You should consider giving the paper a read, although it won't address this particular behavior, which is something we implemented ourselves to ensure that we weren't getting noise clusters output. https://www.maine.gov/geolib/programs/lidar/ForestrySegmenting_Trees.pdf

WhiteSte:
>Might it be that there is high noise in the dataset that is causing problems?
If you want to repost the link to the data, I can take a look. I tried last week and it appeared to be dead.

The dataset is really at high resolution, you can check, I've uploaded it again. There shouldn't be much noise
[las+tif.zip](https://www.swisstransfer.com/d/91e1f9e0-d587-4e59-88ce-bb7bff86387f)

>It ensures that any cluster is at least that many points, but it can certainly continue growing.

Under which circumstances? I will take a deep look to the paper soon

What i'm thinking is that this algorithm works pretty well when you have a lot of trees one near the others, but not when they are isolated . Is that wrong? Because i played a bit with the parameters and i could cluster the tree in the screenshot only increasing the min points required (but of course i miss all the others)

hobu:
#4373 is an attempt to solve the issue.

PR #4369, opened 2024-03-25
use CONFIG for Draco and Arrow. Relax Draco to 1.5+

hobu:


hobu:
picked back to `2.7-maintenance` at 06aee3866fc4f4246addefc3ed24575252da3abc

Issue #4368, opened 2024-03-24
TileDB Config File param "sm.dedup_coords true" not used for array creation


weidinger-c:
Based on the tutorial https://docs.tiledb.com/main/integrations-and-extensions/geospatial/pdal it states that per default TileDB allows duplicates. 
To able to update point cloud  coordinates inside the database, duplicates should be set to false. The guide states it as follows:

_**Ingesting LAS Files**
First, create a TileDB config file tiledb.config where you can set any [TileDB configuration parameter](https://docs.tiledb.com/main/how-to/configuration#configuration-parameters) (e.g., AWS keys if you would like to write to a TileDB array on S3). Make sure you also add the following, as currently TileDB does not handle duplicate points (this will change in a future version)._

I tried that step by creating a file named tiledb.config with the content:
sm.dedup_coords true

But still, after using the PDAL TileDB writer, the schema of the created database reports "allows_duplicates = true".
I guess It comes from the code part https://github.com/PDAL/PDAL/blob/a2c196deec2f5aee1f5b06b5f117ef82967a69d5/plugins/tiledb/io/TileDBWriter.cpp#L292
where the schema is always set to allow duplicates, no matter what config file was provided:
Line 292: schema.set_allows_dups(true);

normanb:
@weidinger-c that is an issue, the main docs on our website lag behind the code, we are addressing that. What is your use case, do you need to set `dups` to false. I will update the code.

I would suggest that you want `duplicates` and that you couple that with deletion - https://docs.tiledb.com/main/how-to/arrays/writing-arrays/deleting Then you can add a point in the same place and delete the previous point.

weidinger-c:
@normanb yes i need `dups = false` 
My usecase is that I load a lot of las files, compute some algorithms on parts of the data and in the end want to update some attributes e.g. Classification for the points. Currently, the points are added a second time instead of replacing the attribute values, which IMO is due to `dups` being true at the moment. Thanks for the help!

normanb:
@weidinger-c to achieve this workflow you will need to replace the point in two stages, delete the point and then write. `dups` is misleading in this case, you can't update the existing point by setting `dups = false`

weidinger-c:
@normanb This is quite confusing for me. TileDB is database where I cannot change the stored data unless I delete a whole entry first to then insert it again?

normanb:
@weidinger-c I am going to create a couple of python examples which I will share here so we can explore what you need.

[...]

PR #4370, opened 2024-03-25
TileDB option to enable/disable duplicate points

normanb:
Removed hard coded setting of allow duplicates property in the TileDB array schema. I have added a test that checks that duplicate points are removed on read.


normanb:
Related issue: https://github.com/PDAL/PDAL/issues/4368

hobu:
picked back to `2.7-maintenance` at ad4eeb09bf68b41a4b3d6cb41ac939fccfd6c061

PR #4371, opened 2024-03-25
Add writers.copc.sort to allow user to change the dimension that chun…

hobu:
Add `writers.copc.sort` to take in a dimension name to allow users to control what dimension the chunks are sorted upon. This defaults to `GpsTime`, which was what has always been used.

There are situations where you might want to control the sorting. For example, if you are dropping the `GpsTime` dimension from your data, the default behavior is going to be random in the chunk (and extra compute). 

## Sorting behavior INCLUDING normal existing `GpsTime` values

``writers.copc`` normally sorts the data on `GpsTime`. As you can see this is usually the right thing to do.

image

## Sorting behavior after ZEROING OUT `GpsTime` values

However, if the `GpsTime` values are `0`, we probably want to sort on something else...

image


hobu:
reverted in `c4fec5d0cceec8f927fef079e9ac7e469780fd87`

PR #4372, opened 2024-03-25
Make sure all stages are initialized before prepared is called.

### LINKED ISSUE #4356: prepared() called before all stages are intialized
Currently prepared() is invoked for each stage right after initialize(). This is wrong. prepared() shouldn't be called until *all* stages have been initialized() to ensure that all dimensions have been added by all stages before prepared() is invoked.

abellgithub:
Close #4356

### DIFF
--- a/pdal/Stage.cpp
+++ b/pdal/Stage.cpp
@@ -159,17 +159,42 @@ QuickInfo Stage::preview()
 
 void Stage::prepare(PointTableRef table)
 {
-    m_args.reset(new ProgramArgs);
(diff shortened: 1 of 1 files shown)


hobu:
Why does tightening this up matter at this point in time?

abellgithub:
It has always been wrong. If you use prepared to check if a dimension exists and the stage that adds that dimension hasn't been processed yet, you'll get wrong behavior.

I don't think I'd merge this right before release, but I think it's pretty low risk.

PR #4377, opened 2024-03-29
we don't need to link GDAL for readers.arrow and writers.arrow

hobu:
#4365 wasn't done enough.


hobu:
picked back to `2.7-maintenance` at 0b9868acbd010d68b232111dbac432f490a0d9ae

PR #4378, opened 2024-03-29
Docs - Update links on Python page

jomey:
Update external links for:
- Numpy
- Conda
- PDAL url on PyPi
- Pip installation

The last was moved to a different url. The other two now point to the https version and 'latest' in the conda case.

[skip ci]


hobu:
Thanks!

PR #4385, opened 2024-04-08
use PROJ cmake CONFIG package instead of homespun find_package

hobu:


hobu:
Pushed back to `2.7-maintenance` at 8a9db9bf9bdadb0613040da210283f397eeebbcc

PR #4388, opened 2024-04-15
PG COPY mode support for pgpointcloud plugin

Nicolasribot:
Allows `writers.pgpointcloud` to use PostgreSQL COPY mode, by adding a new config parameter named `pg_use_copy`.
COPY is significantly faster than INSERT.

Set `pg_use_copy` to `true` to use copy instead of INSERT statements (default to `false`, keeping original INSERT mode).

(a new method was added to `PgWriter` class to avoid messing-up too much with the code. A factorisation should be done between `writeTile` and `copyTile` (the new method))

### DIFF
--- a/doc/stages/writers.pgpointcloud.rst
+++ b/doc/stages/writers.pgpointcloud.rst
@@ -44,7 +44,8 @@ Example
           "connection":"host='localhost' dbname='lidar' user='pramsey'",
           "table":"example",
           "compression":"dimensional",
-          "srid":"26916"
+          "srid":"26916",
+          "pg_use_copy": true
       }
   ]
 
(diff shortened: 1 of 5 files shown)


abellgithub:
What is the benefit of this? I get the use of COPY when you're taking data from postgres to a file, but here we're going to the trouble of using stdin -- things aren't really going to a disk file.

abellgithub:
If this is the case, is there some reason not to use COPY exclusively?

abellgithub:
My strong preference is not to provide options unless there's real value to them. I don't see that here. Also, our postgres driver gets very little use at this point (I think). I'd be happy to go through the code and fix it up if you can test -- I haven't had a postgres setup in years.

PR #4394, opened 2024-04-21
rotation matrix example corrected

ovipaul:
The correct matrix for a 90-degree counterclockwise rotation about the Z-axis is as follows:

| 0  -1  0  0 |
| 1   0  0  0 |
| 0   0  1  0 |
| 0   0  0  1 |

But on the example json file it was showing this:
"matrix":"0  0  -1  0  1  0  0  0  0  0  1  0  0  0  0  1"

Correct matrix would be:
"matrix": "0  -1  0  0  1  0  0  0  0  0  1  0  0  0  0  1"

### DIFF
--- a/doc/stages/filters.transformation.rst
+++ b/doc/stages/filters.transformation.rst
@@ -150,7 +150,7 @@ shown below.
   [
       {
           "type":"filters.transformation",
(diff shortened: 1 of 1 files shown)


hobu:
picked back to `2.7-maintenance` at ad0f1bd370efcb456b7861590ed1ab25e7ada9b3

PR #4397, opened 2024-04-23
use mambabuild for conda package build

hobu:


hobu:
picked back to `2.7-maintenance` at 56d17794fe12d3048ca7363f4b652080e965f330

PR #4398, opened 2024-04-27
CI: conda.yml: change macos-latest to macos-13

rouault:
GHA has changed the macos-latest alias to macos-14 (arm64), hence to get Intel Mac one now needs to explicitly select macos-13 Cf https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -23,7 +23,9 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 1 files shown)


hobu:
I wonder if we even care about x86 OSX at this point.

hobu:
Adapting this in #4397

PR #4400, opened 2024-05-07
Fixed grammer, sentencing, colons and added text to show that autzen should be included in files

TomTheTonk:
Autzen already in the drive so we can add the text to show users have it identifying Clipping data with polygons
dataformat_id missing period at end in colorizing points with imagery
dataformat_id missing period in removing noise
4. Right click on density.sqlite has a period at the end when the other instructions dont in visualizing acquisition density
Many of the exercise mix having and not having periods for the instructions should pick one and only use that and the same with colons
Notes 3. in generating DTM is messed up fixed that


SuggestedChanges.md
```diff
@@ -0,0 +1,9 @@
+
```
hobu, line 1: Let's not add this file to the repo

TomTheTonk: opps I thought I removed it

hobu:
picked back to `2.7-maintenance` at 3727ab632e40fe61b9d71bc54b5e19ef2ee8331e

Issue #4403, opened 2024-05-22
pdal_io_ogr_writer_test is failing with GDAL 3.9


imincik:
<!--
Please note that GitHub issues are meant to report bugs or feature requests, and are not intended for "How to" questions or general support. GitHub issues requesting such support will be closed. Luckily, we have an active community on the GIS StackExchange [1] and the PDAL Mailing List [2] that are more than happy to chime in with advice in these situations!

Also note, that issues installing PDAL via Conda should be directed to https://github.com/conda-forge/pdal-feedstock/issues/new. Similarly, issues with the PDAL Python bindings installed via Conda should be directed to https://github.com/conda-forge/python-pdal-feedstock/issues/new.

[1] https://gis.stackexchange.com/questions/tagged/pdal
[2] https://pdal.io/community.html#mailing-list
-->

**Describe the bug**
`pdal_io_ogr_writer_test` is failing when building with GDAL 3.9.0 in Nixpkgs.

```
$ nix-build -A pdal

...

42/129 Test  #42: pdal_io_ogr_writer_test ......................***Failed    5.75 sec                                                                                                                                                        │
│[==========] Running 15 tests from 1 test suite.                                                                                                                                                                                              │
│[----------] Global test environment set-up.                                                                                                                                                                                                  │
│[----------] 15 tests from OGRWriterTest                                                                                                                                                                                                      │
│[ RUN      ] OGRWriterTest.shapefile                                                                                                                                                                                                          │
│[       OK ] OGRWriterTest.shapefile (421 ms)                                                                                                                                                                                                 │
│[ RUN      ] OGRWriterTest.json                                                                                                                                                                                                               │
│[       OK ] OGRWriterTest.json (431 ms)                                                                                                                                                                                                      │
│[ RUN      ] OGRWriterTest.geopackage                                                                                                                                                                                                         │
│[       OK ] OGRWriterTest.geopackage (448 ms)                                                                                                                                                                                                │
│[ RUN      ] OGRWriterTest.creation_options                                                                                                                                                                                                   │
│/build/source/test/unit/io/OGRWriterTest.cpp:121: Failure                                                                                                                                                                                     │
│Expected equality of these values:                                                                                                                                                                                                            │
│  compare(infofile, outinfofile, 1)                                                                                                                                                                                                           │
│    Which is: 4                                                                                                                                                                                                                               │
│  0                                                                                                                                                                                                                                           │
[...]

PR #4404, opened 2024-05-22
Fixing typo that caused Info to not fill gpstime_maximum on copc reads.

kylemann16:
Info struct was not properly filling `gpstime_maximum` when doing `Info.fill`.


hobu:
Docker test failure here is due to #4403

hobu:
picked back to `2.7-maintenance` at 4519df547444717879ea667f072b3438584cc7c5

PR #4414, opened 2024-05-31
Warn instead of error for when global encoding is not set

### LINKED ISSUE #4412: Change 6-10/WKT error to a warning.
Changing the following error to a warning would allow PDAL to proceed with broken files.

> Global encoding WKT flag not set for point format 6 - 10.

hobu:
Close #4412


hobu:
picked back to `2.7-maintenance` at 87473eb095da22ae5f6009a614f95a647e04c9db

PR #4415, opened 2024-05-31
Add 'dimension' output to filters.sample

hobu:
If the user sets a `dimension` option on `filters.sample`, a Uint8 dimension with that name will be populated 1 or 0 whether or not the point was sampled instead of culling points.

This PR also added the ability to set the voxelization origin explicitly with `origin_x`, `origin_y`, and `origin_z` options.


doc/stages/filters.sample.rst
```diff
@@ -73,6 +73,20 @@ cell
   Voxel cell size. If ``radius`` is set, ``cell`` is automatically computed
   such that the cell is circumscribed by the sphere defined by ``radius``.
 
+dimension
```
abellgithub, line 76: Not sure I love this name. Maybe `indicator_dim` or some such? Names are hard.

filters/SampleFilter.cpp
```diff
@@ -90,8 +103,24 @@ PointViewSet SampleFilter::run(PointViewPtr view)
...
+
             output->appendPoint(*view, point.pointId());
+        } else
+        {
+            // If the user set a dimension, write 0s be cause we didn't
+            // keep the point. Otherwise we drop it on the floor.
+            if (m_dimension != pdal::Dimension::Id::Unknown)
+            {
+                point.setField(m_dimension, 0);
+                output->appendPoint(*view, point.pointId());
```
abellgithub, line 121: `appendPoint` is incorrect here.

===

You could use this litter helper or move its logic into `voxelize`

```
bool func(point)
{
  bool keep = voxelize(point);
  if (m_dimension != pdal::Dimension::Id::Unknown)
  {
    point.setField(m_dimension, keep);
    keep = true;
  }
  return keep;
}
```
In run():

```
if (func(point))
  output->appendPoint(*view, point.pointId());
``` 

In processOne():

```
return func(point);
```

abellgithub (changes requested):
Bug needs fixing.

doc/stages/filters.sample.rst
hobu: We use this name in `filters.overlay` already, so I'm just following suit.

filters/SampleFilter.cpp
hobu: updated as suggested

PR #4416, opened 2024-06-03
fix on gridDecimation filter

alavenant:
some fix on the grid decimation filter


hobu:
@alavenant can you describe with more detail why you added this? What was working incorrectly?

abellgithub:
This needs a test that fails.

alavenant:
First of all, i add a las file wich is square. This allows to determine easily the number of cells, so the number of endpoints. So it's help for the final test.
 
Then I corrected the way to determine the grid (bounds of each cells). In the first version, there was one column too many. And the choice of cell of a point was not always good.

alavenant:
@abellgithub, I will add a tests that fails !

filters/GridDecimationFilter.cpp
```diff
@@ -107,9 +111,9 @@ void GridDecimationFilter::processOne(BOX2D bounds, PointRef& point, PointViewPt
 
 void GridDecimationFilter::createGrid(BOX2D bounds)
 {
-    double d_width = std::floor((bounds.maxx - bounds.minx) / m_args->m_edgeLength) + 1;
-    double d_height = std::floor((bounds.maxy - bounds.miny) / m_args->m_edgeLength) + 1;
-    
+    double d_width = std::round((bounds.maxx - bounds.minx) / m_args->m_edgeLength);
```
abellgithub, line 114: I think the original is correct here. Consider an edge length of 1 and a min/max of [0, 1.1].

abellgithub: I will add that if you have, say, an edge length of 10 and bounds of [0, 30], you need 4 cells, as the boundaries are:
[0, 10), [10, 20), [20, 30), [30, 40).

abellgithub:
I'm confused by this. The test doesn't use data that can be easily inspected so it's hard to tell what case we're actually trying to check.

Can you make a simple, text-based input set with a few points that shows the issue and solution?

alavenant:
add a text example for tests. 

And add fix to be sure to deal with all the points. In the example, the 0.500000000001 resolution causes some division problems ; problems I fix with the test of the cell and with the expension of the map (+2 at the link of +1). Nevertheless, this case doesn't seems to be usual...

hobu:
Is this PR still relevant?

hobu:
I pulled this over to #4880

PR #4421, opened 2024-06-11
E57 plugins need to link against PDAL when building standalone

hobu:


hobu:
It is now possible to build plugins that require external libraries in what I called a "standalone" build configuration. Consider the following:

1. Build and install a PDAL that doesn't have plugins like E57 or CPD or HDF or TileDB.
2. Grab a source tree and go into a plugin directory and configure the plugin:
  ```
   cmake -G Ninja $CMAKE_ARGS \
      -DSTANDALONE=ON \
      -DCMAKE_BUILD_TYPE=Release \
      -DCMAKE_INSTALL_PREFIX=$PREFIX \
      -DCMAKE_PREFIX_PATH=$PREFIX \
      -DPDAL_DIR:PATH=$PREFIX/lib/cmake/PDAL \
      ..
  ```
3. Build and install the plugin to your `PDAL_DRIVER_PATH`:
  ```
      ninja install
  ```

We are using this to [break apart the PDAL Conda package](https://github.com/conda-forge/pdal-feedstock/pull/247) to allow plugins with esoteric dependencies to be independent from the base PDAL library that has otherwise pretty boring stock geospatial library dependencies.

hobu:
picked back to `2.7-maintenance` at d3aa57fe2d30ec1ff18c8847265454cf32544beb

Issue #4418, opened 2024-06-06
Change default behavior of `coord_id` in BPF writer


chambbj:
Currently, the default behavior of the BPF writer is to default to a `coord_id` of 0 when none is provided. This triggers setting the `coord_space` to 0 also, which indicates a Cartesian coordinate system. While this is a valid possibility, it is relatively uncommon (at least IMO). This means that a simple `pdal translate` from LAS to BPF, even with known SRS in the LAS file, will result in a Cartesian output in the BPF. The BPF writer could instead use its existing `auto` capability to attempt to detect the correct SRS from the source.

abellgithub:
Is this a request to have the default value of `coord_id` be `auto` instead of `0`?

abellgithub:
I'm thinking @hobu would have no problem with this.

hobu:
> I'm thinking @hobu would have no problem with this.

✅ 

@hobu asked @chambbj to ticket it 😄

PR #4428, opened 2024-06-26
implement 'auto' as default CRS for BPF writer #4418

hobu:


io/BpfWriter.hpp
```diff
@@ -56,6 +56,9 @@ class PDAL_DLL BpfWriter : public FlexWriter
         CoordId() : m_auto(false), m_val(0)
         {}
 
+        CoordId(bool isAuto, int val) : m_auto(isAuto), m_val(val)
```
abellgithub, line 59: AFAICT, you don't need this or the header change. You can just change `m_auto` to `true` in the above ctor and that's it.

hobu: My compiler complained when I tried to do that without making a ctor.

abellgithub: Where was that code?

hobu: Neither of these work without making the ctor
```
args.add("coord_id", "UTM coordinate ID", m_coordId, "auto");
```

```
args.add("coord_id", "UTM coordinate ID", m_coordId, {true, 0});
```

abellgithub: Just drop the last argument.

hobu: > And change the `false` to `true` in the default constructor.

I didn't think we wanted to do that because it would mean finding all usages of CoordId and making sure that auto wasn't set to `true` when we were using it.

hobu:
Picked back to `2.7-maintenance` at faf623272745c1b8eb7b9d14427a1eb71734ea11

PR #4434, opened 2024-07-01
Fix building on OpenBSD

brad0:
OpenBSD does not have a sys/timeb.h header.
OpenBSD does not have wordexp(3).


abellgithub:
Why does one use OpenBSD? Just curious.

brad0:
> Why does one use OpenBSD? Just curious.

As a 30 yr user and developer of *BSD's I prefer the way things are done. Like the package tools and other ways of doing things.

PR #4435, opened 2024-07-02
Use POSIX endian.h header on OpenBSD

brad0:


abellgithub:
I don't think there are many users of these platforms, but it seems like this will break older releases of these OSes. Do you know how long it has been since this change in location has been made?

hobu:
Picked back to `2.7-maintenance` at 835754c12d0b8aa0933b289febfd08e1f06f82bf

PR #4396, opened 2024-04-23
build config for filters.cpd STANDALONE

hobu:

PR #4439, opened 2024-07-08
Fix CPD compilation of tests when building filters.cpd in non-standalone builds

hobu:
This should fix the issue brought up at https://github.com/PDAL/PDAL/pull/4396#issuecomment-2214091102


hobu:
Cherry-picked back to `2.7-maintenance` at b9c89b401b0ab137c2ca94ffae2bf96e3205decc

Issue #4442, opened 2024-07-11
Enable specification of VLR or EVLR with writers.las


hobu:
`writers.las` supports [setting VLRs on files](https://pdal.io/en/latest/stages/writers.las.html#vlrs), but it doesn't allow the user to specify whether or not it should be at the front (regular VLR) or back (EVLR) of the file.  I propose allowing the user to specify this with a single `evlr: true` element to the VLR specification:

```
            {
            "description": "A description under 32 bytes",
            "record_id": 43,
            "user_id": "hobu",
            "evlr" : true,
            "filename": "path-to-my-file.input"
            },
```

PR #4443, opened 2024-07-11
Allow forcing VLR creation as EVLR 

hobu:
Implements #4442 to allow the user to force the writing of a VLR as an EVLR if they specify it.

### DIFF
--- a/io/private/las/Vlr.cpp
+++ b/io/private/las/Vlr.cpp
@@ -323,9 +323,8 @@ std::istream& operator>>(std::istream& in, las::Evlr& v)
         else if (el.key() == "evlr")
         {
             if (!el.value().is_boolean())
-                throw pdal_error("LAS VLR metadata key must be specified as a string.");
-            bool doEVLR = el.value().get<bool>();
-            v.writeAsEVLR = doEVLR;
+                throw pdal_error("LAS VLR  key must be specified as a boolean.");
+            v.writeAsEVLR = el.value().get<bool>();
         }
         else
             throw pdal_error("Invalid key '" + el.key() + "' in VLR specification.");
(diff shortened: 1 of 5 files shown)


io/private/las/Vlr.cpp
```diff
@@ -320,6 +320,13 @@ std::istream& operator>>(std::istream& in, las::Evlr& v)
 
             v.dataFunc = MetadataFunc(metadataId);
         }
+        else if (el.key() == "evlr")
+        {
+            if (!el.value().is_boolean())
+                throw pdal_error("LAS VLR metadata key must be specified as a string.");
```
abellgithub, line 326: You've required it to be a boolean, not a string, right?

io/private/las/Vlr.hpp
```diff
@@ -85,10 +85,10 @@ struct Vlr
 
     Vlr() = default;
     Vlr(const std::string& userId, uint16_t recordId, const std::string& description) :
-        userId(userId), recordId(recordId), description(description)
+        userId(userId), recordId(recordId), description(description), writeAsEVLR(false)
```
abellgithub, line 88: This attribute belongs in EVLR, I think. It looks like it's only ever used in the context of an Evlr.

hobu: I put it on the VLR because I was worried someone might be using the VLR class by itself. I'm not happy with the arrangement anyway, as it mixes up VLR and EVLR, but I wasn't quite sure what to do here 🤷.

io/private/las/Vlr.cpp
hobu: updated as suggested in 4a59f9dcbfcecec7dfb97f64adec49ac1c783c36

io/private/las/Vlr.hpp
abellgithub: I think the logic, for better or worse, was that you would create an Evlr object in the system and it would get sliced to a Vlr object when you knew you wanted to write to the Vlr section rather than the Evlr section.

hobu: yes, that's probably the behavior you want most of the time too. My situation is I want to attach large metadata documents to the end of files and there's no need for these to be up front (think COPC content).

PR #4459, opened 2024-07-24
Join threads and cleanup on stage teardown.

### LINKED ISSUE #4458: PDAL can crash if readers.las has reader threads active
If readers.las has outstanding threads and some other stage throws an exception, PDAL can crash. PDAL is exiting anyway, but it's disconcerting to get a segfault.

abellgithub:
Close #4458


hobu:
picked back to `2.7-maintenance` at abeeb89ccbb602e929cd08aa7b704f4d1d8b0a0f

PR #4464, opened 2024-07-31
Fix libdir path in pkgconfig file

imincik:
Use single absolute path variable `CMAKE_INSTALL_FULL_LIBDIR` instead of
combination `CMAKE_INSTALL_PREFIX` and `CMAKE_INSTALL_LIBDIR` which, in some
cases, can result in double absolute path value.

For example:
```
libdir=/nix/store/65ikg5m0rsl2p4p22gpsgwkqi8cizkvz-pdal-2.7.2//nix/store/65ikg5m0rsl2p4p22gpsgwkqi8cizkvz-pdal-2.7.2/lib
```

See [Nixpkgs PR](https://github.com/NixOS/nixpkgs/pull/331092) .


hobu:
picked back back to `2.7-maintenance` at cc149530d3c2d0a061cfed813d052ae5f5a5df7b

PR #4468, opened 2024-08-09
Support sorting multiple dimensions at once in `filters.sort`

hobu:
This PR implements `dimensions` option on `filters.sort` to allow the user to do multiple sorts at one time. If multiple dimensions are requested, the first dimension is `std::stable_sort` and the rest of the dimensions are `std::sort`. If only a single dimension is requested, the `algorithm` option is respected.

### DIFF
--- a/filters/SortFilter.cpp
+++ b/filters/SortFilter.cpp
@@ -100,9 +100,9 @@ void SortFilter::filter(PointView& view)
             // If we have multiple dimensions, stable_sort the first one
             // and then sort the rest
             if (i == 0)
-                std::stable_sort(view.begin(), view.end(), cmp);
-            else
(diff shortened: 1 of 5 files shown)


filters/SortFilter.cpp
```diff
@@ -50,31 +50,67 @@ std::string SortFilter::getName() const { return s_info.name; }
...
+    {
+
+
+        m_activeDim = &m_dims[i];
+
+        if (m_dims.size() > 1)
+        {
+            // If we have multiple dimensions, stable_sort the first one
+            // and then sort the rest
+            if (i == 0)
```
abellgithub, line 102: I think this is backwards. The first sort can be a standard sort. All the rest need to be stable in order to keep the relative positions.

PR #4469, opened 2024-08-10
Implement ``filters.label_duplicates``

hobu:
The filter assigns a ``Duplicate`` dimension value to ``1`` if all of the dimensions specified in the ``dimensions`` option are equal. The points **must** be sorted by each dimension before ``filters.label_duplicates`` or it will not behave as expected.

### DIFF
--- a/filters/LabelDuplicatesFilter.cpp
+++ b/filters/LabelDuplicatesFilter.cpp
@@ -95,26 +95,24 @@ void LabelDuplicatesFilter::filter(PointView& view)
     if (view.size() < 2)
         return;
 
-    bool isDuplicatePoint(true);
-    for (PointId idx = 1; idx < view.size(); ++idx)
+    auto isDuplicatePoint = [&view, this](auto idx)
     {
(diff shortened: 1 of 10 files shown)


filters/LabelDuplicatesFilter.cpp
```diff
@@ -0,0 +1,121 @@
...
+    for (PointId idx = 1; idx < view.size(); ++idx)
+    {
+        bool isDuplicateDimension(true);
+        for(auto dimId: m_dims)
+        {
+            double current = view.getFieldAs<double>(dimId, idx);
+            double previous = view.getFieldAs<double>(dimId, idx - 1);
+
+            isDuplicateDimension = current == previous;
+            if (!isDuplicateDimension)
```
abellgithub, line 108: Opinion: this would be more clear if you just wrote `if (current != previous)`.

Opinion: when you want to jump out of a nested loop, a lambda is more readable than indicator variables.

hobu: Updated both as suggested.

PR #4477, opened 2024-08-18
add 'bins' item to filters.stats output

hobu:
For some reason we first started adding 'counts' items that look like this:

```
        "counts":
        [
          "1.000000/108",
          "7.000000/24",
          "9.000000/122704"
        ]
```

But these are hard to use and require custom parsing. They aren't JSON. I don't know why, but it's too late to break them now.

I'm adding a "bins" parameter that at least won't require custom parsing, but we are not simply replacing "counts" because people could now be parsing the silly things.

```
        "bins":
        {
          "1.000000": "108",
          "7.000000": "24",
          "9.000000": "122704"
        }
```

### DIFF
--- a/filters/StatsFilter.cpp
+++ b/filters/StatsFilter.cpp
@@ -101,10 +101,38 @@ void Summary::extractMetadata(MetadataNode &m)
     }
     else if (m_enumerate == Count)
     {
+        // For some reason we first started adding 'counts' items that look like this:
+        //
+        //         "counts":
+        //         [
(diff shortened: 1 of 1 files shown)


abellgithub:
These should be numeric.

hobu:
> These should be numeric.

I guess. I figured it was easier for JSON people to figure out if they're castable to numbers. The issue is we can't reliably convert the keys to integers or whatever because everything is read as doubles. I don't think it is worth the hassle.

PR #4478, opened 2024-08-19
Optional stable sort for filters.sort

abellgithub:
This seems to fix the sorting issue by creating a boolean that never gets moved that indicates an original (non-temporary) PointRef.

### DIFF
--- a/pdal/PointRef.hpp
+++ b/pdal/PointRef.hpp
@@ -309,7 +309,7 @@ private:
     PointId m_idx;
     PointView *m_view;
     PointId m_viewIdx;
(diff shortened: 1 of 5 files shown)


pdal/PointRef.hpp
```diff
@@ -309,6 +309,7 @@ class PDAL_DLL PointRef
     PointId m_idx;
     PointView *m_view;
     PointId m_viewIdx;
+    bool m_orig;
```
hobu, line 312: Can you add a little comment about what this is for?

hobu (approved):

PR #4487, opened 2024-08-27
Add `filters.expressionstats` and `pdal info --breakout` option

hobu:
`filters.expressionstats` allows you to applying a list of expressions to summarize a dimension (typically `Classification` but could be other integer ones like `PointSourceId`).

```
    "filters.expressionstats":
    {
      "dimension": "Classification",
      "statistic":
      [
        {
          "expression": "(Keypoint==1.000000)",
          "position": 0
        },
        {
          "expression": "(Overlap==1.000000)",
          "position": 1
        },
        {
          "bins":
          [
            {
              "count": 154,
              "value": 1
            }
          ],
          "expression": "(Synthetic==1.000000)",
          "position": 2
        },
        {
          "bins":
          [
            {
              "count": 313615,
              "value": 1
            },
            {
              "count": 6847,
              "value": 7
            },
            {
              "count": 4425,
              "value": 18
            }
          ],
          "expression": "(Withheld==1.000000)",
          "position": 3
        }
      ]
    }
```

### DIFF
--- a/doc/stages/filters.expressionstats.rst
+++ b/doc/stages/filters.expressionstats.rst
@@ -1,0 +1,107 @@
+.. _filters.expressionstats:
+
+filters.expressionstats
+===============================================================================
+
+The :ref:`filters.expressionstats` stage computes counting summary for a single
+dimension for a given set of expressions. This is useful for summarizing dimensions
+that are conveniently countable.
+
+.. embed::
+
+.. streamable::
+
+.. warning::
+
+    The ``dimension`` selected should be an integer, not floating point dimension.
+    Additionally, a dimension with lots of unique values is likely to generate a
+    many entries in the map. This may not be what you want.
+
+Example
+................................................................................
+
+.. code-block:: json
+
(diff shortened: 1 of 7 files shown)


hobu:
> Easy to parse, no need to mess with expressions on the part of the user.

As currently implemented, the user doesn't have to do much to get the equivalent output. I've taken `filters.expressionstats` and added it to `pdal info` as the `--breakout` option:

```
pdal info --stats --breakout myfile.las
```

`breakout` implies the $Classification * Flags$ summary, albeit in a more generic and verbose form. Either form would be convenient to shred with an OLAP tool like [DuckDB](https://duckdb.org/).

It would be useful to be able to execute SQL expressions against data inside of PDAL, but that seems like a stretch. This filter allows just a bit of that without embedding an engine. Some things I could think of using it for besides the typical Classification breakdown include:

* Counting cluster information
* Counting PointSourceID discrepancies
* ScanAngleRank filtering

PR #4502, opened 2024-09-19
allow override/set of --breakout dimension

hobu:
This also removes it from `--all` usage because not all point clouds have `Classification` or the `Withheld/Synthetic/Overlap/Keypoint` dimensions.


hobu:
picked back to `2.8-maintenance` at ed3cd0ece1c713f6298dd0d68dd63a04103875d1

PR #4505, opened 2024-09-26
Fix typo in filters.expression docs

fnands:
Classification is spelled as Classifcation (i.e. with a missing **i** between **f** and **c**). 

This PR fixes this. 

This one tripped me up when I copy pasted this example.


chambbj:
Thanks for catching this!

PR #4509, opened 2024-10-01
Hexbin filter - H3 dim support

ibell13:
Added functionality in filters.hexbin to read H3 cell indexes added to points with filters.h3, and use them to create hexagons. Also fixed up OGR code to write the correct CRS to the density/boundary output.


hobu:
rebase master to get fa370cb7e8a2678d36eca6771c5f613ed62d13f3 so the conda builder will succeed

ibell13:
> I don't see any value in this. It adds complication for no advantage. Will you not get exactly the same result without this special processing?

It would be the same result, but the data is already there in the H3 field. The processing I do on it is essentially the same as how we create hexagons in the tests, so it's not really adding much new code other than some conditionals in the filter. I think one reason it's useful is that filters.h3 can run on non-WGS points, so hexbin isn't forcing people to reproject their data.

abellgithub:
1) IMO we should reproject data if necessary in filters.hexbin no matter.
2) AFAICT the ONLY benefit is that you save an X/Y -> IJ/H3 conversion, which is very, very meh.

We shouldn't add code that we don't need. Direct code paths are better.

Maybe we should just ditch the H3 filter and use this.

Also, this PR has too much stuff in it. It's too hard to parse what has been done. Unless there's some good reason, each PR should do address only one thing. I'd recommend creating an issue that explains what change is being made that can be tied to the PR.

PR #4510, opened 2024-10-01
fix filters.geomdistance initialization and distance comparisons

hobu:


abellgithub:
It's not clear here what the actual issue was that you're fixing.

hobu:
the `ogr` option of `filters.geomdistance` requires GDAL. That means we need to have `gdal::registerDrivers()` called to wake things up. We can't call that in `addArgs` because things aren't awake yet, so we need to call it in `ready`. 

We also should prevent comparing distances of null geometries.

hobu:
picked back to `2.8-maintenance` at 04d919ef4b4e65bf0c1dafb6aff7d0ebb1a271e8

PR #4511, opened 2024-10-01
[Bugfix] `CopcReader` maybe has memory leaks

AbhinavAtrishi:
- `CopcReader::inspect()`, initializes `PointTable t` with `initialize(t)` but does not ever call `done(t)`. This seems to be a memory leak and is caught by ASan
- Also added a safer destructor for the threadpools, so that there is no accidental leak.

PS: I'm not sure if this is the best way to do this, or if any other implementation is preferred?

### DIFF
--- a/io/CopcReader.cpp
+++ b/io/CopcReader.cpp
@@ -264,7 +264,16 @@ CopcReader::CopcReader() : m_args(new CopcReader::Args), m_p(new CopcReader::Pri
 
 
 CopcReader::~CopcReader()
-{}
+{
+    if (m_p)
+    {
+        if (m_p->pool)
+        {
+            m_p->pool->stop();
+        }
+        m_p->connector.reset();
+    }
+}
 
 
 std::string CopcReader::getName() const
@@ -638,6 +647,7 @@ QuickInfo CopcReader::inspect()
             qi.m_bounds.clip(b);
     }
     qi.m_valid = true;
+    done(t);
 
     return qi;
 }


abellgithub:
I don't understand the actual leak. What is being allocated that isn't being freed?

Can you explain why the additional code is necessary in the destructor?

AbhinavAtrishi:
> I don't understand the actual leak. What is being allocated that isn't being freed? Under what circumstances?
> 
> Can you explain why the additional code is necessary in the destructor?

I'll have to look into it, I'm also really new to this so will actually do a proper investigation. The code in the destructor is not needed at all, apologies for that I'll remove it. I guess the `Threadpool` does it on its own

abellgithub:
valgrind will report certain static allocations at termination as leaks when they aren't. When built with debugging, I would think valgrind would identify the specific code line. Thanks.

abellgithub:
At this point it doesn't seem like anyone understands what the leak is or
why calling done would fix it. It would be nice to understand.

AbhinavAtrishi:
I'm sorry I could not get to this, I plan to spend more time on this over the weekend. I still could not identify where the memory leak is using a smaller example

Calling `done` though is essential for completely different reasons, wherein if you have a multi-threaded application calling `preview()` then `preview()` should properly manage and close its own threadpool.

abellgithub:
That doesn't make sense to me. When you call `inspect()` it resets the thread pool.

Issue #4451, opened 2024-07-19
Help needed reading EPT from azure blob container with SAS token


trns1997:
I was wondering if you had an example on how to read ept data from an azure blob container which is accessible only via a SAS token.

I have a pipeline that looks like the following (based on https://gist.github.com/hobu/ee22084e24ed7e3c0d10600798a94c31):
```
{
    "pipeline": [
        {
            "bounds": "([-10425171.940, -10423171.940], [5164494.710, 5166494.710])",
            "filename": "https://<AZURE_STORAGE_ACCOUNT>.blob.core.windows.net/<PATH_TO_EPT>/ept.json",
            "type": "readers.ept",
            "tag": "readdata"
        },
        {
            "filename": "test.laz",
            "inputs": [ "readdata" ],
            "tag": "writerslas",
            "type": "writers.las"
        }
    ]
}
```

**Result** :
```
pdal pipeline test.json --debug
(PDAL Debug) Debugging...
(pdal pipeline readers.ept Debug) PDAL: readers.ept: Could not read from <AZURE_STORAGE_ACCOUNT>.blob.core.windows.net/<PATH_TO_EPT>/ept.json
```
Which is completely expected as i have not provided the reader with the sas token to access the file. 

So i tried a very naive approach where i provieded the sas token to the filename directly as shown below:

```
{
    "pipeline": [
        {
            "bounds": "([-10425171.940, -10423171.940], [5164494.710, 5166494.710])",
            "filename": "https://<AZURE_STORAGE_ACCOUNT>.blob.core.windows.net/<PATH_TO_EPT>/ept.json?<SAS_TOKEN>",
            "type": "readers.ept",
            "tag": "readdata"
        },
        {
            "filename": "test.laz",
            "inputs": [ "readdata" ],
            "tag": "writerslas",
            "type": "writers.las"
        }
    ]
}
```
**Result** :
```
pdal pipeline test.json --debug
(PDAL Debug) Debugging...
(pdal pipeline readers.ept Debug) Query bounds: ([-10425171.94, -10423171.94], [5164494.71, 5166494.71], [-1.797693134862316e+308, 1.797693134862316e+308])
Threads: 15
(pdal pipeline Debug) Executing pipeline in stream mode.
PDAL: readers.ept: Could not read from <AZURE_STORAGE_ACCOUNT>.blob.core.windows.net/<PATH_TO_EPT>/ept-hierarchy/0-0-0-0.json
```
[...]

PR #4524, opened 2024-10-10
add header and query options to readers

### LINKED ISSUE #4523: Https header support to readers
For the moment based on usage, I know that `reader.ept` supports http headers that are passed to arbiter to read files in a remote server. I was wondering is there a particular reason why other readers do not have this option. I understand I could programmatically download for example my las file and pass the downloaded file to the LasReader, but I was thinking it would be cool to pass the readers the http headers with token and all necessary info needed for it to fetch the file and read it. Please let me know what you guys think? Or if this functionality already exists and I simply missed it?

trns1997:
Closes #4523 
Closes https://github.com/PDAL/PDAL/issues/4451

- [x] Test with local las
- [x] Test with remote las with key (no query no header just classic downloadable link)
- [x] Test with remote las with header
- [x] Add this logic to base class (Reader class)
- [x] Unit Test Pass
- [ ] Test all readers manually? Or unit testing is enough?
- [ ] Add Unit Test ?
- [ ] Add example doc to close this issue as well https://github.com/PDAL/PDAL/issues/4451

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,7 +150,7 @@ if (WITH_TESTS)
     enable_testing()
 endif()
 
-add_subdirectory(plugins)
+# add_subdirectory(plugins)
 
 if (WITH_TESTS)
     include (${PDAL_CMAKE_DIR}/gtest.cmake)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,7 +150,7 @@ if (WITH_TESTS)
     enable_testing()
 endif()
 
-# add_subdirectory(plugins)
+add_subdirectory(plugins)
 
 if (WITH_TESTS)
     include (${PDAL_CMAKE_DIR}/gtest.cmake)
--- a/plugins/pgpointcloud/CMakeLists.txt
+++ b/plugins/pgpointcloud/CMakeLists.txt
@@ -28,6 +28,8 @@ PDAL_ADD_PLUGIN(reader_libname reader pgpointcloud
         ${PDAL_LIBRARIES}
     INCLUDES
         ${LIBXML2_INCLUDE_DIR}
+        ${NLOHMANN_INCLUDE_DIR}
+        ${PDAL_VENDOR_DIR}
 )
 
 #
--- a/doc/stages/readers.las.md
+++ b/doc/stages/readers.las.md
@@ -120,6 +120,16 @@ threads
 
 : Thread pool size. Number of threads used to decode laz chunk tables (Default: 7)
 
+header
+
+: HTTP headers to forward for remote Las endpoints, specified as a JSON
+  object of key/value string pairs.
+
+query
+
+: HTTP query parameters to forward for remote Las endpoints, specified as a
+  JSON object of key/value string pairs.
+
 [las format]: http://asprs.org/Committee-General/LASer-LAS-File-Format-Exchange-Activities.html
 [las specification]: http://www.asprs.org/a/society/committees/standards/LAS_1_4_r13.pdf
 [laszip]: http://laszip.org
--- a/io/CopcReader.cpp
+++ b/io/CopcReader.cpp
@@ -223,8 +223,6 @@ public:
     bool fixNames;
     bool doVlrs;
 
-    NL::json query;
-    NL::json headers;
     NL::json ogr;
 
     int keepAliveChunkCount = 10;
@@ -299,8 +297,6 @@ void CopcReader::addArgs(ProgramArgs& args)
     args.add("resolution", "Resolution limit", m_args->resolution);
     args.add("polygon", "Bounding polygon(s) to crop requests",
         m_args->polys).setErrorText("Invalid polygon specification. Must be valid GeoJSON/WKT");
-    args.add("header", "Header fields to forward with HTTP requests", m_args->headers);
-    args.add("query", "Query parameters to forward with HTTP requests", m_args->query);
     args.add("ogr", "OGR filter geometries", m_args->ogr);
     args.add("fix_dims", "Make invalid dimension names valid by changing invalid "
         "characters to '_'", m_args->fixNames, true);
@@ -313,30 +309,6 @@ void CopcReader::addArgs(ProgramArgs& args)
 }
 
 
-void CopcReader::setForwards(StringMap& headers, StringMap& query)
-{
-    try
-    {
-        if (!m_args->headers.is_null())
-            headers = m_args->headers.get<StringMap>();
-    }
-    catch (const std::exception& err)
-    {
-        throwError(std::string("Error parsing 'headers': ") + err.what());
-    }
-
-    try
-    {
-        if (!m_args->query.is_null())
-            query = m_args->query.get<StringMap>();
-    }
-    catch (const std::exception& err)
-    {
-        throwError(std::string("Error parsing 'query': ") + err.what());
-    }
-}
-
-
 void CopcReader::initialize(PointTableRef table)
 {
     if (m_args->threads > 100)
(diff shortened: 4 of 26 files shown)


trns1997:
@abellgithub i would like your input. I was thinking of adding the options directly to the Reader class that way all readers have the possibility to pass http headers and query options. What do you think?

abellgithub:
This looks good. I expect it will be useful. Please add the options to the documentation.

I don't see a need to do it for all readers at this point. If there is demand, it can be changed.

trns1997:
> This looks good. I expect it will be useful. Please add the options to the documentation.
> 
> I don't see a need to do it for all readers at this point. If there is demand, it can be changed.

The reason I asked is because I want to use couple of readers with http headers. Such as `reader.las`, `readers.text` and `readers.sbet`. They all seem to have a similar mode of functioning. So I was thinking maybe adding the query and headers option in the base class might be a cleaner approach?

trns1997:
> > So I was thinking maybe adding the query and headers option in the base class might be a cleaner approach
> 
> I support doing this. People expect consistent behavior of the readers in regards to this kind of thing and it would be frustrating to have it activated for some readers and not the others.

Alright @hobu, I'll go with the approach where I will add http headers and query args to the reader class. The logic to parse the input string and maybe create the arbiter object there as well but I need to see in détail. But firstly I will validate that this current implementation works completely and from there I can implement the same in the base class.

abellgithub:
This isn't relevant for all readers. Something will have to be done to support those formats that aren't necessarily self-contained files.

trns1997:
> This isn't relevant for all readers. Something will have to be done to support those formats that aren't necessarily self-contained files.

What readers do you have in mind?

hobu:
This is relevant for all readers being used over HTTP. Maybe it would be better to have an `arbiter` `NL::json` object available for every reader and hand off everything to it that way.

CMakeLists.txt
```diff
@@ -150,7 +150,7 @@ if (WITH_TESTS)
     enable_testing()
 endif()
 
-add_subdirectory(plugins)
+# add_subdirectory(plugins)
```
trns1997, line 153: @hobu @abellgithub  building plugins fail as it is unable to fine arbiter. Not sure why but it is cause i included the connector object tor the reader class

trns1997:
@hobu @abellgithub i would like your input on the latest commit before i apply this to all readers.  The idea is the that the reader class possesses the connector object and the http headers and query. And derived readers will use this connector object to read remote files. I want thinking of maybe adding the remote file check in the reader class as well but firstly i would like your go ahead before i converge and apply this. Also i would like you input on how to fix the plugin compile issue i am facing.

CMakeLists.txt
trns1997, line 153: Cannot find arbiter, it is because i include the `Connector.hpp` to the reader class. The plugins do not seem to have access to `#include <nlohmann/json.hpp>` and `#include <arbiter/arbiter.hpp>`

plugins/pgpointcloud/CMakeLists.txt
```diff
@@ -28,6 +28,8 @@ PDAL_ADD_PLUGIN(reader_libname reader pgpointcloud
         ${PDAL_LIBRARIES}
     INCLUDES
         ${LIBXML2_INCLUDE_DIR}
+        ${NLOHMANN_INCLUDE_DIR}
```
trns1997, line 31: I need to include these to the plugin, as reader class include `Connector.hpp` which require both `nlohann` and `arbiter`

abellgithub: I don't understand this, but I don't have time to look right now. The class can certainly be changed to eliminate these dependencies.

I think the way things are being done isn't great in general (not these changes, as such). `header` and `query` can be of some type that knows how to read itself from a string, like other arguments. Then `setForwards` is unnecessary.

I'll look more tomorrow.

hobu:
Instead of two new options, I think `header` and `query` should be put in an `arbiter` `NL::json` option object that could go on every reader, and it would be up to Arbiter to decide what to do with it. This allows Arbiter to add support for other interesting things without having to touch the options of stages going forward.

Additionally, we should back out the changes to `readers.ept` and forward any settings of `headers` and `query` down into the `arbiter` object. We don't want to break people's pipelines.

trns1997:
> Instead of two new options, I think `header` and `query` should be put in an `arbiter` `NL::json` option object that could go on every reader, and it would be up to Arbiter to decide what to do with it. This allows Arbiter to add support for other interesting things without having to touch the options of stages going forward.
> 
> Additionally, we should back out the changes to `readers.ept` and forward any settings of `headers` and `query` down into the `arbiter` object. We don't want to break people's pipelines.

hmm not sure i understand everything, correct me if i am wrong. What you are suggesting is that the `Reader` Class contains an `arbiter`/`connector` object, and each derived reader in their respective  `addArgs` method will add the http_header and http_query option which is then forwarded to the `arbiter`/`connector` object present in the `Reader` Class? Or i misunderstood entirely ?

abellgithub:
@hobu This would be a change in the way arbiter is working. The API doesn't support what you suggest. I don't have strong feelings,  but perhaps @connormanning does. I don't love the JSON interface, but I'm probably alone on that.

trns1997:
> @hobu This would be a change in the way arbiter is working. The API doesn't support what you suggest. I don't have strong feelings, but perhaps @connormanning does. I don't love the JSON interface, but I'm probably alone on that.

Instead of parsing header and query as a json we can just parse it as a `StringMap` directly (need to see if it is possible) but I feel the gain is minimal. But if there is a preference for this I can take a quick look.

trns1997:
@abellgithub @hobu as far as the current changes go, reading local and remote laz and entwine work with no impact on user side for `readers.ept`, it is just the `las.readee` has 2 new options that it did not have before.

trns1997:
ping @hobu @abellgithub i would really love to get your input and finish this up so i no longer need to patch for my usage.

hobu:
My concern is adding `header` and `query` to only `readers.las` and `readers.copc` does us no good when users want it in `readers.ply` or `readers.text`. So then we should add it to every reader, yeah?

If we add it to every reader, do we add just only `header` and `query`, or could we add an `NL::json` object that simply forwards a tree of whatever down to Arbiter? That's what I was trying to describe above.

trns1997:
> Can we halt this for a bit? I think that this needs more discussion.

Sure thing. I just updated readers that read remote files like the copc reader, that way we have a more holistic view of the feature. Now I guess we can address the concerns. I am guessing one of which is should all readers support remote files and really need to parse header and query args? If we're not a fan of having header and query args in the Reader class then we can have an interface class called maybe the RemoteReader class has the ability to fetch remote files, and Readers such a EptReader and LasReader inherit that class? But not if it's a great idea.

hobu:
See #4625. Let's wire in the capability in this PR in a more holistic way.

trns1997:
> See #4625. Let's wire in the capability in this PR in a more holistic way.

@hobu what is the idea concretely? We have a filespec member in the Reader class and all derived classes will have access to this object to then create the connector to fetch remote files?

abellgithub:
I don't think there is any "one size fits all" solution to this. Sometimes you want to copy a file from a remote source. Sometimes you might want to stream it. Sometimes you might want to fetch various blocks. Each reader needs consideration.

This PR provides a unified way to pass additional file access information to any reader that may use it without adding arguments to all the stages. It's backward-compatible, so changes to enhance reader remote access can be done in steps.

trns1997:
@abellgithub makes sense. I guess once https://github.com/PDAL/PDAL/pull/4625 is merged i'll patch up this PR to incorporate the changes and adapt it :)

hobu:
#4625 is now merged. Please adapt to its usage and we can move forward on this.

trns1997:
Closing This PR as it has diverged significantly, it is simpler to just start fresh. Refer to PR https://github.com/PDAL/PDAL/pull/4633

Issue #2685, opened 2019-08-21
PDAL_DLL pragma poorly named


abellgithub:
PDAL_DLL indicates a symbol export.  We should call it something like PDAL_EXPORT.

dg0yt:
In PDAL 2.9.0, this is now renamed to `PDAL_EXPORT`.

PR #4528, opened 2024-10-20
Rename PDAL_DLL to PDAL_EXPORT (#2685)

enno-t:
This PR renames the PDAL_DLL macro to PDAL_EXPORT, as proposed in Issue #2685.

This was mainly done with the following script:

```
find . -type f | while read -r file
do
	sed -i -e 's|PDAL_DLL|PDAL_EXPORT|g' "$file"
done
```
and then reverting unintended changes.


hobu:
Was this a reaction to a specific problem, or were you just looking to contribute? I think the reason why we haven't made this change is a long time ago `PDAL_DLL` sometimes really had the behavior of `PDAL_IMPORT`. That's no the case now, however.

PR #4529, opened 2024-10-21
fetch the bounds for PlyReader::inspect

hobu:
`readers.ply` was not actually providing bounds in `preview`. It must do so or not implement preview.


hobu:
picked back to `2.8-maintenance` at 6eb7724609ee3e1c6b0033be238591323fd43c0d

PR #4531, opened 2024-10-22
Tile Index creation: add multithreaded support, where query support, new options

ibell13:
Added new options for creating a tile index with the _pdal tindex create_ command:
- _threads_/_requests_ option: specify the number of CPU threads used to create point cloud boundaries in a new tile index; default 1
- _where_ option: PDAL where expression to run on the points being indexed
- _simplify_, _threshold_, _resolution_ and _sample_size_ options: settings for creating point cloud boundaries are now handled by an internal filter in TIndexKernel, rather than being passed as filters.hexbin options. These options are equivalent to their counterparts in filters.hexbin.

Also created new tests for the internal boundary filter in TIndexTest.cpp.

### DIFF
--- a/kernels/TIndexKernel.cpp
+++ b/kernels/TIndexKernel.cpp
@@ -182,12 +182,12 @@ void TIndexKernel::addSubSwitches(ProgramArgs& args,
             "location");
         args.add("ogrdriver,f", "OGR driver name to use ", m_driverName,
             "ESRI Shapefile");
-        args.add("absolute_path", "Convert the tile index filename to "
-            "absolute path when creating a new file with OGR", m_absPath, true);
         args.add("t_srs", "Target SRS of tile index", m_tgtSrsString,
             "EPSG:4326");
         args.add("a_srs", "Assign SRS of tile with no SRS to this value",
             m_assignSrsString, "EPSG:4326");
+        args.add("write_absolute_path",
+            "Write absolute rather than relative file paths", m_absPath);
         args.add("stdin,s", "Read filespec pattern from standard input",
             m_usestdin);
         args.add("path_prefix", "Prefix to be added to file paths when writing "
@@ -246,6 +246,9 @@ void TIndexKernel::validateSwitches(ProgramArgs& args)
         if (m_filespec.size() && m_usestdin)
             throw pdal_error("Can't specify both --filespec and --stdin "
                 "options.");
+        if (m_prefix.size() && m_absPath)
+            throw pdal_error("Can't specify both --write_absolute_path and "
+                "--path_prefix options.");
         if (args.set("a_srs"))
             m_overrideASrs = true;
     }
@@ -324,6 +327,10 @@ void TIndexKernel::createFile()
         m_files = FileUtils::glob(m_filespec);
     else
         m_files = readSTDIN();
+
+    if (m_absPath)
+        for (auto& s : m_files)
+            s = FileUtils::toAbsolutePath(s);
 
     if (m_files.empty())
     {
@@ -636,12 +643,7 @@ bool TIndexKernel::createDataset(const std::string& filename)
         throw pdal_error(oss.str());
     }
 
-    if (m_absPath) {
-        std::string dsname = FileUtils::toAbsolutePath(filename);
-        m_dataset = OGR_Dr_CreateDataSource(hDriver, dsname.c_str(), NULL);
-    }
-    else
-       m_dataset = OGR_Dr_CreateDataSource(hDriver, filename.c_str(), NULL); 
+    m_dataset = OGR_Dr_CreateDataSource(hDriver, filename.c_str(), NULL); 
     return (bool)m_dataset;
 }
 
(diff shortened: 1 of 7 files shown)


kernels/TIndexKernel.cpp
```diff
@@ -654,12 +652,33 @@ pdal::Polygon TIndexKernel::prepareGeometry(const FileInfo& fileInfo)
...
     {
         SpatialReference out(m_tgtSrsString);
         g.transform(out);
     }
+
     return g;
 }
 
+
+std::string TIndexKernel::makeMultiPolygon(const std::string& wkt)
```
hobu, line 676: 😬

kernels/TIndexKernel.hpp
```diff
@@ -34,9 +34,17 @@
...
 #pragma once
 
+#include <pdal/Filter.hpp>
+#include <pdal/Streamable.hpp>
 #include <pdal/Stage.hpp>
 #include <pdal/SubcommandKernel.hpp>
 #include <pdal/util/FileUtils.hpp>
+#include <filters/private/hexer/HexGrid.hpp>
+
+namespace hexer
```
hobu, line 44: why do we need this forward declaration if we're also including the `HexGrid.hpp` header?

kernels/TIndexKernel.cpp
```diff
@@ -118,6 +118,15 @@ void TIndexKernel::addSubSwitches(ProgramArgs& args,
             "Write absolute rather than relative file paths", m_absPath);
         args.add("stdin,s", "Read filespec pattern from standard input",
             m_usestdin);
+        args.add("threads", "Number of threads to use for file boundary creation",
```
hobu, line 121: Maybe we should add a `requests` synonym here too.

kernels/TIndexKernel.cpp
```diff
@@ -118,6 +118,15 @@ void TIndexKernel::addSubSwitches(ProgramArgs& args,
...
+            m_threads, 1);
+        args.add("smooth", "Smooth boundary output", m_doSmooth, true);
```
hobu, line 123: Maybe this should be more accurately called `simplify`?

kernels/TIndexKernel.cpp
```diff
@@ -118,6 +118,15 @@ void TIndexKernel::addSubSwitches(ProgramArgs& args,
...
+        args.add("threshold", "Required cell density, to be used in internal "
```
hobu, line 124: `Number of points a cell must contain to be declared positive space`

kernels/TIndexKernel.cpp
```diff
@@ -118,6 +118,15 @@ void TIndexKernel::addSubSwitches(ProgramArgs& args,
...
+            "hexbin filter (non-fast boundary)", m_density, 15);
+        args.add("edge_length", "cell edge length to be used in internal hexbin "
```
hobu, line 126: This is really a `resolution`, and I think we should call it that.

kernels/TIndexKernel.cpp
```diff
@@ -118,6 +118,15 @@ void TIndexKernel::addSubSwitches(ProgramArgs& args,
...
+            "filter (non-fast boundary)", m_edgeLength);
+        args.add("sample_size", "Sample size for auto-edge length calculation in "
```
hobu, line 128: Do we have to expose this to users? They can do `--filters.hexbin.sample_size` themselves if they want.

kernels/TIndexKernel.cpp
```diff
@@ -654,12 +652,33 @@ pdal::Polygon TIndexKernel::prepareGeometry(const FileInfo& fileInfo)
     using namespace gdal;
 
     Polygon g(fileInfo.m_boundary, fileInfo.m_srs);
+    if (fileInfo.m_gridHeight && m_doSmooth)
+    {
+        double tolerance = 1.1 * fileInfo.m_gridHeight / 2;
```
hobu, line 657: where did this magic come from?

hobu (changes requested):

kernels/TIndexKernel.cpp
```diff
@@ -67,6 +67,78 @@ void setDate(OGRFeatureH feature, const tm& tyme, int fieldNumber)
...
+        }
+    }
+    virtual bool processOne(PointRef& point)
+    {
+        double x = point.getFieldAs<double>(Dimension::Id::X);
+        double y = point.getFieldAs<double>(Dimension::Id::Y);
+        m_grid->addXY(x, y);
+        return true;
+    }
+    virtual void spatialReferenceChanged(const SpatialReference& srs)
```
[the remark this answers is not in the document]
ibell13: If I do Stage::getSpatialReference() on the filter stage without setting it explicitly, the SRS is empty. But maybe there's a different/better way to set it.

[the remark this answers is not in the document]
abellgithub: Yes, I forgot it was being extracted in the FileInfo.

hobu (approved):

PR #4535, opened 2024-10-23
Make use of dimName.

abellgithub:
There was a dimName variable but it was unused.

### DIFF
--- a/filters/NeighborClassifierFilter.cpp
+++ b/filters/NeighborClassifierFilter.cpp
@@ -45,21 +45,23 @@
 namespace pdal
 {
 
-static PluginInfo const s_info = PluginInfo(
+static PluginInfo const s_info
+{
     "filters.neighborclassifier",
     "Re-assign some point attributes based KNN voting",
-    "http://pdal.io/stages/filters.neighborclassifier.html" );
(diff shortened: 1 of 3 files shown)


filters/NeighborClassifierFilter.cpp
```diff
@@ -134,7 +140,7 @@ void NeighborClassifierFilter::doOneNoDomain(PointRef &point, PointRef &temp,
         { return p1.second < p2.second; });
 
     // update point
-    auto oldclass = point.getFieldAs<double>(m_dim);
+    auto oldclass = point.getFieldAs<int>(m_dimId);
```
hobu, line 143: should we be changing the type here? I guess people could use doubles as classifiers. We wouldn't care, but we're going to trim them off here by doing this.

abellgithub: This matches the voting scheme of other parts of the filter. I made a note in the documentation. I don't know why you'd want to do this.

hobu (approved):

hobu:
picked back to `2.8-maintenance` at 988a1745aeffbbf451c07b4dce4598ed04b9d5a4

PR #4537, opened 2024-10-25
include cstdlib for gcc-15

t0b3:
see downstream issue at https://bugs.gentoo.org/937770


abellgithub:
Why? Do you have an error to report? I don't see any use of the contents of this header in this file.

PR #4540, opened 2024-11-08
Optional HEAD request

gui2dev:
Some API might not implement the HEAD verb to get file info.

For streaming reader such as `readers.ept` this means a lot of failed requests.

Similar to `gdal` option `CPL_VSIL_CURL_USE_HEAD` there can be a `PDAL_NO_HEAD` env variable that, if set, would disable

```cpp
std::unique_ptr<std::size_t> pSize = arbiter::Arbiter().tryGetSize(path);
```

for remote files.

There might be better ways to implement that, maybe in `Arbiter` directly ?

```cpp
std::unique_ptr<std::size_t> Http::tryGetSize(const std::string path) const
{
    return getDriver(path)->tryGetSize(stripProtocol(path));
    // return std::unique_ptr<std::size_t>(); if ARBITER_NO_HEAD ?
}
```

### DIFF
--- a/pdal/PDALUtils.cpp
+++ b/pdal/PDALUtils.cpp
@@ -248,7 +248,9 @@ public:
 uintmax_t fileSize(const std::string& path)
 {
     uintmax_t size = 0;
-    if (isRemote(path))
+    std::string disabled;
+    int set = Utils::getenv("PDAL_NO_HEAD", disabled);
+    if (isRemote(path) and set != -1)
     {
         std::unique_ptr<std::size_t> pSize = arbiter::Arbiter().tryGetSize(path);
         if (pSize)


hobu:
@connormanning This is probably something that already exists in Arbiter that isn't bubbled up, correct?

connormanning:
Can you elaborate on the problem this would solve, exactly?  It seems to me that in the *failure* case for a request, we simply move the failure from GET to HEAD, and in the *success* case, we add an RTT to the overhead of the call.

> this means a lot of failed requests

But they're going to fail anyway, right?  Maybe I'm not understanding the distinction about why the HEAD request helps you for a GET request that would fail.

gui2dev:
> Can you elaborate on the problem this would solve, exactly?  It seems to me that in the *failure* case for a request, we simply move the failure from GET to HEAD, and in the *success* case, we add an RTT to the overhead of the call.
> 
> > this means a lot of failed requests
> 
> But they're going to fail anyway, right?  Maybe I'm not understanding the distinction about why the HEAD request helps you for a GET request that would fail.

Some readers, like las reader, can take advantage of knowing the file size.

While it is not mandatory, especially for remote files, it still tries to get it, by making an HTTP HEAD request.

If API does not implement HEAD request, or if one does not want the extra HEAD traffic (let's say for performance reasons), pdal/arbiter should let user disable them.

gui2dev:
> @connormanning This is probably something that already exists in Arbiter that isn't bubbled up, correct?

I did not find anything like that in arbiter.

connormanning:
I think I read this backwards.  You are saying that there are HEAD requests happening in your workflow that you want to skip?  I thought you were wanting to *add* a HEAD request prior to some GET request as some pre-flight thing.

In that case Arbiter internals aren't really involved here, this is business logic in PDAL which is explicitly making the HEAD request.

Ok I think I understand now - but in that case I'm not sure what to suggest.  I'm not even sure what `PdalUtils::fileSize()` would be expected to do in the case of EPT (and things like STAC etc) since it's more of a catalog than a file.  What behavior are you hoping to achieve?

gui2dev:
> I think I read this backwards.  You are saying that there are HEAD requests happening in your workflow that you want to skip?  I thought you were wanting to *add* a HEAD request prior to some GET request as some pre-flight thing.
> 
> In that case Arbiter internals aren't really involved here, this is business logic in PDAL which is explicitly making the HEAD request.
> 
> Ok I think I understand now - but in that case I'm not sure what to suggest.  I'm not even sure what `PdalUtils::fileSize()` would be expected to do in the case of EPT (and things like STAC etc) since it's more of a catalog than a file.  What behavior are you hoping to achieve?

Just trying to disable HEAD requests that are done by arbiter when trying to get file size for remote files.

connormanning:
> It's only called by pdalUtils::fileSize() actually.

This conversation seems to be at the wrong scope.

Arbiter is not making the requests, PDAL is explicitly making this HEAD request.  Also, this happens in the generic `PdalUtils::fileSize()` utility, where you would *expect* this request to happen.  But some of the discussion is talking about LAS and EPT readers etc.

If you are having some problem where there is a specific scenario where you are seeing floods of failed requests, it seems likely that this needs to be addressed by this utility function not being called in the first place, rather than adding a magic switch to disable it under the hood.

Can you provide exactly the scenario that is causing problems?  Are you translating data, are you doing `info` things, running a pipeline, are your endpoints remote, etc.

gui2dev:
> > It's only called by pdalUtils::fileSize() actually.
> 
> This conversation seems to be at the wrong scope.
> 
> Arbiter is not making the requests, PDAL is explicitly making this HEAD request. Also, this happens in the generic `PdalUtils::fileSize()` utility, where you would _expect_ this request to happen. But some of the discussion is talking about LAS and EPT readers etc.
> 
> If you are having some problem where there is a specific scenario where you are seeing floods of failed requests, it seems likely that this needs to be addressed by this utility function not being called in the first place, rather than adding a magic switch to disable it under the hood.
> 
> Can you provide exactly the scenario that is causing problems? Are you translating data, are you doing `info` things, running a pipeline, are your endpoints remote, etc.

Ok, let me put some context.
We exploit `ept` data that is stored remotely behind a custom api. It does not supports `HEAD` requests, so we monitored a lot of failed `HEAD` requests. We asked for `HEAD` requests support, but that might take some time.

In the mean time, I was looking for a way to mitigate that high error rate. Digging deeper, made me look for `HEAD` requests.

As it is purely remote / HTTP related, I looked for `Arbiter` and `Connector`.
`Connecter::headRequest(const std::string& path)` is only used by `stac`.

Arbiter is making `HEAD` requests only in `tryGetSize()` requests, which is only called by `pdalUtils::fileSize()`.

But I guess it's rather actually here that we have the head request :

https://github.com/PDAL/PDAL/blob/461fa742f5a0e6f91b7be9e0846e6bc90f8c4999/vendor/arbiter/arbiter.cpp#L680C13-L680C69

connormanning:
Ok, thanks for the context, that's very helpful.  That particular HEAD request was added in https://github.com/connormanning/arbiter/pull/31 to do streaming downloads rather than buffering the whole download in memory.

I wonder if this chunking behavior should be opt-in, as it only makes sense for really large files which arbiter is not generally optimized for.

gui2dev:
> Ok, thanks for the context, that's very helpful.  That particular HEAD request was added in https://github.com/connormanning/arbiter/pull/31 to do streaming downloads rather than buffering the whole download in memory.
> 
> I wonder if this chunking behavior should be opt-in, as it only makes sense for really large files which arbiter is not generally optimized for.

True, as for `ept` data it's not and it is supposed that `Range` request are also supported, which is not the case in my use case

abellgithub:
Is this applicable for PDAL at all?

connormanning:
It is applicable.  I submitted a PR to Arbiter but have not merged anything.  This would make a big difference if you were to `pdal translate` a large LAS/LAZ file, for example.  Rather than streaming it, it would download the whole thing in one big memory blob first and then dump it to disk/

Really what we'd want is [this](https://github.com/connormanning/arbiter/issues/12), where the drivers would expose a streaming API so we could stream directly, rather than this specialization.  This existing specialization "manually" streams for HTTP-based drivers, by explicitly making range requests following a HEAD to get the size of the file, for which ranges will be fetched.  As opposed to the drivers having a `getReadableStream()` method which could be piped directly to an output `getWritableStream()` thing.

The fact that we are already using Curl's callback methods to build our output data from the TCP stream ([here](https://github.com/connormanning/arbiter/blob/master/arbiter/util/curl.cpp#L44-L57)) would allow us to do this with mostly already-existing code that would need to be wired together for HTTP-based drivers.  And then exposing analogous filestream-based options for the `file://` driver.

gui2dev:
That's for reading. In the mean time some (most) http based driver like azure won't allow you to upload more than a given size without [chunk your blob into blocks](https://learn.microsoft.com/en-us/rest/api/storageservices/put-block?tabs=microsoft-entra-id).


Guilhem Villemin
________________________________

hobu:
@gui2dev now that @normanb landed GDAL VSI support in `master`, can you achieve what you want with that?

gui2dev:
Absolutely

PR #4541, opened 2024-11-10
Tests: add compatibility with GDAL 3.10

rouault:
GDAL 3.10 has deprecated opening a MEM::: dataset


hobu:
picked back to `2.8-maintenance` at d93a850e1dc85b3d9bdd11a2a8ff8d70c9785ddc

PR #4543, opened 2024-11-12
Doc: pdal pipeline: use non-deprecated array JSON syntax in examples

rouault:


hobu:
Thanks for the contribution. I wonder if we shouldn't have one example of both, however. I sometimes use the `pipeline` form so I can stick a `hobu` object at the same level as `pipeline` to inform other processing activities the pipeline defines.

hobu:
picked back to `2.8-maintenance` at 17acc5764c540203af68a2dda67c049f85f46db6

PR #4544, opened 2024-11-12
allow MEM:: data to be opened for GDAL 3.10+ (supercede #4541)

hobu:


abellgithub:
Why do we have to do this awfulness now?

hobu:
> Why do we have to do this awfulness now?

At GDAL 3.10, opening a `MEM::` GDAL raster data source is no longer supported unless you set the magic flag. #4541 changed it so we don't have to do `MEM::` for opening a test data source. On reconsideration, I think the approach there is better.

Issue #4546, opened 2024-11-13
portable_popen swallows /'s on windows


hobu:
[portable_popen](https://github.com/PDAL/PDAL/blob/master/pdal/util/Utils.cpp#L323) eats `/`'s on windows and swaps them all to `\\`. That's normally ok, but in the case of passing GDAL VSI paths like `/vsistdout` we definitely don't want the `/` swapped to a `\\`.

I'm not sure what to do. Should it be the responsibility of portable_popen to be doing the swapping unbeknownst to the caller at this point?

abellgithub:
All of that slash swapping should be killed.

hobu:
> All of that slash swapping should be killed.

🔪

hobu:
Fixed by #4547

PR #4547, opened 2024-11-13
stop quietly replacing / with \ in portable_popen #4546

hobu:


hobu:
picked back to `2.8-maintenance` at e070bcdfbed466c0944a2c11b6b502c82c2436d7

Issue #4548, opened 2024-11-14
Support OGR feature selections for any geometry definitions 


hobu:
## Problem

The definition of geometries for stages is inconsistent. Sometimes we take in `wkt` (which also transparently takes in GeoJSON but this isn't widely advertised), on some stages it is called`polygon`, on others `geometry`, and sometimes we support an `ogr` definition (like in [`readers.ept`](https://pdal.io/en/2.8.1/stages/readers.ept.html) and [`readers.copc`](https://pdal.io/en/2.8.1/stages/readers.copc.html)). It is all a bit messy and arbitrary.

## OGR Definition

The `ogr` definition is the most powerful approach, and it offloads the work of dealing with geometries to OGR. The following example shows how to select a geometry and do some reprojection to create a buffered geometry:

```
    "ogr": {
        "datasource": "../desired.geojson",
        "sql": "SELECT ST_Transform(ST_Union(ST_Buffer(ST_Transform(geometry, 3857), 300)), 4326) as geometry  from desired where HUC8 in ('02050304')",
        "options": {
            "dialect": "SQLITE"
        }
    }
```

![image](https://github.com/user-attachments/assets/e1b1a05a-a3c5-43bc-b7c7-9b6f85b1175d)

## Desire

It should be possible to pass in an `ogr` definition to PDAL stages for any option that would otherwise take in a geometry in WKT or GeeJSON. In almost all instances, this is what you want – figuring out how to manually transform a geometry into WKT or GeoJSON, maybe through some CLI operation, is fraught and annoying. Our current behavior also makes that activity happen outside of the context of a PDAL pipeline operation. It is much easier to point to some data, define a selection with [OGR SQL](https://gdal.org/en/latest/user/ogr_sql_dialect.html), manipulate and transform it as shown above, and let PDAL consume the geometry for its operations.

## Proposal

* Centralize the implementation of the `ogr` option
* Allow any `wkt` or `polygon` or `geometry` option to *also* take an `ogr` definition for geometries in addition to simple WKT or GeoJSON
* Enhance the `ogr` definition with any missing capabilities

ibell13:
Here are all the stages that take geometry options right now:
 
Readers
- **CopcReader**: takes wkt/json polygon (_polygon_) and ogr json block (_ogr_)
- **EptReader**: takes wkt/json polygon (_polygon_) and ogr json block (_ogr_)
[...]

PR #4550, opened 2024-11-14
New tutorial for 'pdal tindex create' and readers.tindex 

ibell13:
Made a new tutorial for some PDAL tile index workflows:
- creating a tile index with an S3-hosted tiled copc dataset, using `pdal tindex create`
- using readers.tindex to read the index where it intersects certain polygons, and writing a raster of HeightAboveGround for each.

When #4548 gets implemented, it could make some of the garbage I do to feed WKT into readers.tindex unnecessary; but right now I think that's the simplest way to do it.

Suggestions for clarification or improvement are welcome!


hobu (changes requested):

hobu:
picked up in #4599

PR #4551, opened 2024-11-15
Support nan and min/max filtering in expressions

hobu:
I opened this as a PR so we don't lose it. It still needs some tests and maybe a doc update.


hobu:
active in #4561 4561

Issue #4493, opened 2024-09-06
Poisson filter: Colors from pointcloud are not reconstructed


simplelife2010:
I am using a pipeline to convert a pointcloud into a mesh.

```
[
    {
        "type" : "readers.las",
        "filename": "ground_shifted.laz"
    },
    {
        "type": "filters.poisson",
        "depth": 9,
        "density": true
    },
    {
        "type":"writers.ply",
        "faces":true,
        "filename":"poisson.ply"
    }
]
```
The colors in the laz pointcloud are 16bit, so rgb values are between 0 and 65535. In the output ply file, rgb values are all set to 255.

```
ply
format ascii 1.0
comment Generated by PDAL
element vertex 30344
property float64 x
property float64 y
property float64 z
property uint16 intensity
property uint8 returnnumber
property uint8 numberofreturns
property uint8 scandirectionflag
property uint8 edgeofflightline
property uint8 classification
property uint8 synthetic
property uint8 keypoint
property uint8 withheld
property uint8 overlap
property float32 scananglerank
property uint8 userdata
property uint16 pointsourceid
property float64 gpstime
property uint8 scanchannel
property uint16 red
property uint16 green
property uint16 blue
property float64 normalx
property float64 normaly
property float64 normalz
property float64 density
element face 60590
property list uint8 uint32 vertex_indices
end_header
2.35787 -4.36061 30.9803 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 255 255 255 -0.0501183 -0.0122663 0.998668 12.4105
2.10782 -4.1586 30.9803 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 255 255 255 0.0830477 0.0251049 0.996229 12.3303
2.0246 -3.70051 30.9803 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 255 255 255 0.0810682 -0.00103172 0.996708 12.2637
2.35787 -3.65725 30.9803 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 255 255 255 0.00213294 -0.0635009 0.99798 12.3934
2.45993 -4.1586 30.9803 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 255 255 255 -0.149958 0.0294945 0.988252 12.4502
2.37781 -3.70051 30.9803 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 255 255 255 0.00213294 -0.0635009 0.99798 12.4045
[...]

PR #4553, opened 2024-11-15
document that filters.poisson doesnt scale colors #4493

hobu:


hobu:
picked back to `2.8-maintenance` at b047f14f8bc0702fb867028104b3bfbb1f196f89

PR #4560, opened 2024-11-22
Detect path to conf.py and set JSON/CSV paths relative

chambbj:
Kind of a stab in the dark at fixing the paths and ensuring that the dimension CSV gets created.


hobu:
picked back to `2.8-maintenance` at 794e99bbe370429d2ee44c440359387053feeabc

PR #4562, opened 2024-11-28
Convert day seconds

gui2dev:
## Extend `filters.gpstimeconvert` to handle day seconds.

In some formats (like `rxp`), `GpsTime` is in day seconds. I extended `filters.gpstimeconvert` to be able to handle that time representation.

* Added `GPS day seconds (gds)` format definition
* Mark `conversion` parameter as deprecated. Introduice `in_time` and `out_time` as new arguments.
*  Update documentation
*  Added tests

### DIFF
--- a/doc/stages/filters.gpstimeconvert.md
+++ b/doc/stages/filters.gpstimeconvert.md
@@ -92,13 +92,19 @@ Convert from wrapped GPS week seconds to GPS time.
 
 ## Options
 
+conversion (deprecated)
+
+: The time conversion. Must follow the pattern: "{in_time}2{out_time}". Can't be used with "in_time" and "out_time" \[Required\]
+
 in_time
 
-: The input time standard ("gt","gst","gws" or "gds"). \[Required\]
+: The input time standard ("gt","gst","gws" or "gds"). Must be used with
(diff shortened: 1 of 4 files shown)


abellgithub:
The argument changes are breaking changes.

hobu:
> If yes, how would you mark the argument deprecated?

See the `.addSynonym` and `setHidden` methods of `ProgramArgs`. Hide the old one and then add it as a synonym to the new one.

gui2dev:
> > If yes, how would you mark the argument deprecated?
> 
> See the `.addSynonym` and `setHidden` methods of `ProgramArgs`. Hide the old one and then add it as a synonym to the new one.

Ok, but I'm replacing one arg (conversion = in2out) by two (in and out). Would that make any sense?

gui2dev:
Ok, I already made changes to accept both in2out or in and out, but I can definitely make throw an error instead when using in2out.

Guilhem Villemin
________________________________

hobu:
> The argument changes are breaking changes.

The subsequent updates now support both argument approaches – `in2out`, `in`+`out`. Let's merge this and go forward. As I said up thread, I think the user base for this is quite small and they can adapt to these updates.

PR #4559, opened 2024-11-20
Automate release creation

hobu:
* Creates a draft release when we push an `x.x.x` tag
* Updates release notes 
* tweaks CI config a little bit

PR #4564, opened 2024-12-01
Validate checksums and version number in automated release CI

hobu:
The release automation builds the tarballs and sha256sum files using the `dist` build command and attaches it to a draft release whenever a tag of `x.x.x` is pushed. See #4559 for more details.

This PR validates the checksums and confirms that the version number that `pdal-config --version` emits from an installed build actually matches the tag that was used to create the draft release.


hobu:
Picked back to `2.8-maintenance` at c2faaeca175dc970e47068aebdd23affb77ed620

PR #4565, opened 2024-12-03
Add Pixi install instructions

chambbj:
Don't have to accept as-is and we can discuss the config of the environment. I had a little fun with the name and dropped in a lot of extras, which maybe we should instead install the bare minimum and teach folks how to add additional dependencies.


hobu:
picked back to `2.8-maintenance` at 387b84480e332473f0a61deacaf009d50b45dc03

PR #4566, opened 2024-12-04
Protect against including null strings when reading PDAL vlr types

hobu:
We were including null characters when reading PDAL JSON VLR types. Because we don't reinterpret the JSON anywhere, these NULLs were being passed on through and causing invalid output.

### DIFF
--- a/io/private/las/Utils.cpp
+++ b/io/private/las/Utils.cpp
@@ -40,6 +40,7 @@
 
 #include <pdal/PointRef.hpp>
 
+#include <pdal/util/Algorithm.hpp>
 #include <pdal/util/Extractor.hpp>
 #include <pdal/util/Inserter.hpp>
 #include <pdal/util/Utils.hpp>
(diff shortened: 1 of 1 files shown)


abellgithub:
There is no null termination of JSON, as such.

JSON that contains NULL bytes is invalid.

I'm not sure what the issue is, but I don't think removing any NULL bytes seems right.

Perhaps we should just throw an exception because the data is bad?

hobu:
> JSON that contains NULL bytes is invalid.

Correct. The issue is the VLR is a null-terminated string and we were reading `dataLength` not `dataLenght-1` and passing it into `Metadata::addWithType (json)`, which simply uses the string as-is. It has nulls in it because we're reading the last character, and then we have a problem.

The reason why I chose to replace any `\0`'s in the string instead of simply doing `dataLength-1` is because VLR `dataLength` regularly lies. I wouldn't expect us to be writing bad lengths, but I also wouldn't have expected us to slurp up JSON as a string without interpreting and validating it in `addWithType(json)`, which is what we do.

hobu:
picked back to `2.8-maintenance` at 814f0c1e59eb11eff454328a26affd49647e3622

PR #4567, opened 2024-12-04
Support remote grid_option_file

hobu:


pdal/Options.cpp
```diff
@@ -166,21 +166,28 @@ StringList Options::toCommandLine() const
 
 Options Options::fromFile(const std::string& filename, bool throwOnOpenError)
```
abellgithub, line 167: This is fine, but there would be less change and it would perhaps be clearer if the function signature was changed to `Options::fromFile(std::string filename, bool throwOnOpenError)`

hobu: I wanted to avoid changing the signature. I don't think it will matter in the end because @msmitherdc has said maybe he doesn't need this at all.

msmitherdc:
No longer needed by GRiD, closing

PR #4568, opened 2024-12-05
Enhance `filters.gpstimeconvert` to make it streamable

gui2dev:
Implement `processOne` to make `Filters.gpstimeconvert` streamable.

### DIFF
--- a/filters/GpsTimeConvert.cpp
+++ b/filters/GpsTimeConvert.cpp
@@ -13,16 +13,16 @@
  *       notice, this list of conditions and the following disclaimer in
  *       the documentation and/or other materials provided
  *       with the distribution.
- *     * Neither the name of Hobunc. or Flaxen Geo Consulting nor the
+ *     * Neither the name of Hobu, inc. or Flaxen Geo Consulting nor the
  *       names of its contributors may be used to endorse or promote
  *       products derived from this software without specific prior
  *       written permission.
  *
  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
- * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIESNCLUDING, BUT NOT
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
  * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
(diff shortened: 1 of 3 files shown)


filters/GpsTimeConvert.cpp
```diff
@@ -85,14 +83,18 @@ void GpsTimeConvert::testTimeType(std::string& type)
...
 {
+    m_first = true;
+    m_lastTime = std::numeric_limits<double>::lowest();
+    m_numSeconds = std::numeric_limits<int>::lowest();
     if (!m_conversion.empty() && (m_inTime.empty()) && (m_outTime.empty()))
     {
         m_conversion = Utils::tolower(m_conversion);
-        std::vector<std::string> s = Utils::split(m_conversion,'2');
+        std::vector<std::string> s = Utils::split(m_conversion, '2');
         m_inTime = s[0];
```
abellgithub, line 93: This should check that the vector contains (at least) two items.

hobu:
This needs a `run` method too. See ReprojectionFilter for inspiration.

hobu:
> This needs a `run` method too. See ReprojectionFilter for inspiration.

@abellgithub we often forget to add the `run` method. I wonder if we shouldn't provide a default implementation.

filters/GpsTimeConvert.cpp
```diff
@@ -193,225 +210,174 @@ int GpsTimeConvert::dayStartGpsSeconds(std::tm date)
...
 
     return duration.count();
 }
 
-void GpsTimeConvert::unwrapWeekSeconds(PointView& view)
+void GpsTimeConvert::unwrapWeekSeconds(PointRef& point)
 {
     // any decrease in time is interpreted as a week rollover
-    for (PointId i = 0; i < (view.size()-1); ++i)
+    while (point.getFieldAs<double>(Dimension::Id::GpsTime) < m_lastTime)
```
gui2dev, line 222: @hobu, @abellgithub we realized that test is somehow too strict. For some point cloud, you might have little decrease of `GpsTime` (namely when multiple returns).
We introduced another parameter called `wrapped_tolerance` that would update the test like this :
`point.getFieldAs<double>(Dimension::Id::GpsTime) + m_wrappedTolerance < m_lastTime`

Is that ok if I put this modification in this PR ?

hobu: > Is that ok if I put this modification in this PR ?

👍

hobu:
Ready to merge @gui2dev ?

gui2dev:
> Ready to merge @gui2dev ?

Absolutely !

PR #4570, opened 2024-12-09
Build docker image also for Apple Silicon (arm64) architecture

ludvik-krvotok:
Right now only Docker image for amd64 architecture is built. That results in poor performance on Apple Silicon macs. This change should build Docker image also for Apple Silicon (arm64) architecture.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -65,7 +65,7 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           file: "./scripts/docker/ubuntu/Dockerfile"
           context: .
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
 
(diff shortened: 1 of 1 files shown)


hobu:
Does the arm64 builder actually build on arm64 architecture now, or is it still emulated? The reason why we weren't doing this is because it was taking so long that it was timing out our builders to build `linux/arm64` emulated.

hobu:
/me looks 58 minutes later and sees the docker containers task is still not complete.

hobu:
2hr 40min later and it still isn't built. We are not merging something like this until there is native arm64 builds in public GHA CI.

ludvik-krvotok:
@hobu Was a bit idealistic PR :). I tried to build it locally, but that failed as used source (FROM) isn't build for arm64 as well.

PR #4573, opened 2024-12-10
New argument to process OGR JSON blocks

### LINKED ISSUE #4548: Support OGR feature selections for any geometry definitions

ibell13:
Creating a new class that resolves #4548; the OGRSpec class processes a JSON specification for reading geometry with OGR at the time an argument is added, rather than during a stage's execution.

Right now, the class 
- takes a string argument 
- parses it
- validates it into a JSON
- gets all OGR-specified polygons as PDAL `Polygons`, stores them

This covers the needs of most stages that could potentially use an OGR datasource, like `readers.ept` and `readers.copc`. It's implemented there, as well as in `filters.geomdistance` (where it only uses a single polygon from the container we create). It still needs to be added to `filters.crop`, `readers.tindex` ~~and `filters.overlay`~~ but that should be trivial.

Right now, it could use some reworking to allow non-`Polygon` geometries to be specified. This might be nice for consistency's sake, for the few times this gets used currently (or maybe future additions, like 2.5D polygons?). I've been having some trouble designing a good solution for that, though.

### DIFF
--- a/pdal/private/OGRSpec.cpp
+++ b/pdal/private/OGRSpec.cpp
@@ -70,7 +70,7 @@ void OGRSpec::parse()
     else if (!(Utils::tolower(m_json.at("type").get<std::string>()) == "ogr"))
         throw error("'ogr' option must have 'type':'ogr' specified!");
 
-    // m_opts should maybe get cleared here
+    m_opts = {};
     for (auto& item : m_json.items())
     {
         std::string key = Utils::tolower(item.key());
@@ -110,7 +110,6 @@ void OGRSpec::parse()
         {           
             std::stringstream out;
             out << "unexpected field '" << key << "' in OGR JSON!";
-            // this is fatal - caught by fromString(). Should just log maybe, if that's possible
             throw error(out.str());
         }
     }
--- a/pdal/private/OGRSpec.cpp
+++ b/pdal/private/OGRSpec.cpp
@@ -1,0 +1,133 @@
+#include "OGRSpec.hpp"
+#include <pdal/private/gdal/GDALUtils.hpp>
+
+#include <nlohmann/json.hpp>
+
+namespace pdal
+{
+
+OGRSpec::OGRSpec()
+{}
+
+
+OGRSpec::OGRSpec(const std::string& ogrJsonStr)
+{
+    validateInput(ogrJsonStr); 
+    initialize();
+}
+
+OGRSpec::OGRSpec(const NL::json& ogrJson)
+{
+    validateInput(ogrJson);
+    initialize();
+}
+
+void OGRSpec::update(const std::string& ogrJsonStr)
+{
+    m_geom.clear();
+    validateInput(ogrJsonStr);
+    initialize();
+}
+
+void OGRSpec::update(const NL::json& ogrJson)
+{
+    m_geom.clear();
+    validateInput(ogrJson);
+    initialize();
+}
+
+void OGRSpec::validateInput(const std::string& ogrJsonStr)
+{
+    try
+    {
+        m_json = NL::json::parse(ogrJsonStr);
+    }
+    catch(NL::json::parse_error& e)
+    {
+        std::string s(e.what());
+        auto pos = s.find("]");
+        if (pos != std::string::npos)
+            s = s.substr(pos + 1);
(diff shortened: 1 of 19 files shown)


doc/stages/ogr.json
```diff
@@ -0,0 +1,13 @@
+
+{
```
hobu, line 2: I think we should have 

```
"type":"ogr"
```

as part of the definition of this thing

ibell13: Optional or required?

hobu: Required. It's not a free-form thing.

Anyone adding them to `readers.ept` or `readers.copc` without the `type` will get an error and have to update their pipelines.

abellgithub: Wouldn't this break existing pipelines?

hobu: Just mine 😄 

(I don't think the `ogr` option in `readers.ept` and `readers.copc` is widely used or known about)

abellgithub (commented):
This is very nice.

pdal/private/OGRSpec.hpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+        catch(const NL::json::exception& e)
+        {
+            std::string s(e.what());
+            auto pos = s.find("]");
+            if (pos != std::string::npos)
+                s = s.substr(pos + 1);
+            std::stringstream msg;
+
+            msg << "Failed to parse JSON field with error: " << s;
```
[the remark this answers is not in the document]
ibell13: I tried to get it to do that, but I think the function would have to take both `item.key()` & `item.value()` for that. unless there's some other way to dereference the json iterator that I missed.

pdal/private/OGRSpec.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include <nlohmann/json.hpp>
+
+namespace pdal
+{
+
+OGRSpec::OGRSpec()
+{}
+
+
+OGRSpec::OGRSpec(const std::string& ogrJsonStr)
```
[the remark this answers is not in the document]
abellgithub: It's fine to have test support. If you're using them, that's OK.

pdal/private/OGRSpec.hpp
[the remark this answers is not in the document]
abellgithub: I guess I should see what an actual error looks like :)

PR #4575, opened 2024-12-11
Use tabs for the OS variants of using Conda

chambbj:
These days, with `jupyter-book` we can use tabs to present OS-specific code samples.


hobu:
picked back to `2.8-maintenance` at a8a33723488ab16e4e4f715414f03b8a89dfeb13

PR #4577, opened 2024-12-12
Handle multiple extrabyte VLR definitions in `readers.las` and `readers.copc`

alavenant:
Hello, 

We have some troubles to read this data ([Uploading test.laz.zip…]()) who contains several extra attributes as extra bytes. 

The one who write this file choose to use several times the (LASF_Spec, 4) keys for added attributes. The Las 
specification encourages to use it one time as a table with all added attributes.  

Extract of Las specification : 
image

Actually, i'm not sure to understand well this sentence ; so i'm not sure of the uniqueness requirement.

In any case, my purpose is here to be able to read several times the (LASF_Spec, 4), so be able to read well the present las. 

Antoine


abellgithub:
I don't know what to think about this. My initial reaction is that the files you have are invalid and that PDAL should tell you that they're invalid.  I tend to frown on adding support to PDAL to read invalid data -- many questions come up.  The specification implies that there should only be a single VLR in a file with a unique user ID/record ID pair (see the "superseded" record ID). The language, though, isn't absolute and should be improved.

I would open an issue requesting clarification from the LAS committee. You can do that [here](https://github.com/ASPRSorg/LAS).

Of course, you can modify your own copy of PDAL to do whatever you like.

hobu:
I really don't want to support this. Doing so without complaint will allow these kind of files to proliferate  I want to throw an error and allow the user to provide their own EB VLR/schema via JSON as an option on `readers.copc` or `readers.las`. 

Of course, when we took a stand over 1.4 files writing GeoTIFF SRS for PDRFs 6-10, we ended up reversing it 🤷.

hobu:
> something like that

Yes, exactly. Complain to the user when we see this VLR configuration, mention in the error they can define their own schema for the EB VLRs via option, and let them use that to overcome the data configuration challenge so any data they write won't proliferate the issue.

alavenant:
I'm ready to purpose a new draft (at least for Las, not for copc). 

But I encounter a problem with one of my test. The code will read a json table with extended dimension as mentioned in the last message, but i'm not able to use it in the Options methods. This doesn't works :

```    
    Options ops1;
    ops1.add("filename", Support::datapath("laz/las_with_extra_byte.laz"));
    
    ops1.add("extra_bytes_vlr", "[{\"description\":\"RIEGL Extra Bytes\",\"values\":[{\"name\":\"Deviation\",\"type\":\"uint8\"}]},\
                                  {\"description\":\"\",\"values\":[{\"name\":\"confidence\", \"type\":\"uint8\"}]}]");
````
Could you help me to add correctly the table ? 

My code works well when I'm using pdal in command line.

Thanks

**Comment by abellgithub** [comment]
We already have the option `extra_dims` in readers.las. The only issue with using that is that the reader currently prefers the file-based VLR over the data specified in `extra_dims`. I think there are easy ways around this. Is there some advantage to a new JSON-like option other than the fact that it's JSON-like?

**Comment by hobu** [comment]
> Is there some advantage to a new JSON-like option other than the fact that it's JSON-like?

No, not really. As long as there's a consistent way to specify all of the bits you would need for a combined EB VLR entry...

**Comment by abellgithub** [comment]
There is no combining because you're ignoring the VLRs in this case, right? You just use the list as provided in the option to interpret the data.

**Comment by hobu** [comment]
> There is no combining because you're ignoring the VLRs in this case, right?

Yes, that's what I meant. User would need to specify a description that rolls up the definition of all of the EB dimensions and PDAL uses that in its pipeline going forward.

**Comment by alavenant** [comment]
Sorry, I'm a little bit confused.

If I understand you well : 
 * the actual "extra_dims" option will be transformed and will waiting for JSON description of inside extra byte VRL 
 * While reading the JSON description, a little warning if there is more than one VRL ("not a good understanding of the specification") 
 * same while reading VRL of the las
 * the code start to check that the extra byte VRL description of the las is equivalent to the JSON description
 * if check are ok, we read all the extra dimensions. 

The first "*" could be brittle for users. But I agree this is not useful to have 2 same kind of attribut. 

An other option could be to preserve the  "extra_dims" option as it is, and try to read it in more than one extra byte VRL (with warning if more than one extra byte VRL  and error if added dimensions are not same as inside VRL extra byte). 

Tell me what you think the right way, and I will code it ! _(Maybe after Xmas...)_

**Comment by hobu** [comment]
>  I doubt anyone will notice the change, though we can emit some sort of a warning.

Agree

**Comment by alavenant** [comment]
hello, 
The draft is updated.

**Comment by abellgithub** [comment]
I think we're out of sync. I was advocating for a minimal change. This changes the format of the `extra_dims` argument among other things.

**Comment by alavenant** [comment]
Okay, I think I understand what you mean : 
 * not change the actual "extra_dims" options (not json structure)
 * warning if more than one EB VLRs ; error if "extra_dims" option is not specified
 * check "extra_dims" are well included in EB VLRs
 * "extra_dims" override all the EB VLRs 
 * add a warning if there is one (or more) EB VLR and "extra_dims" option are not specified (or if all EB VLR content are not specified in "extra_dims" option)

Is it correct ?

**Comment by abellgithub** [comment]
I don't know what you mean by [check "extra_dims" included in EB VLRs]. Other than that, yes. I don't think you need to to any checking of the EB VLR if `extra_dims` is specified.  We're making it an override.

**Comment by alavenant** [comment]
Ok, I update the code. If you want to have a look. 
Alpine test do not pass, but it doesn't seems to be due to my code.

**Comment by abellgithub** [comment]
Sorry to take so long to look at this. I think this looks good. If it's OK with you I'm going to make some small changes (English and formatting) and you can look it over before I merge it.

**Comment by hobu** [comment]
> When going through this it seems that multiple extra dim VLRs are being combined.

This is fine with me, although we should make sure to warn when this is encountered. If we only take the first one, the user isn't going to be able to do anything about it for the others. I guess it is implemented so we can be liberal in what we accept 🤷

**Comment by alavenant** [comment]
@abellgithub, format and correct my bad english as you want ! 
I agree with @hobu, it leaves users the possibility to use all the extradims.

File: io/LasReader.cpp [hunk 1]
```diff
@@ -622,23 +620,39 @@ void LasReader::queueNextStandardChunk()
...
+            log()->get(LogLevel::Warning) << "Bad size for extra bytes VLR.  Ignoring.\n";
+            continue;
+        }
+        
+        lVrlEB.push_back(vlr);
     }
 
-    std::vector<las::ExtraDim> extraDims =
-        las::ExtraBytesIf::toExtraDims(vlr->data(), vlr->dataSize(), d->header.baseCount());
+    if (lVrlEB.size()>1)
```
abellgithub, line 637: What does `lVrlEB` stand for?

io/LasReader.cpp
```diff
@@ -622,23 +620,39 @@ void LasReader::queueNextStandardChunk()
...
+    {
```
abellgithub, line 638: I don't understand this. If `extra_dims` isn't specified you're going to throw. But my proposal, anyway, was that if `extra_dims` dims was specified, it overrode any VLRs:
[here](https://github.com/PDAL/PDAL/pull/4577#issuecomment-2557410820)

I still think that if you specify `extra_dims` you should ignore all the EB VLRs. Perhaps what we're now saying is that if you DON'T specify `extra_dims` then you should merge the EB VLRs and print a warning?

alavenant: This is the fact via the line who follows this test : 

```
    // extra_dims ovveride extra_bytes by default
    if ( d->extraDims.size() != 0 ) return;
```

I prefer to collect extra byte contained in the las in order to let a warning in case of several extra byte. 

But I agree that the function could be stopped early if the d->extraDims is specified. 

I'm gonna change the code

alavenant: I update the code. Is it ok for you ?

hobu:
Please rebase to master.

alavenant:
> Please rebase to master.

Done

PR #4589, opened 2024-12-17
Handle unknown data types in `readers.arrow` and `writers.arrow`

hobu:
When given parquet or arrow files with `STRUCT` or `STRING` types in them, we were failing in `readers.arrow`.

This PR also adds `ConstPointTableRef PointRef::table() const` to provide convenient access to the PointRef's active table. `writers.arrow` was maintaining its own map of `<dimName,dimId>` because this was not available.


hobu:
Picked up in #4611

PR #4593, opened 2024-12-18
Add glob functionality for local and remote files

### LINKED ISSUE #4574: arbiter support for tindex
I had asked about this in a different channel and Howard described it roughly as "not currently supported but would be useful". I recently ran into this again and thought it was worth opening an issue here.

I have a very large bucket of laz files on `s3` I want to make a geometry layer of tile extents of. Currently, `tindex` does not seem to support the normal arbiter abilities you get with other `pdal` calls. 

```bash
$ pdal tindex create "pdal-tindex.gpkg" -f geopackage "s3://somebucket/someprefix/*.laz"

PDAL: Couldn't find files to index: s3://somebucket/someprefix/*.laz.
```
It would be very helpful to be able to point `tindex` at arbitrary paths rather than having to download/manually glob everything out.

Old but related: https://github.com/PDAL/PDAL/issues/2153

ibell13:
Resolves #4574. In a new `pdal::Utils` function, Arbiter's `resolve()` is integrated to read remote files recursively with a glob pattern. Uses existing FileUtils glob function instead if file is local. Replaced `FileUtils::glob()` in  `PipelineReaderJSON`, `TIndexKernel` and `TileKernel`. 

I should probably make some tests for this, but it seems like it works so far. `PDALUtils.cpp` seemed like the best place for the new glob function, but it could go anywhere.


WY-CGhilardi:
I tested the conda package (ubuntu-conda-package) from the CI artifacts for this branch and still get a globbing failure locally. I can't seem to get any kind of verbose output to show the actual arbiter traffic like a regular `pdal info` call or something

```
$ VERBOSE=1 pdal tindex create --tindex "test-output.gpkg" --filespec s3://somebucket/someprefix/laz/*.laz -f GeoPackage
PDAL: Couldn't find files to index: s3://somebucket/someprefix/laz/*.laz.
```

WY-CGhilardi:
OK, confirming that this is working for me now locally

PR #4599, opened 2024-12-23
`pdal tindex` tutorial

hobu:
Continuing work on #4550


hobu:
@ibell13 I would expect this composition to work for creating the tindex:

```
curl https://s3.amazonaws.com/hobu-lidar/WI_Adams_2019/filelist.txt | pdal tindex create WI_Adams.parquet -f Parquet -s
```

But I get the error:
> PDAL: No driver for /Users/hobu/https://s3.amazonaws.com/hobu-lidar/WI_Adams_2019/USGS_LPC_WI_Statewide_2019_A19_0001.copc.laz

It must be prepending the pwd of the cwd to each file entry. I think we need to stop doing that.

PR #4600, opened 2024-12-27
support setting writers.copc.levels and writers.copc.node_threshold

hobu:


hobu:
I know you're not enthusiastic about this one @abellgithub, but do you have any issues with this PR as implemented?

abellgithub:
You can derive the levels from the count. Seems like a conflict and potentially confusing to have both.

I don't see how this works, generally speaking. You may never get to the reprocessor.

There are some strange hardcoded values.

hobu:
> You can derive the levels from the count. Seems like a conflict and potentially confusing to have both.

The PR currently uses the `level` directly and ignores the `node_threshold` if it is set by the user. I agree it could be confusing, but people tend to think about this in terms of point counts. I added both to give someone the option to be very explicit with `level`

> You may never get to the reprocessor

What should be implemented for limits in that case?

> There are some strange hardcoded values.

My thinking was to give people the ability to control the ranges around what had been our previous default, but still keep the possible ranges between $`2^{13}`$ (8192) and $`2^{17}`$ (131072).

hobu:
We're going to do this a different way.

Issue #4607, opened 2025-01-08
Forward OGR type declarations are fragile


rouault:
In GDAL master (and 3.10.1RC1, but now reverted for 3.10.1RC2), https://github.com/OSGeo/gdal/pull/11608 made sure that the expansion of the OGRSpatialReferenceH handle was consistent with whether GDAL was built in debug or non-debug mode, since the definition of some opaque types is different in both modes (which affect GDAL C++ ABI, but not the C one).
However PDAL's pdal/Geometry.hpp has a ``using OGRSpatialReferenceH = void *;`` which is only valid for non-GDAL debug builds
It would be safer if PDAL wouldn't redefine GDAL's types.

cf https://lists.osgeo.org/pipermail/gdal-dev/2025-January/060023.html

abellgithub:
I think we were trying to prevent people from needing GDAL headers to build with PDAL. But it has been a long time so I don't remember.

rouault:
> because GDAL is missing `gdal_fwd.h`

fixed per https://github.com/OSGeo/gdal/pull/11613

PR #4608, opened 2025-01-08
Use gdal_fwd.hpp when available for forward declarations

hobu:
Addresses #4607

### DIFF
--- a/pdal/Geometry.hpp
+++ b/pdal/Geometry.hpp
@@ -39,9 +39,16 @@
 
 #include <memory>
 
+#include <gdal_version.h>
+
+// Forward decl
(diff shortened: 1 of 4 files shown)


pdal/Geometry.hpp
```diff
@@ -39,9 +39,16 @@
 
 #include <memory>
 
+#include <gdal_version.h>
```
abellgithub, line 42: See #4 here: https://en.cppreference.com/w/cpp/preprocessor/include

abellgithub:
Looks fine. I might check around to see if we're doing that same thing in other places -- it wouldn't surprise me.

PR #4617, opened 2025-01-13
fixes md5 link to 2.8.3

smellman:
- fixes md5 link to 2.8.3 from 2.8.0 in [download page](https://pdal.io/en/2.8.3/download.html#download)


hobu:
doh!

Thanks for the fix. We should automate this part of the release process too.

PR #4624, opened 2025-01-17
upload .conda files from conda package build CI

hobu:


scripts/ci/conda/upload.sh
```diff
@@ -4,17 +4,19 @@
...
 
 ls
 pwd
 find .
 
-if [ -z "${ANACONDA_TOKEN}" ]
-then
-    echo "Anaconda token is empty, not uploading"
-    exit 0;
+set PACKAGE_PATTERN="*pdal*.conda"
```
rouault, line 14: As far as I can see your current conda-build version still produces .bz2 files. For some reason GDAL must use a more recent one than PDAL's one

hobu: Thanks. We want it to be producing `.conda` files going forward. I will update

PR #4625, opened 2025-01-17
New `FileSpec` class to handle HTTP headers and query in filename

ibell13:
Extended the specification for the `filename` option in reader stages. Along with taking a single string for `filename`, a json object can be used instead with the following options:

- path field: string filename
- headers field: json object of http header key-value pairs
- query field: json object of http query key-value pairs

```json
[
  {
    "type": "readers.copc",
    "filename": {
      "path": "https://some/remote/file.laz",
      "headers": {
        "headerkey": "headerval"
      },
      "query": {
        "querykey": "queryval"
      }
    }
  }
]
```
The `FileSpec` class will parse and ingest these options, to be accessed by readers. This hasn't yet been implemented in any stages.

### DIFF
--- a/pdal/FileSpec.cpp
+++ b/pdal/FileSpec.cpp
@@ -58,54 +58,31 @@ bool extractStringMap(NL::json& node, StringMap& map)
 
 } // unnamed namespace
 
-namespace Utils
-{
-    template<>
-    StatusWithReason fromString(const std::string& s, FileSpec& spec)
-    {
-        NL::json json;
-        // parseJson working with plain strings would be nice
-        if (s[0] == '{' || s[0] == '[')
-        {
-            auto status = Utils::parseJson(s, json);
-            if (!status)
-                return status;
-        }
-        else
-        {
-            // maybe there's a way to make a null json string type and add
-            // a value, but I didn't see it
-            // Or we can just assume it's the spec's filename and ingest it without
-            // doing anything w/ json, maybe?
-            NL::json stringJson(s);
-            json = stringJson;
-        }
-
-        return spec.parse(json);
-    }
-}
-
 
 FileSpec::FileSpec(const std::string& path)
 {
-    // same as fromString. This is used in PipelineManager to take a spec in ReaderCreationOpts
-    NL::json json;
-    Utils::StatusWithReason status;
+    Utils::StatusWithReason status = ingest(path);
+    if (!status)
+        throw pdal_error(status.what());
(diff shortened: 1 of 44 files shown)


pdal/FileSpec.cpp
```diff
@@ -0,0 +1,183 @@
...
+            return { -1, "'filename' sub-argument 'headers' must be an object of "
+                "string key-value pairs." };
+    }
+    node.erase(it);
+    return true;
+}
+
+std::ostream& operator << (std::ostream& out, const FileSpec& spec)
+{
+    // some weird stuff was happening in some stages w/o this
```
abellgithub, line 164: We should figure this "weird stuff" out.

pdal/PipelineManager.cpp
```diff
@@ -325,23 +325,31 @@ Stage& PipelineManager::makeReader(const std::string& inputFile,
...
     std::string driver, Options options)
 {
-    StageCreationOptions ops { inputFile, driver, nullptr, options };
+    ReaderCreationOptions ops { inputFile, driver, nullptr, options };
 
     return makeReader(ops);
 }
 
-
+// not removing this so stuff doesn't break
```
abellgithub, line 333: Make "for backward compatibility" ?

abellgithub (commented):
Looks pretty good.

connormanning:
Neat.

pdal/FileSpec.cpp
ibell13: It was about the string value inside an `Option` object, I'm pretty sure. It looked kind of wonky if only the filename was specified. It should work either way though.

pdal/Stage.hpp
```diff
@@ -387,6 +387,8 @@ class PDAL_EXPORT Stage
     void setupLog();
     void handleOptions();
     void countElements(const PointViewSet& views);
+    // set subclass-specific options after they've been processed
+    virtual void assignParsedOptions();
```
hobu, line 391: A consequence of adding this virtual is previous plugins will need to be recompiled. Just putting this in so I have a reminder when I make release notes.

abellgithub: If this is at the end of the list of virtual functions, it should probably be fine. We can try to think of another way to deal with this, too.

hobu: I don't think it is too much burden. We just need to highlight it when we release.

PR #4628, opened 2025-01-17
Implemented new filter separateflightline, edited docs of related filters

j-tenny:
The purpose of this filter is to separate points by flight line before going into filters.trajectory, thus an independent trajectory can be created for each flight path over an area of interest. This filter considers each return sequentially in order of GPS time and separates points into a new view if the difference in GPS time between two time-adjacent returns is greater than threshold `time_gap`.

### DIFF
--- a/doc/stages/filters.separateflightline.md
+++ b/doc/stages/filters.separateflightline.md
@@ -17,7 +17,7 @@ be sorted by ascending `GpsTime`.
 
 ```json
 [
-    "input.text",
+    "input.las",
     {
         "type": "filters.sort",
         "dimension": "GpsTime",
--- a/doc/stages/filters.separateflightline.md
+++ b/doc/stages/filters.separateflightline.md
@@ -1,0 +1,41 @@
+(filters.separateflightline)=
+
+# filters.separateflightline
+
+The **separate flight line filter** takes a single `PointView` as its input and
+creates a new `PointView` for each pass of an aircraft based on `GpsTime`. 
+`PointView` must contain the `GpsTime` dimension. If two sequential returns are
+separated by a difference in `GpsTime` greater than the threshold `time_gap`, it
+is assumed that the returns came from different flight lines. Points must first
+be sorted by ascending `GpsTime`.
+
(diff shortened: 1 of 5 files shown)


j-tenny:
Related to issue #4628

hobu:
@chambbj see any way this could just be a pipeline? 

It seems kind of specific and special.

j-tenny:
@hobu splitting using filters.groupby PointSourceID works in many cases but in many cases it produces a different number of segmentations compared to splitting by gaps in GPS time. Splitting by GPS time seems more reliable. Splitting by GPS time could also probably be integrated into filters.trajectory but I'm scared to touch that code because there is a lot going on in there.

chambbj:
> @chambbj see any way this could just be a pipeline?

I haven't been able to think of much we could do with pipelines to make this work. Only thing that really came to mind would be to generalize it and have it looks for gaps in any dimension requested. But I don't know how often that would get used.

hobu:
> looks for gaps in any dimension requested

I would rather have a `filters.gapdetect` that did this generally for any user-specified dimension when combined with `filters.sort`. This PR only works for one kind of dimension and scenario. There is a cost to having a filter, and specific ones like this generate maintenance load on the software while providing low generally applicability. 

Would you be willing to reformulate this to `filters.gapdetect`  that could be combined in a pipeline @j-tenny? We appreciate the contribution, and it is an interesting use case, but I'm concerned that you're its only user.

hobu:
I think it might make sense to add this as another mode to [`filters.divider`](https://pdal.io/en/stable/stages/filters.divider.html) instead of naming another filter. 

## Example `filters.divider` usage that divides input when a count/run of `threshold` points matches the expression provided by `break` 
```
    "example.las",
    {
        "type":"filters.divider",
        "break":"EdgeOfFlightLine == 1",
        "threshold":1
    },
    {
        "type":"writers.las",
        "filename":"out_#.las"
    }
```

hobu:
See if you can make #4689 work for you to break when `EdgeOfFlightLine == 1`

PR #4643, opened 2025-02-03
`tindex create` SRS validation

### LINKED ISSUE #4637: Add validation for multiple SRSs in `tindex create` input
`pdal tindex create` should behave like [gdaltindex](https://gdal.org/en/stable/programs/gdaltindex.html) when dealing with input files that have multiple spatial references, instead of silently accepting them: an option equivalent to `-skip_different_projection` should specify whether different SRSs will be accepted. This could throw an error, or just skip the file and give a warning (like gdal does) depending on what we want.

ibell13:
Closes #4637. Gives a warning when multiple SRSs are found in `tindex create` input, and adds `--skip_different_srs` option to skip files with a different SRS than the first one that's read.


abellgithub (commented):
LGTM

Issue #4648, opened 2025-02-14
Crash at runtime for example program [with fix]


Alex-Kent:
**Describe the bug**
Due to changes in the PDAL library one of the included examples compiles but fails at run-time.  Specifically, compiling and running `examples/reading-streamer/reading_streamer.cpp` fails:

```
user@host:~/PDAL-master/examples/reading-streamer$ ./reading_streamer
terminate called after throwing an instance of 'pdal::pdal_error'
  what():  sample streamer: Missing value for positional argument 'filename'.
Aborted (core dumped)
```
**Expected behavior**
The points in `input.las` should be displayed to the console, e.g.:

```
user@host:~/PDAL-master/examples/reading-streamer$ ./reading_streamer | head
Process point: 595500, 4.49094e+06, -0.55
Process point: 595500, 4.49095e+06, -0.67
Process point: 595711, 4.491e+06, -0.37
Process point: 595711, 4.491e+06, -0.36
Process point: 595711, 4.491e+06, -0.33
Process point: 595711, 4.491e+06, -0.35
Process point: 595711, 4.491e+06, -0.36
Process point: 595712, 4.491e+06, -0.45
Process point: 595712, 4.491e+06, -0.46
Process point: 595712, 4.491e+06, -0.43
```

**System/installation information:**
```
$ uname -a

Linux host 6.5.0-14-generic #14~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Nov 20 18:15:30 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

```
$ pdal --version

--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
pdal 2.9.0 (git-version: Release)
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

**Fix:**
Changing the `StreamProcessor` class to inherit from `NoFilenameWriter` instead of `Writer` fixes the issue.
(The program writes directly to STDOUT using `cout`, not to a file.)

**Related:**
OSGeo/grass/pull/3567, OSGeo/grass/issues/3552, and PDAL/PDAL/pull/4342.

PR #4649, opened 2025-02-14
Fixes the run-time "Missing value for positional argument 'filename' error in reading-streamer example

Alex-Kent:
Resolves https://github.com/PDAL/PDAL/issues/4648

Due to PDAL API changes the `reading-streamer` example crashed at run-time.  This change fixes that.


Alex-Kent:
Odd that those tests are failing, the example code is completely separate from the code that's failing (PDAL even builds fine if the `examples` directory is deleted [either before or after running CMake]).

hobu:
`pdal_io_nitf_writer_test` <-- unrelated or semi-related failure.

PR #4653, opened 2025-02-21
Add `writers.spz` and `readers.spz` (open 3DGS format from Niantic)

ibell13:
Vendored [nianticlabs/spz](https://github.com/nianticlabs/spz), creating a reader and writer for the `.spz` compressed 3D gaussian splat format.

Each point in a `.spz` file must contain the following attributes:
```
- XYZ position
- XYZ scale
- XYZW rotation
- RGB colors
- opacity
- 0/9/24/45 spherical harmonics
```
I've tried to bake in interchange to/from PLY, which is the standard format for storing uncompressed gaussian splats; most attributes, other than XYZ positions and RGB/Alpha, are assigned to proprietary dimensions that use a PLY naming convention (like `f_rest_*` for spherical harmonics). Dimensions which correspond to these naming conventions are expected to be present in the writer -- if not, it writes zeroes.


io/SpzWriter.cpp
```diff
@@ -0,0 +1,204 @@
...
+{}
+
+std::string SpzWriter::getName() const { return s_info.name; }
+
+void SpzWriter::addArgs(ProgramArgs& args)
+{
+    args.add("antialiased", "Mark the data as antialiased", m_antialiased);
+}
+
+void SpzWriter::initialize()
```
abellgithub, line 31: Is writing to a remote destination a thing? I didn't know this was a thing.

io/SpzWriter.cpp
```diff
@@ -0,0 +1,204 @@
...
+    {
+        m_scaleDims.push_back(layout->findDim("scale_" + std::to_string(i)));
+        if (i != 3)
+            m_rotDims.push_back(layout->findDim("rot_" + std::to_string(i)));
+    }
+    m_plyAlphaDim = layout->findProprietaryDim("opacity");
+
+    // find spherical harmonics dimensions and float RGB
+    //!! these are still assumed to be in the correct order
+    for (const auto& dim : layout->dimTypes())
```
abellgithub, line 56: Order isn't a good assumption.

abellgithub (commented):
I didn't spend that much time with the writer.

My general thought is that this should all use `UnpackedGaussian`. When you read, call `unpack()` and work with the unpacked data. This eliminates the need to copy the "unpacking" code from the library.

I would do the same when writing. Currently all the points are in the "cloud" at once, then they are packed into another data structure when `saveSpz` is called. It's a ton of data duplication. In the write case there is no interface for `UnpackedGuassian`. I would make one and make a PR against the SPZ library. That way you can just fill up the `UnpackedGaussian` for each point and then add it to the cloud through a `pack()` call or some such. The worst that can happen is that they don't want the PR and then you just use that internally.

Also, In the existing writer code, there are calls to `unpackXXX` which I don't think belong. When writing, you'd be packing, not unpacking, but I'm pretty sure that's happening in the library when you call `saveSpz` anyway -- see `saveSpz` and `packGaussians`. I'm guessing this doesn't currently pass a round-trip test.

vendor/spz/src/cc/load-spz.h
```diff
@@ -0,0 +1,85 @@
...
+GaussianCloud loadSpz(const std::vector<uint8_t> &data);
+
+// Loads Gaussian splat from a vector of bytes in packed format.
+PackedGaussians loadSpzPacked(const std::string &filename);
+PackedGaussians loadSpzPacked(const uint8_t* data, int size);
+PackedGaussians loadSpzPacked(const std::vector<uint8_t> &data);
+
+// PDAL: add for write support. Put this in another header?
+bool saveSpzPacked(const PackedGaussians &g, std::vector<uint8_t> *output);
+bool saveSpzPacked(const PackedGaussians &g, const std::string &filename);
```
dg0yt, line 71: This is not from upstream :disappointed:

dg0yt (commented):
Oops, another vendored library.

hobu:
> Oops, another vendored library.

They don't release it, don't really merge patches, and don't seem interested in supporting it as a community library. SPZ support in PDAL is highly experimental as of 2.9.0. No promises.

PR #4658, opened 2025-03-01
Add float-encoded RGB and Alpha as built-in dimensions

ibell13:
Adding new built-in dimensions that support R/G/B/Alpha as 0-1 floats (`RedF`, `GreenF`, `BlueF`, `AlphaF`). The alternate names are the standard for their equivalent .ply vertex attributes (at least for 3DGS data).

### DIFF
--- a/pdal/Dimension.json
+++ b/pdal/Dimension.json
@@ -675,5 +675,29 @@
     "name": "Duplicate",
     "type": "uint8",
     "description": "Indicator of whether or not a point is a labeled duplicate"
+    },
+    {
+    "name": "RedF",
+    "alt_names": "f_dc_0",
+    "type": "float",
+    "description": "Red channel represented as 0-1 decimal value"
+    },
+    {
+    "name": "GreenF",
+    "alt_names": "f_dc_1",
+    "type": "float",
+    "description": "Green channel represented as 0-1 decimal value"
+    },
+    {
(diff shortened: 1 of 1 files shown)


hobu:
I think these should be `RedD`, etc. instead of `RedF` and with type `float32`'s. I don't understand the `alt_name` of `f_dc_1` and the like though. I think that is maybe just special to SPZ and no one else.

ibell13:
I think those `alt_name`s are used for most/all gaussian splat data; the [web viewers](https://superspl.at/editor/) I've tried require those names. I found [some stuff describing it](https://github.com/adobe/USD-Fileformat-plugins/blob/main/ply/README.md?plain=1#L51) but couldn't find an official spec or anything.

abellgithub:
What's the reason for the additional RGB dims? You can make the dimensions any type you like. Do we need to support two RGB dim sets simultaneously?

hobu:
> Do we need to support two RGB dim sets simultaneously?

Maybe. The `readers.spz` returns RGB dimensions from 0-1.0f while PDAL's dimensions are [`uint16_t`](https://github.com/PDAL/PDAL/blob/master/pdal/Dimension.json#L82) by default. The code that marshals these would be much easier to read if it was clear which kind of RGB dimension we were working with when.

PR #4679, opened 2025-04-10
de-vendor gtest

### LINKED ISSUE #4673: Build issues with GTest
**Describe the bug**

On windows there is an issue with the CMake definition of Global for gtest, commenting this out uses the vendored gtest.

```
CMake Error at cmake/gtest.cmake:7 (find_package):
  find_package called with invalid argument "GLOBAL"
Call Stack (most recent call first):
  CMakeLists.txt:156 (include)
```

And

```
   Creating library test\unit\pdal_stage_factory_test.lib and object test\unit\pdal_stage_factory_test.exp
StageFactoryTest.cpp.obj : error LNK2019: unresolved external symbol "class testing::TestInfo * __cdecl testing::internal::MakeAndRegisterTestInfo(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> >,char const *,char const *,char const *,struct testing::internal::CodeLocation,void const *,void (__cdecl*)(void),void (__cdecl*)(void),class testing::internal::TestFactoryBase *)" (?MakeAndRegisterTestInfo@internal@testing@@YAPEAVTestInfo@2@V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@PEBD11UCodeLocation@12@PEBXP6AXXZ4PEAVTestFactoryBase@12@@Z) referenced in function "void __cdecl `dynamic initializer for 'private: static class testing::TestInfo * pdal::StageFactoryTest_Load_Test::test_info_''(void)" (??__E?test_info_@StageFactoryTest_Load_Test@pdal@@0QEAVTestInfo@testing@@EA@@YAXXZ)
bin\pdal_stage_factory_test.exe : fatal error LNK1120: 1 unresolved externals
``` 

Updating the vendored gtest library fixes the issue with the compilation.

hobu:
A limited version of #4674 to only de-vendor GTest.

This PR changes our default behavior with GTest to the following:

* Try to find GTest in the local environment and use them if `USE_EXTERNAL_GTEST` are ON (defaults to ON if either is found in the compilation environment)
* If we didn't find it locally, grab and build a fixed version GTest from GitHub as part of the PDAL build


hobu:
@abellgithub I am frustrated to have issues like #4673 and the repeated cycle of re-vendoring of GTest as new releases are made, and I would rather we pull and install a fixed version from GitHub every time we build instead of vendoring a frozen copy into our source tree. Additionally, the PR allows you to build from a GTest that's in your environment. This scenario has been happening in some of the packaging situations. 

Do you have any objection to this approach going forward?

abellgithub:
This the the first time that cmake has deprecated this in at least six years (maybe 10?)  and it took me about 5 minutes to modify the version numbers in the files, so to me it's not a big deal. The stability afforded the build process by having a known quantity is valuable (at least to me).

I have three concerns with downloading files. 1) I don't particularly like a build system putting things onto my computer that I didn't explicitly request 2) Any changes in the downloaded files can lead to breakage that is bad for users and maintainers 3) cmake is very automated at locating packages, which can be good for those who don't care or aren't changing things but bad if you want to make modifications to packages and then build with them -- in my experience it can be a problem to get all the package finding to work as you want.

Those are my thoughts. Do what you like.

hobu:
I'm confused by Number 1. We're currently vendoring an old version GTest and using it whether or not they already have it installed. I would say we're already acting contrary to the developer's wishes at least in regards to a potentially self-installed GTest that they may desired to be used (pretty sure we have a ticket or two on this too). When we started using GTest it was not a widely available dependency. It is now, but of course how it is compiled impacts whether or not a locally available version is suitable to use. 

Number 2 is handled by https://github.com/PDAL/PDAL/pull/4679/files#diff-2e8eeba3972935a4ffffbc2331c81f9a74e62ecb67b29919897fd8c7cbca7ef8R63 which always sets the remote version to 1.15.2

This PR specifically improves Number 3. It uses a locally installed GTest if it is available, it opts in to using it if found, and the behavior is controllable via CMake parameter.

PR #4680, opened 2025-04-10
guard pragma usages only relevant to msvc

hobu:


hobu:
We might consider something like https://stackoverflow.com/a/18463996/498396 if we need to do more of this kind of thing.

hobu:
picked back to `2.8-maintenance` at 2ff485231f434ede97c1fc026f340fabbadb249d

PR #4681, opened 2025-04-11
floor values before cast when creating grid in `filters.voxelcentroidnearestneighbor`

### LINKED ISSUE #4678: Unexpected Voxel Grid Alignment in filters.voxelcentroidnearestneighbor
## Input:
Image

## Command
```
pdal translate las_with_several_extra_byte_bloc.laz sub.laz filters.voxelcentroidnearestneighbor --filters.voxelcentroidnearestneighbor.cell=2
```

## Output
Image

## Issue
As you can see in the output, there's an unusual and unexpected pattern introduced after voxel-based subsampling.

This appears to be related to the following section of the PDAL source code: 
https://github.com/PDAL/PDAL/blob/204d9d1de4262a04f081b6746a026c716968fb37/filters/VoxelCentroidNearestNeighborFilter.cpp#L71

Specifically, the filter initializes the voxel grid using the coordinates of the first point in the cloud (seen in red in the image below, index 0):
Image

## Question / Discussion:

Is this voxel origin behavior intentional?

Most voxel-based filters (at least in my experience) typically align the voxel grid using the minimum bounds of the point cloud, rather than the first point. I understand that computing bounds upfront can be more expensive, so I’m guessing this was a deliberate optimization?

Still, the current behavior can lead to inconsistent and non-deterministic voxel alignments, depending on point ordering — which can affect reproducibility and visual quality.

If this behavior is unintentional or considered suboptimal, I’d be happy to contribute a patch or MR that uses the cloud’s minimum bounds as the voxel grid origin.

Shoutout to @JulesSanchez for helping identify and investigate this behavior.

trns1997:
Closes #4678

I dug into the issue and found the cause. The problem seems to be that negative values were being cast to 0—this inadvertently causes a “centerline” to form, stemming from the point at index 0. In other words, when a coordinate falls into a negative voxel based on its position, the plain cast was pushing it into the voxel at 0.

Using std::floor ensures that values on the negative side are correctly mapped to their respective negative voxel rather than being rounded up to 0. This change removes the symmetric pattern observed near the point at index 0.

### Visual Comparison

Before (Symmetric Pattern):
image

After (Corrected Mapping):
image

As seen in the images above, with the floor function, points that should belong to the negative voxel are now assigned properly, thereby eliminating the symmetric “+” effect that we were observing.

### DIFF
--- a/filters/VoxelCentroidNearestNeighborFilter.cpp
+++ b/filters/VoxelCentroidNearestNeighborFilter.cpp
@@ -81,9 +81,9 @@ PointViewSet VoxelCentroidNearestNeighborFilter::run(PointViewPtr view)
         double y = view->getFieldAs<double>(Dimension::Id::Y, id);
         double x = view->getFieldAs<double>(Dimension::Id::X, id);
         double z = view->getFieldAs<double>(Dimension::Id::Z, id);
-        ssize_t r = static_cast<ssize_t>((y - y0) / m_cell);
-        ssize_t c = static_cast<ssize_t>((x - x0) / m_cell);
-        ssize_t d = static_cast<ssize_t>((z - z0) / m_cell);
+        ssize_t r = static_cast<ssize_t>(std::floor((y - y0) / m_cell));
+        ssize_t c = static_cast<ssize_t>(std::floor((x - x0) / m_cell));
+        ssize_t d = static_cast<ssize_t>(std::floor((z - z0) / m_cell));
         populated_voxel_ids[std::make_tuple(r, c, d)].push_back(id);
     }
 
--- a/pdal/PointView.hpp
+++ b/pdal/PointView.hpp
@@ -66,7 +66,7 @@ typedef std::set<PointViewPtr, PointViewLess> PointViewSet;
 
 class PDAL_EXPORT PointView
 {
-    FRIEND_TEST(VoxelTest, center);
+    FRIEND_TEST(VoxelCenterNearestNeighborFilterTest, center);
     friend class Stage;
     friend class PointRef;
     friend class PointViewIter;
--- a/test/unit/CMakeLists.txt
+++ b/test/unit/CMakeLists.txt
@@ -448,7 +448,8 @@ target_compile_definitions(pdal_filters_hexbin_test
         H3_PREFIX=PDALH3
 )
 
-PDAL_ADD_TEST(pdal_filters_voxel_test FILES filters/VoxelTest.cpp)
+PDAL_ADD_TEST(pdal_filters_voxel_center_nearest_neighbor_test FILES filters/VoxelCenterNearestNeighborFilterTest.cpp)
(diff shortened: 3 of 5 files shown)


abellgithub:
This is the correct solution IMO. I think what happened was that this code previously used the minimum coordinate of the bounding box as the origin. In that case, simply casting was correct. When the code changed to use an arbitrary origin, the computation should also have been changed as you have done.

This does suggest a few tests are in order. I'll do them if nobody gets to them before me.

trns1997:
@abellgithub I totally agree with you. The subtle grid changes when choosing an arbitrary starting point seem negligible. That said, I'd suggest a quick test: generate a synthetic PointCloud with known output and simply assert that the computed voxel assignments match the expected ones after running the filter. What do you think?

trns1997:
@hobu @abellgithub , i added a simple test, which should test the algo more or less. I tested the orginal filter without the floor and the test fails due to the miassignement of the points in the voxel. Hopefully  everything looks good.

hobu:
The backport to `2.8-maintenance` failed:
```
The process '/usr/bin/git' failed with exit code 1
```
<details>
<summary>stderr</summary>

```
error: could not apply ec25bf783... add test
hint: After resolving the conflicts, mark them with
hint: "git add/rm <pathspec>", then run
hint: "git cherry-pick --continue".
hint: You can instead skip this commit with "git cherry-pick --skip".
hint: To abort and get back to the state before "git cherry-pick",
hint: run "git cherry-pick --abort".
hint: Disable this message with "git config set advice.mergeConflict false"

```
</details>

<details>
<summary>stdout</summary>

```
[backport-4681-to-2.8-maintenance eed770de3] floor values before cast when creating grid
 Author: trns1997 <trns1997@gmail.com>
 Date: Fri Apr 11 07:59:44 2025 +0200
 1 file changed, 3 insertions(+), 3 deletions(-)
Auto-merging pdal/PointView.hpp
Auto-merging test/unit/CMakeLists.txt
CONFLICT (content): Merge conflict in test/unit/CMakeLists.txt

```
</details>

To backport manually, run these commands in your terminal:
```bash
# Fetch latest updates from GitHub
git fetch
# Create a new working tree
git worktree add .worktrees/backport-2.8-maintenance 2.8-maintenance
# Navigate to the new working tree
cd .worktrees/backport-2.8-maintenance
# Create a new branch
git switch --create backport-4681-to-2.8-maintenance
# Cherry-pick the merged commit of this pull request and resolve the conflicts
git cherry-pick 89f91b7740dbc397b5392645c1d6ce86c9e0f7e6,ec25bf7832145e81f30dee12c4616e2cd1afbe37
# Push it to GitHub
git push --set-upstream origin backport-4681-to-2.8-maintenance
# Go back to the original working tree
cd ../..
# Delete the working tree
git worktree remove .worktrees/backport-2.8-maintenance
```
Then, create a pull request where the `base` branch is `2.8-maintenance` and the `compare`/`head` branch is `backport-4681-to-2.8-maintenance`.

PR #4688, opened 2025-04-15
`filters.hexbin`: allow sampling with small point counts

ibell13:
Changed the default behavior of `filters.hexbin`: now sets sample sizes for edge length calculation based on the point count being processed in `filter()`. No longer throws errors if the point count is smaller than the `sample_size` arg. Also added a test.


filters/HexBinFilter.cpp
```diff
@@ -159,6 +153,14 @@ void HexBin::filter(PointView& view)
 {
     PointRef p(view, 0);
 
+    // If we still need to calculate the edge length/resolution, set the sample size.
+    if (m_grid->sampling())
+    {
+        // Make sure the sample size set in addArgs isn't larger than the number of points
+        m_sampleSize = std::min((point_count_t)m_sampleSize, view.size());
```
abellgithub, line 160: Need `(std::min)` to stop Windows macro expansion.

hobu:
> PDAL: filters.hexbin: Sampling for hexbin auto-edge length calculation failed!

The windows build is failing with this error.

PR #4689, opened 2025-04-15
Implement `filters.divider.expression` to break on expression

hobu:
This implements the behavior desired in #4628 in a more generic way.

### DIFF
--- a/doc/stages/filters.divider.md
+++ b/doc/stages/filters.divider.md
@@ -3,12 +3,12 @@
 # filters.divider
 
 The **Divider Filter** breaks a point view into a set of smaller point views
-based on simple criteria.  The number of subsets can be specified explicitly,
-or one can specify a maximum point count for each subset.  Additionally,
-points can be placed into each subset sequentially (as they appear in the
-input) or in round-robin fashion.
+based on criteria.  The filter supports three modes – `partition`, `round_robin`, and
+`expression`. With `partition` mode, the number of subsets can be specified explicitly.
+`round_robin` allows you to specify a maximum point count for each subset. Finally,
+`expression` mode breaks up the point view based on a given `expression` and `count`.
 
-Normally points are divided into subsets to facilitate output by writers
+Points can be divided into subsets to facilitate output by writers
 that support creating multiple output files with a template (LAS and BPF
 are notable examples).
 
@@ -16,7 +16,7 @@ are notable examples).
 .. embed::
 ```
 
-## Example
+## `partition` Example
 
 This pipeline will create 10 output files from the input file readers.las.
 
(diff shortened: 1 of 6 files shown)


doc/stages/filters.divider.md
```diff
@@ -34,13 +34,56 @@ This pipeline will create 10 output files from the input file readers.las.
...
 mode
 
-: A mode of "partition" will write sequential points to an output view until
-  the view meets its predetermined size. "round_robin" mode will iterate
-  through the output views as it writes sequential points.
+: A mode of `partition` will write sequential points to an output view until
+  the view meets its predetermined size. `round_robin` mode will iterate
+  through the output views as it writes sequential points. `expression` will
+  iterate write sequential points into an output view until the view meets
+  the expression and `count` size.
```
chambbj, line 86: Is count the overall size? Or the number of occurrences of the expression?

filters/DividerFilter.cpp
```diff
@@ -85,27 +114,63 @@ void DividerFilter::addArgs(ProgramArgs& args)
     args.add("mode", "A mode of 'partition' will write sequential points "
         "to an output view until the view meets its predetermined size. "
         "'round_robin' mode will iterate through the output views as it "
-        "writes sequential points.", m_mode, DividerFilter::Mode::Partition);
-    m_cntArg = &args.add("count", "Number of output views", m_size);
-    m_capArg = &args.add("capacity", "Maximum number of points in each "
-        "output view", m_size);
+        "writes sequential points. A mode of 'split' will output new "
+        "views every time a 'capacity' number of points with the given 'expression'"
+        "are visited", m_args->m_mode, DividerFilter::Mode::Partition);
```
chambbj, line 119: Here I'm seeing `expression` combined with `capacity` and not `count` (which I think was the intent), but it's the first I've heard of a mode called `split`!

chambbj (commented):
I think your code is (probably) fine after a cursory glance. The purpose or intended behavior is what's a little confusing.

doc/stages/filters.divider.md
```diff
@@ -34,13 +34,56 @@ This pipeline will create 10 output files from the input file readers.las.
...
+expression `UserData == 122`. It can be used in combination with `count` to
+enforce a _run_ of points that must satisfy the expression before splitting
+the view.
+
+```json
+[
+    "example.las",
+    {
+        "type":"filters.divider",
+        "expression":"UserData == 122"
```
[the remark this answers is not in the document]
hobu: `mode` is implicitly `expression` if an `expression` is set. It seemed kind of redundant to have to write both, but it is ok to do so (see tests for an example).

filters/DividerFilter.cpp
```diff
@@ -85,27 +114,63 @@ void DividerFilter::addArgs(ProgramArgs& args)
...
-        throwError("Must specify either option 'count' or option 'capacity'.");
-    if (m_cntArg->set())
+    if (m_args->m_splitExpressionArg->set())
     {
-        m_sizeMode = SizeMode::Count;
-        if (m_size < 2 || m_size > 1000)
+        m_args->m_mode = DividerFilter::Mode::Expression;
+
+        if (!m_args->m_cntArg->set())
+            m_args->m_size = 1; // Default to 1 if the user didn't specify a break count
```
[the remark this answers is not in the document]
hobu: The idea of `expression` mode is that you break every time the `expression` matches `count` number of times. For unsorted data, this might make you the same number of views as there are points in the worst case scenario.

filters/DividerFilter.cpp
hobu: after thinking about this some more, I dropped the `count` thing entirely. All this now does is split based on the expression passing. You can control it by stacking together other filters.

filters/DividerFilter.cpp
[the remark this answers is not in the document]
hobu: `count` was dropped. It splits every time the `expression` passes.

doc/stages/filters.divider.md
hobu: `count` was dropped. See other comments.

PR #4668, opened 2025-04-02
Geometry.hpp: fix compatibility with GDAL master in debug mode

rouault:
Cf https://github.com/qgis/QGIS/pull/61316#issuecomment-2773077876

PR #4693, opened 2025-04-23
Use gdal_fwd for OGRLayerH if available

hobu:
This might address #4668 @dg0yt


dg0yt:
It is not enough. But I can start from here.

Issue #57, opened 2011-11-09
TerraSolid .bin files can have alternative color/time header indicators 


hobu:
From a private email:

```
Based on experience, and confirmed by TerraSolid itself, the header file of Terrascan 
binary file can content an error on the 'gps time' flag.
The flag value to define if records include 'gps time' can be 
1 or 8 (depending of some obscure reasons).
Your reading header code (src/drivers/terrasolid/Reader.cpp) is:

105 m_haveColor = (m_header->Color==1);
106    m_haveTime = (m_header->Time==1);

if m_header->Time is set to 8, m_haveTime is FALSE but should be TRUE.

I propose you (as did to my code) to change this line to:

105 m_haveColor = (m_header->Color!= 0 );
106 m_haveTime = (m_header->Time  != 0);

if you don't do that, you'll read binary records with shift of size_of(unsigned int).


knowing the header size and the individual record size, you also can check, 
according the file size, if:
- the number of point is correct,
- the color and gps time flags are correct.
```

Issue #14, opened 2011-08-27
Document oper== for Swig usage


mpgerlek:
We need to warn in our Notes that the function "operator==" isn't swig-friendly.

For classes likely to be swig'd, an equals() function should be provided (which forwards to oper==, or vice versa).

PR #4715, opened 2025-05-22
Segfault bug

oleg-alexandrov:
I managed to get PDAL QuickInfo to segfault. Apparently this function really does not like to be called before the reader performs an "execution".

That makes sense, but then it should not crash, but rather throw an error about uninitialized reader.

How to test:

Pull and build tests. Run:

ctest -V -R copc   

It will do:

```
51: [ RUN      ] CopcReaderTest.boundedRead2d
1/2 Test #51: pdal_io_copc_reader_test .........***Exception: SegFault  0.14 sec

```

### DIFF
--- a/test/unit/io/CopcReaderTest.cpp
+++ b/test/unit/io/CopcReaderTest.cpp
@@ -183,6 +183,7 @@ TEST(CopcReaderTest, resolutionLimit)
 
 TEST(CopcReaderTest, boundedRead2d)
 {
+
     BOX2D bounds(515380, 4918350, 515400, 4918370);
 
     // First we'll query the EptReader for these bounds.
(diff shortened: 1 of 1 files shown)


abellgithub:
I don't understand what you've done. Does the test crash as it currently is written? If so, can you provide a traceback?

oleg-alexandrov:
Yes, the test crashes as it is written. You can run it yourself and see.

oleg-alexandrov:
What I did is comment out a line, which says "execute". Without "execute", quickinfo cannot gather enough data to run.

oleg-alexandrov:
Looks like it managed to crash your regressions too.

https://github.com/PDAL/PDAL/actions/runs/15198614646/job/42748360485?pr=4715

#14 51.88  57/155 Test  #57: pdal_io_copc_reader_test ............................***Exception: SegFault  0.62 sec

hobu:
The segfault happens at https://github.com/PDAL/PDAL/blob/master/io/CopcReader.cpp#L691 because `reader.copc`'s pool isn't active yet. I suppose we need to update CopcReader.cpp's `preview()` implementation to wake the pool.

Presumably this issue is specific to `readers.copc`. Did you have a scenario where other readers were able to `preview()` without issue and are now having trouble with `readers.copc`?

oleg-alexandrov:
Thank you for looking into this. Regular LAS did not crash.

PR #4728, opened 2025-06-06
Move JSON out of FileSpec header

abellgithub:


hobu:
I no-op'd JsonFwd.hpp and of warnings squirt out where we need to remove the JsonFwd.hpp and the esri PageManager stuff still has `NL::json` references.

PR #4735, opened 2025-06-13
Add `filters.supervoxel` stage

bramton:
This PR adds support for the supervoxel algorithm. This is an unsupervised over-segmentation algorithm for point cloud data. The algorithm aims to preserve boundries. The algorithm is described in the paper "[Toward better boundary preserved supervoxel segmentation for 3D point clouds](https://doi.org/10.1016/j.isprsjprs.2018.05.004) and the implementation is provided [here](https://github.com/yblin/Supervoxel-for-3D-point-clouds). This PR is based on the implementation of the original authors.


hobu:
Neat!

Any chance for some kind of small test? The PDAL 2.9 release is imminent, and I'm going to include this once it compiles on all platforms, but having a test to prevent it from rotting would be prudent.

PR #4737, opened 2025-06-17
Add test for filters.supervoxel

bramton:
By popular demand, I have added a simple test for the filters.supervoxel stage 😉


hobu:
@bramton every test fails. Is that expected? 😉

bramton:
It seems the culprit is 43db5cd8f. Only the `.second` should have been removed as its return value is not used. Without any inserts the size of populatedVoxels will not change, and all tests will fail.

hobu:
gah, sorry. Good thing we had a test!

hobu:
reverted in 86f53a62bd5d9ab822eaa8b467ec829c65012285

PR #4738, opened 2025-06-17
`readers.stac` & `readers.tindex`: forward HTTP headers & queries to readers

ibell13:
Since `FileSpec` was added to replace the `header` and `query` options (specified in place of the filename), `readers.stac` and `readers.tindex` haven't been able to forward these options to their sub-readers. Allowed a partial `FileSpec` object (without the `path` field) to be specified in `reader_args` for these drivers. Consolidated all of the option setting for readers into `setReaderOptions()` so it can be aware of the filepath when iterating through the reader args. Documented it in the tindex & stac docs.

`--readers.stac.reader_args = '{"type": "readers.copc", "filespec": {"headers": {"header_key": "header_val"}, "query": {"query_key": "query_val"}}}'`

There are a couple issues with this: 

- The `filespec` reader arg that's being used isn't _actually_ sent to the reader, and would be an invalid `FileSpec` in a normal case (since it doesn't include the path)
- Since we aren't aware of whether the reader supports headers & queries, they are silently ignored if the reader can't use them

This seems like the best option to fix this without remaking `FileSpec` or messing with the base Reader class. I can add some more validation of the `filespec` in `setReaderOptions()` if it's needed.

### DIFF
--- a/doc/stages/readers.stac.md
+++ b/doc/stages/readers.stac.md
@@ -114,6 +114,14 @@ reader_args
 '[{"type": "readers.ept", "resolution": 100}, {"type": "readers.las", "nosrs": true}]'
 ```
 
+  HTTP headers & queries can be forwarded to the reader using a partial {ref}`filespec`
+  JSON object, omitting the `path` component:
+
+```bash
+--readers.stac.reader_args \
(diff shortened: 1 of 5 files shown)


hobu (approved):
This is a bit of copypasta for both the `reader.tindex` and `readers.stac` `reader_args.file_spec` and the documentation of them. A future PR should pull these together.

PR #4757, opened 2025-07-04
shut off nrvo for copy/move elison in windows

hobu:


abellgithub:
Why does conda windows work?

hobu:
> Why does conda windows work?

It doesn't run the tests, just the build and making sure that `pdal.exe` executes.

PR #4766, opened 2025-07-16
Update CMakeLists.txt

houyongkuo:
Adapt for GCC/GXX 8.x


hobu:
Please adapt something like this https://github.com/PDAL/PDAL/blob/master/CMakeLists.txt#L268 instead of hardcoding `stdc++` and `stdc++fs`, which are GNU-centric.

Issue #4769, opened 2025-07-18
Segfault in subsequent stages when upstream reader returns no points


trns1997:
I’ve encountered a segmentation fault in a PDAL pipeline where the upstream reader returns zero points — due to a bounding box with no points — and the downstream stages don’t account for this. While I first hit this with `filters.voxelcentroidnearestneighbor`, the root cause seems broader: many pipeline stages assume non-empty input, and fail (e.g., via segfault).

Example Pipeline:
```
{
    "pipeline": [
        {
            "bounds": "([-100, -100], [10, 10])",
            "filename": "https://na-c.entwine.io/nyc/ept.json",
            "type": "readers.ept",
            "tag": "readdata"
        },
        {
            "cell": 0.1,
            "type": "filters.voxelcentroidnearestneighbor",
            "tag": "voxelcentroidnearestneighbor"
        },
        {
            "filename": "test_fail.laz",
            "type": "writers.las",
            "tag": "writelas"
        }
    ]
}
```

This pipeline causes a segfault because the bounding box doesn’t include any points, so readers.ept emits an empty point view, which is then passed into the filter.

### GDB Backtrace
```
Thread 1 "pdal" received signal SIGSEGV, Segmentation fault.
0x0000728d3e025460 in pdal::ColumnPointTable::getFieldInternal(pdal::Dimension::Id, unsigned long, void*) const () from /home/builder/workspace/install_cpu/conda/envs/entwine/bin/../lib/libpdalcpp.so.17
(gdb) bt
#0  0x0000728d3e025460 in pdal::ColumnPointTable::getFieldInternal(pdal::Dimension::Id, unsigned long, void*) const () from /home/builder/workspace/install_cpu/conda/envs/entwine/bin/../lib/libpdalcpp.so.17
#1  0x0000728d3dc1a8c9 in double pdal::PointView::getFieldAs<double>(pdal::Dimension::Id, unsigned long) const () from /home/builder/workspace/install_cpu/conda/envs/entwine/bin/../lib/libpdalcpp.so.17
#2  0x0000728d3de2fcd1 in pdal::VoxelCentroidNearestNeighborFilter::run(std::shared_ptr<pdal::PointView>) () from /home/builder/workspace/install_cpu/conda/envs/entwine/bin/../lib/libpdalcpp.so.17
[...]

PR #4774, opened 2025-07-23
Protect some filters against empty views

hobu:


hobu:
This hopefully fixes #4769 and potentially a few others.

hobu:
I don't like this approach but it gets the job done. I wonder if we could have some kind of decorator-like thing we put on the filter that says "I don't work on zero-point PointViews". Or maybe *every* filter simply returns its `filter` and `run` methods in the face of zero-point PointViews.

trns1997:
> I don't like this approach but it gets the job done. I wonder if we could have some kind of decorator-like thing we put on the filter that says "I don't work on zero-point PointViews". Or maybe _every_ filter simply returns its `filter` and `run` methods in the face of zero-point PointViews.

@hobu. I think centralizing this behavior in a common base class makes a lot of sense. Maybe adding an “empty-check” in the constructor or entry point of Stage (or possibly Filter) so that any PointView with zero points skips calling filter() and run(). By default, this behavior ensures all filters gracefully handle empty views. And if a custom filter really does want to operate on an empty view, it can override or opt out of the check. That way we can avoid repetitive empty–view checks in every filte. Ensure consistent behavior for future filters without additional effort and it would make the exception explicit - "Previous stage passed empty view" or something like that.

abellgithub:
I think you should search history. If I recall, there are cases where you
want processing to continue.


Andrew Bell
***@***.***

hobu:
>  I think centralizing this behavior in a common base class makes a lot of sense. 

I agree. I don't like the approach of the PR, but once I started, I wanted to see how many filters I would need to touch and in which ways. There's three types of "zero point protection" happening:

1. A streamable filter with a `filter` method that has secondary effects like manipulating member variables.
2. A non-streamable filter with its own `run` implementation that may have secondary effects returns either an empty PointViewSet or the same PointView in a PointViewSet (what's the rules on this, btw? We have filters that do both, and the default `Filter::run` method returns the same PointView)
3. A filter that starts by updating/making a kdtree (which is implicit zero point protection)

## Proposed implementation approaches

> I think you should search history. If I recall, there are cases where you want processing to continue.

I think the following approaches would not disrupt the desired behavior while allowing us to remove the duplication of this PR (and many other filters that already had something)
 
### 1. streamable filters

Check if the PointView is empty in `Filter::run` and skip passing it to the implementation's `filter` method.

### 2. non-streamable filters

I think the place to affect this would be in `StageRunner::run`. If both the `m_keeps` and `m_skips` are empty, we simply don't `run` the stage and update `StageRunner`'s `m_viewSet` with the (empty) PointView.

### 3. kdtrees

I guess we can just leave these alone for now, but implementation of 1. is going to preempt things.

trns1997:
It sounds good to add an “empty-check” in the entry point of `Filter::run()`, and similarly in `StageRunner::run()` for non-streamable filters, so that any PointView with zero points simply skips `filter()` or `run()` by default. This ensures all filters gracefully handle empty views and eliminates repetitive empty-view checks.

By default, filters would skip empties and downstream stages would continue propagating the empty PointView (like hotwiring the pipeline). If a custom filter truly wants to operate on an empty view, it could override a virtual method like `skipEmpty()` or explicitly opt out of the check (this could also be a flag?).

hobu:
015b14f5b6d60c54067aad88473a18057f80f54d attempts to implement what I described in https://github.com/PDAL/PDAL/pull/4774#issuecomment-3113678754

hobu:
@abellgithub when a filter does `run()` that does nothing on a PointView with no points, is it it supposed to return the PointViewSet _with_ the PointView in it or _without_ it? We have filters doing both...

pdal/private/StageRunner.cpp
```diff
@@ -55,7 +55,22 @@ PointViewPtr StageRunner::keeps()
 void StageRunner::run()
 {
     point_count_t keepSize = m_keeps->size();
-    m_viewSet = m_stage->run(m_keeps);
+
+    // Don't run the stage if we're a filter and we're empty
+    bool bDoRunStage(true);
+    Filter* isFilterType = dynamic_cast<Filter*>(m_stage);
```
trns1997, line 61: @hobu I like that this logic is now in StageRunner, but I wonder if it makes sense for StageRunner to stay agnostic to the specific type of stage. Right now it’s checking dynamic_cast<Filter*>, which couples StageRunner to filters specifically.

Maybe instead we add a virtual method like skipIfEmpty() on Stage (default true), and have Filter override it if needed. Then StageRunner just calls that method and doesn’t need to know whether it’s dealing with a filter or something else. That way the skip logic stays centralized but type-agnostic.

abellgithub: I don't think there are use cases right now for running filters if there are no points, so adding a virtual call that does exactly what this does doesn't add anything. This is simple and can be changed later if needs change.

hobu: > Maybe instead we add a virtual method like skipIfEmpty() on Stage (default true), and have Filter override it if needed. 

I was trying to avoid adding another method because doing so would necessitate a major release. Waiting for feedback from @abellgithub about the entire approach, however. I know we've gone back and forth about this topic over the years.

PR #4784, opened 2025-08-09
OpenDroneMap support

NathanMOlson:
Hello! I've recently started working with OpenDroneMap and I'm hoping I can help reduce the reliance on custom forks of other open source projects (including PDAL). 

This PR exports some methods from `GDALGrid`. THese methods are used by OpenDroneMap's RenderDEM project here: https://github.com/OpenDroneMap/RenderDEM/blob/08590374d4db07e6053ad40f528674caca965ae2/render.cpp

RenderDEM uses `pdal::GDALGrid` as follows:

1. A `GDALGrid` is constructed.
2. Points are added using `GDALGrid::addPoint()`.
3. `GDALGrid::finalize()` is called.
4. A data pointer is obtained using `GDALGrid.data()`. This pointer is used to loop through all the points in the grid and check if there is data present in the grid or not.
5. If there is data in the grid, the data is written using `pdal::gdal::Raster::writeBand()`.

Can we export these methods in PDAL? Or can someone suggest a change to RenderDEM to achieve the same result without requiring PDAL to export these methods?


abellgithub:
These are in a "private" subdirectory to allow changes to be made without impacting others.

If you find this code useful, I'd suggest just copying it to your own repo and make it your own.

hobu:
@NathanMOlson At this point, I think we'd rather you grab a copy of the GDALGrid code in its entirety  and plop it into ODM and bury it in your own namespace or whatever. 

As @abellgithub mentioned, we'd like to reserve the right to change it without breaking anything but PDAL. If we push it up to public or export it, we're on the hook for it. One immediate thing I can think of that might break things is if we were to add alternative interpolation methods to it. We haven't done that yet, although it's been a desire.

NathanMOlson:
OK, thanks for your attention. As far as I can tell, ODM is using `GDALGrid` directly instead of the public `GDALWriter` interface because it is faster. At least part of the speedup is due to some bounds checking ODM is doing before `addPoint()`. This speedup could potentially be useful to PDAL.

I'll look into this a little more and see if there is anything I can contribute without adding to the public API.

abellgithub:
PDAL doesn't know the bounds before running GDALWriter unless you supply them, so it has to move things around as new points come in. There is a `bounds` argument that lets the user supply these bounds and eliminates all the data shuffling.

PR #4797, opened 2025-08-18
Provide arm version of docker image

valentinschabschneider:
I tested the arm version of this docker image for my project. As far i can see, pdal is designed to work on arm systems so why not provide a docker image for it.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -30,6 +30,11 @@ jobs:
           echo "are we pushing packages" ${{ env.PUSH_PACKAGES }}
           echo "event_name" ${{ github.event_name }}
           echo "ref" ${{ github.ref }}
+      - name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v3
+        with:
+          version: latest
       - name: Setup Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@v3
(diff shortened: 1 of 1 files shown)


hobu:
> pdal is designed to work on arm systems so why not provide a docker image for it.

It is indeed, but we don't provide arm containers because it takes 3+ hours for the CI to build. I don't understand why that is the case, but we can't have such a long delay as part of our normal CI operations. 

If you have any insight and can debug the build to discover why, I'll happily merge this update.

valentinschabschneider:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

arm runners are available for free in public repos. i changed the currently used `ubuntu-latest` to `ubuntu-24.04-arm`. I'm not sure how long the amd64 build will take, but based on experience, there shouldn't be a big difference.

hobu:
I had attempted to follow https://docs.docker.com/build/ci/github-actions/multi-platform/ without much success.

To use the arm runners and keep our existing x86 builds, we would need to have _two_ CI jobs and merge them together with a docker manifest, correct?

hobu:
A more complete implementation of this was added in #4810 by @j9ac9k which doesn't take so long to build.

valentinschabschneider:
very nice! thank you @j9ac9k

PR #4810, opened 2025-09-01
Create ARM64 PDAL docker image

j9ac9k:
Using docker multi-platform build, create both arm64 and amd64 platform docker images of PDAL.

The manifest generation is tied the `PUSH_PACKAGES` variable which makes it tough to test.  Open to feedback.

Supersedes #4797

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -12,8 +12,6 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
-env:
-  PUSH_PACKAGES: ${{ github.repository_owner == 'PDAL' && github.event_name != 'pull_request'  }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -86,3 +127,40 @@ jobs:
...
+      - name: Create and push multi-platform manifest
+        run: |
+          docker manifest create pdal/pdal:latest \
+            pdal/pdal:latest-amd64 
+            pdal/pdal:latest-arm64
+          docker manifest push pdal/pdal:latest
+        
+          docker manifest create ghcr.io/pdal/pdal:latest \
+            ghcr.io/pdal/pdal:latest-amd64 \
+            ghcr.io/plda/pdal:latest-arm64
```
j9ac9k, line 165: just noticed this typo here, will resolve after I test on my local branch

PR #4811, opened 2025-09-02
Alias PDAL_DLL => PDAL_EXPORT and warn users to upgrade if it is defined

hobu:


hobu:
Old pdal-python still uses `PDAL_DLL`, and we need a bit more transition time. 

We should be able to remove this by PDAL 2.10 or 2.11

PR #4850, opened 2025-10-28
Update download.md

danlessa:
fix typo


kasparas-k:
Was this really a typo? `pdal` was already listed before `python-pdal`, now it is listed twice in the same install command

hobu:
eh, good point. `python-pdal` is going to pull in `pdal` and `libgdal` anyway. I'll fix it.

PR #4852, opened 2025-10-29
Fix Connector segfault on 404 from /vsis3/ path

craigds:
`FileUtils::openFile()` returns nullptr when the file doesn't exist or isn't good.

The code then immediately tries to call `in->fail()` on the null pointer.

This fixes the issue by handling nullptr there and returning the appropriate not-found error


abellgithub:
Thanks!

PR #4855, opened 2025-10-29
Eliminate PLY property strdup memory leaks

elidwa:
Replace the static _strdup usage in vendor/kazhdan/Ply.h with statically
allocated buffers for all property names. The previous approach performed
heap allocations during static initialization and AddressSanitizer flagged
them as leaks on shutdown.

This mirrors how other PDAL components handle PLY registration strings and
keeps ASan clean without relying on suppressions.

### DIFF
--- a/vendor/kazhdan/Ply.h
+++ b/vendor/kazhdan/Ply.h
@@ -201,9 +201,26 @@ typedef struct PlyFace
 	int *vertices;
 	int segment;
 } PlyFace;
+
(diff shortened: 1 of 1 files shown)


abellgithub:
AFAICT `ReadProperties` and `WriteProperties` are entirely unused. Is there some reason we shouldn't just remove them?

abellgithub:
Thanks for this but I'm going to close and instead remove things as most of this code is unused.

Issue #4858, opened 2025-11-01
Specifying a custom classs ID for ground in ground classification and hag filters


kasparas-k:
Currently, ground classification filters `csf` `pmf` `skewnessbalancing` and `smrf` always assign `Classification = 2` to ground points, and height above ground filters `hag_nn` and `hag_delaunay` always expect the ground class to be 2, as per [ASPRS Standard LIDAR Point Classes](http://www.asprs.org/a/society/committees/standards/LAS_1_4_r13.pdf).

However, the same strict adherence to ASPRS Stanard classes is not observed in the noise filters `elm` and `outlier`, where the `class` parameter defaults to the ASPRS noise class 7, but can be specified by the user.

I would like to propose adding the additional class flexibility for the ground classification task as well, while keeping the default behavior compliant with the ASPRS standard.

In terms of the hag filters, I often work with pre-classified point clouds, which have custom class ID mappings, so I'd need to juggle class assignments in order to stick to the custom scheme. Up until now I was using [lidR](https://github.com/r-lidar/lidR) for height normalization (which allows specifying the ground class) but it's slower, hacky to integrate into non-R projects and infects my code with the GPL-3.0 license.

hobu:
> I would like to propose adding the additional class flexibility for the ground classification task as well, while keeping the default behavior compliant with the ASPRS standard.

If you are proposing to make PRs that implement this capability for these filters, please proceed. We can review and merge them as they are completed.

If you are proposing that someone else complete this effort, my firm is available for hire.

kasparas-k:
Thanks for the prompt response! I was going to work on the PR, but wanted to check in before writing any code. There may have been a very good reason for strictly disallowing custom ground class that I wasn't aware of.

hobu:
> There may have been a very good reason for strictly disallowing custom ground class that I wasn't aware of.

No good reason. Some of these filters pre-date more sophisticated things like `filters.expression`. A hacky way to make things work is to use `filters.ferry` to copy stuff into the `Classification` dimension and then copy it back out, but proper dimension selectivity on the filters themselves would be much better.

PR #4861, opened 2025-11-05
Add class argument for hag_delaunay, hag_dem and hag_nn filters

kasparas-k:
Takes care of the first half of issue #4858 
Height above ground for pre-classified point clouds that do not conform to the ASPRS standard ground class 2

### DIFF
--- a/doc/stages/filters.hag_dem.md
+++ b/doc/stages/filters.hag_dem.md
@@ -103,5 +103,9 @@ zero_ground
   `Z` value to raster DEM.
   \[Default: true\]
 
+class
+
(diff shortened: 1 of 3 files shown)


kasparas-k:
I should also add the same `class` option for `filters.hag_dem`
While it does not have such a note in the documentation, it also expects ground class == 2 when the `zero_ground` (default true) option is given

https://github.com/PDAL/PDAL/blob/c6519ad97efcd48c54b8c5cc22eed50d26c151f5/filters/HagDemFilter.cpp#L134-L141

abellgithub:
Thanks!

PR #4867, opened 2025-11-10
Add custom class arguments for ground classification filters

kasparas-k:
Custom ground and off-ground classes for `smrf`, `pmf`, `csm`, `skewnessbalance` and `sparsesurface` filters. 

When `only_ground` is set to true (default false to keep original behavior) only points classified as ground get their Classification value modified. I also tested with a .ply point cloud without a Classification field and confirmed that when the field is created, it is initialized to all 1 values, so it's not a dangerous setting even for unclassified point clouds. If the user specified ground_class = 1 and only_ground = true, this would make the filters a no-op on unclassified point clouds, but I think at that point it's the responsibility of the user, since it's a setting meant for pre-classified point clouds.

This pull request closes the issue #4858


hobu:
@abellgithub do we have a clean example of a mix-in that we could use to model this PR? This appears to do what is needed in an unintrusive way but it duplicates a bunch of code to do it. 

We definitely want an include for the doc bits – see `filter_opts.md` and its inclusion in the documentation to model that.

kasparas-k:
I think a mix-in would be nice, but I'm not immediately sure how I could write one as of now.

I need to have its `l_prepared` method to fire off in `Stage::prepare`
https://github.com/PDAL/PDAL/blob/0348903570150d0933b3bc8646027d55a90add9d/pdal/Stage.cpp#L195-L202

And to have its `l_addArgs` fire in `Stage::addAllArgs`
https://github.com/PDAL/PDAL/blob/0348903570150d0933b3bc8646027d55a90add9d/pdal/Stage.cpp#L106-L112

I could achieve this by adding another layer of inheritance onto `Filter` (`GroundFilter` for example, which calls Filter::l_* methods just like `Filter` does for `Stage`, and has additional member variables for the class ID control) but a mix-in feels much better. At the moment I'm just not sure how multiple inheritance works in C++, whether I can do it without messing with the base class

abellgithub:
My vote would be to merge this as is and then improve in another PR.

But does this need a test or two?

kasparas-k:
I added a common test file for all ground classification filters. Also, to avoid code repetition I used gtest's `INSTANTIATE_TEST_SUITE_P`. I didn't provide it a name generator, so the test name suffixes are indices of the filter name list. This is not a problem since when a test fails, gtest also prints the value of `GetParam()`.

I added a tiny point cloud specifically for testing ground filters, this is what the data looks like:
image

It is a heavily downsampled crop of a point cloud from [Poland's national geoportal](https://www.geoportal.gov.pl/en/data/lidar-measurements-lidar/), which has an Open license allowing any use case. 

It was my first time creating a new text file in this repository, so referencing other files and the contributing code guidelines guide, I added a copyright notice at the top. Am I supposed to add my name? I'm not super concerned about being directly credited for the test file, so I can update the copyright header as required.

PR #4882, opened 2025-11-26
Find neighbors in `filters.normal` with radius, move processing

ibell13:
Added the `radius` option to `filters.normal`; allows the neighbors required for calculating normal vectors to be specified as all points within a radius rather than always using k-nearest. Added test & updated docs.

Moved the calculation for single-point normals to a new file (`NormalUtils.hpp`) & results into a new class (`NormalResult`).


filters/private/NormalResult.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+    NormalResult() : m_normal(Vector3d::Zero()), m_curvature(0.0)
+    {}
+
+    Vector3d m_normal;
+    double m_curvature;
+
+    bool empty() const { return m_normal.isZero(); }
+
+    void calcNormal(PointView& view, PointIdList neighbors)
```
abellgithub: I'm not a fan of throwing exceptions like this. It obligates every user to catch and they may not even be aware that this function throws. I can't think of any reason not to just return some status.

The only places we throw that I'm aware of is when the use is entirely internal to something local and it would be really cumbersome to return an error.

abellgithub: The name of this function is weird. The class name is `NormalResult`. `calcNormal` seems redundant.

test/unit/filters/NormalFilterTest.cpp
```diff
@@ -251,4 +251,64 @@ TEST(NormalFilterTest, RampPlane2)
...
+    PointViewSet viewSet = filter.execute(table);
+    PointViewPtr outView = *viewSet.begin();
+
+    Dimension::Id nx = table.layout()->findDim("NormalX");
+    Dimension::Id ny = table.layout()->findDim("NormalY");
+    Dimension::Id nz = table.layout()->findDim("NormalZ");
+    Dimension::Id c = table.layout()->findDim("Curvature");
+
+    double expected = std::sqrt(2.0) / 2.0;
+    for (auto const& p : *outView)
```
abellgithub: Not a fan of `auto const` here. What is `p`? I have no idea. What's gained by `auto`? I know people do this all over, but I still think it's bad practice unless the alternative is worse.

filters/NormalFilter.cpp
```diff
@@ -119,9 +123,20 @@ void NormalFilter::prepared(PointTableRef table)
         m_args->m_up = false;
     }
 
-    // The query point is returned as a neighbor of itself, so we must increase
-    // k by one to get the desired number of neighbors.
-    ++m_args->m_knn;
+    if (m_radiusArg->set() && m_args->m_knn)
+    {
```
abellgithub: This seems like a confusing warning. It's one or the other (radius or count). If someone sets both, I would think that's a hard error because they don't know what they're doing. Just going on seems an invitation to trouble.

test/unit/filters/NormalFilterTest.cpp
```diff
@@ -251,4 +251,64 @@ TEST(NormalFilterTest, RampPlane2)
...
+
+    BufferReader reader;
+    NormalFilter filter;
+    Options filterOps;
+    filterOps.add("radius", 2.0);
+    filter.setInput(reader);
+    filter.setOptions(filterOps);
+    filter.prepare(table);
+
+    PointViewPtr view(new PointView(table));
```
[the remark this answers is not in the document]
ibell13: This stuff in the normal filter test was mostly a modification of the `RampPlane2` test (L#206); I copied the data from that and made a duplicate plane that was offset from the original data. If it was calculating it with 8 neighbors (default) I would get junk, but using a radius would consider each set of points separately.

filters/NormalFilter.cpp
abellgithub: You already have code that is checking if an argument is set...

filters/NormalFilter.cpp
```diff
@@ -130,12 +145,34 @@ void NormalFilter::compute(PointView& view, KD3Index& kdi)
...
     {
         // Perform eigen decomposition of covariance matrix computed from
-        // neighborhood composed of k-nearest neighbors.
-        PointIdList neighbors = kdi.neighbors(p.pointId(), m_args->m_knn);
-        auto B = math::computeCovariance(view, neighbors);
+        // neighborhood composed of k-nearest neighbors, or within radius.
+        PointIdList neighbors;
+        if (m_radiusArg->set())
+        {
+            neighbors = kdi.radius(p.pointId(), m_args->m_radius);
```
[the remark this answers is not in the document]
abellgithub: When you're doing radius check, it's almost inevitable that you're going to have points without neighbors. This seems like noise to me.  If you feel compelled to check this, you can do something in the other class to let you retrieve the error. I would just use some error codes.

filters/private/NormalProcessor.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+    void findNormal(PointView& view, PointIdList neighbors)
+    {
+        // Check if the covariance matrix is all zeros
+        auto B = math::computeCovariance(view, neighbors);
+        if (B.isZero())
+            return;
+    
+        SelfAdjointEigenSolver<Matrix3d> solver(B);
+        if (solver.info() != Success)
```
abellgithub: This is utility code. You shouldn't assume that someone wants their program to crash when calling it. You have a perfectly good return code you can use or you can implement some more detailed error checking, but throwing from some utility function is yuck.

PR #4971, opened 2026-03-07
Fix PointGrid zero-cell construction for small inputs

ClayWarren:
Ensure PointGrid never computes zero cells for small inputs.

Treat invalid approxPerCell as programmer error (assert(approxPerCell > 0)).

Add runtime guard for oversized cell counts before assignment to uint16_t.

Use Windows-safe (std::numeric_limits<uint16_t>::max)() form and include pdal_error.hpp for the throw path.


abellgithub:
I'd like to talk to you about this next week when you're working. I'm curious about a few things.

abellgithub (commented):
I find the PR description way too verbose. The title pretty-much says it all. AI seems create lots of text -- maybe because many people pay per word ;)

PR #5001, opened 2026-04-03
adapt dependabot grouping

hobu:


hobu:
taken from https://github.com/qgis/QGIS/pull/65676

PR #5006, opened 2026-04-16
fix curl ca bundle support (refs qgis/QGIS#65835)

jef-n:


hobu (approved):
Thanks for the fix! Amazing this one has been out there so long.

This should be upstreamed to https://github.com/connormanning/arbiter/ but @connormanning can do that.

jef-n:
> Thanks for the fix! Amazing this one has been out there so long.
> 
> This should be upstreamed to https://github.com/connormanning/arbiter/ but @connormanning can do that.

file layout changed there - and looking into that I thought a shorter and maybe better fix would have been to just move "CURL_CA_BUNDLE" into `caInfoKeys` instead of adding `caBundle`.

connormanning (approved):

hobu:
Upstreamed to Arbiter at https://github.com/connormanning/arbiter/pull/61 and pulled back to PDAL at #5027

PR #5007, opened 2026-04-17
`tindex create`: add support for writing STAC Geoparquet

### LINKED ISSUE #4870: Add option to write STAC GeoParquet from `pdal tindex`
Writing a tile index with [STAC GeoParquet](https://github.com/stac-utils/stac-geoparquet) schema would be very convenient.

* Add a `base_url` option to `pdal tindex` to concatenate to the item's path
* Add `id` from basename of resource
* Add `bbox` from whatever user specified (fast or slow) bbox

`pdal info myfile.las --stac` produces a GeoJSON stack item for a resource when read. The code from this should be combined to a common place and used for both the GeoParquet and GeoJSON STAC variants.

ibell13:
Closes #4870. Adds the `--stac-geoparquet` option when creating a tile index, allowing a STAC item to be created in the [STAC-geoparquet](https://stac-utils.github.io/stac-geoparquet/latest/spec/stac-geoparquet-spec/) format using the OGR Parquet driver. This generally mirrors the GDAL tile index implementation (https://github.com/OSGeo/gdal/pull/13431). The `base-url` option seen there is functionally equivalent to our `path_prefix` option. I haven't added their more granular options for construction of the `id` field (`id-method`, `id-metadata-item`).

A complete refactor of `TIndexKernel` was also completed, moving all file processing (for `create` & `merge` mode) into the `TindexProcessor` class (with `TileIndex` & `StacIndex` subclasses). OGR functions were moved into new `tindex::Dataset` & `tindex::Feature` classes.

I have added a number of fields from the STAC [projection](https://github.com/stac-extensions/projection) & [point cloud](https://github.com/stac-extensions/pointcloud) extensions to make the output more usable. Some of these may be unnecessary or outside the scope of `tindex`, and more or fewer fields could be written if necessary. The `pc:schemas` & `pc:statistics` fields are written as JSON strings, rather than native parquet structs. Generally, the fields written to the stac item conform to those created with `pdal info --stac`.

Possible issues/considerations:

- The bbox field (required in the spec) is created using the `WRITE_COVERING_BBOX` layer creation option. This was added in GDAL 3.9, so the PDAL required version (3.8) can be bumped to avoid this. I have a check in the `StacIndexBuilder` constructor for this.
- The `proj:geometry` field isn't added at the moment.
- Whether to add GDAL's `id` derivation methods


pdal/PDALUtils.cpp
```diff
@@ -167,6 +167,27 @@ std::string toJSON(const MetadataNode& m)
     return output;
 }
 
+std::string toJSON(const MetadataNodeList& mList)
```
abellgithub, line 170: A metadata node list isn't necessarily a list in JSON-land. It's just a bunch of nodes. I don't know where this is used, but a comment to that effect is at least warranted.

kernels/TIndexKernel.cpp
```diff
@@ -45,6 +45,8 @@
 #include <pdal/private/gdal/GDALUtils.hpp>
 #include <pdal/private/gdal/SpatialRef.hpp>
 #include <filters/private/hexer/HexGrid.hpp>
+#include <nlohmann/json.hpp>
```
abellgithub: Is this done? There seem to be comments that indicate that it's not.

This seems to be screaming for some sort of subclassing in some way rather than a bunch of `if` cases for STAC support.

pdal/PDALUtils.cpp
ibell13: This was used in `setStacFields()` because the `schemas` & `statistics` nodes (arrays of objects) weren't getting converted correctly (being output as arrays of values without names). I could probably change the behavior of the standard `toJSON` function to handle those cases but I didn't want to break things.

doc/apps/tindex.md
```diff
@@ -25,6 +25,7 @@ $ pdal tindex create <tindex> <glob>
 --lyr_name             OGR layer name to write into datasource
 --tindex_name          Tile index column name
 --ogrdriver, -f        OGR driver name to use
+--stac-geoparquet      Whether to write tile index as a STAC GeoParquet file
```
abellgithub: I'm still not a fan of this. Conficts with `--ogrdriver` which I think already serves this purpose.

kernels/private/tindex/Dataset.cpp
```diff
@@ -0,0 +1,241 @@
...
+    }
+}
+
+void Feature::setField(Field *field, const StringList& values)
+{
+    char** csl = NULL;
+    for (const std::string& s : values)
+        csl = CSLAddString(csl, s.c_str());
+    OGR_F_SetFieldStringList(m_feature, field->m_index, csl);
+}
```
abellgithub, line 82: Pretty sure there is a leak here. This is why IMO this is better:

```
std::vector<char *> toGDALStringList(const StringList& l)
{
   std::vector<char *> out;
    for (const std::string& s : l)
        out.push_back(s.data());
    out.push_back(nullptr);
    return out;
}
```

kernels/private/tindex/Dataset.cpp
```diff
@@ -0,0 +1,241 @@
...
+{
+    gdal::registerDrivers();
+
+    if (m_driverName == "ESRI Shapefile")
+        m_maxFieldSize = 254;
+}
+
+Dataset::~Dataset()
+{
+    if (m_dataset)
```
abellgithub: This line can go away.

kernels/private/tindex/Dataset.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+bool Dataset::openDataset()
+{
+    m_dataset = OGROpen(m_idxFilename.c_str(), TRUE, NULL);
+    return (bool)m_dataset;
+}
+
+bool Dataset::openLayer(const std::string& layerName)
+{
+    if (OGR_DS_GetLayerCount(m_dataset) == 1)
```
abellgithub: This is confusing to me. You take a name, but in the first case you don't use it. Should we check that the layer you get matches the name you pass?

kernels/private/tindex/Dataset.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+bool Dataset::createLayer(const std::string& layerName, const std::string& srsString,
+    const StringList& lcOptions)
+{
+    gdal::SpatialRef srs(srsString);
+    // This used to just write to log w/ LogLevel::Error
+    if (!srs)
+        throw TIndexError("Unable to import srs for layer creation");
+
+    char** papszOptions = NULL;
```
abellgithub: You could use the function I wrote above here too.

kernels/private/tindex/Dataset.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+    m_dataset = OGR_Dr_CreateDataSource(hDriver, m_idxFilename.c_str(), NULL);
+    return (bool)m_dataset;
+}
+
+bool Dataset::createLayer(const std::string& layerName, const std::string& srsString,
+    const StringList& lcOptions)
+{
+    gdal::SpatialRef srs(srsString);
+    // This used to just write to log w/ LogLevel::Error
```
abellgithub: Why the comment? Also why the change? The caller is checking the return.

kernels/private/tindex/Dataset.cpp
```diff
@@ -0,0 +1,241 @@
...
+        OGR_Fld_Destroy(hFieldDefn);
+    }
+    // Once all the fields have been created, the final layerDefn is ready
+    // & we can get the field indices
+    getFieldIndexes();
+}
+
+void Dataset::getFieldIndexes()
+{
+    if (!m_layerDefn)
```
abellgithub, line 227: Seems like you should grab the layer definition handle when you create the layer.

kernels/private/tindex/Dataset.cpp
```diff
@@ -0,0 +1,241 @@
...
+    OGRErr err = OGR_L_SetAttributeFilter(m_layer, query.c_str());
+    if (err != OGRERR_NONE)
+    {
+        std::ostringstream oss;
+        oss << "Unable to set attribute filter with OGR error '" <<
+            CPLGetLastErrorMsg() << "'";
+        throw TIndexError(oss.str());
+    }
+
+    bool output(false);
```
abellgithub: This variable is equivalent to `hFeat` if I'm not mistaken.

kernels/private/tindex/StacIndex.cpp
```diff
@@ -0,0 +1,80 @@
...
+    if (runBoundary(stats, stacFileInfo, manager))
+    {
+        MetadataNode readerMeta = reader.getMetadata();
+        MetadataNode statsMeta = stats.getMetadata();
+        MetadataNode infoMeta = info.getMetadata();
+        stacFileInfo.addMetadata(statsMeta, readerMeta, infoMeta, m_pcType);
+    }
+}
+
+void StacIndexBuilder::createExtraFields(const FileInfoPtr& fileInfo,
```
abellgithub, line 145: If you want a reference, seems like you should pass one in rather than convert.

kernels/private/tindex/StacIndex.cpp
```diff
@@ -0,0 +1,80 @@
...
+    feature.setField(m_linksField, stacFileInfo.links());
+    feature.setField(m_idField, FileUtils::getFilename(stacFileInfo.m_filename));
+    feature.setField(m_stacVersionField, STAC_VERSION);
+    feature.setField(m_stacExtensionsField, m_extensions);
+    feature.setField(m_srsField, SpatialReference(stacFileInfo.m_srs).getPROJJSON());
+    feature.setField(m_pcCountField, stacFileInfo.count());
+    feature.setField(m_pcEncodingField, stacFileInfo.encoding());
+    feature.setField(m_pcTypeField, m_pcType);
+    feature.setField(m_datetimeField, stacFileInfo.datetime());
+    // Not sure if schema and statistics need to be native parquet lists or if json is ok
```
abellgithub: Should we figure this out? Where is the reference?

kernels/private/tindex/StacIndex.hpp
```diff
@@ -0,0 +1,108 @@
...
+    StacFileInfo(const std::string& filename) : FileInfo(filename)
+    {
+        // stacProjection & stacPointCloud need these set in the root metadata
+        // for it to work. Should refactor to be less metadata dependent
+        m_root.add("filename", filename);
+        MetadataNode self = m_root.addList("links");
+        self.add("rel", "derived_from");
+        self.add("href", filename);
+        m_properties = m_root.add("properties");
+        m_extensions = { "https://stac-extensions.github.io/projection/v1.1.0/",
```
abellgithub: I don't see that this is used. There's a function that returns it, but I don't see it being called. For other attributes it's used in the builder, but in this case these are hardcoded in the builder.

kernels/private/tindex/TileIndex.cpp
```diff
@@ -0,0 +1,41 @@
...
+    m_mtimeField = m_dataset->defineField("modified", OFTDateTime);
+    m_ctimeField = m_dataset->defineField("created", OFTDateTime);
+}
+
+FileInfoPtr TileIndexBuilder::makeFileInfo(const std::string& filename)
+{
+    return std::make_unique<FileInfo>(filename);
+}
+
+void TileIndexBuilder::getFileInfo(FileInfoPtr& fileInfo)
```
abellgithub: The name here is a little confusing. Usually "get..." returns the thing you're gettings. Also, seems like you should be passing in a reference to the file info.

kernels/private/tindex/TIndexProcessor.cpp
```diff
@@ -0,0 +1,297 @@
...
+#include "Dataset.hpp"
+#include "TIndexError.hpp"
+
+#include <pdal/Polygon.hpp>
+#include <pdal/util/ThreadPool.hpp>
+#include <pdal/private/gdal/GDALUtils.hpp>
+#include <pdal/private/gdal/SpatialRef.hpp>
+#include <pdal/Polygon.hpp>
+
+#include <ogr_api.h>
```
abellgithub: Are these GDAL includes still necessary?

kernels/private/tindex/TIndexProcessor.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+//
+// Base class
+//
+
+TIndexProcessor::TIndexProcessor(const Args& args, const std::string& tileIndexColumnName,
+        const std::string& srsColumnName, const std::string& driverName, const std::string& tgtSrs,
+        const std::string& assignSrs)
+    : m_dataset(new Dataset(args.idxFilename, driverName)),
+      m_args(std::move(args)),
```
abellgithub: Seems confusing/dangerous to move here.

kernels/private/tindex/TIndexProcessor.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+TIndexProcessor::~TIndexProcessor()
+{}
+
+void TIndexProcessor::create(const StringList& files, PipelineManager& mgr)
+{
+    const std::string filename = files.front();
+    m_infos.reserve(files.size());
+
+    //!! could just pass these in directly
```
abellgithub: I would remove the comments or do what they say.

kernels/private/tindex/TIndexProcessor.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    bool output(false);
+    try
+    {
+        output = m_dataset->queryLayer(query);
+    }
+    catch(TIndexError& e)
+    {
+        std::ostringstream out;
+        out << "Unable to query layer for file '" << fileInfo->m_filename << "': "
```
abellgithub: I think this error string would confuse me if I were a user, but I can't think of what it should say.

kernels/private/tindex/TIndexProcessor.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+#include <ogr_api.h>
+
+#include <pdal/util/FileUtils.hpp>
+#include <pdal/private/gdal/GDALUtils.hpp>
+#include <pdal/Options.hpp>
+
+#include "TIndexBoundary.hpp"
+
+// Get GDAL's forward decls if available
```
abellgithub: Don't think these definitions are necessary here anymore.

kernels/private/tindex/TIndexProcessor.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+    // boundary options
+    bool fastBoundary;
+    bool doSmooth;
+    int32_t density;
+    double edgeLength;
+    uint32_t sampleSize;
+    std::string boundaryExpr;
+
+    // stac-specific
```
abellgithub: What's this?

kernels/private/tindex/TIndexProcessor.hpp
```diff
@@ -0,0 +1,139 @@
...
+    std::string m_originalSrs;
+
+    Field *m_tindexColumnNameField;
+    Field *m_srsColumnNameField;
+
+    //
+    // Only used in create()
+    //
+    LogPtr m_log;
+    //StringList m_files;
```
abellgithub: What's this?

kernels/TIndexKernel.cpp
```diff
@@ -153,9 +65,7 @@ std::string TIndexKernel::getName() const { return s_info.name; }
 TIndexKernel::TIndexKernel() : SubcommandKernel()
 //ABELL - need to option this.
```
abellgithub: Can probably remove this comment if we're not doing it.

kernels/TIndexKernel.cpp
```diff
@@ -318,185 +226,33 @@ StringList readFileList(const std::string& filename)
...
-
-    FieldIndexes indexes = getFields();
-
-    if (!m_wkt.empty())
+    std::vector<tindex::FileInfo> files;
+    try
     {
-        pdal::Polygon g(m_wkt, m_tgtSrsString);
-        OGR_L_SetSpatialFilter(m_layer, g.getOGRHandle());
+        // this should just be the base class but it's pure virtual
```
abellgithub: I don't understand the comment.

pdal/PDALUtils.cpp
abellgithub: Didn't we address this another way?

pdal/PDALUtils.hpp
```diff
@@ -264,6 +264,8 @@ inline void writeProgress(int fd, const std::string& type,
 
 std::string dllDir();
 std::string PDAL_EXPORT toJSON(const MetadataNode& m);
+std::string PDAL_EXPORT toJSON(const MetadataNodeList& mList);
```
abellgithub: Yeah, I thought I did something to avoid these. No?

kernels/private/tindex/Dataset.cpp
abellgithub: Actually, probably better to do something like this as there's no way for the strings to go away before being set:
```
void featureSetField(OGRFeatureH feature, int field, const StringList& l)
{
   std::vector<char *> sl;
    for (const std::string& s : l)
        sl.push_back(s.data());
    sl.push_back(nullptr);
    OGR_FSetFieldStringList(feature, field, sl.data());
}
```

kernels/TIndexKernel.cpp
ibell13: Wouldn't be difficult to if we wanted to add it.

kernels/private/tindex/Dataset.cpp
ibell13: Not since we destroy it, but I changed things to be closer to that.

kernels/private/tindex/Dataset.cpp
ibell13: I don't think the layerDefn gets updated to include all the field information one they're added unless you redefine it. So for me it seems best for it not to exist until all the fields are defined.

kernels/private/tindex/Dataset.cpp
ibell13: I think it's fine to take out the comment - the issue is that either we don't return info about it being an srs-related failure (by returning false), or (as it is now) it'll only throw that error and not the `Couldn't open or create layer` one in `TIndexProcessor::create`.

doc/apps/tindex.md
ibell13: It's not quite the same as the driver, since you should be able to create a non-STAC tile index with the geoparquet driver. But I agree that it should be done differently.

kernels/TIndexKernel.hpp
```diff
@@ -40,6 +40,8 @@
 #include <pdal/SubcommandKernel.hpp>
 #include <pdal/util/FileUtils.hpp>
 
+#include <nlohmann/json.hpp>
```
abellgithub: This can't be included here.

kernels/private/tindex/Dataset.cpp
```diff
@@ -164,10 +198,11 @@ bool Dataset::createLayer(const std::string& layerName, const std::string& srsSt
 
 void Dataset::createFields()
 {
-    for (Field& field : m_fields)
+    for (auto field : m_fields)
```
abellgithub, line 212: So curious why you would change the type to `auto`. Why do you prefer that?

doc/apps/tindex.md
abellgithub: Why should you be able to create a non-STAC tile index? Who would care? We define the output format from tindex.

kernels/private/tindex/TIndexBoundary.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+    std::string getName() const
+    { return "tindex-boundary"; }
+    double height()
+    { return m_grid->height(); }
+    std::string toWKT()
+    {
+        std::ostringstream out;
+        out.setf(std::ios_base::fixed, std::ios_base::floatfield);
+        out.precision(10);
```
abellgithub, line 32: Why the precision stuff? Does the output of hexer not have enough precision? Too much?

doc/apps/tindex.md
hobu: A good point, but we have had the older format out there for quite a while. Presumably people are using it, and the older format could be any OGR-writeable format, not just GeoParquet..

We can't write STAC GeoParquet-style metadata in any OGR-writeable format. Not all of the formats support the metadata storage, etc.

kernels/private/tindex/TIndexBoundary.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <pdal/PointRef.hpp>
+#include <pdal/PointView.hpp>
+#include <filters/private/hexer/HexGrid.hpp>
+
+namespace pdal
+{
+namespace tindex
+{
+
+class TindexBoundary : public Filter, public Streamable
```
abellgithub, line 14: This needs a comment as to why you have re-worked the boundary stuff. What's special about it?

kernels/private/stac/StacInfo.hpp
```diff
@@ -147,6 +149,25 @@ inline void stacProjection(MetadataNode& root, MetadataNode& statsMeta,
...
+inline void addDatetime(MetadataNode& properties, MetadataNode& readerMeta)
+{
+    //TODO make sure these are available
+    //For now, if there isn't date similar to laz/las/copc then use now.
+    try
+    {
+        std::string doy = getChild(readerMeta, "creation_doy").value();
+        std::string year = getChild(readerMeta, "creation_year").value();
+        properties.add("datetime", getDateStr(year, doy));
+    } catch (std::exception &)
```
abellgithub, line 163: This is a very large exception block (meaning, std::exception is everything). Is that necessary? I have no idea what exceptions might occur in those functions, but if it's some system-like thing you probably don't want to roll on.

kernels/private/tindex/Dataset.cpp
```diff
@@ -0,0 +1,280 @@
...
+{
+    if (layerName.size())
+        m_layer = OGR_DS_GetLayerByName(m_dataset, layerName.c_str());
+    else if (OGR_DS_GetLayerCount(m_dataset) == 1)
+        m_layer = OGR_DS_GetLayer(m_dataset, 0);
+
+    return (bool)m_layer;
+}
+
+bool Dataset::createDataset()
```
abellgithub, line 173: Should this check that `m_dataset` is null?

kernels/private/tindex/Dataset.cpp
```diff
@@ -0,0 +1,280 @@
...
+    {
+        std::ostringstream oss;
+        oss << "Unable to set attribute filter with OGR error '" <<
+            CPLGetLastErrorMsg() << "'";
+        throw TIndexError(oss.str());
+    }
+
+    OGR_L_ResetReading(m_layer);
+    auto hFeat = OGR_L_GetNextFeature(m_layer);
+    bool output = (bool)hFeat;
```
abellgithub, line 273: This is kinda confusing. Can't you drop `output` and just use `hFeat`?

kernels/private/tindex/Dataset.hpp
```diff
@@ -0,0 +1,127 @@
...
+    {
+        if (m_value.is_number_integer())
+            m_type = OFTInteger64;
+        else if (m_value.is_number_float())
+            m_type = OFTReal;
+        else if (m_value.is_array())
+            m_type = OFTStringList;
+        else if (!m_value.is_string())
+            m_subtype = OFSTJSON; // Dump the value when writing
+    }
```
abellgithub, line 59: Do you know the JSON is one of those types and valid?

doc/apps/tindex.md
abellgithub: I don't understand unless you're saying people were using Parquet in the past. Was that even supported? What would be incompatible if so? If that's the concern, can we not detect that case when reading?

kernels/private/tindex/StacIndex.cpp
```diff
@@ -0,0 +1,167 @@
...
+        dim.add("name", dimName);
+        Dimension::Id id = Dimension::id(dimName);
+        Dimension::Type t = Dimension::defaultType(id);
+        dim.add("type", Dimension::toName(Dimension::base(t)));
+        dim.add("size", Dimension::size(t));
+        root.addList(dim);
+    }
+    stacFileInfo.addSchema(root);
+
+    // If the optional stats filter was added in fillFileInfo, we still need
```
abellgithub, line 130: I still don't understand this comment. Why do you only run the pipeline if there are *more* than one stages on it? It's unclear to me why.

kernels/private/tindex/StacIndex.hpp
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    std::string statistics()
+    {
+        return jsonElement(m_root.children("pc:statistics"));
+    }
+
+    std::string datetime()
+    {
+        return getChild(m_root, "datetime").value();
```
abellgithub: I *think* this is coming from `private/stac/StacInfo.hpp`. This is pretty confusing. If you're going to borrow this it seems like it would be good to know where it's coming from -- maybe stick that function in a stac namespace or something? Or you could jsut copy it.

kernels/private/tindex/TileIndex.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+struct field;
+
+class TileIndexBuilder : public TIndexProcessor
+{
+public:
+    TileIndexBuilder(const Args& args, const std::string& tileIndexColumnName,
+        const std::string& srsColumnName, const std::string& driverName, const std::string& tgtSrs,
+        const std::string& assignSrs);
+    //void create(const StringList& files) override;
```
abellgithub: Remove?

kernels/private/tindex/TileIndex.cpp
```diff
@@ -0,0 +1,55 @@
...
+    return std::make_unique<FileInfo>(filename);
+}
+
+void TileIndexBuilder::fillFileInfo(FileInfoPtr& fileInfo)
+{
+    PipelineManager manager;
+    manager.commonOptions() = m_commonOptions;
+    manager.stageOptions() = m_stageOptions;
+
+    Stage& reader = manager.makeReader(fileInfo->m_filename, "");
```
abellgithub: `reader` doesn't appear to be used.

kernels/private/tindex/TIndexBoundary.hpp
ibell13: I think this is all carried over 1:1 from the previous implementation - just moved to its own file. But I can put some stuff in about why we aren't using an actual hexbin filter for it

kernels/private/stac/StacInfo.hpp
ibell13: I didn't write this but I'm assuming it's to guard for empty strings being passed into `getDateStr()`. I can add a check for the strings existing instead. I think there are lots of other ways the code in that file can be improved, though, so I might save it for another PR

kernels/private/tindex/Dataset.cpp
ibell13: I don't think so, since it needs to be destroyed before we return

kernels/private/tindex/Dataset.hpp
ibell13: We know it's valid when it gets parsed, and if it's a type that isn't listed (mostly `bool` & `object` type) we assign the subtype to JSON which calls `value.dump()`. The only major issue I can think of is if an array of non-strings or a mixed type array is the value, I couldn't think of a good check for that.

kernels/private/tindex/StacIndex.cpp
ibell13: Because when there is only a reader in the manager, we can get all the information we need from the `QuickInfo`; but when there is a stats filter (I could also check with `m_writeStats`) we need the metadata from the filter in `filllFileInfo` so we have to run the stages.

kernels/private/tindex/TIndexBoundary.hpp
ibell13: I think this was interpreted from `HexBinFilter`; it's an option with a default of 8 there, I can't remember why we decided on 10 here. I'm assuming we want it as precise as is reasonable before it gets handed to OGR.

kernels/private/tindex/Dataset.cpp
abellgithub: I don't understand the above comment.

PR #5018, opened 2026-05-15
Include <unordered_map>

nilason:
Include <unordered_map> where used, compilation fails on macOS 12 (AppleClang 13) and older with:

```
/opt/local/var/macports/build/pdal-7a63b7db/work/PDAL-2.10.1-src/filters/RelaxationDartThrowing.cpp:108:37: error: no template named 'unordered_map' in namespace 'std'
    using PointIdNeighborMap = std::unordered_map<PointId, KD3Index::RadiusResults>;
                               ~~~~~^
```

### DIFF
--- a/filters/RelaxationDartThrowing.cpp
+++ b/filters/RelaxationDartThrowing.cpp
@@ -42,6 +42,7 @@
 #include <numeric>
 #include <random>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 namespace pdal


io/private/copcwriter/Output.cpp
```diff
@@ -34,6 +34,7 @@
 
 #include <iostream>
 #include <limits>
+#include <unordered_map>
```
abellgithub: unordered_map is included in `Output.hpp`, so I don't know why it would be necessary here.

abellgithub (commented):
Strange that this only up here. Usually Apple Clang sucks in almost all the standard header files and this isn't an issue on Apple Clang 14.

Can you confirm the issue with Output.cpp?

nilason:
> Strange that this only up here. Usually Apple Clang sucks in almost all the standard header files and this isn't an issue on Apple Clang 14.

With AppleClang 14 and newer it works fine.

> Can you confirm the issue with Output.cpp?

No I can't. I searched for files using unordered_map but didn't include, and found this. I overlooked to check the header files. I now see it has been in this file for a long time, so that is most likely not an issue. I removed this change.

abellgithub:
This seems fine, but it seems like we're having some CI problems.

hobu:
yes, I'm working the CI churn

nilason:
> ... but it seems like we're having some CI problems.

#5015 aims to address some of them.

PR #5019, opened 2026-05-17
Make the use of Ccache optional for building libE57Format

nilason:
Configuration of libE57Format opportunistically uses Ccache if found, without possibility to opt out.
This is a rather unfortunate decision, especially for packaging systems (see eg. [bug report for MacPorts](https://trac.macports.org/ticket/74004)).

This introduces an option for "use Ccache for libE57Format", which defaults to 'OFF'.

An alternative solution could be to hard code the `WITH_CCACHE_E57` variable to OFF  in PDAL instead of making it optional.

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -86,8 +86,6 @@ option(BUILD_PLUGIN_E57
     "Choose if e57 support should be built" FALSE)
 add_feature_info("E57 plugin" BUILD_PLUGIN_E57
     "read/write data to and from e57 format")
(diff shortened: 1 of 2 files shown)


hobu:
I think I'd rather drop ccache usage here entirely. We don't want to have the effect of it being required to compile PDAL.

nilason:
> I think I'd rather drop ccache usage here entirely. We don't want to have the effect of it being required to compile PDAL.

Perfectly fine by me.

PR #5026, opened 2026-05-24
filters: fix UB in DimRange sort comparator

ncopa:
Using element addresses as a tie-breaker does not provide a valid strict weak ordering for std::sort once the elements are being reordered. Replace it with a deterministic field-based comparison so sorting the range list no longer relies on undefined behavior.

This fixes a segfault observed on Alpine Linux riscv64:

```
99% tests passed, 1 tests failed out of 144

Total Test time (real) = 127.50 sec

The following tests FAILED:
	  9 - pdal_io_arrow_writer_test (SEGFAULT)
Errors while running CTest
```

### DIFF
--- a/filters/private/DimRange.cpp
+++ b/filters/private/DimRange.cpp
@@ -170,22 +170,11 @@ void DimRange::parse(const std::string& r)
         throw error("Invalid characters following valid range.");
 }
 
-
+// Dim ranges sort so that all ranges of a dimension are contiguous.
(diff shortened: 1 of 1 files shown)


abellgithub:
@ncopa: I think this should still work and is more straightforward. I'd like to change the way all of this is done but this seems OK for now.

abellgithub:
Close #5028

Issue #5037, opened 2026-06-07
PDAL_CREATE_PLUGIN macro misnaming output


quillaja:
The plugin documentation clearly states that plugins must be named `libpdal_plugin_{TYPE}_{NAME}` but the PDAL_CREATE_PLUGIN macro creates them without the leading **lib**. I've also found this to be the reality recently when getting started with the plugin tutorial--pdal does not recognize the dll without **lib**.

https://github.com/PDAL/PDAL/blob/723d1b2ea03bc10deb1369a851c5046edee0a81e/cmake/pluginmacros.cmake#L19

I suspect this is the only line that needs to be altered.

hobu:
> I suspect this is the only line that needs to be altered.

I think it is the documentation that's wrong here (for windows). IIRC, the windows DLL loader look for a specific naming pattern and the "lib" part in the beginning is dropped off.

Regardless, this code has been untouched for more than a decade, and we aren't changing at this point.

quillaja:
I literally had this problem yesterday. On v2.10.
- **lib**pdal_plugin_filter_myfilter.dll = pdal on windows lists the filter when running `pdal -drivers`
- pdal_plugin_filter_myfilter.dll = pdal on windows does **not** list the filter when running `pdal -drivers`

Plus the code hasn't even existed for a decade. The blame says it was added 5 years ago by @abellgithub in PR #3588

hobu:
ok, so maybe some more investigation then. Please reopen your PR.

quillaja:
Should I make a new PR? It appears I can't reopen or take any actions with the original #5038

quillaja:
Thank you.

The macro appears to be a convenience wrapper around the couple of cmake lines required to create the plugin target. The [docs](https://pdal.org/en/2.10.0/development/plugins.html) [state](https://pdal.org/en/2.10.0/development/writing-filter.html#the-source) the format _with_ lib, but the included [cmake example](https://pdal.org/en/2.10.0/development/writing-filter.html#compilation) is missing that.

I'm using [vcpkg](https://vcpkg.io/en/package/pdal) to get pdal v2.10.1 for windows-x64, and building my filter against that. I'm using MS Build Tools v18.2.0.

[...]

PR #5039, opened 2026-06-07
Correct PDAL_CREATE_PLUGIN target name (take 2)

quillaja:
See #5037

Prepends "lib" to plugin target name, as expected by pdal.

This is a resubmitted clone of #5038  as requested by @hobu

### DIFF
--- a/cmake/macros.cmake
+++ b/cmake/macros.cmake
@@ -126,7 +126,7 @@ macro(PDAL_ADD_PLUGIN _name _type _shortname)
     set(multiValueArgs FILES LINK_WITH INCLUDES SYSTEM_INCLUDES)
     cmake_parse_arguments(PDAL_ADD_PLUGIN "${options}" "${oneValueArgs}"
         "${multiValueArgs}" ${ARGN})
-    if(MSVC)
+    if(WIN32)
         set(${_name} "libpdal_plugin_${_type}_${_shortname}")
     else()
         set(${_name} "pdal_plugin_${_type}_${_shortname}")
--- a/cmake/pluginmacros.cmake
+++ b/cmake/pluginmacros.cmake
@@ -18,9 +18,9 @@ macro(PDAL_CREATE_PLUGIN)
     set(PROJECT_NAME ${NAME}_${TYPE})
 
     if(MSVC)
-        set(${_name} "libpdal_plugin_${TYPE}_${NAME}")
+        set(TARGET "libpdal_plugin_${TYPE}_${NAME}")
     else()
-        set(${_name} "pdal_plugin_${TYPE}_${NAME}")
+        set(TARGET "pdal_plugin_${TYPE}_${NAME}")
     endif()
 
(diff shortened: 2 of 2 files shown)


hobu:
ok, the reason why I said 10 years ago is this behavior in `PDAL_ADD_PLUGIN` has been around forever. `PDAL_CREATE_PLUGIN` is similar, but was missing the branch for for the DLL/.so name difference on windows. I think what I just pushed is the right fix.

hobu:
@quillaja please confirm my changes still work for you and then I will merge and include in the 2.10.2 release. Thanks!

quillaja:
It's naming the filter dll with _lib_ when I use MSVC and not _lib_ when I use Clang--both from MS Build Tools/Visual Studio Installer--as one would expect from the change. I'm not really sure if it's a compiler thing or not, since the filter was detected and functioning whether I used MSVC or Clang, as long as the name started with _lib_. Maybe more of windows thing? Perhaps branching on `WIN32` instead of `MSVC` would be more appropriate if it's OS-related? I tried that as well and it always prepends _lib_ regardless of whether I use MSVC or Clang.

Anyway, it's ultimately no big deal to rename the file afterward now that I know that's a thing. It was just very confusing when I tried to do the filter tutorial for the first time. Thank you.

hobu:
> Perhaps branching on `WIN32` instead of MSVC would be more appropriate if it's OS-related?

done!

hobu:
picked back to `2.10-maintenance` at 82bda33ec94666b75e73d65c52d652c20aff2e16

MehdiChinoune:
On MinGW libraries are already prefixed with "lib", with this change we ended up with libraries/plugins called liblibpdal_...

hobu:
@MehdiChinoune can you provide an updated patch that appropriately conditionalizes the CMake configuration for MinGW? We have no way to test

MehdiChinoune:
> @MehdiChinoune can you provide an updated patch that appropriately conditionalizes the CMake configuration for MinGW? We have no way to test

Revert it to MSVC (which is true for VS and clang in MSVC mode) https://cmake.org/cmake/help/v4.3/variable/MSVC.html

hobu:
@quillaja please confirm https://github.com/PDAL/PDAL/pull/5039#issuecomment-4712060073 and make a new PR implementing it

quillaja:
@hobu @MehdiChinoune 
Would also be true of the the change to [PDAL_ADD_PLUGIN (macros.cmake)](https://github.com/PDAL/PDAL/blob/1ed9bd24d9982421b14f3fd95f0d52e6d6d81d57/cmake/macros.cmake#L134)? That and PDAL_CREATE_PLUGIN were both changed in e38162abee1e06268a16dbaf62c99df200a3fbea.

quillaja:
Sorry about that! I think I was using clang++ when I thought I was using clang-cl.  Using clang-cl does enter the MSVC case in the macro.

PR #5040, opened 2026-06-08
Make sure StageExtensions.hpp is only included #once

hobu:


hobu:
picked back to `2.10-maintenance` at 50d60996096e1c60bb9ab80b2a45df888d508624

PR #5046, opened 2026-06-13
groupby multiple dimensions

quillaja:
### Summary
It would be useful to be able to create unique groups from multiple dimensions using `filters.groupby`. This adds that functionality and an addition unit test to support it.

### Detail
For example, imagine you overlay zones and want to extract each flightline into a PointView. This was more difficult before and required a more complex pipeline. With the proposed modification, this is all that would be required (assuming you overlay values into ClusterID):
```json
{
        "type": "filters.groupby",
        "dimension": "ClusterID,PointSourceId"
}
```

These modifications should be **backwards compatible** with existing pipelines where only a single groupby dimension was supported.

In addition to the filter modification, I have added 1 utility hashing function to `pdal/Utils/Utils.hpp`. The hash function was taken from [boost](https://www.boost.org/doc/libs/latest/libs/container_hash/doc/html/hash.html#ref_hash_combine) and a document written by [Nicolai Josuttis (nico@josuttis.de)](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0814r0.pdf). Fundamentally it uses `std::hash` and a mixing function. I have added it to PDAL utils as an exported function because it may very well be useful in user-created plugins. No such functionality is available in the cpp stdlib.

An alternative approach to hashing, as seen in the commits, is using a `std::vector<int64_t>` of the dimension values as keys. In my testing I saw no human perceptible differences in performance, though I strongly suspect the hashing approach to be more performant. I did not perform official benchmarking, however.

### Other
I have tried to follow formatting and other conventions observed in the existing codebase, and have formatted using the project .clang-format where appropriate.

I will gladly update the documentation on pdal.org if the PR is accepted.

### DIFF
--- a/doc/stages/filters.groupby.md
+++ b/doc/stages/filters.groupby.md
@@ -34,7 +34,7 @@ By default the groups are ordered according to the order of first occurance with
 
 dimension
 
-: The dimension containing data to be grouped.
+: A list of 1 or more dimensions by which to group points.
 
 ```{include} filter_opts.md
 ```
--- a/doc/stages/filters.groupby.md
+++ b/doc/stages/filters.groupby.md
@@ -3,7 +3,7 @@
 # filters.groupby
 
 The **Groupby Filter** takes a single `PointView` as its input and
-creates a `PointView` for each category in the named [dimension] as
+creates a `PointView` for each category in the named {ref}`dimensions` as
 its output.
 
 ```{eval-rst}
--- a/doc/stages/filters.groupby.md
+++ b/doc/stages/filters.groupby.md
@@ -26,8 +26,21 @@ only points of a single `Classification`.
 ]
 ```
 
-```{note}
-By default the groups are ordered according to the order of first occurance within the input. To change this, use `filters.sort` first to order the points according to `dimension`.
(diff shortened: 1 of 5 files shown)


chambbj (approved):
I have no issues with this addition. Thanks!

quillaja:
Thank you. I'll add doc updates to this PR.

abellgithub (commented):
This looks good, but it needs a documentation update.

quillaja:
2 questions.
1. There was an old comment from Chambers (~9 years ago) `also need to check that we have a dimension with discrete values`. I reworded it slightly and added TODO. Is this something you'd like me to do? I presume the comment was originally made since it's not only mostly meaningless to group by a float dimension, but the fact that it's being read into an int and implicitly quantized is "dubious" behavior.
2. `filters.sort` has a "dimension**s**" synonym for its option. Do you want that for `filters.groupby` as well? I personally don't find it necessary, and it seems sort is the only filter that uses a synonym for pluralization. But I'm happy to add that if it's desired.

hobu:
> There was an old comment from Chambers (~9 years ago) also need to check that we have a dimension with discrete values. I reworded it slightly and added TODO. Is this something you'd like me to do? I presume the comment was originally made since it's not only mostly meaningless to group by a float dimension, but the fact that it's being read into an int and implicitly quantized is "dubious" behavior.

I don't think we need to do anything special. @chambbj ?

> filters.sort has a "dimensions" synonym for its option. Do you want that for filters.groupby as well? I personally don't find it necessary, and it seems sort is the only filter that uses a synonym for pluralization. But I'm happy to add that if it's desired.

No need.

PR #5047, opened 2026-06-13
application order of multidimensional sort in filters.sort

quillaja:
### Summary
Reverses the order, in code, by which sort dimensions are applied to points. This makes using `filters.sort` more natural when sorting by multiple dimensions.

### Detail
In applications that allow one to sort by multiple columns/fields, normally the user lists them in order of significance. For example, if I specify to sort by "A,B" I expect to see the values of A in order, then values of B sorted _within_ equal values of A, etc. This is how Excel and pandas work.

This is also how sorting works with regular integers, from a binary perspective, where the least significant bits are ordered _within_ the 'groups' of their more significant bits.
```
0: 0000
1: 0001
2: 0010
3: 0011
4: 0100
5: 0101
```

Expected (excel) when I would list dimensions `Dim1,Dim2`:
Screenshot 2026-06-13 152206

Screenshot 2026-06-13 152228

However, the `filters.sort` currently does the opposite. This is "programmer" order, for lack of better words. Again, if I were to list dimensions `Dim1,Dim2`, then `filters.sort` will produce this sorting. Note the 2nd image. :
Screenshot 2026-06-13 152206
Screenshot 2026-06-13 125813
Screenshot 2026-06-13 152249

To achieve the desired result with `filters.sort` right now, one must list dimensions in _reverse_ order: `Dim2,Dim1`.

The proposed change simply iterates through the dimensions using the reverse iterator instead of the forward iterator.

I have added a test `multidims_in_order_of_significance` to the filter's test suite.

### Other
If PR #5046 is accepted, the proposed change in `filters.sort`'s application of `dimension` would be a natural fit with multi-dimensional grouping. For example:
```json
{
    "type": "filters.sort",
    "dimension": "ClusterID,Classification"
},
{
    "type": "filters.groupby",
    "dimension": "ClusterID,Classification"
}
```
I suppose you might do this to follow the advice to sort before grouping described in [the note in `filters.groupby`](https://pdal.org/en/stable/stages/filters.groupby.html#example). You might want to influence the output PointViews for filters such as `filters.icp` where the order matters. Though possible now, it is not intuitive to have to remember that the dimensions specified for `filters.sort` must be _the reverse_ of other filters.


abellgithub:
I see no reason to change this behavior. The code has operated this way for at least ten years and such a change may break existing pipelines/usage.

quillaja:
The git blame shows that multiple dimension support in filters.sort was added only 1.75 years ago in 9258353e282feb2aeff9a6d39a9a3a4668429db1

quillaja:
Sorry about the force pushes to the branch. I was trying to fix [gh from thinking the commits that are already in your master are part](https://github.com/orgs/community/discussions/39031) of this PR.  🫤 

Edit: I got my fork and the feature branch correctly rebased to isolate the commits for this PR.

hobu:
I'm a little fuzzy, but I believe the ordering was this way to support allowing the user to have multiple `filters.sort` stages each with different `dimension` options in a pipeline. 

I don't have a strong opinion about changing the default ordering for `filters.sort` on multiple dimensions, but this PR would also need to clean up and make clear the documentation on all of this, as this is more of a documentation problem than anything.

quillaja:
@hobu I'm pretty certain this change shouldn't affect the case you describe where there are multiple sequential single-dimension sort filters in a pipeline. 

I agree about the documentation. I will add updates to doc/stages/filters.sort.md (assuming that's the correct file) to this PR.

hobu:
The comments are helpful, but the user docs also need to be updated with information about how this works, a note that the behavior is changing at PDAL 2.11 for multi sorting, and some kind of linkage to the groupby behavior which this PR aligns to.

abellgithub:
The code looks fine to me, but it might be cleaner to add the dimensions to the list (m_dims) in the order you want to sort, then you don't have to mess with reverse iterators and such.

hobu:
> add the dimensions to the list in the order you want to sort

Yeah, that would make the scenario with multiple `filters.sort` filters behave as you expect too.

quillaja:
> add the dimensions to the list (m_dims) in the order you want to sort

That works for me. One way or another the dimensions that the user enters in the pipeline have to be reversed to sort with respect to (what I think is) the expected significance of the listed dimensions. That reversal can happen in `filter()` or `prepared()`.

quillaja:
@hobu @abellgithub  I had some errors building the docs locally with the instructions in `doc\README.md`, but I was able to use jupyter book directly to generate the docs enough to check them.

PR #5052, opened 2026-06-16
address warnings from gcc 14.3

hobu:
GCC 14.3 complains about some stuff. This PR tries to shut it up.


abellgithub:
Do we have this (GCC 14.3) on CI somewhere? My preference is to fix the warnings rather than turn them off unless they're problematic.

But this is fine if you want to roll with it.

hobu:
> Do we have this (GCC 14.3) on CI somewhere? 

yes, the `ubuntu-latest` Conda builder is using it.

PR #5054, opened 2026-06-17
Remove PDAL_ADD_FREE_LIBRARY and provide better STATIC build support

hobu:


hobu:
`filters.trajectory` no longer building with this PR. We didn't have that in any build path except for the Conda builder

hobu:
We should probably leave a dummy `PDAL_ADD_FREE_LIBRARY` in the cmake configuration that is a no-op in case some external .cmake files were using it for whatever reason.

hobu:
@mwestphal Can you please test this PR? I'm trying to provide more convenient STATIC build support while accounting for @abellgithub removing some dead CMake configuration that probably isn't needed.

mwestphal:
> @mwestphal Can you please test this PR? I'm trying to provide more convenient STATIC build support while accounting for @abellgithub removing some dead CMake configuration that probably isn't needed.

Sure, ill test it

mwestphal:
@hobu it looks like this branch (cf9358e07fec211ce1ac1f62d006ebdc58309cef) is not working and Im not able to load any file after building statically.

hobu:
> not working and Im not able to load any file after building statically.

can you provide a little more detail on this? Is this on Windows? Something else? The tests are all passing STATIC or SHARED on Alpine at least.

mwestphal:
> > not working and Im not able to load any file after building statically.
> 
> can you provide a little more detail on this? Is this on Windows? Something else? The tests are all passing STATIC or SHARED on Alpine at least.

ArchLinux:

 - install gdal locally using pacman
 - build and install pdal with `BUILD_SHARED_LIBS=OFF`
 - build VTK against that pdal
 - build F3D against that VTK
 - use F3D to open file, not working

Issue #3791, opened 2022-06-17
PDAL MBIO driver fail to initieate with latest MB-System release


epifanio:
**Describe the bug**

Unable to use the MB-System reader  when building  PDAL from source `git: master`  using the latest MB-System release `v5.7.8`

The build returns no errors, but when attempting to run `pdal info` on an MB-System supported file, I got the following error:

```
pdal info --driver readers.mbio --readers.mbio.format="MBF_HYSWEEP1" /opt/mb-data/170_001_1942.hsx
PDAL: readers.mbio: Can't initialize mb-system reader: 

```
 

**Expected behavior**

I expect MB.IO PDAL driver to be able to read MB-Syem-supported files

Note:

Using MB-System `v5.7.5` -  the driver  works as expected. 

The initialization issue happen with MB-System >5.7.5 -- `v5.7.6/7/8` and with `git: master`

 
**System/installation information:**

```
$ uname -a
Linux 642537381f8d 5.15.0-37-generic #39-Ubuntu SMP Wed Jun 1 19:16:45 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

```
$ pdal --version
-----------------------------------------
pdal 2.4.1 (git-version: ceb023)
-----------------------------------------

```
**Reproduce the issue**

The following [`dockerfile`](https://gist.github.com/epifanio/6ca36e56e8a5f19464bd121166396688) builds a machine running PDAL from `git: master` and MB-System based on the version argument.

Example: 
```
docker build --build-arg MBSYS_VERSION=5.7.5 -t pdal-mbio:5.7.5 .
```

To reproduce the PDAL-MB-Sytem machine used to debug the issue - I added 3 lines at the end of the dockerfile to download and test an example file

epifanio:
@hobu  here it is  an updated [`Dockerfile`](https://gist.github.com/epifanio/f163c75e5de017af4f64cbc42dbc3a87)

epifanio:
Will mbsystem driver moved from source and/or marked as deprecated?

hobu:
Potentially. We can move it to https://github.com/PDAL/unsuppported-plugins

epifanio:
I guessed it will soon happen if nobody can look into why stopped working.
What will be the best way to avoid that and get the plug-in working again :)
Is there any party that can be involved? who was to propose/financing it in the first place, in the best scenarion they can maybe still be interested in bringing it back to life. 

As it is now:
[...]

PR #5056, opened 2026-06-19
readers.mbio: fix initialization + zero-point read with MB-System >= 5.7.6 (#3791)

epifanio:
## Summary

`readers.mbio` fails against **MB-System >= 5.7.6** (and current 5.8.x). This fixes the regression reported in #3791.

MB-System 5.7.6 added strict date validation in `mb_get_time()`, which `mb_read_init()` now applies to the begin/end times. The plugin passes **invalid hardcoded dates** (year 0 / `INT_MAX`, month 0, day 0), so:

1. **Initialization aborts** with the new `MB_ERROR_BAD_TIME` (17). That code is missing from `MbError`'s table, so the message renders **blank**: `Can't initialize mb-system reader:` (the empty text after the colon is the tell).
2. Even past init, the read loop treats any `mb_read()` `status == 0` as end-of-data and **bails on the first non-fatal record** (the comment/header records that precede survey data) → **zero points read**.

## Fix (3 small changes, `plugins/mbio/io/`)

- **`MbReader.cpp::ready()`** — use MB-System's canonical default epoch (`1962/2/21 .. 2062/2/21`) for `btime_i`/`etime_i`. It spans any real survey and passes the new validation.
- **`MbError.cpp`** — add code `17` (`"Bad time value."`) so the diagnostic is no longer blank.
- **`MbReader.cpp::loadData()`** — stop only on a **fatal** error (`error > 0`, incl. EOF); skip non-fatal records; extract only successful survey pings (`status == MB_SUCCESS && kind == MB_DATA_DATA`).

```diff
-    int btime_i[7] { 0, 0, 0, 0, 0, 0, 0 };
-    int etime_i[7] { (std::numeric_limits<int>::max)(), 0, 0, 0, 0, 0 };
+    int btime_i[7] { 1962, 2, 21, 10, 30, 0, 0 };
+    int etime_i[7] { 2062, 2, 21, 10, 30, 0, 0 };
...
-        if (status == 0)
+        if (error > 0)
         {
-            if (error > 0 && error != MB_ERROR_EOF)
+            if (error != MB_ERROR_EOF)
                 throwError("Error reading data: " + MbError::text(error));
             return false;
         }
-        if (kind == 1)
+        if (status == MB_SUCCESS && kind == 1)
```

## Testing

Verified on Ubuntu 24.04 with MB-System `master` and PDAL: `readers.mbio` reads bathymetry from MB-System's bundled `mb24`/`mb43` test files (58 points) and from real **Reson 7k** (`.s7k`, format 89) survey data. Before: init aborted with the blank error; after fixing only init: 0 points; after all three fixes: full data.

## Notes / possible follow-ups

- On modern glibc, building the plugin also needs `<rpc/rpc.h>` from **libtirpc** (MB-System's installed `mb_define.h` includes it under `CMAKE_BUILD_SYSTEM`); that's a build-environment concern handled outside this source change.
- Separate, smaller bugs observed in the same plugin (not part of #3791): a hand-rolled `timeconvert()` that skews `GpsTime`, a stray all-zero point emitted on empty reads, and `MB_SIDESCAN_NULL` padding in `datatype=sidescan`. Happy to send these as follow-up PRs if welcome.

### DIFF
--- a/plugins/mbio/io/MbError.cpp
+++ b/plugins/mbio/io/MbError.cpp
@@ -66,6 +66,7 @@ std::map<int, std::string> errors =
     { 14, "Bad system." },
     { 15, "Bad data." },
     { 16, "Missing data." },
+    { 17, "Bad time value." },
     { -1, "Time gap." },
     { -2, "Position out of bounds." },
     { -3, "Time out of bounds." },
--- a/plugins/mbio/io/MbReader.cpp
+++ b/plugins/mbio/io/MbReader.cpp
@@ -96,8 +96,8 @@ void MbReader::ready(PointTableRef table)
     int pings = 0;  // Perhaps an argument for this?
     int lonflip = 0; // Longitude -180 -> 180
     double bounds[4] { -180, 180, -90, 90 };
-    int btime_i[7] { 0, 0, 0, 0, 0, 0, 0 };
-    int etime_i[7] { (std::numeric_limits<int>::max)(), 0, 0, 0, 0, 0 };
+    int btime_i[7] { 1962, 2, 21, 10, 30, 0, 0 };
+    int etime_i[7] { 2062, 2, 21, 10, 30, 0, 0 };
     char *mbio_ptr;
     double btime_d;
     double etime_d;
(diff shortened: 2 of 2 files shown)


epifanio:
@hobu this fixes the PDAL-mbio initialization failure we tried to look at in Florence. It was a change on mbsystem side. With this PR and the one about level2 MBIO - the driver is back alive. Currently working on some example usage.

PR #5057, opened 2026-06-19
readers.mbio: fix GpsTime, drop phantom point, skip sidescan nulls

epifanio:
## Summary

Three small, independent correctness fixes in `readers.mbio`, separate from the MB-System ≥ 5.7.6 initialization regression (#3791, see PR #5056). Each can stand alone.

### 1. `GpsTime` was wrong by ~1900 years + one month
`mb_read()` already returns the ping time as a Unix epoch (UTC) in `pingTimeT`, but `loadData()` ignored it and used a hand-rolled `timeconvert()` that **adds 1900 to MB-System's already-4-digit year** and indexes the cumulative-days table with a **1-based month**. Result: `GpsTime` skewed by ~1900 years and a month.
**Fix:** use `pingTimeT` directly; remove the dead `timeconvert()` / `leapSeconds()` helpers. (`GpsTime` now matches `mbnavlist`.)

### 2. Phantom `(0,0,0)` point on empty reads
`read()` bound a `PointRef` to `view->point(0)` before any data was produced, which can materialize a stray all-zero point when a file yields no matching records.
**Fix:** use the canonical streamable pattern `PointRef point(*view)` + `setPointId` (cf. `io/TextReader.cpp`).

### 3. `datatype=sidescan` emitted null-fill pixels
`extractSidescan()` queued every processed-sidescan slot, including the `MB_SIDESCAN_NULL` (`-1e9`) null fill, as points.
**Fix:** skip null pixels.

## Diff
+13 / −90 in `plugins/mbio/io/MbReader.cpp` (the deletions are the dead `timeconvert`/`leapSeconds`).

## Testing
Compiles against PDAL + MB-System (Ubuntu 24.04). `GpsTime` verified against `mbnavlist`; sidescan output no longer contains `-1e9` fill points; an empty/comment-only file yields zero points.

## Relationship to #5056
Independent of #5056 (the #3791 init/zero-point crash fix) — different lines, no overlap; either can merge first.

### DIFF
--- a/plugins/mbio/io/MbReader.cpp
+++ b/plugins/mbio/io/MbReader.cpp
@@ -132,94 +132,6 @@ void MbReader::ready(PointTableRef table)
         (void **)&m_sslat, &error);
 }
 
-namespace
-{
-
-int leapSeconds(int y, int m)
-{
-    struct leap
-    {
-        int year;
-        int month;
-    };
-
-    const std::vector<leap> leaps
-    {
-        {1981, 7}, {1982, 7}, {1983, 7}, {1985, 7}, {1988, 1}, {1990, 1},
-        {1991, 1}, {1992, 7}, {1993, 7}, {1994, 7}, {1996, 1}, {1997, 1},
-        {1999, 1}, {2006, 1}, {2009, 1}, {2012, 7}, {2015, 7}, {2017, 1}
-    };
-
-    int secs = 0;
-    for (size_t i = 0; i < leaps.size(); ++i)
-    {
-        if (y > leaps[i].year)
-            secs++;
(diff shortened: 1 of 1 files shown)


abellgithub:
Has this PR been written by AI? It seems like it.

epifanio:
Hi @abellgithub - most of the work has been done with AI -  - I used a Opus 4.8 model (hi tier) - but even with its help took me a week of testing and prompt-eng ... going back and fwd with real survey data to test all the assumptions made - so short answer is yes.

hobu:
AFAIK, no one is using `readers.mbio`. Any activity including more test code to get it in working shape is welcomed.

PR #5063, opened 2026-06-25
`readers.ept`: add `ept.json` metadata items to reader metadata

ibell13:
Added the point count and other items from an EPT dataset's `ept.json` to the reader's metadata. This is helpful in #5007 where I need to read point count based on reader metadata. `hierarchyType` is not added currently.

### DIFF
--- a/io/EptReader.cpp
+++ b/io/EptReader.cpp
@@ -299,7 +299,7 @@ void EptReader::initialize()
     }
 
     // Adding point count info from ept.json to metadata
-    m_metadata.add<point_count_t>("points", m_p->info->pointCount(),
+    m_metadata.add<point_count_t>("points", m_p->info->points(),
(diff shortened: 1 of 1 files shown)


io/EptReader.cpp
```diff
@@ -298,6 +298,10 @@ void EptReader::initialize()
         throwError(e.what());
     }
 
+    // Adding point count info from ept.json to metadata
+    m_metadata.add<point_count_t>("points", m_p->info->points(),
```
hobu: Side point: this count has the potential to be out of sync with the actual point count of the EPT data. PDAL will still incrementally read it in that case.

Let's follow suit with what we're doing for `readers.copc` as far as metadata. https://github.com/PDAL/PDAL/blob/master/io/CopcReader.cpp#L331

PR #5064, opened 2026-06-26
examples: Align with main project CMake minimum 3.13

t0b3:
Since version 3.31, CMake warns that compatibility with CMake < 3.10 will be removed from a future version of CMake. Moreover, the project already has 3.13 baseline.

Closes: https://bugs.gentoo.org/976077


hobu:
Thanks. I missed these.

PR #5066, opened 2026-06-27
Sort stages in TOC alphabetically

quillaja:
This reorganizes the documentation TOC for stages/readers, stages/writers, and stages/filters alphabetically.

As a user, it's a lot easier to find the stage I'm looking for when I know its name and I can use my ABCs.

### DIFF
--- a/doc/_toc.yml
+++ b/doc/_toc.yml
@@ -38,8 +38,8 @@ parts:
       - file: stages/readers.buffer
       - file: stages/readers.copc
       - file: stages/readers.draco
-      - file: stages/readers.ept
       - file: stages/readers.e57
+      - file: stages/readers.ept
       - file: stages/readers.faux
       - file: stages/readers.fbi
       - file: stages/readers.gdal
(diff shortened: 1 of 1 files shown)


kasparas-k:
About the order of filters, in the current documentation there's some level of grouping by function, which I'd argue is more useful than alphabetical sorting. For example all four ground classification filters are listed in order (csf, pmf, skewnessbalancing, smrf) which is helpful for new users exploring documentation. If you know what you're looking for, you can always ctrl+f for it regardless of ordering.

quillaja:
Yes, I noticed that in the toc's yaml. But those groups do not appear on the website in the left-hand toc. Visible groups by function would be a fine alternative to pure alphabetical sort.

hobu:
> Visible groups by function would be a fine alternative to pure alphabetical sort.

It would be fine to fix the TOC so it groups thing accordingly, but we are not going to break up the grouping for pure alphabetical sorting.

quillaja:
To be clear, my intention was to list only the left-hand TOC alphabetically, not the "filters.md" page.

image


I created a few additional "sections" for the filters toc. Doing so at least makes it obvious why the filters are listed in the order they are.

image

That said, it appears each section must have an additional file for it. In my opinion, that adds no useful content (ie those files would be filled with "fluff") but creates more files to maintain and manage.

Thank you for considering this PR.

PR #5073, opened 2026-07-13
CI: back up one for our curl build

hobu:


hobu:
No longer needed. Fixed by https://github.com/conda-forge/icu-feedstock/pull/51

PR #5076, opened 2026-07-16
Fix COPC writer UTF-8 output paths on Windows

### LINKED ISSUE #5075: writers.copc crashes on Windows when output filename contains UTF-8 non-ASCII characters
## Summary

On Windows, `writers.copc` crashes when its output filename contains UTF-8 non-ASCII characters. The issue is limited to the COPC writer's output path handling.

## Environment

- PDAL 2.9.3 (`git-version: 5f0a5f`)
- Windows x64, MSVC/vcpkg build

## Reproduction

With an ASCII input path and a UTF-8 output path:

```powershell
pdal translate ascii-input.las "中文 路径/output.copc.laz" --writer=writers.copc
```

The process exits with `-1073740791` (`0xC0000409`).

## Observed matrix

| Input path | Output path | Writer | Result |
| --- | --- | --- | --- |
| UTF-8 non-ASCII | UTF-8 non-ASCII | `writers.las` | succeeds |
| ASCII | ASCII | `writers.copc` | succeeds |
| UTF-8 non-ASCII | ASCII | `writers.copc` | succeeds |
| ASCII | UTF-8 non-ASCII | `writers.copc` | crashes |

The COPC output cannot be read back because writing fails. ASCII COPC output is successfully readable via `readers.copc`.

## Suspected cause

Two COPC writer code paths pass PDAL's UTF-8 `std::string` filename directly to narrow C++ file-stream overloads:

- `io/private/copcwriter/Output.cpp`
- `io/private/copcwriter/Processor.cpp`

On Windows, those overloads use the active code page rather than reliably converting UTF-8 to the native UTF-16 path.

A fix and regression test are available in a follow-up PR.

AsterZ99:
Fixes #5075.

## What changed

- Add a Windows-specific UTF-8-to-native-path helper for COPC writer file streams.
- Use it for the initial output stream and random-access compressed-chunk writes.
- Add a COPC writer regression test that writes and reads back a filename containing UTF-8 non-ASCII characters.

## Why

The COPC writer passed PDAL's UTF-8 filename string to narrow C++ file-stream overloads. On Windows these use the active code page, so a non-ASCII output filename can fail or crash. Other platforms retain the existing byte-path behavior.

## Validation

- Reproduced the prior Windows failure with `writers.copc` and a UTF-8 output filename.
- Verified the regression exercises initial output creation and chunk writes using more than 100,000 points, then reopens the result with `readers.copc`.
- Confirmed the patch has no whitespace errors with `git diff --check`.

### DIFF
--- a/io/private/copcwriter/Common.hpp
+++ b/io/private/copcwriter/Common.hpp
@@ -36,7 +36,6 @@
 
 #include <stdint.h>
 #include <array>
-#include <filesystem>
 #include <map>
 #include <set>
 #include <string>
@@ -74,19 +73,6 @@ enum class Index
 
 namespace copcwriter
 {
-
-// std::fstream's narrow-string overload uses the active Windows code page.
-// PDAL filenames are UTF-8, so use the filesystem path overload on Windows to
-// preserve non-ASCII characters.  Other platforms retain their byte-path
-// behavior.
-inline std::filesystem::path nativePath(const std::string& filename)
-{
-#ifdef _WIN32
-    return std::filesystem::u8path(filename);
-#else
-    return std::filesystem::path(filename);
-#endif
-}
 
 // These are hopes, not absolutes.
 const int MaxPointsPerNode = 100000;
--- a/io/private/copcwriter/Common.hpp
+++ b/io/private/copcwriter/Common.hpp
@@ -36,6 +36,7 @@
 
 #include <stdint.h>
 #include <array>
+#include <filesystem>
 #include <map>
 #include <set>
 #include <string>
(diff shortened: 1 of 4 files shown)


io/private/copcwriter/Common.hpp
```diff
@@ -74,6 +75,19 @@ enum class Index
 namespace copcwriter
 {
 
+// std::fstream's narrow-string overload uses the active Windows code page.
```
abellgithub: There is already code to convert to UTF-16. See FileUtils::toNative()

AsterZ99: Thanks — updated the PR to use `FileUtils::toNative()` for both COPC writer streams. I also adjusted the regression test to focus on successful COPC output to a UTF-8 filename. It passes in a Windows Release build, and I verified the rebuilt CLI can write and read back a COPC file at a Chinese path.

test/unit/io/CopcWriterTest.cpp
```diff
@@ -259,6 +259,33 @@ TEST(CopcWriterTest, scaling)
     FileUtils::deleteFile(filename);
 }
 
+TEST(CopcWriterTest, unicodeFilename)
```
abellgithub, line 262: This test doesn't test the error condition. If you comment out the changes in Processor.cpp and Output.cpp the test passes.

AsterZ99: Thanks for the catch. This is Windows-specific: on my Windows/MSVC build, commenting out the two `FileUtils::toNative()` calls makes this test fail because the UTF-8 output file is not created (and its size is 0). Restoring the calls makes it pass. I have updated the test to skip on non-Windows platforms so it does not imply coverage there.

abellgithub: This was my fault. I forgot this was a Windows issue when testing.

PR #5090, opened 2026-07-31
writers.arrow: do not write xyz column to GeoParquet because it is redundant

wonder-sk:
Currently, writers.arrow writes point geometry both as `wkb` column, and also as `xyz` column - but the `xyz` column seems to be redundant. This PR changes the behavior so that `xyz` is only written to Feather output, and `wkb` is written to GeoParquet.

As a result:
- a test output geoparquet file is 345 MB instead of 476 MB - avoiding redundant data
- it is necessary to pass `geoarrow_dimension_name=wkb` to readers.arrow, otherwise such file gets read without XYZ dimensions

I am not an expert on all the GeoParquet stuff, so I am curious what you think...

### DIFF
--- a/plugins/arrow/io/ArrowWriter.cpp
+++ b/plugins/arrow/io/ArrowWriter.cpp
@@ -194,7 +194,8 @@ private:
 class WkbHandler : public BaseDimHandler
 {
 public:
-    WkbHandler(arrow::MemoryPool *pool) : m_builder(arrow::fixed_size_binary(29), pool)
+    WkbHandler(arrow::MemoryPool *pool, const std::string& pipelineMetadata = std::string()) :
+        m_pipelineMetadata(pipelineMetadata), m_builder(arrow::fixed_size_binary(29), pool)
     {}
 
     FieldPtr field() override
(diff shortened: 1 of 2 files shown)


plugins/arrow/io/ArrowWriter.cpp
```diff
@@ -194,7 +194,8 @@ class XyzHandler : public BaseDimHandler
 class WkbHandler : public BaseDimHandler
 {
 public:
-    WkbHandler(arrow::MemoryPool *pool) : m_builder(arrow::fixed_size_binary(29), pool)
+    WkbHandler(arrow::MemoryPool *pool, const std::string& pipelineMetadata = std::string()) :
```
ibell13, line 197: Maybe `WkbHandler` should accept `m_geoArrowDimensionName` if we want the WKB field name to be modifiable (even though it wasn't before); the defaults for it could be set in `initialize()` rather than `addArgs()`, conditioned on the output format.

ibell13 (approved):
Looks good to me, the changes from my comment aren't necessary and could be added in a new PR.

abellgithub:
What does the pipeline metadata have to do with the writing/not writing of XYZ data?

wonder-sk:
> What does the pipeline metadata have to do with the writing/not writing of XYZ data?

@abellgithub The `XyzHandler` class contains code that writes the pipeline metadata. Now that XyzHandler is not used for parquet output anymore, the same code was added to `WkbHandler` to write those metadata...

PR #5091, opened 2026-07-31
fix(readers.stac): resolve relative hrefs against remote source URLs

### LINKED ISSUE #5088: `readers.stac` incorrectly resolves relative asset URLs in a remote VPC
### Description

When reading a remote VPC whose assets use relative `href` values, `readers.stac` resolves them as local filesystem paths instead of URLs.

For example:

```json
{
  "type": "readers.stac",
  "filename": "https://data.geopf.fr/chunk/telechargement/download/LiDARHD-NUALID/NUALHD_1-0__LAZ_LAMB93_PQ_2025-03-14/index.vpc",
  "bounds": {
    "minx": 890250,
    "miny": 6262500,
    "maxx": 890750,
    "maxy": 6263500,
    "crs": "EPSG:2154"
  }
}
```

The VPC contains relative asset links such as:

```text
./LHD_FXX_0890_6263_PTS_LAMB93_IGN69.copc.laz
```

According to the STAC Item Specification, an Asset Object's `href` may be either a relative or an absolute URI.  

PDAL incorrectly resolves the asset to:

```text
/home/esgn/local_directory/https://data.geopf.fr/.../LHD_FXX_0890_6263_PTS_LAMB93_IGN69.copc.laz
```

and fails with:

```text
PDAL: No driver for /home/ESeguin2/test_gpf/https://data.geopf.fr/.../LHD_FXX_0890_6263_PTS_LAMB93_IGN69.copc.laz
```

### Expected behavior

The relative asset should be resolved against the remote VPC URL:

```text
https://data.geopf.fr/chunk/telechargement/download/LiDARHD-NUALID/NUALHD_1-0__LAZ_LAMB93_PQ_2025-03-14/LHD_FXX_0890_6263_PTS_LAMB93_IGN69.copc.laz
```

### Environment

```text
PDAL 2.10.2
git-version: da2cd8
```

Using absolute asset URLs in a local copy of the VPC works correctly.

esgn:
Fixes #5088.

`readers.stac` was resolving relative asset/link hrefs of a remote STAC/VPC document against the local filesystem CWD, producing paths like `/home/user/https:/data.geopf.fr/…/x.copc.laz`.

`handleRelativePath()` now resolves references against the source URL (RFC 3986) when the source is actually remote. Local sources remain untouched.

Side effects on remote sources: root-relative hrefs (`/x.laz`) now resolve against the authority, dot-segments are normalized, and a reference's own query/fragment is preserved.

Unit-testing without network (`StacReaderTest.relative_path_resolution`) was added. 
Also verified end-to-end against the VPC listed in #5088.


test/unit/io/StacReaderTest.cpp
```diff
@@ -565,6 +565,87 @@ TEST(StacReaderTest, wrench_test)
...
+
+    // Absolute hrefs pass through untouched.
+    EXPECT_EQ(handleRelativePath(vpc, "s3://bucket/x.laz"),
+        "s3://bucket/x.laz");
+    EXPECT_EQ(handleRelativePath(vpc, "https://other/x.laz"),
+        "https://other/x.laz");
+
+    // A GDAL VSI path is not a URL (its "://" follows a '/'): under a
+    // remote base it resolves as root-relative rather than passing through.
+    EXPECT_EQ(handleRelativePath(vpc, "/vsicurl/https://other/x.laz"),
```
kylemann16: Are people really adding `/vsicurl/` to their asset paths? I've never seen that. This test does the job either way, but we might consider just making it a root path with a url in a query string since I think that's the more likely use case for this logic.

kylemann16 (approved):
Looks good to me. Thanks for fixing!

test/unit/io/StacReaderTest.cpp
esgn: You're right. This case is removed and any mention of vsicurl is also removed from the commit.

test/unit/io/StacReaderTest.cpp
```diff
@@ -565,6 +565,87 @@ TEST(StacReaderTest, wrench_test)
     EXPECT_EQ(view->size(), 111065u);
 }
 
+TEST(StacReaderTest, relative_path_resolution)
```
ibell13: I think `test/unit/StacUtilsTest.cpp` should be created and this should be moved there

ibell13 (commented):
Looks good, thanks! Only some minor changes that I'll push myself.

PR #5096, opened 2026-08-05
readers.arrow: auto-detect primary geometry column from GeoParquet files

wonder-sk:
Previously users would have to specify the geometry column explicitly, e.g.
```
pdal translate --readers.arrow.geoarrow_dimension_name="geometry" <input_file.parquet> <output_file.copc.laz>
```

Now the reader uses the `geo` -> `primary_column` metadata and user does not need to pass the geometry column on input.


ibell13 (approved):
Looks good to me

PR #5112, opened 2026-08-26
install gdal-master for windows

hobu:
Testing the latest curl fixes on windows to make sure there's nothing left on the PDAL side.


hobu:
The new failures on windows are something with `writers.ogr` that happens on windows. It will need to be investigated separately, but the curl issue appears to be fixed ❇️ 🎊

PR #5120, opened 2026-08-31
Make EptInfo header self-contained

ClayWarren:
## Summary

The formatting pass exposed a latent include-order dependency: `io/private/ept/EptInfo.hpp` stores `NL::json` by value but relied on a transitive include for the complete `nlohmann/json.hpp` definition. Include it directly to make the header self-contained.

This is kept separate from the formatting changes so the review stays focused.

## Validation

- Compiled a standalone translation unit that includes `io/private/ept/EptInfo.hpp` with the configured build include roots
- Ran `git diff --check`

### DIFF
--- a/io/private/ept/EptInfo.hpp
+++ b/io/private/ept/EptInfo.hpp
@@ -38,6 +38,8 @@
 #include <pdal/SpatialReference.hpp>
 #include <pdal/util/Bounds.hpp>
 
+#include <nlohmann/json.hpp>
+
 #include "FixedPointLayout.hpp"
 
 namespace pdal


abellgithub:
This doesn't harm anything, but it also doesn't seem to help anything - change for change sake. Is there some issue that's being fixed?

ClayWarren:
There is no reported user-facing issue. I separated this because it was the only semantic change from the formatting and static-analysis work, and it currently forms the base of [ClayWarren/PDAL #8](https://github.com/ClayWarren/PDAL/pull/8). If this kind of header self-containment and static-analysis cleanup is not useful to PDAL, I can close this PR and the stacked fork PR rather than pursue either further upstream.

PR #5125, opened 2026-09-06
Add a mode implementation to filters.stats

### LINKED ISSUE #5123: Mode in filters.stats
It would be really good to have a real mode in the filters.stats. For making maps like classification maps, this si really important (you cannot just average class values).

hobu:
@mazingaro here's an attempt at implementation of `mode` for `filters.stats`. 

Can you please suggest with specificity what you were talking about in regard to corners? I don't remember exactly what you were asking for.

### DIFF
--- a/filters/StatsFilter.hpp
+++ b/filters/StatsFilter.hpp
@@ -55,14 +55,31 @@ public:
 
 typedef std::map<double, point_count_t> EnumMap;
 typedef std::vector<double> DataVector;
+typedef uint32_t GlobalStats;
+
+    enum GlobalStat
+    {
+        GlobalNone = 0,
+        GlobalMedian = 1,
+        GlobalMad = 2,
+        GlobalMode = 4,
+        GlobalAll = GlobalMedian | GlobalMad | GlobalMode
+    };
 
 public:
-    Summary(std::string name, EnumType enumerate, bool advanced = true) :
-        m_name(name), m_enumerate(enumerate), m_advanced(advanced)
-    { reset(); }
+    Summary(std::string name, EnumType enumerate, bool advanced = true,
+            GlobalStats globalStats = GlobalNone) :
+        m_name(name), m_enumerate(enumerate), m_advanced(advanced),
+        m_globalStats(globalStats)
+    {
+        if (m_enumerate == Global && m_globalStats == GlobalNone)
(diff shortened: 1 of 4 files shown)


mazingaro:
> @mazingaro here's an attempt at implementation of `mode` for `filters.stats`. 
> 
> Can you please suggest with specificity what you were talking about in regard to corners? I don't remember exactly what you were asking for.
> 

Thank you very much!
I've opened a new issue #5126 where I described it a bit better and tried to give some suggestions. I checked a lot if such things are alredy implemented, but couldn't find anything... So I hope I don't spam.

filters/StatsFilter.hpp
```diff
@@ -55,14 +55,31 @@ class PDAL_EXPORT Summary
 
 typedef std::map<double, point_count_t> EnumMap;
 typedef std::vector<double> DataVector;
+typedef uint32_t GlobalStats;
```
abellgithub, line 58: Don't think this type adds anything, but meh.

abellgithub (approved):

test/unit/filters/StatsFilterTest.cpp
```diff
@@ -407,35 +407,151 @@ TEST(Stats, enum)
...
-	EXPECT_DOUBLE_EQ(statsZ.minimum(), 0.0);
-	EXPECT_DOUBLE_EQ(statsZ.maximum(), 1000.0);
+    EXPECT_DOUBLE_EQ(statsZ.median(), 20.0);
+    EXPECT_DOUBLE_EQ(statsZ.mad(), 10.0);
+    EXPECT_DOUBLE_EQ(statsZ.minimum(), 0.0);
+    EXPECT_DOUBLE_EQ(statsZ.maximum(), 50.0);
+    EXPECT_DOUBLE_EQ(statsZ.mode(), 10.0);
+}
+
+TEST(Stats, globalMode)
```
abellgithub, line 464: This seems like test bloat. Maybe just use the values from this test in the above and call it a day. All the metadata checking seems unnecessary.

hobu: The test above doesn't have a real mode. I can trim the other stuff down, but I wanted to make sure we properly gathered the mode.

ibell13 (approved):
Looks good to me

PR #5129, opened 2026-09-10
Fix u8string usage for C++20 compatibility

ibell13:
PDAL wasn't compiling with C++20 due to `u8string()` no longer returning std::string.


hobu:
> PDAL wasn't compiling with C++20 

Was this some other package that was compiling as C++20?

> `__cpp_lib_char8_t` 

It wouldn't seem to me that a dirty ifdef is the proper fix for this. Is it?

ibell13:
@hobu

> Was this some other package that was compiling as C++20?

When I was trying to link libarrow for my updates to #5007 (haven't pushed yet) it was forcing everything to compile with C++20. I'm not sure whether we would want/need to update the required c++ version for those fixes yet, though.

> It wouldn't seem to me that a dirty ifdef is the proper fix for this. Is it?

It's what was used by #4979 so I went with it, but I can take another look.

abellgithub:
Seems strange that you'd have a link issue because arrow used C++20.

abellgithub:
This stuff is a freaking mess (thanks C++ committee), but I think this fix is OK in general, though we should probably stick it in a util function so as not to bleed all over.

hobu (approved):

